PR #16, opened 2017-06-03
Custom babel config support

CompuIves:


CompuIves:
I decided to merge this in, but to keep the custom babel config behind a feature flag that's disabled for now. I want to include custom babel support when we support proper templating and ejecting. Otherwise the data in the db will be too polluted as soon as we move over to the template system.

PR #24, opened 2017-06-17
Package Refresh Icon

donavon:
Add a refresh icon to each npm package dependency. This is more discoverable that entering the package name and clicking "Replace Package".

codesandbox-refresh

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -25,6 +25,15 @@
         "test",
         "tool"
       ]
+    },
+    {
+      "login": "donavon",
+      "name": "Donavon West",
+      "avatar_url": "https://avatars0.githubusercontent.com/u/887639?v=3",
+      "profile": "http://donavon.com",
+      "contributions": [
+        "code"
+      ]
     }
   ]
 }
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # [CodeSandbox](https://codesandbox.io)
-[![All Contributors](https://img.shields.io/badge/all_contributors-1-orange.svg?style=flat-square)](#contributors)
+[![All Contributors](https://img.shields.io/badge/all_contributors-2-orange.svg?style=flat-square)](#contributors)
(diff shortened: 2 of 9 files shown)


CompuIves:
This is awesome! Thanks a lot for this pull request! 😄 😄 

PR looks good, I have 3 small nitpicks:

- Could you swap the two icons (so that the cross icon is on the right side), I believe this will reduce misclicks for users that are used to have the cross icon on the right side.
- I think it's better for the refresh icon to have a white color on hover, I associate red with deletion and this may cause some confusion.
- Could you add some spacing between the icons (something like 0.2em)?

Again, this is really great, thanks @donavon!

donavon:
Yeah, I thought that the icons needed to be swapped as well, but only _after_ I submitted. 😉  About the color... I think you're right that it should be white (I reused the code), but that brings up a question of consistency. The remove icon for a Files is white. Why is is red(-ish) for npm packages? The code could be cleaner if they were the same. Do you mind if I make remove white as well (i.e. the same as with Files)

image

And yes, I'll add some spacing between icons.

BTW, look for separate PR for "refresh all" coming soon!

image

CompuIves:
Ah, good points, I agree.

Let's make everything white in that case for the sake of consistency, we should then do this for 'External Resources' remove icon as well.

Another small thing, could you put a `transition: 0.3s ease color;` on the icons so they transition more slowly to the white color?

That "refresh all" function is also really great!! This will make it so much easier for users to stay up to date with dependencies, thanks a lot!

donavon:
OK, ready.

I DRYed things up a bit by adding a common `Icon` component. "Files" had a `.0.25em` padding and the transition that you are looking for. I've use the common `Icon` for "External Resources" as well, making it white.

However, since External Resources has jest tests, I needed to update the snap.

CompuIves:
This is really great @donavon!

I'll merge it in now 👍.

PR #26, opened 2017-06-22
Git support

CompuIves:


src/app/pages/Sandbox/Editor/Workspace/Project/Statistics.js
```diff
@@ -0,0 +1,47 @@
...
+
+const CountDescription = styled.span`
+  color: rgba(255, 255, 255, 0.6);
+`;
+
+const CountContainer = styled.div`
+  margin: .5rem 1rem;
+  font-size: .875rem;
+`;
+
```
donavon: You could DRY this up a bit...

```js
const CountContainer = ({className, count, singular, plural}) => (
  <div className={className}>
    <Count>{`${count} `}</Count>
    <CountDescription>
      {` ${count === 1 ? singular : plural}`}
    </CountDescription>
  </div>
);

const StyledCountContainer = styled(CountContainer)`
  margin: .5rem 1rem;
  font-size: .875rem;
`;
```

then below...

```js
export default ({ likeCount, forkCount, viewCount }: Props) => (
  <div>
    <WorkspaceSubtitle>Statistics</WorkspaceSubtitle>
    <StyledCountContainer count={likeCount} singular="like" plural="likes" />
    <StyledCountContainer count={forkCount} singular="fork" plural="forks" />
    <StyledCountContainer count={viewCount} singular="view" plural="views" />
  </div>
);
```

donavon: I hate seeing a repeating pattern like this 😏

CompuIves: Awesome, very good point! Thanks for reviewing this PR and keeping me sharp 😄 .

donavon: It's fun. I'll try and find actual logic errors too and not just style issues :)

CompuIves: Great! I'll go for a run and go through your feedback after 🏃

src/app/pages/Sandbox/Editor/Workspace/Project/index.js
```diff
@@ -155,26 +147,23 @@ export default class Project extends React.PureComponent {
...
-        </ViewCountContainer>
-        <ViewCountContainer>
-          <ViewCount>{viewCount}{' '}</ViewCount>
-          <ViewCountDescription>
-            unique
-            {viewCount === 1 ? ' view' : ' views'}
-          </ViewCountDescription>
-        </ViewCountContainer>
+
+        {githubUrl &&
```
donavon: if `githubUrl` is an empty string (i.e. not a `null` or `undefined`), this will insert an HTML comment into the DOM. Not a big deal, but if you care, change the line to:
```js
{!!githubUrl &&
```

src/app/pages/Sandbox/Editor/Workspace/Project/index.js
```diff
@@ -155,26 +147,23 @@ export default class Project extends React.PureComponent {
...
+            <WorkspaceSubtitle>Source Control</WorkspaceSubtitle>
+            <Item>
+              <a href={githubUrl} rel="noopener noreferrer" target="_blank">
+                {repoName}
+              </a>
+            </Item>
+          </div>}
+
+        <Statistics
+          viewCount={viewCount}
```
donavon: Man... It sure will be nice when React allows "props shorthand syntax" so we can just do this:
```html
<Statistics viewCount likeCount forkCount />
```
The only way around it currently is to do this:
```js
const StatisticsProps = { viewCount, likeCount, forkCount };
...
<Statistics {...StatisticsProps} />
```
I don't expect you to change anything, just venting.

src/app/store/entities/sandboxes/boilerplates/default-boilerplates.js
```diff
@@ -1,6 +1,7 @@
 export const JS = {
   id: 'js',
-  condition: '\.js$', // eslint-disable-line
+  extension: '.js',
+  condition: '.js$', // eslint-disable-line
```
donavon: Why? As a general rule, IMO, you should add the eslint rule that you are disabling on the line. Ex:
```js
// eslint-disable-line no-use-before-define
```
otherwise you could be masking a different eslint error

donavon: But does this mean that we can now use `.jsx`? Yay! 🎉

src/app/utils/url-generator.js
```diff
@@ -30,6 +30,18 @@ export const profileSandboxesUrl = (username: string, page?: number) =>
...
+  username,
+  path,
+}: {
+  repo: string,
+  branch: string,
+  username: string,
+  path: string,
+}) => `https://github.com/${username}/${repo}/tree/${branch}/${path}`;
+
 export const optionsToParameterizedUrl = (options: Object) => {
```
donavon: You need to encode your keys and values. You should also sort so that your cache is more predictable.
```js
export const optionsToParameterizedUrl = (options: Object) => {
  const keyValues = Object.keys(options)
     .sort()
     .map(key => `${encodeURIComponent(key)}=${encodeURIComponent(options[key])}`)
     .join('&');
  return keyValues ? `?${keyValues}` : '';
 };
```

CompuIves: Ah this is great! Really good feedback!

src/app/pages/Sandbox/Editor/Workspace/Project/index.js
CompuIves: Ah this is also a really good point, thanks!

src/app/pages/Sandbox/Editor/Workspace/Project/index.js
CompuIves: Hmm you're right, do this would also clash with the boolean implicit =true. Interesting

src/app/store/entities/sandboxes/boilerplates/default-boilerplates.js
CompuIves: Ah, this is a leftover from the 286th iteration of CodeSandbox. It's called boilerplates, if you enable module view and view a component that only exports itself it will run the boilerplate for the `.js` file and render the component.

The issue was that I've changed the caching logic to use a special caching strategy on `.js`, but this wasn't propagated to boilerplates.

I'm not planning to add `.jsx` support as `create-react-app` doesn't support this either, I want to keep compatibility with `create-react-app`.

src/app/utils/url-generator.js
```diff
@@ -9,19 +9,35 @@ export const host = () => {
...
+  path: string,
+}) =>
+  'github/' + // eslint-disable-line prefer-template
+  encodeURIComponent(git.username) +
+  '/' +
+  encodeURIComponent(git.repo) +
+  '/tree/' +
+  encodeURIComponent(git.branch) +
+  '/' +
+  encodeURIComponent(git.path);
```
donavon: does this work? I'm not sure that `path` should be encoded. The others still need to be.

donavon: I have a cleaner version of `sandboxGitUrl` running. It will need to be converted to TS.
```js
const sandboxGitUrl = ({ username, repo, branch, path }) => (
  buildEncodedUri`/s/github/${username}/${repo}/tree/${branch}/` + path
);
```

It uses a [Tagged Template Literal](https://www.styled-components.com/docs/advanced#tagged-template-literals) (the same thing used in `styled-components`) that I wrote called `buildEncodedUri` to call `encodeURIComponent` with each value in the template.

```js
const buildEncodedUri = (strings, ...values) => (
  strings[0] + values.map((value, i) => (
    `${encodeURIComponent(value)}${strings[i + 1]}`
  )).join('')
);
```

It might be little overkill for this, but it's reusable for other URI building applications.

see it here on [codesandbox](https://codesandbox.io/s/RoNWKDJqK) (you may have heard of it :)

CompuIves: Ahh this is genius! Great work! 

Really, this is awesome

CompuIves: Updated the PR

src/app/pages/Sandbox/Editor/Workspace/Project/index.js
donavon: I guess the proposal is this: https://github.com/facebook/jsx/issues/23 (I forgot to put inside of curly braces):
```html
<Statistics {viewCount} {likeCount} {forkCount} />
```

PR #48, opened 2017-07-11
fix($urlGenerator): use localhost:3000 if LOCAL_SERVER

faceyspacey:


CompuIves:
This is great, thanks for adding this!

CompuIves:
Oh by the way, can you open another PR where you add yourself as a contributor?

You can do that by running

```
npm run add
```

PR #53, opened 2017-07-13
Add .jsx extension support

### LINKED ISSUE #51: .jsx file extension support
This one should be easy to do, I found out that create-react-app does support the `jsx` extension. We should update the sandbox to handle these cases.

All the code for this resides in `src/sandbox/`, if you'd search for `js` you'll find where we check for the extension.

I *believe* all logic is in `src/sandbox/eval/js/index.js` and `src/sandbox/utils/resolve-module.js`.

bengummer:
Fixes #51


src/sandbox/eval/index.test.js
```diff
@@ -3,223 +3,225 @@ import { clearCache } from './js';
 
 describe('eval', () => {
```
bengummer: The changes in this file look big, but it's really just indented one level (because of the new `forEach` loop) and using the loop's `jsExtension` instead of static `.js`.

src/sandbox/utils/resolve-module.js
```diff
@@ -3,7 +3,7 @@ import type { Module, Directory } from 'common/types';
 
 const compareTitle = (original: string, test: string) => {
   if (original === test) return true;
-  if (original === `${test}.js`) return true;
+  if (original.match(new RegExp(`^${test}.jsx?$`))) return true;
```
bengummer: Used RegExp here, but happy to switch back to `===` if you prefer

CompuIves: Hmm, I'd prefer to let this still be a strict equal, since the case is very clear and building Regexes can be quite expensive.

bengummer:
Thanks for the feedback @CompuIves - I've reverted the regex and run the contributor task.

CompuIves:
Thanks a lot! I'll deploy it now.

PR #56, opened 2017-07-14
handle different GitHub URL input formats

### LINKED ISSUE #54: "Import from GitHub" page sometimes shows invalid output URL
If I go to https://codesandbox.io/s/github and enter `http://github.com/test/123` it output a sandbox URL of `https://codesandbox.iohttp://github.com/test/123` (or `https://www.github.com/test/123`) it outputs something like `https://codesandbox.iohttps://www.github.com/test/123`.

Should be a relatively uncommon edge case if people are copy + pasting GH URLs directly from their browser into the page.

I'm happy to take a look at fixing this :)

image

bengummer:
fixes #54


CompuIves:
Super! Also thanks for the tests!

PR #78, opened 2017-07-23
[WIP] Add font family to preferences

viankakrisna:
Add font family for preference. AFAIK this should be all from the client side, haven't got to test it yet until we have staging server :)

Also, why do we need to `injectGlobal` here @CompuIves ? I think we can move it to `CodeContainer`?

### DIFF
--- a/src/app/components/Preference/PreferenceSwitch.js
+++ b/src/app/components/Preference/PreferenceSwitch.js
@@ -18,6 +18,7 @@ export default class PreferenceSwitch extends React.PureComponent {
       <Switch
         onClick={this.handleClick}
         small
+        style={{ width: '3rem' }}
         offMode
         secondary
         right={value}
--- a/src/app/components/Preference/PreferenceText.js
+++ b/src/app/components/Preference/PreferenceText.js
@@ -17,11 +17,12 @@ export default class PreferenceText extends React.PureComponent {
   };
 
   render() {
-    const { value } = this.props;
+    const { value, placeholder } = this.props;
     return (
       <Input
         style={{ width: '9rem' }}
         value={value}
+        placeholder={placeholder}
         onChange={this.handleChange}
       />
     );
--- a/src/app/components/Preference/PreferenceText.js
+++ b/src/app/components/Preference/PreferenceText.js
@@ -1,34 +1,26 @@
 import React from 'react';
-import styled from 'styled-components';
 
 import Input from 'app/components/Input';
-
-const StyledInput = styled(Input)`
-  text-align: center;
-`;
 
 type Props = {
   value: boolean,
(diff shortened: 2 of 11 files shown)


CompuIves:
Really nice, I was able to get 'Arial' in my code editor hahaha. Okay great, I had to add one preference option to work with strings (only supported boolean and number until now), check this commit: https://github.com/CompuIves/codesandbox-client/commit/9942d3fe5f65d0cf9544328a2a3a7597df1ee2c1.

Would be great to move the styles to `CodeContainer`, I never liked `injectGlobal` very much.

CompuIves:
Can you also add a description underneath the Preference option?

viankakrisna:
ah yes, I'm not aware that it's only accepted boolean and number. 😄  Updated with https://github.com/CompuIves/codesandbox-client/commit/9942d3fe5f65d0cf9544328a2a3a7597df1ee2c1, description, and the moving of style to CodeContainer

CompuIves:
Haha, a coincidence. I'm currently implementing Prettier Preferences and now need the Preference type refactor too. Let me take a look.

CompuIves:
Looks good! I made one small change to add specify a placeholder ('Source Code Pro') in the input, makes it more clear on what to input. I think that's enough for the user to see what needs to happen, so we can remove the description.

I added the last required commit here: https://github.com/CompuIves/codesandbox-client/commit/bf114d81a85375c153bfef3f49951c06f0f6d6dc.

And this is how it looks like:
screen shot 2017-07-23 at 20 11 21


Then it'll be ready to merge! Thanks a lot for this PR, this is very valuable.

CompuIves:
Great! Let's ship this :shipit:  🚀

CompuIves:
Oh oops, forgot to change the name.  Well, 🤷‍♂️

viankakrisna:
😅  you move really fast! thank you!

CompuIves:
Haha, thanks. I'm thinking of replacing the preferences system for something that takes less boilerplate, now you need to change 4 things before it works. Very confusing and tedious.

viankakrisna:
yup, I missed some things when I first opened the PR. Today I just use this for forms [shameless plug](https://github.com/viankakrisna/react-reusable/blob/master/src/Form.js) plain react and lodash 😄

CompuIves:
Hmm, interesting. If we could convert the preferences to something like that it would be really nice.

Best case would be if we only had to define preferences in the store, and then it automatically gets synced with localStorage when updated.

viankakrisna:
maybe under one action? SET_PREFERENCE, and just validate the preference schema for that action payload.

CompuIves:
Yap, that's even better.

PR #80, opened 2017-07-23
simplify preferences store

viankakrisna:
again, not tested, but this should make adding a preference less boilerplate-y


CompuIves:
Holy macaroni, we were just talking about this less than two hours ago! 😱

CompuIves:
This looks *great*! Let me take it for a test run 🚗 .

CompuIves:
Hmm, it doesn't work yet, don't know why yet. 

I have a way for you to enable the preferences 😄 .

Run this in the console:

```
store.dispatch({type: 'SET_CURRENT_USER', data: { jwt: 'aefawef'}})
```

viankakrisna:
ah yes, that's what I needed! debugging....

viankakrisna:
should be working now, thanks for the preferences enabler!

CompuIves:
Niiiice! There is one last change you need to make in `Prettier.js`. That one is still using the old system.

CompuIves:
Oooh this is so great, now settings will be a breeze to add!

CompuIves:
Nice! Okay, one last quick review and it'll be ready to merge.

CompuIves:
Yup, this is great. Thanks again (for the 3rd time today!).

Issue #25, opened 2017-06-18
I18n support


CompuIves:
CodeSandbox is used by a lot of users in China and Russia, I think it'll be very valuable to have an i18n system in place that people could contribute to.

This will make CodeSandbox much more accessible.

Thinks that need to happen:

- [ ] Change build system to use [i18n-next](https://github.com/i18next/react-i18next/)
- [ ] Add a world icon to the header with a dropdown to select language, store preference in `localStorage` for now using preferences reducer
- [ ] Move all text to `t()` calls
- [ ] Move existing text to `en.json`

Arthelon:
I've previously used react-intl (FormatJS) for as an i18n library and transifex for managing translations (free for OSS). React-intl is more popular at the moment and has better in-built formatting but it's great that i18n-next comes with its own compatible translation management system.

I'd love to give this a try!

Arthelon:
Their docs are also much nicer! This [essentials section](https://www.i18next.com/essentials.html) is quite neat.

CompuIves:
That would be really awesome @Arthelon! This is a major feature and would benefit a lot for CodeSandbox.

I propose that we first store the translations in the repo as json files, because most developers that will translate already know github and so the barrier is lower to contribute in translations. 

I'm open to any suggestion though, let me know what you think or if you have any questions!

haraldur12:
Hey guys, I would love to help you out with i18n as well. Wouldn't it be good to first strip all the text from the code and then use a localisation service that offers free service to open source projects ?

steppefox:
@CompuIves what is the status of this issue? Is it still relevant? Looks like related PR was closed because it was too outdated.

vblazenka:
@CompuIves @steppefox I would also like to help with integrating i18n into the codesandbox.
I think the project will benefit because languages can open new markets in China and Russia (like mentioned before)

Saeris:
Adding this to our public backlog. Closing this out for now, please refer to this link for the status of this request:

https://www.notion.so/Add-i18n-Support-8be1d832f709428ebe5c06bd85bc7640

PR #84, opened 2017-07-25
Add i18n support using react-i18next

Arthelon:
Addresses #25 

This PR does the following:
- Adds a language dropdown that indicates the current site language and allows users to switch languages.
- Initializes an i18next instance with the language codes `en` (english) and `zh` (chinese). Translation source files (json) are pulled from the default path `/locales/{{language}}/{{namespace}}.json`
- Configures the gulp build so that translation source files are included in the build output
- Translates the `Sandbox` page title (using react-i18next's t function)
- Initializes flow type stubs for i18next-related libraries
- Adds browser language detector plugin

I've added Chinese as a second language for demonstration purposes, this can be removed if see fit.

### Next steps?
- [ ] Add a script to extract translation keys from source code into translation files
- [ ] Add a script to display translation coverage
- [ ] Define namespace/translation conventions
   - For this PR, I've separated namespaces into `common` and `sandbox`. Other than commonly used text, I think that it makes sense to split translation namespaces by different pages.
- [ ] Migrate to a translation management system (locize?) 
- [ ] Replace all text with `t` functions and move existing text to translation source files
- [ ] Add a notice (in the language dropdown?) asking for help with translations
- [ ] Improve dropdown styling
   - Add flags?
   - Disappear when focus is lost


CompuIves:
This is reaaally nice! I love the implementation. Also great to open an early PR so we can discuss the initial groundwork.

> Define namespace/translation conventions

Like you said, let's use the namespaces based on pages. `sandbox` can be a bit confusing (for me at least), since `sandbox` is the name of the sub application (the preview window). Maybe we can do something like `editor`, `profile`, `github`, etc. and a `common`.

> Add a notice (in the language dropdown?) asking for help with translations

Really great idea!

> Migrate to a translation management system (locize?)

Let's do this as last step, so we can focus on having a base to work from first.

> Replace all text with `t` functions and move existing text to translation source files

Hmm, I've been thinking a bit about this. Maybe we can also make `t` a global function or a singleton. Normally I avoid globals but there are some advantages in this case:

1. Translations almost never change, we can do a page reload on locale change.
2. We don't have to wrap every component in a HoC.

But I'm not sure how much support for something like this is, and maybe this is completely unnecessary and wrapping components with `translate` will not be a lot work. Hm, the more I think about this the more I'm leaning on just using HoC functions, would still like to hear your opinion on this 🤔 .

> I've added Chinese as a second language for demonstration purposes, this can be removed if see fit.

Chinese will be really important for us. Especially now I'm going to focus on other cli's like `vue`.

I'd love to help with this, though I need to check if I can find some time. I'm on a vacation to Vietnam right now and am not sure how often I can find time to work on CodeSandbox, we'll see 😄.

But like I said, really great work so far, I'm very impressed. Thanks a lot for this!

Arthelon:
As you said, there's no reason that `t` can't be used as a singleton but the HoC does provide reasonable defaults and configuration options (specifically for loading namespaces). 

In addition, the react-i18next library also provides two components for interpolating React code in translations so it might be best to use their HoC as well for consistency. Personally I don't think it's a lot of work (just an extra line more than directly importing the i18next instance). Maybe it can be added as a decorator for clarity?

CompuIves:
You're right, I think I'm overthinking the amount of boilerplate. Let's go with the HoC, seems like that is the most conventional one. I'll try to send in some PRs during the coming week for moving text to translation keys.

Arthelon:
Added some translations to the Header on the Editor page, lmk what you think @CompuIves

CompuIves:
Thanks!! This is perfect, exactly how it should be.

So this means that we're very close to finishing I18n support! Only the sidebar and the preferences need to be replaced with tags and we can start accepting translations.

This aligns perfectly with the Vue release, which should happen soon too.

CompuIves:
Awesome work @Arthelon! Thanks for all the effort, this is great!

PR #86, opened 2017-07-25
Add/preferences/line height

viankakrisna:


CompuIves:
Nice! This is cool, especially with the new preferences system this is so easy to review. I'll almost merge this in, there's one quick thing: I think we can remove the 'Editor' name from the last 3 options, so it will be:

- Font size
- Font family
- Line height

As we're already in the editor settings. After that I'll merge it in!

viankakrisna:
>especially with the new preferences system this is so easy to review.

That's the goal! 😄

CompuIves:
Nice! I'll try to deploy this today, but the connection here is 10 kbps. So if it won't work I'll try to deploy tomorrow from another location.

Issue #83, opened 2017-07-24
resizer shows over the settings modal


vueu:
When in settings, the bar that resizes the file tree, and the editor, is shown above the modal. Seen here:
![Resizer over](https://i.gyazo.com/c67182d4d621936aa2619def372c8175.png)

The modal had a z-index of 20. Bumping it up to 30 fixed it, as seen here:
![Resizer under](https://i.gyazo.com/4459edce5aee60dd6e72ad552448f03f.png)

I can put in a pull request on this tonight, let me know what you think!

CompuIves:
This is great! I always thought my screen was broken when I saw this line, good that we can finally fix this issue.

vueu:
> I always thought my screen was broken ...
😆

Arthelon:
@CompuIves Seems like this issue can be closed

ValentinH:
This issue still exists:
![image](https://user-images.githubusercontent.com/2678610/31586828-2ca84df2-b1d7-11e7-8ced-15629a477122.png)

I'll submit a PR now.

PR #90, opened 2017-07-28
Fix resizer showing over modals #83

vueu:


CompuIves:
Thanks! I can test this somewhere this evening, when I have a laptop. I'll keep you posted!

CompuIves:
Ah, I remember doing the `z-index: 30` there because I wanted to show this highlight on hover:

screen shot 2017-08-02 at 10 32 38

There must be a way to make the hover still show and not overlap on the other UI elements. 🤔

vueu:
Ah, yes. My issue was with the share modal. For some reason I could not get the share modal to show over the separator unless z-index on the separator was 0, although the settings modal worked fine. I'll definitely look into this more when I get the chance.

CSS is weird 😛

Arthelon:
@CompuIves Isn't it just a matter of sequencing the respective z-indexes so that the resizer appears above other elements but below the modal?

PR #93, opened 2017-07-31
remove extra question from COPPA section

crosscompile:
Not sure how I ended up reading the privacy policy so thoroughly, but this seems like a placeholder question that was never removed.

### DIFF
--- a/src/app/pages/Terms/PrivacyPolicy.js
+++ b/src/app/pages/Terms/PrivacyPolicy.js
@@ -279,10 +279,6 @@ export default () =>
       <div className="innerText">
         We do not specifically market to children under the age of 13 years old.
       </div>
-      <div className="innerText">
-        Do we let third-parties, including ad networks or plug-ins collect PII
-        from children under 13?
-      </div>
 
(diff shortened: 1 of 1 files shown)


CompuIves:
Thanks for the contribution @jayjaycross! I'm not an expert, but I always thought that everyone requires the COPPA section? If we don't require it we can certainly remove it 😃.

jayjaycross:
Ah sorry I'm not sure why the diff looks so funky, but I only deleted the line that says:

> Do we let third-parties, including ad networks or plug-ins collect PII from children under 13?

It seemed weird to have this unanswered question in the privacy policy, but [upon googling further](https://www.google.com/search?q=Do+we+let+third-parties%2C+including+ad+networks+or+plug-ins+collect+PII+from+children+under+13) it looks like this is present in a lot of privacy policies 🤔

lbogdan:
Closing this for now. I think what we could do is adding a "No." answer to the "Do we let third-parties [...]?" question, like [here](https://tedxmannheim.com/privacy), for example.

PR #120, opened 2017-08-16
fix error overlay

### LINKED ISSUE #117: Error Overlay styling
The error indicator is positioned absolute which makes is very hard to see the indicator at all. Also because of the positioning it doesn't scroll.

problem

My suggestion would be to style it the following way which makes it very clear where the error is and also no special positioning is needed.

suggestion

I could formalize an PR if you approve the suggestion or have better ideas to visualize that case.

johann-sonntagbauer:
Error overlay now uses normal flow and inserts an additional line for error information. So no overlay of text and info caret occur any longer.

fixes #117


CompuIves:
Wow that was quick! Thanks a lot!

Could you add yourself to the contributors? You can do this by running

```
npm run add
```

src/sandbox/react-error-overlay/utils/dom/absolutifyCaret.js
```diff
@@ -8,15 +8,6 @@
...
-function removeNextBr(parent, component: ?Element) {
-  while (component != null && component.tagName.toLowerCase() !== 'br') {
-    component = component.nextElementSibling;
-  }
-  if (component != null) {
-    parent.removeChild(component);
-  }
-}
-
 function absolutifyCaret(component: Node) {
```
CompuIves: I think we could remove this whole function, right?

CompuIves:
Just tested it, works great! The only standing thing is that I think we could delete `src/sandbox/react-error-overlay/utils/dom/absolutifyCaret.js` altogether.

src/sandbox/react-error-overlay/utils/dom/absolutifyCaret.js
johann-sonntagbauer: you are right, no additional logic remains. Removed the file and cleaned up. Thx for the detailed review.

CompuIves:
Great, thanks for the contribution!

PR #128, opened 2017-08-17
add editorconfig

Tushkiz:
annoyed by IDE warnings, adding this to maintain consistent styles between different editors 😅


CompuIves:
Oh good one! For me it's already in the monorepo, but not not for others of course.

PR #129, opened 2017-08-17
fix vim mode switch

### LINKED ISSUE #118: When disable Vim mode the code disappears
Hello! When I click `preferences` and select `Vim mode` everything is great, **but when open `preferences` again and disable the Vim mode**, the code _**disappears**_... View:

image

Video: https://youtu.be/KOo8K0i_Uog
I'm using Windows 10 x64bit, Chrome 60

johann-sonntagbauer:
ugly hack to work around eslint dependency which redefines the require instance on window
this breaks the monaco loader.js
the following workaround can be removed after migration to monaco (vim mode) is done

That is the most basic solution I could came up with. Please let me know what you think :)

fixes #118

### DIFF
--- a/src/app/utils/codemirror/eslint-lint.js
+++ b/src/app/utils/codemirror/eslint-lint.js
@@ -375,8 +375,9 @@ export default (async function initialize() {
   while (!window.eslint) {
     // eslint-disable-next-line
     await delay(100);
-    window.require = origRequire;
   }
+
+  window.require = origRequire;
 
   linter = new window.eslint();
   linter.defineRules(allRules);
--- a/src/app/utils/codemirror/eslint-lint.js
+++ b/src/app/utils/codemirror/eslint-lint.js
@@ -354,7 +354,15 @@ export function fix(source: string) {
 }
 
 export default (async function initialize() {
-  if (!window.eslint && linter === null) {
+  // ugly hack to work around eslint dependency which redefines the require instance on window
+  // this breaks the monaco loader.js
+  // the following workaround can be removed after migration to monaco (vim mode) is done
+  const origRequire = window.require;
+  if (
+    !window.eslint &&
+    linter === null &&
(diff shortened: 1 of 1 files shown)


src/app/utils/codemirror/eslint-lint.js
```diff
@@ -354,7 +354,15 @@ export function fix(source: string) {
 }
 
 export default (async function initialize() {
-  if (!window.eslint && linter === null) {
+  // ugly hack to work around eslint dependency which redefines the require instance on window
+  // this breaks the monaco loader.js
+  // the following workaround can be removed after migration to monaco (vim mode) is done
+  const origRequire = window.require;
```
CompuIves: Out of curiosity, why did you add this line?

CompuIves:
Really great, thanks again! Left one question, after that it's ready to merge.

src/app/utils/codemirror/eslint-lint.js
johann-sonntagbauer: Thats the hacky part of the change. Because eslint will override the window.require reference. So I store the original reference, load eslint and afterwards restore the original reference.

src/app/utils/codemirror/eslint-lint.js
```diff
@@ -367,6 +375,7 @@ export default (async function initialize() {
   while (!window.eslint) {
     // eslint-disable-next-line
     await delay(100);
+    window.require = origRequire;
```
johann-sonntagbauer: damn as I see now the placement is wrong here. I have to move it out of the loop. *shame*

src/app/utils/codemirror/eslint-lint.js
CompuIves: Good one! Ready to merge 😄

CompuIves:
Thanks!!

PR #131, opened 2017-08-18
Make '+' action into a dropdown

### LINKED ISSUE #106: Make "+" button in Sandbox header a dropdown
To reduce space occupied in the header, the github import, default new sandbox, and cli upload behaviour should be combined into a single dropdown.

Tushkiz:
Fixes #106 

Todo
- [x] add menu
- [x] add `/s/cli` link to menu
- [x] add `/s/vue` link to menu
- [x] fix state transition issue
- [x] add menu to `Navigation`
- [x] add `/s/preact` link to menu


CompuIves:
**Great**! Thanks! Are you comfortable with making the CLI page? It only has to look like the Github import and contain an example on how to instart the cli + a link to the npm package page.

Tushkiz:
Yup, definitely doing it.

Tushkiz:
@CompuIves I have added a cli page, review it.
Should we also change the '+' action in the Navigation?

CompuIves:
Super great! I'll review it after I have merged this: https://github.com/CompuIves/codesandbox-client/pull/137. You're right, we should alter the Navigation as well.

CompuIves:
Well, we suddenly got support for Vue 😄 . https://codesandbox.io/s/vue

I think it would be great if we would reflect this in the dropdown, something like:

- New React Sandbox
- New Vue Sandbox

Arthelon:
Maybe add an additional select input in the dropdown to specify the sandbox type!

Tushkiz:
@Arthelon I see, you are suggesting this because we might add support for more templates. In that case, I think we should display a modal which will show all the supported templates for selection.

CompuIves:
Hmm, yes I see the purpose of that as we get more sandboxes. Though right now I'd say that it would be best to show all solutions in the dropdown, because we don't have that many templates yet and it will save a click.

CompuIves:
Just tested this, looks very nice! I have some small suggestions before merging this:

1. It seems like I forgot to put a wrapper around the 'Import from GitHub' page, I think we could add it in this PR:

```js
      <MaxWidth>
        <Margin vertical={1.5} horizontal={1.5}>
          {/* Page Here */}
        </Margin>
      </MaxWidth>
```

2. The border-bottom highlight on the profile page is not shown properly, we can fix this by just removing the bottom highlight on these pages because that's the new design we want to follow.
screen shot 2017-08-23 at 16 50 22

3. It would be nice to add a chevron next to it, to indicate that it's a dropdown. We already have a prop for it called `moreInfo`. Something like the share view:

screen shot 2017-08-23 at 16 51 29

4. We just added Preact support (https://codesandbox.io/s/preact). So it'd be nice if that one could be added too (sorry! 😅 )

After that we're ready for takeoff!

Tushkiz:
Roger that 😅. I will update the PR

src/app/utils/url-generator.js
```diff
@@ -16,13 +16,16 @@ export const host = () => {
 export const protocolAndHost = () => `${location.protocol}//${host()}`;
 
 export const newSandboxUrl = () => `/s/new`;
```
jseminck: Should this be `/react` now instead of `/new` ? Or is there some backwards compatibility that could break by changing it?

Tushkiz: It would break the links which people have been sharing, but I agree with you.
May be we can add support for `/react` and keep `/new` for backwards compatibility.

CompuIves: Yep, we could add a redirect (in the pages component) from new to React. Let's do this in a separate PR to keep things separate.

CompuIves:
This is so great @Tushkiz! I'll merge this in when I get home, including some other changes to make our transition official.

Tushkiz:
Cool 🤘

CompuIves:
Is this PR done? In that case I'm eager to merge it in 😃 !

Tushkiz:
Yes, we can. Let me resolve the merge conflicts

CompuIves:
Oh hmmm it seems like prettier got its own life 😅 . I think updating prettier to 1.6 should fix it.

CompuIves:
Whoop, fixed it!

CompuIves:
Made a small change so we still see the bottom highlight on hover in the editor, I changed my mind about that. Waiting for tests and then merging it in.

Thank you very much for all the work and effort you put in this, this is a very important change!

Tushkiz:
Yay! 🍻 Let me know, what's next. It is fun.

PR #132, opened 2017-08-18
Add development support for people on Windows

jseminck:
Currently running `yarn start` will not work for people on Windows machines. (yes - we are a rare breed 😄 I just forgot my mac charger at work 😢 )

Anyway - using https://github.com/kentcdodds/cross-env makes this pretty straight forward. We use it in multiple projects at work and haven't run into any issues.

Without this, `yarn start` fails with the following error:

```
> yarn start
yarn start v0.27.5
warning ..\package.json: No license field
$ LOCAL_SERVER=1 node scripts/start.js
'LOCAL_SERVER' is not recognized as an internal or external command,
operable program or batch file.
error Command failed with exit code 1.
```


CompuIves:
Ahh this is really great @jseminck. Yesterday I was debugging this with someone on Windows, good to have support for Windows now.

Only thing left is running `npm run add`.

jseminck:
Added. Turned out a bit problematic as the CLI was giving errors for me. Created an issue with details on the all-contributors-cli project.

CompuIves:
Great! Thanks!

PR #135, opened 2017-08-20
Make action tooltips positions more consistent

jseminck:
Current behaviour:
current

With this PR:
after

Now all the tooltips are aligned in the same position and the blue bottom border is visible on hover on the actions. I believe this is the original idea behind the hover behaviour for these actions. If I assumed wrong, please let me know how you'd like it to be! 😄

Created an issue as well: https://github.com/CompuIves/codesandbox-client/issues/134


CompuIves:
Oh nice!! I remember fighting with this, seems like you found the solution 😄 .

I'll merge this in after this monster PR: https://github.com/CompuIves/codesandbox-client/pull/137

CompuIves:
Merging now, thanks for solving this!

PR #147, opened 2017-08-27
Introduce delete file modal

jseminck:
This PR adds a modal for confirming file deletion, instead of using `alert()`. If you're happy with it, we can also do something similar for deleting sandboxes.

I'm not particularly happy with the design. It's not something I'm very good at. I tried using elements that exist on other places already. Actually, the red colour for the Delete button doesn't feel right...

I'm not sure if you want this. Maybe you want to keep `alert()` but I assumed it was done this way initially because it was the easiest approach. 😄 

deletemodal

As usual, I learned quite a lot by looking at your Modal implementation (1 instance of `Modal` and keeping everything else in the state), and I'll probably use the same approach for another project where we recently started adding more modals. 

**PS:** I noticed my diff includes a lot of trailing comma removals. This was done by the pre-commit hook, is it expected?

### DIFF
--- a/src/app/pages/Sandbox/Editor/Workspace/SandboxActions/index.js
+++ b/src/app/pages/Sandbox/Editor/Workspace/SandboxActions/index.js
@@ -1,20 +1,16 @@
-import React from 'react';
-import styled from 'styled-components';
+import React from "react";
+import { connect } from "react-redux";
+import { bindActionCreators } from "redux";
+import styled from "styled-components";
 
-import Button from 'app/components/buttons/Button';
+import Button from "app/components/buttons/Button";
 
-import WorkspaceInputContainer from '../WorkspaceInputContainer';
+import modalActionCreators from "app/store/modal/actions";
+import Alert from "app/components/Alert";
 
-import WorkspaceSubtitle from '../WorkspaceSubtitle';
+import WorkspaceInputContainer from "../WorkspaceInputContainer";
 
-type Props = {
-  id: string,
-  deleteSandbox: (id: string) => void,
-  newSandboxUrl: () => void,
-  setSandboxPrivacy: (id: string, privacy: number) => void,
-  isPatron: boolean,
-  privacy: 0 | 1 | 2,
-};
+import WorkspaceSubtitle from "../WorkspaceSubtitle";
 
 const PrivacySelect = styled.select`
   background-color: rgba(0, 0, 0, 0.3);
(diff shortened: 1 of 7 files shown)


CompuIves:
Nice @jseminck, this is indeed something we desperately needed! I like this implementation, but think it would be even better if we have a general 'Alert' component that we can use on all places we use alert. A solution to the buttons is to make the right button blue (primary color) and the left one yellow (secondary color).

jseminck:
Great idea. I'll tackle it tonight, re-use it for delete sandbox and post new screenshots 👍

jseminck:
Sorry! I have been a bit busy. I looked for the secondary colour the other day but I couldn't find it. Did you mean the same yellow when you hover over one of the current primary buttons?

The secondary colour inside the theme object is also some shade of blue.

That yellow would look a bit odd as a secondary cancel button because it has the same colour as hovering over the primary one. So when you hover over primary, both buttons have the same colour.

lbogdan:
Maybe something like this? (from [here](https://ux.stackexchange.com/questions/49991/should-yes-delete-it-be-red-or-green/))

CompuIves:
Good points @jseminck, and that looks beautiful @lbogdan! Though I don't think we should go with a red primary button, since we a simpler UX/UI if we use the same type of modal on the whole website (without exceptions). A red primary button only makes sense for deletion modals.

Maybe we could do something like this (quick mockup):

screen shot 2017-09-02 at 13 05 13

lbogdan:
Oh, I thought we were only talking about *destructive* modals. :smile:

jseminck:
Updated: changed button styles, renamed component and moved it to /components/ folder, also used Alert for sandbox deletion. 

Maybe the hover for cancel still needs some adjustment 😄

src/app/pages/Sandbox/Editor/Workspace/SandboxActions/index.js
```diff
@@ -28,18 +24,43 @@ const PrivacySelect = styled.select`
...
+
+    modalActions.openModal({
+      Body: (
+        <Alert
+          title="Delete File"
+          body={<span>Are you sure you want to delete this sandbox?</span>}
+          onCancel={modalActions.closeModal}
+          onDelete={async () => {
+            await this.props.deleteSandbox(this.props.id);
+            await this.props.newSandboxUrl();
```
jseminck: Do these function calls really need to `await`? Right now the modal wouldn't be closed until `deleteSandbox()` and `newSandboxUrl()` are completed. - not sure if that is intended...

CompuIves:
Super great work! Nice!

src/app/pages/Sandbox/Editor/Workspace/SandboxActions/index.js
CompuIves: Hmm, yes, seems like we can remove the `await` statementes.

CompuIves:
Nice, works perfectly! Merging it in now...

PR #149, opened 2017-08-28
Dockerfile: MAINTAINER is deprecrated

robbyoconnor:


CompuIves:
Another thank you! Could you add yourself as collaborator to CodeSandbox? You can do that by running:

```
npm run add
```

robbyoconnor:
Done, I'm not sure what I can help out with but I do ops stuff so infra questions are always welcome.

CompuIves:
Nice, thanks! Merging in now.

PR #150, opened 2017-08-28
Dockerfile: use COPY vs ADD; use of MAINTAINER is deprecated

robbyoconnor:


CompuIves:
Thanks @robbyoconnor! This is appreciated.

PR #154, opened 2017-08-31
Fix plugin order to re-enable using decorators

mweststrate:
For example https://codesandbox.io/s/2vmzpM0wK is now broken. See also: https://github.com/mobxjs/mobx/issues/1154

Fixes https://github.com/CompuIves/codesandbox-client/issues/151


CompuIves:
Great! Thanks @mweststrate, merging in now.

PR #163, opened 2017-09-02
ShareView: make inputs and textareas explicit readonly.

lbogdan:
Fixes https://github.com/CompuIves/codesandbox-client/issues/162.


CompuIves:
Wow you're quick, thanks again!

PR #169, opened 2017-09-03
Fix all eslint warnings and errors.

lbogdan:


src/app/components/Preference/PreferenceNumber.js
```diff
@@ -22,10 +23,9 @@ export default class PreferenceInput extends React.PureComponent {
   };
 
   render() {
-    const { value, style, step } = this.props;
+    const { value, style } = this.props;
     return (
       <StyledInput
-        step={step}
```
CompuIves: Why was this prop removed?

src/app/components/sandbox/CodeEditor/Monaco.js
```diff
@@ -46,14 +48,6 @@ const Container = styled.div`
   z-index: 30;
 `;
 
-const fontFamilies = (...families) =>
```
CompuIves: Let's comment this part of the code out instead, I'm hoping to enable it next week again (it's not used because of a bug).

src/app/components/sandbox/CodeEditor/index.js
```diff
@@ -3,12 +3,19 @@ import Loadable from 'react-loadable';
 
 import Monaco from './Monaco';
 
+type Props = {
+  preferences: {
```
CompuIves: You can use the type `Preferences` for this, it resides in `/common/types`.

src/app/components/sandbox/Preview/utils/should-update.js
```diff
@@ -24,7 +24,7 @@ function compareArrays(m1, m2, fields: Array<String>) {
       return true;
     }
 
-    return checkFields(first, second, MODULE_FIELDS);
+    return checkFields(first, second, fields);
```
CompuIves: Oh this is a good one! So all this time I've been passing in fields that don't get checked?

src/app/pages/Sandbox/Editor/Workspace/EntryContainer.js
```diff
@@ -6,7 +6,7 @@ export const getContainerStyles = props => {
     ? theme.primary
     : theme.templateColor || theme.secondary;
   let styles = `
-    ${props => props.noTransition || 'transition: 0.3s ease all;'}
+    ${props.noTransition || 'transition: 0.3s ease all;'}
```
CompuIves: I think we should keep this the same (maybe rename `props`), because the passed in props are different from the component props in the case of styled-components.

src/app/pages/Sandbox/Editor/Workspace/Files/DirectoryEntry/Entry/EntryIcons.js
```diff
@@ -69,23 +69,29 @@ type Props = {
   root: ?boolean,
   error: boolean,
 };
-export default ({
+export default function EntryIcon({
```
CompuIves: I think it would be cleaner if we'd use the existing way, but do `isOpen = false` in the prop arguments.

src/app/store/api/actions.js
```diff
@@ -51,7 +51,6 @@ const getMessage = (error: Error & { response: ?Object }) => {
 const showError = error => dispatch => {
   const errorMessage = getMessage(error);
   dispatch(notificationActions.addNotification(errorMessage, 'error'));
-  error.apiMessage = errorMessage;
```
CompuIves: We need to keep this one for error reporting later on.

src/app/store/entities/sandboxes/utils/create-zip/vue-cli/index.js
```diff
@@ -63,20 +63,22 @@ export default function createZip(
 ) {
   zip.file('README.md', README);
 
-  return zip.loadAsync(files).then(zip => {
```
CompuIves: Is there a reason you renamed this from `zip` to `src`?

lbogdan: `EsLint: 'zip' is already declared in the upper scope. (no-shadow)`

CompuIves: Ah, right 👍

src/app/components/Preference/PreferenceNumber.js
lbogdan: Mostly because I forgot there's an `<input>` `step` property, and it wasn't used anywhere.

src/app/pages/Sandbox/Editor/Workspace/EntryContainer.js
lbogdan: I didn't know that. There are a few more, I'll refactor them. `styleProps` is ok?

CompuIves: Hmm, yes, sounds good!

src/app/components/Preference/PreferenceNumber.js
CompuIves: I didn't know either when this was added, haha 😅 . I think we need to keep this one, it's important for the `line-height` preference.

src/app/pages/Sandbox/Editor/Workspace/Files/DirectoryEntry/Entry/EntryIcons.js
lbogdan: I've tried with
```js
export default function EntryIcon({
  ...
  isOpen = true,
  ...
: Props) { ...
```
, but I still get `EsLint: propType "isOpen" is not required, but has no corresponding defaultProp declaration. (react/require-default-props)`.

src/app/store/api/actions.js
lbogdan: Restored.

src/app/components/sandbox/CodeEditor/Monaco.js
lbogdan: Restored and commented.

src/app/components/sandbox/CodeEditor/index.js
lbogdan: Nice, fixed.

src/app/components/Preference/PreferenceNumber.js
lbogdan: Right now it uses the default value - 1. I restored anyway. :smile:

src/app/pages/Sandbox/Editor/Workspace/Files/DirectoryEntry/Entry/EntryIcons.js
CompuIves: Ah, just read this from [here](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-default-props.md): 

> The same also holds true for stateless functional components: default function parameters do not behave the same as defaultProps and thus using defaultProps is still preferred.

Let's use `defaultProps` then.

lbogdan:
So is there anything left for me to do? What about the failing tests?

CompuIves:
This looks good to me! I'll test the code later today (in about 4 hours), after that we can merge. Regarding the tests, it seems like the only failing tests are style related, so let's update the snapshots.

CompuIves:
Looks good!! Could you fix the merge conflicts and update the snapshots? Then I'll merge this in.

PR #172, opened 2017-09-03
Fix eslint errors for src/embed.

lbogdan:


CompuIves:
Nice! Luckily this one wasn't that big 😄 .

PR #175, opened 2017-09-04
Preserve resource order

lbogdan:
The order in which resources (especially `<script>`s) are loaded is important in cases where there are dependencies between them (most notable case is jQuery). That's the reason for which sorting the resources by name in the Dependencies > External Resources section in the left pane can be misleading.
Also the reason for which I removed `sort()` from `getExternalResourcesConcatenation()` (`['a.js', 'b.js']` should not be equivalent to `['b.js', 'a.js']`).

There was another issue with the way `async` was set on resources `<script>`s, as a HTML attribute instead of a DOM element attribute (see https://developer.mozilla.org/en/docs/Web/API/HTMLScriptElement#async_property).

After these changes, adding jQuery and a plugin that depends on it works, regardless of the time they load in (I was having issues with [MaterializeCSS](http://materializecss.com/) loading faster than jQuery, executing first and failing not finding jQuery).


CompuIves:
This is reaaaaally great @lbogdan, thank you so much!

PR #182, opened 2017-09-05
Fix eslint issues in the sandbox editor.

lbogdan:
Fixes https://github.com/CompuIves/codesandbox-client/issues/142.

### DIFF
--- a/src/app/components/sandbox/CodeEditor/CodeMirror.js
+++ b/src/app/components/sandbox/CodeEditor/CodeMirror.js
@@ -410,8 +410,9 @@ export default class CodeEditor extends React.PureComponent<Props, State> {
     }
 
     if (preferences.lintEnabled) {
+      const initialized = 'eslint' in window;
       System.import('app/utils/codemirror/eslint-lint')
-        .then(initializer => initializer.default())
+        .then(initializer => !initialized && initializer.default())
         .then(() => {
           this.codemirror.setOption('lint', true);
         });
--- a/src/app/utils/codemirror/eslint-lint.js
+++ b/src/app/utils/codemirror/eslint-lint.js
@@ -307,6 +307,7 @@ const defaultConfig = {
 };
 
 let linter = null;
+let eslintRequire = null;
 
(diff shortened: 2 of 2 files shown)


src/app/components/sandbox/CodeEditor/CodeMirror.js
```diff
@@ -410,8 +410,9 @@ export default class CodeEditor extends React.PureComponent<Props, State> {
     }
 
     if (preferences.lintEnabled) {
+      const initialized = 'eslint' in window;
       System.import('app/utils/codemirror/eslint-lint')
-        .then(initializer => initializer.default())
+        .then(initializer => !initialized && initializer.default())
```
lbogdan: `eslint-lint.js`.`initialize()` (the default export) should execute only when importing eslint the first time. This gets triggered whenever switching between CodeMirror and Monaco.

src/app/utils/codemirror/eslint-lint.js
```diff
@@ -377,6 +382,7 @@ export default (async function initialize() {
     await delay(100);
   }
 
+  eslintRequire = window.require;
```
lbogdan: Save eslint's require.

src/app/utils/codemirror/eslint-lint.js
```diff
@@ -336,7 +337,11 @@ function getSeverity(error) {
 
 function eslintValidate(text) {
   if (!linter) return [];
-  return linter.verify(text, defaultConfig);
+  const saveRequire = window.require;
+  window.require = eslintRequire;
+  const result = linter.verify(text, defaultConfig);
+  window.require = saveRequire;
+  return result;
```
lbogdan: When switching from Monaco to CodeEditor, `window.require` refers to Monaco's loader, clashing with `linter.verify()` which calls `require()` for some reason. So we save eslint's require (see below) and use it every time we call `linter.verify()`.

CompuIves:
GREAT! Thanks!! 👍

PR #186, opened 2017-09-05
Fix issue when clicking prettify when the sandbox contains syntax errors

jseminck:
Check that newCode is actually defined before switching the output. I hope the issue description and fix speak for itself. 😄 

https://github.com/CompuIves/codesandbox-client/issues/185

**Edit:** I don't understand why my commits contain stylistic changes to other parts of the code in the same file. I suppose it's some issue from my side with Prettier?


CompuIves:
Nice @jseminck, this is a really important issue. I think you should update prettier to 1.6 to fix the styling issue. I'll merge this in asap, though I can deploy tomorrow morning. Tomorrow I'll have more time in general 😅

jseminck:
Re-installed dependencies which updated prettier. Should be good now 👍

PR #189, opened 2017-09-06
Sandbox: change address bar url on hash change

lbogdan:
Fixes https://github.com/CompuIves/codesandbox-client/issues/115.

### DIFF
--- a/src/sandbox/url-listeners.js
+++ b/src/sandbox/url-listeners.js
@@ -9,32 +9,32 @@ function sendUrlChange(url: string) {
 
 export default function setupHistoryListeners() {
   const pushState = window.history.pushState;
-  window.history.pushState = function(state) {
+  window.history.pushState = function(state, ...args) {
     if (typeof history.onpushstate === 'function') {
       window.history.onpushstate({ state });
     }
     // ... whatever else you want to do
     // maybe call onhashchange e.handler
-    return pushState.apply(window.history, arguments);
+    return pushState.apply(window.history, [state, ...args]);
   };
 
   const replaceState = window.history.replaceState;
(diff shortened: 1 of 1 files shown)


lbogdan:
Hm, seems like there is a bigger issue with the sandbox history, when in iframe. I'll update tomorrow.

lbogdan:
OK, so.

It seems that when a page contains an iframe and you're changing the url or doing `pushState()` inside that iframe, the url change is stored in the main page's history - you can see that by clicking browser's back / forward buttons, it will navigate through the iframe's history. That's also causing a bug, because when using browser's navigation buttons, the url in the sandbox's address bar won't update (steps to reproduce: load https://codesandbox.io/embed/zq380r0wqx , go to page 1, go to page 2, press browser's back button, it will navigate back to page 1, but the url won't change).

You can read more in [this article](http://www.webdeveasy.com/back-button-behavior-on-a-page-with-an-iframe/).

The suggested solution is to use `replaceState()` instead of `pushState()` in the iframe, and manage the iframe's history by ourselves. We will probably have to detect if we're in an iframe (I see there's an exported `isStandalone` from `codesandbox-api`), and replace the whole `window.history` object with our version, that will only do `replaceState()`.

Opinions?

CompuIves:
You're right, a big limitation for the simulated browser is that it has to share history with its parent. I've been struggling for some time with it, it causes a lot of limitations.

I'm not sure if pushState can be replaced with replaceState, since this will remove the native back button functionality for the preview (or am I wrong?). I hope I'm wrong, because it will lift a lot of limitations for us.

lbogdan:
Replaced with #198.

PR #190, opened 2017-09-08
Trim value for Dependency's input field

bogas04:
Was copy-pasta-ying when I encountered this issue.

Adding `preact ` would give an error `Could not find the dependency 'preact '`, expected, but could be fixed by this change.


CompuIves:
Ah, good one @bogas04! Thanks!

Could you add yourself to the contributors by running:

```
npm run add
```

bogas04:
Done @CompuIves ! Was bit busy last week, sorry for the delay.

PR #196, opened 2017-09-10
Sandbox preview: make navigator's back button inactive on first load

lbogdan:
`historyPosition` starts at 1 (because of the initial `sendUrlChange()` -> `commitUrl()` with the sandbox URL), so the condition `historyPosition > 0` is always true.

Fixes https://github.com/CompuIves/codesandbox-client/issues/195.


CompuIves:
Niice! Will merge this in in about 1-2 hours

PR #197, opened 2017-09-12
Svelte support

CompuIves:


Rich-Harris:
This is amazing, thanks for doing this. I have a couple of thoughts, all things that could easily be addressed at a later stage if it's necessary to address them at all:

* What's the easiest way to keep the compiler version up to date? Is it practical to use `@1` instead of `@1.39.0`, or are we better off updating it manually as new versions are released?
* The compiler emits warnings for things that it's not sure about (e.g. [missing methods, unused styles — check the console](https://svelte.technology/repl?version=1.39.1&gist=f7fcb207f58527e4fac667c91c4fdb22)). Is there any way to present those to the user?
* The compiler has a `dev` mode (which the Svelte REPL doesn't currently use, but will soon) which emits more warnings (and errors in some cases), and does things like adding sourcemaps to `<style>` blocks so that it's easy to locate the source of a style you're tweaking with devtools. Should dev mode be enabled here, or not?
* Speaking of styles, the object returned from `svelte.compile` contains `css` and `cssMap` properties. In production it's recommended to use those to create `.css` files (and pass `css: false` to the compiler so that the component doesn't add the styles with JS). That's obviously a lot more fiddly when you're doing it by hand in a situation like this, so the easiest thing by far is to leave it as-is — just thought I'd raise it in case you have strong opinions about that sort of thing.

Let me know if there's anything I can do to help.

CompuIves:
Thanks!! Great feedback 😄 

> What's the easiest way to keep the compiler version up to date? Is it practical to use @1 instead of @1.39.0, or are we better off updating it manually as new versions are released?

Yep, I think the @1 would be the best solution. I'll change it.

> The compiler emits warnings for things that it's not sure about (e.g. missing methods, unused styles — check the console). Is there any way to present those to the user?

Very good one, we generally need a better way to show warnings. I'll work on that now.

> The compiler has a dev mode (which the Svelte REPL doesn't currently use, but will soon) which emits more warnings (and errors in some cases), and does things like adding sourcemaps to <style> blocks so that it's easy to locate the source of a style you're tweaking with devtools. Should dev mode be enabled here, or not?

Really good one! I'll enable dev mode. Is it possible to add inline sourcemaps to the scripts by the way? This would make it easier to make create-react-app error overlay work with Svelte.

> Speaking of styles, the object returned from svelte.compile contains css and cssMap properties. In production it's recommended to use those to create .css files (and pass css: false to the compiler so that the component doesn't add the styles with JS). That's obviously a lot more fiddly when you're doing it by hand in a situation like this, so the easiest thing by far is to leave it as-is — just thought I'd raise it in case you have strong opinions about that sort of thing.

Hmm, good one. For the Vue templates we emit css files manually. I'll do some testing with HMR, maybe I'll move it over to manual emitting. This will make sure that we clean up CSS styles if a file gets unloaded.

Rich-Harris:
> Is it possible to add inline sourcemaps to the scripts by the way?

Yes, absolutely — the object returned from `svelte.compile` looks like this:

```js
const { code, map, ast, css, cssMap } = svelte.compile(...);
```

(Note that we plan to change this in version 2 — it will return a promise, so that we can accommodate async preprocessors, that will resolve to a `{js: {code, map}, css: {code, map}, ast}` object.)

`map` is a v3 sourcemap with a couple of convenience methods — `toString()` (which basically just returns JSON that you'd write to a `.js.map` file), and `toUrl()` (which generates a data URI you can use as an inline sourcemap).

Here's an example of appending a sourcemap to the generated code: https://runkit.com/rich/59b821b2faacb50011ab56c9

CompuIves:
I added the warnings to the code and the console, turned on dev mode and added sourcemaps. I'm planning to add an integrated console in the future that will show the warnings more clear though. 

It looks already very cool with sourcemaps and warnings:

screen shot 2017-09-12 at 22 10 31

Rich-Harris:
that's badass!

PR #198, opened 2017-09-13
Sandbox preview: replace window.history with our own emulated history

lbogdan:
…, when inside the editor.

To do:
- [x] explain the approach
- [x] remove `console.log()`s, maybe some refactoring
- [x] (a lot) more testing
- ~~debug Firefox issue where `window.history.__proto__` is reset when bluring and focusing the window (tab?) again~~

I tested in latest versions of Chrome and Firefox, with both a React app and a Vue app using their corresponding routing libraries, in both `history` and `hash` mode.

Fixes https://github.com/CompuIves/codesandbox-client/issues/115 (and probably other issues related to history navigation behavior).

### DIFF
--- a/src/app/components/sandbox/Preview/index.js
+++ b/src/app/components/sandbox/Preview/index.js
@@ -65,7 +65,7 @@ export default class Preview extends React.PureComponent<Props, State> {
     this.state = {
       frameInitialized: false,
       history: [],
-      historyPosition: 0,
+      historyPosition: -1,
       urlInAddressBar: frameUrl(props.sandboxId, props.initialPath || ''),
       url: null,
     };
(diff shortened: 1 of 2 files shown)


CompuIves:
Very interesting! This would indeed solve a lot of problems, we have to really make sure that it works accordingly. If we have that this would solve *a lot* of problems.

lbogdan:
Ok, what happens here:

As stated [here](https://github.com/CompuIves/codesandbox-client/pull/189#issuecomment-327727829) the history in an iframe is intertwined with the main window's history, so that using the back/forward browser buttons will act on the iframe history, too. The solution would be to change `window.history` inside the iframe to use `history.replaceState()`, which will not have any effect on the main window history, while managing the history inside the iframe ourselves.
- **In /history mode**

This is quite straightforward, we change all history navigation methods - `go()`, `back()`, `forward()`, `pushState()` and `replaceState()` - to use `replaceState()`. Also, `go/back/forward()` will have to emit a `popstate` event.

- **In #hash mode**

This is a bit trickier, as there's no standard way of intercepting url changes when pressing on `<a>` elements with a hash `href`. So we add a `click` event handler on `document`, check if the pressed element is an `<a>` with a hash `href`, and simulate a hash url change by doing `replaceState()` and emitting a `hashchange` event.

CompuIves:
Okay! Testing has engaged 🤓

lbogdan:
Should be ready to merge.

PR #210, opened 2017-09-27
Added react-jss to the landing

kof:


kof:
I took the freedom to add it to the list and to the picked projects. I don't know if react-jss package qualifies for the later, let me know if you don't want to have it there)

PR #215, opened 2017-09-30
Change babel preset to env preset.

jsonnull:


CompuIves:
Great! Thanks! One question: shouldn't we add `babel-preset-env` to the devDependencies in package.json? Not sure, maybe it works differently.

Could you also add yourself to the contributors by running:

```
npm run add
```
?

jsonnull:
Good call, I also added the preset to the devDependencies section.

The overlap shouldn't hurt—I much prefer to keep dependencies up-to-date with runtime deps and devDependencies up-to-date with build-time deps, so this is good.

Also, contributors updated, thank you! The script did have a bad exit status but the update went through—is this something that needs to be looked into?

CompuIves:
Great, thanks! I think the author (Kent) is aware of the exit code of contributors, maybe I need to update dependencies 😅 .

CompuIves:
Oh hmm, it seems like the commit of adding to `devDependencies` is empty (https://github.com/CompuIves/codesandbox-client/pull/215/commits/7fa8249cdfd27db79e40bec33aa01f9e81c1014b).

jsonnull:
Strange!

I added another commit which includes the preset in devDependencies.

CompuIves:
Okay! Thanks for the PR!

PR #216, opened 2017-10-03
Show modal instead of alert

jseminck:
Add a modal (the same that is used in the editor when deleting a sandbox) when deleting sandbox from profile page, instead of alert.

Locally could only test it by forcing `isCurrentUser` to `true` in the code. But otherwise I could not manage to log in to Github when running locally. Though it seems to be working, but please do double-check 😄

### DIFF
--- a/src/app/pages/Sandbox/Editor/Workspace/SandboxActions/index.js
+++ b/src/app/pages/Sandbox/Editor/Workspace/SandboxActions/index.js
@@ -50,7 +50,7 @@ class SandboxSettings extends React.PureComponent {
     modalActions.openModal({
       Body: (
         <Alert
-          title="Delete File"
+          title="Delete Sandbox"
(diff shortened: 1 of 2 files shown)


src/app/pages/Sandbox/Editor/Workspace/SandboxActions/index.js
```diff
@@ -50,7 +50,7 @@ class SandboxSettings extends React.PureComponent {
     modalActions.openModal({
       Body: (
         <Alert
-          title="Delete File"
+          title="Delete Sandbox"
```
CompuIves: Good find!

CompuIves:
Tested it, works good! One thing that didn't work as expected is that the modal doesn't close after deletion. Think this can be fixed by adding `modalActions.closeModal();`.

jseminck:
Oops, will add it.

CompuIves:
Cool!

PR #229, opened 2017-10-10
Upgrade eslint-plugin-react to 7.4.0...

lbogdan:
... as they released a new version since using the master branch in order to fix all the eslint errors. Also fixed a few small issues in the process.


CompuIves:
Cooool, thanks!

PR #239, opened 2017-10-12
More tooltips

brumm:
Hey,

I went ahead and added a couple of tooltips for some icon-only buttons that confused me the first time I used codesandbox.

I didn't investigate https://github.com/CompuIves/codesandbox-client/commit/2b9209fd6f00238dc4ebf2745fa51d017f08dbdf for too long, but it fixed the scrollbar issue for me.


CompuIves:
Niiice! Will test it once I get home.

src/app/pages/Sandbox/Editor/Content/Header/index.js
```diff
@@ -297,6 +314,7 @@ export default class Header extends React.PureComponent<Props> {
             ) : (
               <Action
                 onClick={userActions.signIn}
+                tooltip="Sign in with Github"
```
Tushkiz: I think a tooltip here will be redundant as the button has text 🤔

src/app/components/sandbox/ModeIcons.js
```diff
@@ -131,22 +131,28 @@ const getCurrentMode = ({
   setPreviewView,
 }: Props) => {
   const both = (
-    <ViewIcon onClick={setMixedView} active={showEditor && showPreview}>
-      <EditorIcon half />
-      <PreviewIcon half />
-    </ViewIcon>
+    <Tooltip title="Show both">
```
Tushkiz: Maybe you can make these ModeIcon tooltip titles consistent with those ones [below](https://github.com/CompuIves/codesandbox-client/pull/239/files#diff-1f317ba5bd25c442a32513566177d0c3R264). Currently we have them as `Editor view`, `Split view` and `Preview view`

src/app/pages/Sandbox/Editor/Content/Header/index.js
brumm: Good point, will remove.

src/app/components/sandbox/ModeIcons.js
brumm: Ah, I didn't notice those.

I'd like to suggest moving from _this is the name of the thing you're clicking on_ to a more verb-oriented  _this is what will happen when you click_ style of descriptions. I'd like to change the existing Tooltips to reflect my changes, if possible.

Tushkiz: Let's keep the existing unchanged and change ours to `Show Editor view`, `Show Split view` and `Show Preview view`. What do you say?

brumm: I like it! Will do.

CompuIves:
**Great**! Thanks @brumm!

PR #255, opened 2017-10-23
Refactor Search box - Remove the use of state as the query keyword is required only after user hits enter - Move the component to simple function component

Anenth:


CompuIves:
Nice, looks much better! I found one small issue while testing, it seems like there is a full refresh when you insert a search in the header on `/search` and click on the magnifying glass. You know what could cause this?

Anenth:
@CompuIves oh ya, I know the issue

Anenth:
@CompuIves Fixed!

CompuIves:
Perfect! Thanks @Anenth, these kind of changes are very important and really welcome 😄

PR #259, opened 2017-10-27
Run test before git push

Anenth:
Run `yarn test` on every push

After pushing the code, we have to remember to run `yarn test`. 

**What is the new behavior?**
Before git push, `yarn test` will run.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/package.json
+++ b/package.json
@@ -182,7 +182,8 @@
     "lint:embed": "eslint src/embed",
     "add-contributor": "all-contributors add",
     "generate": "all-contributors generate",
-    "precommit": "lint-staged"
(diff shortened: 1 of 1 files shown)


CompuIves:
Hmm, I think it's good to make sure all checks pass, but doesn't CI already solve this for us? I'm mainly afraid that this will slow my workflow down since pushes will take ~30 seconds then.

Anenth:
@CompuIves There is a checklist on PR Template ` [ ] Tests ` So everytime after pushing the I run yarn test.

PR #265, opened 2017-10-29
Move to workspaces

CompuIves:
We are getting more and more shared modules. With the new homepage we'll have another sub-package.json. It seems like a logical choice to move to a workspace approach.


src/homepage/src/screens/home/NPMFeature/SearchInput/index.js
```diff
@@ -0,0 +1,71 @@
...
+} from 'app/utils/config';
+
+import Input from './Input';
+import Hit from './Hit';
+
+import algoliaImage from './algolia.svg';
+import { searchFacets } from '../../../../utils/algolia';
+import media from '../../../../utils/media';
+
+const Legenda = styled.div`
```
Tushkiz: Dutch! 😬

CompuIves: Hah, whoops! I let some Dutch slip in 😅 . Fixing.

PR #267, opened 2017-10-31
DevTools / console enhancements

lbogdan:
See commit messages.


src/sandbox/console/index.js
```diff
@@ -1,16 +1,18 @@
 import { dispatch } from 'codesandbox-api';
 
-import massageJSON from './massage-json';
+import messageJSON from './message-json';
```
CompuIves: Haha, this wasn't a typo, but a bit cryptic. Maybe rename it to transform-json?

src/app/components/sandbox/Preview/DevTools/index.js
```diff
@@ -230,7 +232,11 @@ export default class DevTools extends React.PureComponent<Props, State> {
         // happen when pointer events are disabled and in turn disables scroll.
         // It's hacky, but it's to fix a bug in the browser.
         setTimeout(() => {
-          this.setState({ height: this.state.height + 1 });
+          const height = this.state.height + 1;
+          this.setState({ height });
+          if (height > 64) {
+            store.set('devtools.height', height);
```
CompuIves: Is this to inspect the height of the devtools?

src/app/utils/debug.js
```diff
@@ -15,6 +17,12 @@ const getDebugger = () => {
     };
   }
 
+  if (store.get('debug.disabled')) {
```
CompuIves: Hmm, maybe we can solve this differently. Debug has the functionality to check based on env var DEBUG=<prefix> what messages to show. Maybe we can put this in localStorage and derive it from that.

src/app/components/sandbox/Preview/DevTools/index.js
CompuIves: Nevermind, read the commit. Reaaaaally cool!

src/sandbox/console/index.js
lbogdan: Haha, **now** I get it!

src/app/utils/debug.js
lbogdan: So is it possible to disable debug messages by running `DEBUG=*something* yarn start`? And if yes, what should `*something*` be?

src/sandbox/console/index.js
lbogdan: Renamed it to `transform-json.js` / `transformJSON`.

src/app/utils/debug.js
lbogdan: (removed it for now)

CompuIves: We do by default DEBUG=*, but if you omit DEBUG you see no messages. You can for example do DEBUG=cs:compiler:* to see all compiler debug messages.

src/sandbox/console/index.js
CompuIves: Great! Yeah massage sounds a bit strange. Heard it last week at the conference haha

src/app/utils/debug.js
lbogdan: That's strange, because I can see all `cs:` messages by default. Maybe because I'm on Windows? :thinking:

CompuIves: I enable it by default 😅  (https://github.com/CompuIves/codesandbox-client/blob/master/src/app/utils/debug.js#L19-L20). We can derive this from the localStorage maybe.

lbogdan: I commented out that line, and it still showed all messages. That's why I conditionally (based on localStorage) did `debug.disable()` in the first place.

CompuIves: Hmm, okay that's strange. Then their logic changed to enable by default.

CompuIves:
Ahhh I created a merge conflict. Shouldn't be too hard to fix.

src/app/utils/debug.js
lbogdan: I'll just disable it manually for now. (and be careful to not commit it! :laughing: )

PR #272, opened 2017-11-01
Fix missing lineheight setting key and enable custom font

donysukardi:
Bug fix

- https://github.com/CompuIves/codesandbox-client/issues/271  Line Height setting is getting persisted to localStorage as undefined instead due to missing key
- https://github.com/CompuIves/codesandbox-client/issues/126 Monaco gets updated with prev preferences instead

**What is the new behavior?**
- Line Height setting gets persisted correctly under `settings.lineheight` key.
- Changes in settings are reflected correctly now and custom font for Monaco is re-enabled with ligatures support.

**Checklist**:

- [x] Ready to be merged 


donysukardi:
I'm not sure if we should unset/remove the `undefined` entry that has been previously created from users' localStorage.

CompuIves:
This is great @donysukardi! There is [this](https://github.com/CompuIves/codesandbox-client/issues/126) current issue with the line height, for windows users it makes the editor unusable. Could you maybe add the functionality to hide the preference for windows users?

Hmm, good one about the `undefined`. I the editor will see it as an unset value, so we don't have to do anything about it.

donysukardi:
Yeah. I'm just nitpicking about the `undefined` :grin:

image

CompuIves:
Oooh I see. Well, yeah, that's something we can miss haha. I'm still a bit worried about the editor, I remember that it was a big issue and that people weren't able to do anything. I want to be sure that it can't happen again.

CompuIves:
Ah, I found the issue! https://github.com/Microsoft/monaco-editor/issues/392

donysukardi:
@CompuIves Yay! Finally found out the problem and pushed the fix to this branch!

The issue isn't with the custom font as described in https://github.com/Micrsoft/monaco-editor/issues/392, because we're not actually loading custom font off the network. So the delay is not applicable to Codesandbox. 

Here's where the real problem lies. `preferences` from `this.props` is used to update the editor config in shouldComponentUpdate. At this point the new props haven't been merged into `this.props` yet.

So I changed the signature of the function to accept props instead, and we're passing in nextProps for the update.

P.S. I actually wouldn't do this in sCU lifecycle, prob better in componentWillReceiveProps or componentDidUpdate, more semantically correct.

CompuIves:
Ahhhh, thank you very much for the investigation! I can test it this Monday, then I'll merge it in 😄

CompuIves:
Hey @donysukardi! Could you quickly do another merge with master? Thanks to @lbogdan we have auto deploy for new PRs, so I can test it on my windows pc.

lbogdan:
@donysukardi You'll have to make sure you rebase your branch off of master, so that you have the "deploy to staging" changes.

donysukardi:
@CompuIves done.. Awesome work @lbogdan!

CompuIves:
Sorry for the late reply, I wanted to test it on Windows but completely forgot about the PR. I now tested it and it looks great, I merged it with master and will now put it in master. Thanks a lot for this PR!

PR #284, opened 2017-11-07
Binary support

CompuIves:


CompuIves:
CircleCI error for react-ts is expected, will be fixed with a server API change.

PR #290, opened 2017-11-09
Console enhancements

### LINKED ISSUE #286: Make "Clear Console" preference on by default
We have a preference to clear the console between executions, this is off by default, but I think it'd be better if we turn this on by default. This will probably clear up some confusion for beginners.

lbogdan:
- enable "Clear console" by default;
- also clear devtools console when "Clear console" is enabled;
- make inspector's theme background transparent, fixes warnings and errors showing like this in the console:

Fixes #286.


CompuIves:
Great! Thanks :D

PR #297, opened 2017-11-16
Homepage: visual switch for web applications section on mobile

duivvv:

homepage change

changed position of Tailored for web applications intro as discussed here https://twitter.com/CompuIves/status/931253159871295488


duivvv:
Fixed it this way, do you have a suggestion so we don't need to change text twice.
Should I create a new component (maybe <Intro/>) in /Frameworks so we can change the text once?

Let me know and I'll do the changes + the all contributors step.

CompuIves:
Super! Hmm yes, I can think of 2 approaches we can do here:

1. Wrap the two elements in a `display: flex` block. We can then do `flex-direction: column-reverse` for mobile and `flex-direction: column;` for higher. This feels like the best solution, but I'm not sure if it's possible with the current layout.
2. Extract the selection to a component and use the `Media` like you do here.

Would be great to have option 1, but only if it makes sense with the current layout. Otherwise 2 is the best option, as it's very clear.

Thanks a lot for the contribution! Great to have a new contributor for CodeSandbox 😄

duivvv:
I also agree 1 would be the best option, taking a look how much impact it has.

duivvv:
@CompuIves Think I nailed it.  🎉

CompuIves:
That's great!! I can merge it in tomorrow, when I'm back at my MacBook. Thanks a lot!

duivvv:
Sorry about the multiple contributor commits.

I thought it didn't work, but apparently it did :)

CompuIves:
This is great, @duivvv. Thanks! Merging it in now, welcome to the contributors 😄 .

duivvv:
No problem, glad I could help

PR #298, opened 2017-11-17
Improve formatting of string arguments in integrated console

viankakrisna:
enhancement

with 
```
console.log('BEFORE', JSON.stringify(variable, null, 2));
console.log('AFTER', JSON.stringify(variable, null, 2));
```
codesandbox io_s_x7ll8n372p

**What is the new behavior?**
codesandbox io_s_x7ll8n372p 2
codesandbox io_s_x7ll8n372p 1

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged 

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -67,7 +67,7 @@
       "login": "viankakrisna",
       "name": "Ade Viankakrisna Fadlil",
       "avatar_url": "https://avatars1.githubusercontent.com/u/9636410?v=4",
-      "profile": "https://musify.id",
+      "profile": "https://github.com/viankakrisna",
       "contributions": [
         "code"
       ]
--- a/packages/app/src/app/components/sandbox/Preview/DevTools/Console/Message.js
+++ b/packages/app/src/app/components/sandbox/Preview/DevTools/Console/Message.js
@@ -133,6 +133,10 @@ function getMessage(message: IMessage) {
     );
   }
 
+  if (message.arguments.every(argument => typeof argument === 'string')) {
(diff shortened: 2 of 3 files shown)


lbogdan:
You can simply do `console.log('some text', someVariable)` to get a much nicer display of `someVariable`.

viankakrisna:
@lbogdan yea, but it's a nested object and i'm too lazy to click on the arrows :D

CompuIves:
Hmm, I noticed that we need to change the logic altogether. I just checked in Chrome console and it seems that if all vars are strings, then we need to concatenate them:

screen shot 2017-11-18 at 16 15 28

Maybe we can do that combined with this change?

viankakrisna:
i see, will see if i can implement that 👍

viankakrisna:
pushed incorrect implementation because i can't test it locally (bad connection, need to upgrade yarn for workspace and install dependencies) 😅

viankakrisna:
weird, can't run it locally even after upgrading yarn and installing dependencies. I'll see this at a later time.

CompuIves:
Yep, I got the same thing. I'm fixing it now, should be fixed in 1.5 hours.

CompuIves:
Fixed @viankakrisna! 😄

viankakrisna:
@CompuIves updated!

with this:
```
const styles = {
  fontFamily: 'sans-serif',
  textAlign: 'center',
};
console.log('APP:', JSON.stringify(styles, null, 2))
```
comes this:
screenshot from 2017-11-20 11-41-26

CompuIves:
Great!! Merci 😄

PR #310, opened 2017-11-19
Simplified Exit logic in deploy.sh

nagamalli9999:
'set -e' will force bash to exit if there is a non zero exit code

This PR makes the deploy.sh more readable and easier to maintain

There is no change in expected behavior

**Checklist**:
- [ ] Documentation "N/A"
- [ ] Tests "N/A"
- [x] Ready to be merged 

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,14 +1,15 @@
 #!/bin/bash
+set -e
 
-rm -rf .deliver && \
-mkdir .deliver && \
-cp -R www .deliver && \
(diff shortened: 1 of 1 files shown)


CompuIves:
Oh wow, today I learned! Thanks for the contribution, I just noticed that we didn't add the script for adding contributors while refactoring the structure. I just fixed that. Could you add yourself to the contributors by running `npm run add-contributor`? 😄

PR #311, opened 2017-11-20
Simplified Exit logic in deploy.sh

nagamalli9999:
'set -e' will force bash to exit if there is a non zero exit code

This PR makes the deploy.sh more readable and easier to maintain

**What is the new behavior?**
There is no change in expected behavior

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 


deploy.sh
```diff
@@ -1,14 +1,15 @@
...
-echo "Untarring www files" && \
+rm -rf .deliver
+mkdir .deliver
+cp -R www .deliver
+echo "Tarring build files"
+tar -C .deliver -zcvf .deliver/code_sandbox_client.tar.gz www
+echo "Cleaning remote tar"
+ssh nginx@ssh.codesandbox.io "rm -f code_sandbox_client.tar.gz"
+echo "Moving tar to server"
+scp .deliver/code_sandbox_client.tar.gz nginx@ssh.codesandbox.io:./
```
lbogdan: Do we need the whole `.deliver` thing? Why not just do
```
tar -zcvf code_sandbox_client.tar.gz www
[...]
scp code_sandbox_client.tar.gz nginx@ssh.codesandbox.io:
```

nagamalli9999: This is a 0 logic change pull aimed at improving maintainability. I don't have enough context to make that change

lbogdan: Sorry, the question was more for @CompuIves, i should have mentioned. 😄

CompuIves: Hmm, yes. Yesterday we moved to the new deployment flow and this is kind of a leftover. We can just move to `www` directory now 😄 . Let's merge this one in first.

PR #322, opened 2017-11-23
Adjust styling of files entry containers

CompuIves:
Some styling changes:

1. Adjusted the height of the entries, this will go hand in hand with the redesign we're currently doing.
2. Changed inactive color to match with our other colors.
3. Fixed width of editing name fields of files.

Looks like this:

screen shot 2017-11-23 at 16 36 21


lbogdan:
Snapshots out of date. :smile:

PR #338, opened 2017-11-29
Replaced condition with regex

Piterden:
Regex https://regex101.com/r/xKBMVQ/1

Conditions

**What is the new behavior?**
Matching

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 

@CompuIves please tell me, if you need such of changes, I can do it in some time...

### DIFF
--- a/packages/app/src/sandbox/eval/transpilers/vue/loader.js
+++ b/packages/app/src/sandbox/eval/transpilers/vue/loader.js
@@ -32,11 +32,9 @@ const getStyleLoaders = (attrs, id, scoped) => {
     scoped: !!scoped,
   })}`;
 
-  if (attrs.lang === 'scss') loader += '!sass-loader';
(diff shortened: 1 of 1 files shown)


CompuIves:
Thanks for the PR! 

It's good to make this condition smaller, but I think this will make it harder to read/understand. The if statement version is more clear in which loaders are added for which lang attribute. Would like to hear the opinion of others here as well.

Piterden:
That's why I am asking )

lbogdan:
I was just about to say the same. Especially there's no performance (or any other) benefit here.

Piterden:
Ya, RegEx'es are just awesome. We could place a link to the RegEx editor in a comment near each RegEx, then it might play the teaching role for those who interested it...

lbogdan:
Closing this as no noticeable benefits, also the file belongs to the [vue-loader](https://github.com/vuejs/vue-loader) fork.

PR #350, opened 2017-12-01
fix: remove scrollbar flickering on mode select

maciej-ka:
Fix

https://github.com/CompuIves/codesandbox-client/issues/349

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged 

While details behind this scroll flickering remain mysterious to me, I realized that current layout has separate scrolls for each column, so the definition of page scroll width can be removed.


CompuIves:
Hmm, interesting. So the scrollbar is the origin of the issue.

With this change we change the visuals of Windows though, maybe there is another way to fix it?

With fix:
screenshot-pr350 cs lbogdan tk 2017-12-02 17-31-43-161

Without fix:
screenshot-codesandbox io 2017-12-02 17-30-14-919

maciej-ka:
I would like to replicate not so nice scrollbar in the leftmost column from your screenshot. 

But on Chrome 62.0.3202.94 things look fine:
screen shot 2017-12-02 at 18 27 14

And on Firefox 57.0 (64-bit) are rather smooth too:
screen shot 2017-12-02 at 18 26 52

Which browser did you use?

CompuIves:
Ah, yes. It only happens on windows PCs. I used Windows Chrome 62.

maciej-ka:
I hope to have some windows machine to improve this soon.

lbogdan:
Closing this since there was a major UI refactor meanwhile.
@maciej-ka If you're still having issues and find a fix, please open a new PR, it should be easier than fixing all conflicts for this one. :slightly_smiling_face:

PR #375, opened 2017-12-08
Update algolia.js

Haroenv:
SFFV already returns a promise


CompuIves:
😱  thanks for the PR!

PR #388, opened 2017-12-13
Fix overflowing MaxWidth__Container

donysukardi:
Bug fix

MaxWidth__Container exceeds 100% width because of the additional paddings in content-box box-sizing

image

**What is the new behavior?**
Use box-sizing: border-box for the Container, hence the paddings are incorporated into the actual width.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged 

**Additional Comments**
Ideally we should use global border-box box-sizing as the following. It will make the sizing more predictable.

However, introducing that at this stage kinda messes up the sizing of some layouts. We could have another PR to refactor it.

```
html {
  box-sizing: border-box;
}

*, *:before, *:after {
  box-sizing: inherit;
}
```

### DIFF
--- a/packages/common/components/flex/MaxWidth.js
+++ b/packages/common/components/flex/MaxWidth.js
@@ -2,6 +2,7 @@ import React from 'react';
 import styled from 'styled-components';
 
 const Container = styled.div`
+  box-sizing: border-box;
   display: flex;
 
   padding: 0 2rem;


CompuIves:
Great! Thank you @donysukardi. I wish we had done this on the first commit, all border box. Now it's a PITA to build it.

PR #389, opened 2017-12-13
Codesandbox refactor to Cerebral and Mobx-state-tree

christianalfoni:
So we have gotten to a point now where we can officially start sharing the progress and give more information about "the big refactor". This video introduces some background and conceptual changes: https://youtu.be/7Z4hInb7LmA .

The following remaining refactors are identified:

- [x] pages/CLI - simple
- [x] pages/Profile - medium
- [x] pages/index.js - simple
- [x] containers/integrations/Github - simple
- [x] containers/integrations/Zeit - simple
- [x] containers/modals/SignOutNotice - simple
- [x] containers/modals/ZenModeIntroduction - simple
- [x] containers/Navigation - simple
- [x] containers/UserMenu - simple
- [x] containers/ContextMenu - simple
- [x] containers/SignInButton - simple
- [x] Key bindings - complex
- [x] Codemirror - complex
- [x] Embed - complex
- [x] local storage handling - complex

When done, let the bug hunt begin :-)

PR #419, opened 2017-12-24
Handle .prettierrc file at root

ValentinH:
It adds support for `.prettierrc` file placed at the root of the sandbox. It updates the Prettier preferences when the `.prettierrc` is saved.

Idea from @kentcdodds at https://twitter.com/kentcdodds/status/944817103814692864

Staging demo: http://pr419.cs.lbogdan.tk/s/rj733y8y6o

This is a kinda raw experiment but it's working fine so far but I already have some improvements in mind: 
- keep the default values for the keys not provided by the `.prettierrc` so the preferences UI is not broken. Like below for the "Tab width":
image

### DIFF
--- a/packages/app/src/app/containers/Preferences/CodeFormatting/Prettier.js
+++ b/packages/app/src/app/containers/Preferences/CodeFormatting/Prettier.js
@@ -44,6 +44,19 @@ class Prettier extends React.PureComponent {
     return (
       <Container>
         <PreferenceContainer>
+          <Description>
+            This configuration can be overridden by a{' '}
+            <a
+              href="https://prettier.io/docs/en/configuration.html"
+              target="_blank"
+              rel="noopener noreferrer"
+            >
+              .prettierrc
+            </a>{' '}
+            JSON file at the root of the sandbox.
+          </Description>
+          <Rule />
+
           <PaddedPreference
             title="Print width"
             type="number"
--- a/packages/app/src/app/containers/Preferences/CodeFormatting/index.js
+++ b/packages/app/src/app/containers/Preferences/CodeFormatting/index.js
@@ -10,7 +10,7 @@ export default () => (
     <Title>
       Prettier Settings{' '}
       <a
-        href="https://github.com/prettier/prettier#options"
+        href="https://prettier.io/docs/en/options.html"
         target="_blank"
         rel="noopener noreferrer"
       >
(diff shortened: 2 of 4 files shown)


CompuIves:
Wow that was fast! Thank you so much!

I'm thinking that we could maybe do the parsing of `.prettierrc` right before we prettify. If there's a valid file we use that and otherwise we use the preference. I think the function of prettify is in the same file.

ValentinH:
You mean that we should not overwrite the preferences with the file content but just check if the file exists when prettyfying?

ValentinH:
I wasn't sure about how to implement this so I'm open to all ideas :)

CompuIves:
> You mean that we should not overwrite the preferences with the file content but just check if the file exists when prettyfying?

Yes exactly, I think that's the most intuitive way.

ValentinH:
Yes I thought about that but I wasn't sure where do to this? Could you point the file to edit? :)

ValentinH:
And then maybe show a message in the preferences panel to indicate that the prettier preferences are currently overwritten by the .prettier.

CompuIves:
> Yes I thought about that but I wasn't sure where do to this? Could you point the file to edit? :)

We run prettify here: https://github.com/CompuIves/codesandbox-client/blob/master/packages/app/src/app/store/entities/sandboxes/modules/actions.js#L49-L50. You can get the `.prettierrc` by getting all modules for the sandbox and doing something like this:

`modules.find(m => m.directoryShortid == null && m.title === '.prettierrc')`. Then use that instead if it exists.

> And then maybe show a message in the preferences panel to indicate that the prettier preferences are currently overwritten by the .prettier.

This is a **great** idea!!

ValentinH:
@CompuIves I have updated following your advice.

However, about my suggestion for showing a message in the case we are overwriting the Prettier config, I'm not sure how to achieve this.
I'm seeing in the state that there can be several sandboxes so how can I get the one currently displayed (except looking at the url)?

CompuIves:
Perfect! Implementation is also very good.

> However, about my suggestion for showing a message in the case we are overwriting the Prettier config, I'm not sure how to achieve this.

Hmm, yes. We will move to a new state system this month per #389, then that would be possible. For now I'd suggest adding a message like: "This configuration can be overwritten by a .prettierrc in the sandbox." to the prettier preferences window. Then this PR is ready to be merged and deployed!

ValentinH:
Ok, sounds good ! 
I'll complete this tomorrow, now it's Christmas time! 

Merry Christmas!!

ValentinH:
Done ;)

CompuIves:
Works perfectly!! Thank you very much!

ValentinH:
\o/

PR #437, opened 2018-01-08
Fixes bug with dependencies having no 'versions' field

### LINKED ISSUE #435: Dependency dialog closes after typing "cq-d"
screenshot from 2018-01-07 17-24-28
👆
 This is what I see in the console.
Pretty random. 😅

lbogdan:
For some reason the `cq-demo` package has no 'versions' field, but it has a 'version', so we use that.
Fixes #435.


CompuIves:
Very interesting, thanks for solving this @lbogdan!

PR #438, opened 2018-01-08
Fix browser false scenario

### LINKED ISSUE #436: postcss<6.0.10 works, postcss>=6.0.10 fails transpiling with missing module
Based on: https://twitter.com/ZeeCoder/status/950051949952618496

screenshot from 2018-01-07 17-07-21

I get this error when I try to use `postcss@>=6.0.10`.
Works fine with 6.0.9, and as far as I can tell the module (supports-color) codesandbox complains about is the exact same dep in both postcss versions.

A link that reproduces the issue:
https://codesandbox.io/s/1vk6lz431q

Downgrading to 6.0.9 fixes the issue. 🤔

Also: locally all postcss runs fine bundled.

CompuIves:
Should fix #436

### DIFF
--- a/packages/app/src/sandbox/eval/manager.js
+++ b/packages/app/src/sandbox/eval/manager.js
@@ -373,7 +373,7 @@ export default class Manager {
         this.cachedPaths[pathId] = resolvedPath;
       }
 
-      if (NODE_LIBS.includes(shimmedPath)) {
+      if (NODE_LIBS.includes(shimmedPath) || resolvedPath === '//empty.js') {
         return {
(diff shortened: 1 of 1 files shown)


ZeeCoder:
🤞
Thank you so much for the swift response! 🎉
When do these things usually go live? 🤔

CompuIves:
Deployed!

ZeeCoder:
Ah mate that's great it works exactly as it's expected: https://codesandbox.io/s/qzlwr68vqj
(All I did is if you resize the window to small enough, the text becomes red.)

Thanks for the quick fix! 🎉

Issue #353, opened 2017-12-01
'Expand console' share option broken in editor (works in embed)


happylynx:
Expand console in 'Share' dialog doesn't work. It adds `?expanddevtools=1` suffix to the url. However the suffix is not honored when the page is loaded.

Test page: https://codesandbox.io/s/1319n0kv44?expanddevtools=1

Steps to reproduce:
1. Create new sandbox and save it
2. Click 'Share'
3. Enable 'Expand console' option
4. Copy url form 'Editor url (also works on Medium)' field
5. Open the url in browser

Expected result: 'Console' panel is opened in the bottom part of Preview view.

Actual result: 'Console' panel is closed.

![image](https://user-images.githubusercontent.com/8846779/33507243-f8573c60-d6f3-11e7-91f7-c81c9106f90e.png)

PR #439, opened 2018-01-08
Fix expanding of devtools in the editor...

lbogdan:
…when url contains expanddevtools=1 parameter.

It's kinda naive, but it works.

### DIFF
--- a/packages/app/src/app/components/sandbox/Preview/index.js
+++ b/packages/app/src/app/components/sandbox/Preview/index.js
@@ -55,7 +55,7 @@ type Props = {
   inactive: ?boolean,
   shouldExpandDevTools: ?boolean,
   entry: string,
-  devToolsOpen: ?boolean,
+  shouldExpandDevTools: ?boolean,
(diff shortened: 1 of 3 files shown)


lbogdan:
Btw, while fiddling with Redux devtools, I've noticed we're not sending a `SET_DEV_TOOLS_OPEN` action on open, only on close.

CompuIves:
Hmm, I think this breaks the `Open Console` quick action. When you open Quick Actions and select 'Toggle Console' it doesn't open the console anymore.

lbogdan:
Well, I was kinda expecting it to break something. :man_shrugging: Maybe we should unify them?

gautamarora:
@lbogdan - This is really cool, and I look forward to this landing.

Just dropping an issue in here as it seems related and possibly one that will be fixed by your PR: https://github.com/CompuIves/codesandbox-client/issues/353

lbogdan:
Closing this for now as it's full of conflicts and we probably need another approach. Left #353 open and labeled it as a bug.

Issue #440, opened 2018-01-09
Support pug in .vue files


zephraph:
It was mentioned in #199, but having support for [pug](https://pugjs.org/api/getting-started.html) files in vue's single file components would be nice. 

Generally vue-loader supports pug out of the box when it's installed... but it looks like it might be a little more involved here. 

I can do a PR if desired, but I'm not sure I'll catch all the references.

Files to update:
- .../Frameworks/index.js
- .../Frameworks/icons.js
-  /packages/app/src/sandbox/eval/presets/vue-cli/index.js
- /packages/app/src/app/components/sandbox/CodeEditor/Monaco.js

Files to add:
- .../Frameworks/pug.svg [logo](https://cdn.rawgit.com/pugjs/pug-logo/eec436cee8fd9d1726d7839cbe99d1f694692c0c/SVG/pug-final-logo-_-colour-128.svg)
- .../Frameworks/pug.png
- /packages/app/src/sandbox/eval/transpilers/pug/...

Did I miss anything?

tristan-shelton:
This is a huge one for me. I hardly ever write HTML components any more.

zephraph:
I'm gonna give it a go. 🤞

pbastowski:
Is it possible to add code colouring for the PUG code in Vue components? 

Currently it looks like this:

![image](https://user-images.githubusercontent.com/5649024/40544579-17f45012-6029-11e8-9ae9-5e055b606ae2.png)

I would expect the code to be coloured, something like this (the colours are from another editor):

![image](https://user-images.githubusercontent.com/5649024/40544678-70ffda78-6029-11e8-9f34-09abd8cfa11c.png)

uNickLin:
@pbastowski 
try to turn on the setting 'Use CodeMirror' and it will looked better
![image](https://user-images.githubusercontent.com/22793771/42255762-eed62590-7f7f-11e8-833a-e524e87cde34.png)
![image](https://user-images.githubusercontent.com/22793771/42255791-20196b30-7f80-11e8-9eb4-dba1f72c33d8.png)

pbastowski:
That works. Thanks :) the color scheme is not as nice, though.

PR #443, opened 2018-01-11
[WIP] Add .pug file support generally and for .vue file templates

zephraph:
Feature; adds support for [pugjs](https://pugjs.org/) templates in vue single file components. 

Addresses #440. The below code fails in a codesandbox instance.

```vue
<template lang="pug">
  h1 test
</template>
```

it fails even when `pug` and `pug-loader` are added as explicit dependencies.

**What is the new behavior?**
`.pug` files will supported generally and in the `template` attribute of `.vue` files. 

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 


zephraph:
Gonna be totally honest... this is a lot more difficult than I thought it would be. 

The hardest part is getting pug to run in the browser. It's not necessarily made for that. I created a rollup bundle that pulls in all its resources buuuut... you can definitely tell it wasn't designed to run in the browser. Or be small. 

I've pushed my progress, but it's currently in a very broken state. I created a project called [pug-static](https://github.com/zephraph/pug-static) that does the rollup build and packaging for pug itself. 

If pug was written w/ esmodules it probably would be going a lot better. There's so much that could be filtered out. Maybe I'll do a PR over there... not sure how deep this rabbit hole goes though and if I really want to follow it all the way down.

tristan-shelton:
As I understand codesandbox uses babel, right? Could you just feed the Vue component into [babel-plugin-transform-pug-html](https://github.com/dumconstantin/babel-plugin-transform-pug-html) and then on to vue-loader as normal?

zephraph:
@triston-shelton thanks for the idea! That might work, but I don't think that it makes it easier. All this stuff is ran at runtime in the browser. I mean, in a web worker, but still. Point being, that just adds another dependency and more weight. That's one of the issues. 

@Compulves, hey thanks! 

> So if I understand you correctly you already have a transpiler from `pug` to `html`?

Well, sort of. I'm working off of the stylus transformer. In principle I know how to hook that up to your version of vue-loader, I'm just concerned about the size and all the node apis that pug relies on. I haven't finished the transpiler yet though.

egoist:
Not sure if this works for you, I have also built a browser version of pug here: https://github.com/egoist/browserified

CompuIves:
This looks perfect to me! The only remaining thing I'd like to have is https://unpkg.com/browserified-pug moved to `/static/js/pug.js`, this way we don't rely on an external service for speed and can use service worker caching.

zephraph:
I'll do that. So far as I can tell though, it's not really working. Did I miss something?

CompuIves:
Hey! Sorry for the late response! I think this code should work. The only thing left is adding the transpiler to the vue loader, that could be done like this:

Adding 

```js
pug: ['pug-loader']
```

to

https://github.com/CompuIves/codesandbox-client/blob/master/packages/app/src/sandbox/eval/transpilers/vue/loader.js#L117

And adding

```js
import pugTranspiler from '../../transpilers/pug';

vuePreset.registerTranspiler((m) => m.path.endsWith('pug'), [{ transpiler: pugTranspiler }]);
```

at the end of https://github.com/CompuIves/codesandbox-client/blob/master/packages/app/src/sandbox/eval/presets/vue-cli/index.js#L18. The last part registers the transpiler to the Vue template, so it can be used by other loaders.

I think that should do it!

zephraph:
Sorry I've been AWOL for a bit. Work is a bit nuts at the moment. I'll make those changes tonight. Let's see what happens. 🤞

zephraph:
Holy heck it works. The only thing that's really missing now is syntax highlighting. 

... but still.... IT WORKS. 🎉 🎊 

http://pr443.cs.lbogdan.tk/s/qxk7vr6qjw

CompuIves:
That is soo cool!! Maybe @rebornix has an idea on how we can best build support for this?

CompuIves:
Whoop! Merged 😄. Thank you very much! It'll be live in ~15 min.

rebornix:
if we are talking about syntax highlighting for pug in Monaco, you probably want to read https://microsoft.github.io/monaco-editor/monarch.html @zephraph

Issue #429, opened 2018-01-07
"could not find the Sandbox you're looking for..." when importing from github


marcinczenko:
I have a very basic project created with create-react-app in a public repository with URL:

https://github.com/marcinczenko/personas.git

## Intended

1. Create a new sandbox by importing from github:

    ```
    URL: https://github.com/marcinczenko/personas.git
    Sanbox URL: https://codesandbox.io/s/github/marcinczenko/personas.git
    ```

2. After clicking `Open Sandbox` I expect to see a sandbox that reflects the content of the imported repository and is associated with that repository (meaning I can commit, PR, etc). I also expect that if I push to the repo externally, the sandbox will pick up the changes (at least after it is reopened if not automatically).

## Actual
1. Create a new sandbox by importing from github:

    ```
    URL: https://github.com/marcinczenko/personas.git
    Sanbox URL: https://codesandbox.io/s/github/marcinczenko/personas.git
    ```

2. After clicking `Open Sandbox` I receive the message:

    ```
    We could not find the Sandbox you're looking for...
    Create Sanbox
    ```

3. After clicking on `Create Sanbox` a new clear sanbox is created that has no association with the github repository provided in the 1st step above.

CompuIves:
Ah, good one! We currently aren't parsing the `.git` correctly, removing that should work:

https://codesandbox.io/s/github/marcinczenko/personas

I'll fix that.

CompuIves:
Well, it got fixed by @arthurdenner 😄 .

Issue #425, opened 2017-12-31
Improve import from GitHub url validation


Hum4n01d:
There are a few improvements that could be made to the import from GitHub page. I will attempt to make these changes and submit a PR.

- The page should accept `hum4n01d/hogwash-client`, `github.com/hum4n01d/hogwash-client`, and `https://github.com/hum4n01d/hogwash-client` as valid GitHub repo urls
- The validator should check for `github.com/username/repo` rather than just `github.com`

Right now, it won't accept `hum4n01d/hogwash-client` as it doesn't contain `github.com`. 

If I put the github link without `https://` this happens:

![screen shot 2017-12-31 at 1 56 19 pm](https://user-images.githubusercontent.com/17228477/34464411-7889c646-ee32-11e7-8869-5f40be0594dd.png)

When I click `Open Sandbox`, it sends me to `https://codesandbox.io/s/github.com/hum4n01d/hogwash-client` which is invalid.

CompuIves:
Yes, we could definitely use better parsing!

cantyjeffrey:
@Hum4n01d
This would be great, as I have been trying (but unable) to import `https://github.com/cantyjeffrey/react-boilerplate` and have tried the following combinations with no such luck:

* https://github.com/cantyjeffrey/react-boilerplate.git
* https://github.com/cantyjeffrey/react-boilerplate
* github.com/cantyjeffrey/react-boilerplate
* cantyjeffrey/react-boilerplate

Am I missing something obvious as to why I can't import this repository?

<img width="1131" alt="screen shot 2018-01-05 at 12 30 15 pm" src="https://user-images.githubusercontent.com/2856292/34622811-3ea7f728-f214-11e7-9aaa-5b2bfce528d5.png">

cc: @CompuIves

Hum4n01d:
@cantyjeffrey I’ve had the same problem, though I don’t think it’s related to this issue :)

cantyjeffrey:
I agree. I tried uploading via CLI and think that the culprit is the fact that it’s an ejected CRA repo, which puts dependencies over the 40 dependency limit. Just a guess.

cantyjeffrey:
CLI threw error, but for GH import it’s a generic error message.

evgenyrodionov:
Also not working with branches with `/` 

https://github.com/mrzvdesign/aviasales-demo-frontend/tree/feat%2Fmain-page

imported as
[...]

PR #448, opened 2018-01-13
Improving GitHub URL parse

arthurdenner:
This PR aims to improve the URL parse in the "Import from GitHub" page. It relates with #425 and #429.

The input only accepts the following URLs:
```
http://github.com/user/repo
http://www.github.com/user/repo
https://github.com/user/repo
https://www.github.com/user/repo
```

**What is the new behavior?**
I'm not  very good with regex, but I tried my best and right now the input also handles:
```
github.com/user/repo
www.github.com/user/repo
```
and recognizes the following as invalid:
```
github.com/
github.com/user/
http://github.com/
http://github.com/user/
http://www.github.com/
http://www.github.com/user/
https://github.com/user/
https://www.github.com/user/
www.github.com/
www.github.com/user/
```

I added a regex to remove the `.git` in the `gitHubToSandboxUrl` function from the URL, if it contains.

Even though I've added tests, I'm not sure if this changes are enough for a better parsing, so let me know if you can see any improvements.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged 


CompuIves:
Ahh, this is perfect! I really like that you added the `.git` removal, we had a lot of related questions to that. Thank you very much, I'll merge it in now.

PR #454, opened 2018-01-14
Multi-line input for DevTools Console

xyc:
Multi-line input support for DevTools Console like VSCode console / Chrome console. It's implemented using Monaco (similar to VSCode implementation).

Also fixes issues where
- command cursor is not cleared on evaluating the command (for example if you press up arrow and edit the content, then press enter, the next up arrow should still bring you the immediate previous command)
- in firefox height is not correct when you have many messages.

Console input cannot have multiple lines.

**What is the new behavior?**
Console input has multi-line support. Shift + Enter goes to the next line. Enter evaluates the command. Up and Down arrow works as expected. You can paste multiple lines into the prompt as well.
multi-line input

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged 


packages/app/src/app/components/sandbox/Preview/DevTools/Console/Input.js
```diff
@@ -1,95 +1,188 @@
...
-  outline: none;
-  color: rgba(255, 255, 255, 0.8);
-  font-family: Menlo, monospace;
-  font-size: 13px;
 `;
 
-type Props = {
-  evaluateConsole: (command: string) => void,
+const monacoOptions = {
+  // language: 'javascript',
```
CompuIves: Why did you comment this line?

xyc: Thanks @CompuIves ! Pleasure to help. 

Do you want syntax highlighting and autosuggestion as well 😃 ?
syntax-highlight

But for that we would need to detect if we are in CodeMirror mode.ecause it looks that in Monaco there's only one theme for editors. (https://github.com/Microsoft/monaco-editor/issues/338)

xyc: The above changes (extracted to a separate file): https://github.com/xyc/codesandbox-client/commit/8b4f76d099b65080c80bf66acc32867f58c89825. Please let me know what you think @CompuIves

CompuIves: > Edit: Alternative approach is to extract defineTheme logic from Monaco.js into MonacoReactComponent/a separate file and have a guard that prevents the theme from being defined twice. That way even in code mirror mode we shall still have the theme available. (defining theme on input component mounting will sometimes result in text re-colorized briefly in other Monaco editors so it's best that theme is only defined once).

I like this approach, sounds good!

Syntax Highlighting and autosuggest are super. I love to work with this console now!

xyc:
Cool I've merged in the syntax highlighting changes and merged with `upstream/master`. Should be ready to merge now. Thanks for taking your time to review this!

CompuIves:
Maaan you deserve **all** the thanks! I love this feature, and the amount of polish you've added is impressive. Thanks for your contribution!

xyc:
Thanks for the opportunity to contribute! Love what you've been doing with CodeSandbox!

PR #461, opened 2018-01-17
Update Select.js

Haroenv:
fix how it looks in Safari by using non-browser appearance (now consistent in all browsers)

styling

Safari: 

screen shot 2018-01-17 at 11 08 56

**What is the new behavior?**

screen shot 2018-01-17 at 11 07 19

**Checklist**:

- [x] Documentation N/A
- [x] Tests N/A
- [x] Ready to be merged 

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -284,6 +284,16 @@
         "code",
         "ideas"
       ]
+    },
+    {
+      "login": "haroenv",
+      "name": "Haroen Viaene",
+      "avatar_url": "https://avatars3.githubusercontent.com/u/6270048?v=4",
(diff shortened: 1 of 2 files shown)


CompuIves:
This is great! I did some testing and it seems like the arrow on the right side is removed from Chrome:

screen shot 2018-01-17 at 12 00 33

After fixing that it's ready for merge 😄

Haroenv:
I only tested my changes in the devtools actually, I think that the style here didn't apply correctly, is that possible?

Haroenv:
The image was behaving weirdly so I base64 encoded it (even though that's usually not necessary for svg)

CompuIves:
Looks good now! 😄

Issue #368, opened 2017-12-06
"Relative" paths in the code preview panel


Nufflee:
Not sure if this is in development and/or reported before but I encoutered this issue earlier today while working on a 720p screen. It is not really an issue, but it was kind of annoying to scroll all the way till the en d of the url bar just to change the current route. There should be an option to only show the part of the url, for instance https://sandbox.io/ [my route here] and not the mile long randomly generated one. 
Cheers!

SaraVieira:
Closed due to inactivity

Feel free to reopen @Nufflee

PR #463, opened 2018-01-17
Add relative URLs (and address issue #368)

Nufflee:
New feature - relative URLs, addresses the issue #368.

There's no current behavior because the feature didn't exist before.

**What is the new behavior?**
Now there's a Preference item (Preferences -> Preview -> last two items) where you can enable "Relative URLs" and change the actual URL. This makes it possible that when the user (if this is enabled) types <wanted relative url>/someRouteInTheApp, it redirects to the full CodeSandbox URL. I'm introducing this change because it's pretty annoying to scroll to the end of the usual long URL every time when working on a smaller screen.

**Checklist**:
- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged

There's still some stuff I can do. For instance disabling the relative URL text box unless relative URLs are enabled for instance. If you have any other ideas I'm more than happy to hear them! Thank you for making this possible @CompuIves!


lbogdan:
Hey @Nufflee , I'm doing a routine spring cleanup here; you were off to a very good start, what happened? :slightly_smiling_face:

SaraVieira:
Closed because of lack of activity

Issue #475, opened 2018-01-26
CodeSandbox not loading inline .d.ts files in node modules


nicknisi:
For TypeScript examples, CodeSandbox is not currently looking for inline type definition files that may be distributed in the npm package, and that are not called out in the `types` property in package.json.

As an example, all of the Dojo 2 modules ([core](https://github.com/dojo/core) as an example) distribute UMD modules for each TypeScript module, along with a .d.ts file for each, rather than specifying a single .d.ts file in package.json. TypeScript can resolve modules in this format, but CodeSandbox currently cannot.

PR #476, opened 2018-01-26
Check for inline type dependencies

nicknisi:

This fixes the issue described in #475. This solution first searches for `types` in package.json, then checks for inline .d.ts files in the packages, then falls back to looking at DefinitelyTyped. It may not be the most efficient way to solve this, so any guidance you can provide would be welcomed!

Currently, type definitions are looked for in the `types` property of package.json. If they are not there, the code checks for the definitions in DefinitelyTyped and then silently fails if they are not found.

**What is the new behavior?**

This change adds a step in the middle that will search the package for inline .d.ts files and, if found, will load them before resorting to checking DefinitelyTyped.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 

I did not find an existing test suite for fetch-dependency-typings.js


CompuIves:
Wow, this is perfect @nicknisi! Thanks for the contribution 😄

Issue #484, opened 2018-01-30
Module not found although is added as a dependency 


bogdansoare:
<img width="1439" alt="screen shot 2018-01-30 at 23 26 03" src="https://user-images.githubusercontent.com/2750170/35592267-1f955d84-0615-11e8-97db-b182728826b5.png">

CompuIves:
This one was fixed by #490 😄

PR #490, opened 2018-01-31
Fix dynamic package.json resolving from unpkg

### LINKED ISSUE #487: [bug] Importing projects with "proxy" directories fails
I suspect this is a bug in a module resolver, wasnt sure in which repo I should report it.

**Input**
```js
import { deferred } from 'redux-saga/utils'
```

**Output**
```
Could not find module in path: 'redux-saga/utils'
```

[This](https://github.com/redux-saga/redux-saga/tree/master/utils) is how `redux-saga/utils` is structured. It is a directory which proxies to appropriate files (so main/module can be appropriately resolved in cjs/es environments) using private `package.json`.

CompuIves:
Fixes #487 and #484

### DIFF
--- a/packages/app/src/sandbox/eval/npm/fetch-npm-module.js
+++ b/packages/app/src/sandbox/eval/npm/fetch-npm-module.js
@@ -97,6 +97,8 @@ function downloadDependency(depName: string, depVersion: string, path: string) {
   return packages[path];
 }
 
+const requestedPaths = [];
+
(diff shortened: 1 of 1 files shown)


Andarist:
Did it get deployed? Or will be later? Thanks for a quick fix!

CompuIves:
Being deployed now!

Andarist:
Superb! 😍

PR #508, opened 2018-02-07
Fix error.captureStacktrace for safari

### LINKED ISSUE #502: Test failures don't have the correct stack trace
Amazing new feature with the tests!!

I noticed that if your test assertion is failing, there's an error, instead of the stack trace and failed assertion

```
Error.captureStackTrace is not a function. (In 'Error.captureStackTrace(p,s)', 'Error.captureStackTrace' is undefined)
```

screen shot 2018-02-07 at 15 32 36

This only happens in Safari, there might be something wrong with the reloading of the service worker somehow, but I'm not sure what really the issue is

CompuIves:
Fixes #502 (hopefully). Needs testing

### DIFF
--- a/packages/common/load-dynamic-polyfills.js
+++ b/packages/common/load-dynamic-polyfills.js
@@ -23,9 +23,18 @@ function detectOpera() {
 }
 
 export default function requirePolyfills() {
+  const promises = [];
+
   if (detectIE() || detectOpera()) {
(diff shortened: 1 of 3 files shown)


lbogdan:
Rebuilt PR http://pr508.cs.lbogdan.tk/s/new .

PR #511, opened 2018-02-08
Return the proper eval-result type

samdenty99:
Bug fix

##### All console return values that are falsy return `undefined` (`false`, `0`, `null`   =>  `undefined`)
image

**What is the new behavior?**
Console return values, display the correct type (`false`  =>  `false`...)

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests
- [X] Ready to be merged 

### DIFF
--- a/packages/app/src/app/components/Preview/DevTools/Console/index.js
+++ b/packages/app/src/app/components/Preview/DevTools/Console/index.js
@@ -54,11 +54,7 @@ class Console extends React.Component {
         const parsedJson = result ? CircularJSON.parse(result) : result;
 
         if (!error) {
-          if (parsedJson) {
-            this.addMessage('log', [parsedJson], 'return');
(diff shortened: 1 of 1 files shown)


samdenty99:
Additionally, it might be worth using something like [`typeof`](https://www.npmjs.com/package/typeof--) to extract the constructors name and apply specialized styles for the different types.

For example, `new Promise(()=>{})` currently returns an empty object in the console.

CompuIves:
Wow, this is perfect @samdenty99! Thank you very much!

> It might be worth using something like typeof to extract the constructors name and apply specialized styles for the different types.

I agree, let's merge this one in already and look at that then 😄

samdenty99:
I'll get round to adding the console data types tomorrow, I did it for another project and they look awesome.

image

PR #521, opened 2018-02-08
Add links to npm dependencies

tomkuehl:
Hi,

i've added an a-tag to the npm dependencies in the sidebar, so you can easily check out the packages used in a sandbox.

**Checklist**:
- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged


CompuIves:
This is really good to have, thank you very much @tomkuehl!

I was wondering if it would be possible to style the URL color based on the template (eg everything is green on https://codesandbox.io/s/vue). You can create such an effect like this:
https://github.com/CompuIves/codesandbox-client/blob/fix/dynamic-dep-resolve/packages/app/src/app/pages/Sandbox/Editor/Workspace/Files/DirectoryEntry/Entry/elements.js#L11

I also think that it would be nicer if we'd remove the underline from the URL, I'm afraid that it will distract too much otherwise.

After that it's ready for merge!

tomkuehl:
Here we go! I have also styled the link for external resources

CompuIves:
**Perfect**! Thank you so much!

PR #536, opened 2018-02-11
Scrolllock

### LINKED ISSUE #492: Scroll position is not preserved when switching between tabs
When I switch forth and back between editors in tabs, the cursor position is preserved, but the scroll position isn't. The editor is always scrolled to the top after switching tabs.

chakrihacker:
closes #492

### DIFF
--- a/packages/app/src/app/store/modules/profile/actions.js
+++ b/packages/app/src/app/store/modules/profile/actions.js
@@ -13,7 +13,7 @@ export function deleteSandbox({ api, state }) {
         id: sandboxId,
       },
     })
-    .then(() => undefined);
+    .then(() => state.set(`profile.current.sandboxCount - 1`));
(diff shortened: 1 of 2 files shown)


packages/app/src/app/store/modules/profile/actions.js
```diff
@@ -13,7 +13,7 @@ export function deleteSandbox({ api, state }) {
         id: sandboxId,
       },
     })
-    .then(() => undefined);
+    .then(() => state.set(`profile.current.sandboxCount - 1`));
```
CompuIves: I think that this one is still from the other PR.

chakrihacker: I think github recognises it won't be a problem, If you want I can create new commit

Issue #538, opened 2018-02-11
Errors updates from other module does not update after module changed in monaco editor


br1anchen:
I am fairly new user to codesandbox.io, it might be an user error after all :P

**Bug reproduce**:
If user removed template file `Hello.js`, and recreated an new `Hello.js` file, and export correct default module.
Original `index.js` will still have error info about `Can not find module in path './Hello'`

**Screenshot**:
<img width="1918" alt="screen shot 2018-02-11 at 23 03 30" src="https://user-images.githubusercontent.com/1086461/36079152-3973c35c-0f80-11e8-84b5-70b89afbff45.png">

After some digging, I think the reason causes it is that, after new `Hello.js` is created, observable `store.editor.erros` is notified through reaction `disposeErrorsHandler` ([here](https://github.com/CompuIves/codesandbox-client/blob/master/packages/app/src/app/pages/Sandbox/Editor/Content/index.js#L119)), since the `modue` now is `Hello.js` instead of `index.js`, it will not call `setModelMarkers` with corresponding `model` ([here](https://github.com/CompuIves/codesandbox-client/blob/master/packages/app/src/app/components/CodeEditor/Monaco/index.js#L352)).

Then I think if any errors updates for other module will be retrieved by not-current module at all. It seems working fine with `CodeMirror`.

CompuIves:
Fixed by you! :D

PR #539, opened 2018-02-11
Fix errors not updated for other module

br1anchen:
Bug fix for #538: 
Update errors info when module changed

Error changes caused by current editor module will not be notified by other module

**What is the new behavior?**
When module changed, it will try to update with current erros

**Checklist**:

- ~~[ ] Documentation~~
- ~~[ ] Tests~~
- [x] Ready to be merged 

Current implementation is to update errors for new module when module changed. I think it is not necessary to update all errors for all related module except current edit module since codesandbox.io is always editing (single?) module.
It is efficient enough to update errors info when module changed.


CompuIves:
Oh this is really great! Thank you very much @br1anchen!! Could you fix the remaining lint errors? Then it's ready for merge.

packages/app/src/app/pages/Sandbox/Editor/Content/index.js
```diff
@@ -185,15 +185,16 @@ class EditorPreview extends React.Component<Props, State> {
     );
     const disposeModuleHandler = reaction(
       () => [store.editor.currentModule, store.editor.currentModule.code],
-      () => {
+      ([newModule, _]) => {
```
CompuIves: I think you can just change this line to:

`([newModule]) => {`

br1anchen: True, thanks for your input.

br1anchen:
@CompuIves by the way, is there any reason we are using forked version `monaco-editor` here? For package stability?

CompuIves:
Great! Thanks for this PR 😄 

> is there any reason we are using forked version monaco-editor here? For package stability?

I use a version fo monaco-editor that is a bit more integrated with CodeSandbox, for example, you can open files by CMD+Clicking on the path in the code.

PR #549, opened 2018-02-13
Add ability to enable the Cerebral debugger

christianalfoni:
Gives the ability to enable the Cerebral debugger in production.

This can be used to help debugging the app, but also just very awesome being an open source project for developers :-)

### DIFF
--- a/packages/app/src/app/controller.js
+++ b/packages/app/src/app/controller.js
@@ -3,10 +3,16 @@ import store from './store';
 
 let Devtools = null;
 
-if (process.env.NODE_ENV !== 'production') {
+if (
+  process.env.NODE_ENV !== 'production' ||
+  (typeof window !== 'undefined' &&
+    'localStorage' in window &&
+    JSON.parse(window.localStorage.getItem('settings.connectDebugger')))
+) {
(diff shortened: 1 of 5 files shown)


packages/app/src/app/controller.js
```diff
@@ -3,10 +3,16 @@ import store from './store';
...
 let Devtools = null;
 
-if (process.env.NODE_ENV !== 'production') {
+if (
+  process.env.NODE_ENV !== 'production' ||
+  (typeof window !== 'undefined' &&
+    'localStorage' in window &&
+    JSON.parse(window.localStorage.getItem('settings.connectDebugger')))
+) {
   Devtools = require('cerebral/devtools').default; // eslint-disable-line
```
CompuIves: Could we make this a dynamic `import`? That saves bundle size for others.

packages/app/src/app/controller.js
```diff
@@ -3,10 +3,16 @@ import store from './store';
...
 }
 
+// preferences.settings.${props`name`}
```
CompuIves: This one can be removed.

PR #561, opened 2018-02-17
Refactor to TypeScript and Cerebral Fluent - Step 1

christianalfoni:
As part of a 3 step process the Codesandbox project will be refactored to Typescript and the upcoming Cerebral Fluent addon. This will give type safety on all state and application logic. 

### Step 1
Run the project with the **fluent** addon. This replaces mobx-state-tree with plain mobx under the hood. 

### Step 2
Add Typescript loader, define state types and rename files to `.ts(x)`

### Step 3
Refactor the sequences and actions to the **fluent API** for complete type safety

This PR currently needs help testing. Just identify any errors.


christianalfoni:
@CompuIves Not sure why jest fails on the button here? 😂 Have not done anything to that button!

christianalfoni:
From my tests this seems ready actually :)

packages/app/package.json
```diff
@@ -1,217 +1,210 @@
 {
```
CompuIves: Could we get this file to two spaces as spacing again, that will make the diff easier to read 😄

packages/app/src/app/pages/Profile/Showcase/index.js
```diff
@@ -18,8 +18,8 @@ class Showcase extends React.Component {
 
   render() {
     const { store } = this.props;
-    const sandbox = store.profile.showcasedSandbox;
-    const isCurrentUser = store.profile.isProfileCurrentUser;
+    const sandbox = store.profile.showcasedSandbox.get();
```
CompuIves: So if I understand correctly these `.get()` calls will have the right typings returned in the future? What's the reason of using `get`?

packages/app/src/app/store/actions.js
```diff
@@ -1,10 +1,18 @@
 import axios from 'axios';
-
+import { Dictionary } from '@cerebral/fluent';
```
CompuIves: What does `Dictionary` do?

packages/app/src/app/store/modules/editor/index.js
```diff
@@ -43,18 +32,27 @@ export default Module({
...
-    currentSandbox,
-    currentModule,
-    mainModule,
-    currentPackageJSON,
-    currentPackageJSONCode,
-    parsedConfigurations,
-  },
-  computed: {
-    isModuleSynced,
+    get isAllModulesSynced() {
```
CompuIves: Will this allow you to also set a `set`?

packages/app/package.json
christianalfoni: Ah yeah, it is prettier messing up... how do I make prettier only save on change if there is a prettier config in the project? You know? :)

christianalfoni: So weird... my editor refuses to format it correctly... anyways, fixed it :)

packages/app/src/app/pages/Profile/Showcase/index.js
christianalfoni: Yeah, you can read about how fluent is used in an app here https://github.com/cerebral/cerebral/blob/canary/docs/guides/typescript.md

There are currently two reason for using computed.

1. To compute sorting, filtering and other kinds of performance affected code. Computed caches the value

2. Only computed has access to parent state

Ideally we would just use a `getter` on this situation, but because we need access to parent state we need to use a computed. I am not sure how to solve `getters` with parent state in a good way. Like technically we can solve everything, the challenge is the typing :)

This is still under evaluation though, so maybe we find a way and can refactor this in the next round :)

packages/app/src/app/store/actions.js
christianalfoni: Dictionary is the same as `ObservableMap`... it basically IS an `ObservableMap`. Using `mobx-state-tree` this is managed "under the hood" because you use the type: `types.map`, but now we explicitly create them. They are observable dynamic maps... basically objects where you add/remove keys.

packages/app/src/app/store/modules/editor/index.js
christianalfoni: Yeah, we can allow for this. You mean doing a:

```js
set isAllModulesSynced () {
  this.changedModuleIds = []
}
```

Type of thing?

christianalfoni: Let me try to get this running, it is nice thing to have :)

christianalfoni: Oh, haha, this actually already works :) It requires a getter set... but yeah, of course you would have a getter first :)

Guria:
Can't start http://pr561.cs.lbogdan.tk/s/new in IE11.
Current codesandbox works fine

christianalfoni:
@Guria found issue with typescript target on build. I think we should hold this until we move `canary` into `next`, which will happen later today most likely, or tomorrow. Will fix the build then as well :)

packages/app/src/app/store/modules/editor/index.js
CompuIves: Awesome!

Guria:
@christianalfoni ok, just checked, now it works in IE11

CompuIves:
I'm getting this when testing on staging:

screen shot 2018-03-09 at 15 56 01

Maybe something due to recent commits?

CompuIves:
Hmm, this is weird. It seems like our Parcel templates broke? http://pr561.cs.lbogdan.tk/s/vanilla

CompuIves:
Ah maybe a dependency issue? I'll generate a new `yarn.lock`.

CompuIves:
Really weird, I think this is a dependency issue but don't know where to look.

CompuIves:
Well, the diff is much smaller now 😄

CompuIves:
Hmm interesting, for me it blocks when loading http://pr561.cs.lbogdan.tk/s/vanilla.

My console:

screen shot 2018-03-13 at 12 30 46

lbogdan:
Tried to build the PR and it failed with
```
src/manager/index.ts(107,27): error TS2345: Argument of type 'Window | null' is not assignable to parameter of type 'Window'.
  Type 'null' is not assignable to type 'Window'.
```

PR #596, opened 2018-02-27
Error in Current Module View for TypeScript React Template

sdaves:
https://codesandbox.io/s/react-ts

Bug Fix

Currently, when I select "Current Module View" for Hello.tsx, I get the error: "Error No boilerplate found for /src/Hello.tsx, you can create one in the future".

**What is the new behavior?**

I would like to see the exported class from the module in the preview window. The code change I made is purely to demonstrate the goal, let me know if you feel something else would better resolve the error.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 

Thank you so much for CodeSandbox!

### DIFF
--- a/packages/app/src/sandbox/boilerplates/default-boilerplates.js
+++ b/packages/app/src/sandbox/boilerplates/default-boilerplates.js
@@ -13,6 +13,12 @@ export default function(module) {
 `,
 };
 
+export const TS = Object.assign(JS, {
+  id: 'ts',
+  extension: '.ts',
+  condition: '.tsx?$',
+})
+
 export const HTML = {
   id: 'html',
   extension: '.html',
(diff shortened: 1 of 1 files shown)


CompuIves:
Thanks for the PR! For some reason I noticed that the boilerplate evaluation fails for the TS template, I think the fix would be changing the TS boilerplate to `.js` instead of `.ts`. But keep the condition on `.tsx?`. That should (maybe) fix it 😄

sdaves:
Do you mean like that?

lbogdan:
Rebuilt PR, but I still get a `No boilerplate found for /src/Hello.tsx, you can create one in the future` error in http://pr596.cs.lbogdan.tk/s/react-ts when in `Current Module View` and I click on `Hello.tsx`.

PR #602, opened 2018-02-28
Fix/documentation nav links

mdxprograms:
Adds a fix for documentation links and tidies up the string concatenation on links

issue: https://github.com/CompuIves/codesandbox-client/issues/585

I noticed that when clicking on a left nav link instead of going to https://codesandbox.io/docs/importing it goes to https://codesandbox.io/docs/docs/importing

After being on docs/importing and clicking on configuration it goes to:
https://codesandbox.io/docs/importing/docs/configuration

**What is the new behavior?**
This corrects the links to not use relative paths but to append sub link navigation links correctly.

**Checklist**:
- [x ] Documentation
- [ x] Tests
- [ x] Ready to be merged
- [x ] Added myself to contributors table


CompuIves:
Perfect, thank you very much!

PR #610, opened 2018-03-02
Add "ROOT_URL" build environment variable

lbogdan:
Needed for staging environment. Default value is `https://codesandbox.io`.

Usage:
`ROOT_URL=http://staging.cs.lbogdan.tk yarn build`.

### DIFF
--- a/packages/app/config/env.js
+++ b/packages/app/config/env.js
@@ -18,6 +18,10 @@ const getHost = () => {
     return `http://${process.env.STAGING_BRANCH}.cs.lbogdan.tk`;
   }
 
+  if ('ROOT_URL' in process.env) {
+    return process.env.ROOT_URL;
+  }
+
   return 'https://codesandbox.io';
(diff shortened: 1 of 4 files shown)


CompuIves:
Perfect!

packages/app/config/env.js
```diff
@@ -18,6 +18,10 @@ const getHost = () => {
     return `http://${process.env.STAGING_BRANCH}.cs.lbogdan.tk`;
   }
 
+  if ('ROOT_URL' in process.env) {
+    return process.env.ROOT_URL;
```
CompuIves: What's the difference between `ROOT_URL` and `CODESANDBOX_HOST`?

lbogdan: `ROOT_URL` basically overrides the logic (and the default value of `https://codesanbox.io`) of `CODESANDBOX_HOST`.

PR #620, opened 2018-03-03
Pr/shorten error messages

joeynimu:
This PR is for shortening error messages in the console.
It simply wraps the errors in a <detail /> tag and gets the first line of the error and inserts it in a <summary /> tag. And if there are more lines in the error apart from the first line, it spits then just after the <summary /> tag inside the <detail /> tag. This enables the user to expand on the error and avoids the console being clattered.


joeynimu:
@CompuIves please have a look if this is good to go, thanks.

CompuIves:
This is super! This looks good, I found one minor issue:

screen shot 2018-03-03 at 14 47 13

It seems the line breaks are gone in the detail message. When that goes in it's good to go!

packages/app/src/app/components/Preview/DevTools/Console/Message/index.js
```diff
@@ -50,6 +59,24 @@ function getMessage(message) {
...
+      const firstLine = splitMessage(msgLine);
+      const msgArray = msgLine.split('\n');
+      if (msgArray.length > 1) {
+        otherErrorLines = msgArray.slice(1);
+      }
+      return (
+        <InnerItem>
+          <details>
+            <summary>{firstLine}</summary>
+            {otherErrorLines}
```
CompuIves: I think this should become `otherErrorLines.join('\n\r')` to make it work again 😄

packages/app/src/app/components/Preview/DevTools/Console/Message/index.js
```diff
@@ -50,6 +59,24 @@ function getMessage(message) {
   }
 
   if (message.arguments.every(argument => typeof argument === 'string')) {
+    if (message.logType === 'error') {
```
CompuIves: Would also be good to put a comment here what the purpose of this code is. For new readers.

CompuIves:
Works perfectly! Thank you very much 😄

joeynimu:
Awesome!

PR #634, opened 2018-03-08
Fix overridden history.go() to sendUrlChange()

### LINKED ISSUE #633: Reverting history pop not reflected in address bar
If an application "blocks" a navigation that is triggered using the browser's back button, the URL in the address bar is inconsistent with the actual location.

Clicking the back (or forward) buttons immediately update the location, which an application must detect by listening for `popstate` events. "Blocking" in this case consists of using `history.go` to revert to the previous location. This behavior is different than clicking a link, which can avoid the browser knowing about the navigation altogether.

### Actual Behavior

https://codesandbox.io/s/x2wr3p21qo

Click on the `/other` link and the pathname is now `/other`.

If you click on the `Home` link, then cancel the prompt, the pathname is still `/other`.

If you click on the fake browser's back button, then cancel the prompt, the pathname is `/`. However, if you open the console and type `window.location.pathname`, you will see that the actual pathname is `/other`.

### Expected Behavior

The app detects the `popstate` and the URL in the fake browser's address bar reflects the true state.

----

I haven't looked into how CodeSandbox's fake browser works, so I'm not sure whether this has a quick fix or if it will require something more involved. I can probably take a look at it sometime soon.

lbogdan:
Fixes #633.

### DIFF
--- a/packages/app/src/sandbox/url-listeners.js
+++ b/packages/app/src/sandbox/url-listeners.js
@@ -37,6 +37,7 @@ export default function setupHistoryListeners() {
           const oldURL = document.location.href;
           origHistoryProto.replaceState.call(window.history, state, '', url);
           const newURL = document.location.href;
+          sendUrlChange(newURL);
           if (newURL.indexOf('#') === -1) {
             window.dispatchEvent(new PopStateEvent('popstate', { state }));
           } else {


CompuIves:
Whoop! Thank you @lbogdan!

PR #700, opened 2018-04-06
Default clearConsoleEnabled to true

### LINKED ISSUE #698: Clear console on change
When using the Console Preview mode, it would be nice to be able to clear it automatically on each code change.

I was thinking about adding a configuration item like `clearConsoleOnChange`
 in the `sandbox.config.json` file.
Another solution could be to add a toggle directly in the Console view header (both devtools and preview) so it's more obvious to find this feature.

What do you think about this? Can I provide a PR?

ValentinH:
Closes #698.


CompuIves:
Whoop! Perfect 😄.

PR #683, opened 2018-04-01
CxJS template

mstijak:
**What kind of change does this PR introduce?**
Adds a new template for CxJS (#628)

**What is the current behavior?**

**What is the new behavior?**

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 
- [ ] Added myself to contributors table 

It took me a while, but it actually wasn't that hard. Please let me know what's the next step.

PR #665, opened 2018-03-26
Add Dojo 2 as a Preset Template fixes #447

nicknisi:
**What kind of change does this PR introduce?**

This PR adds a new template for Dojo 2.

**What is the current behavior?**
This is a new template, requested and described in #447 

**What is the new behavior?**

Dojo 2 template

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 
- [ ] Added myself to contributors table 

- add dojo2 template
- add temporary project as template to load (github)
- add type definition creation to styles transpiler
- add dojo 2 to home page
- add default loading of dojo template src/index.html and src/main.css

**Note:** This has a backend change in order to set the correct template value coming from the API. To work around this in development, I added the following line to packages/app/src/app/store/providers/Api.js:

```diff
diff --git i/packages/app/src/app/store/providers/Api.js w/packages/app/src/app/store/providers/Api.js
index 35248ccb..db410a04 100644
--- i/packages/app/src/app/store/providers/Api.js
+++ w/packages/app/src/app/store/providers/Api.js
@@ -78,6 +78,7 @@ function handleResponse(response, { shouldCamelize = true } = {}) {
     camelizedData.data.npmDependencies = response.result.data.npm_dependencies;
   }
 
+  camelizedData.data.template = '@dojo/cli-create-app';
   return camelizedData.data ? camelizedData.data : camelizedData;
 }
```

PR #709, opened 2018-04-08
Community templates

CompuIves:
This is a combination of #665 and #683. I'm using this to see the complete diff and do testing. I will deploy this to staging too 😄


CompuIves:
cc @nicknisi and @mstijak. I think it would be great if both templates would be based of a GitHub repository. Then I'll change the logic of the importer to detect `CxJS` and `Dojo` so we can do final proper testing on staging.

nicknisi:
Yeah, that makes sense!

mstijak:
@CompuIves The CxJS template is available at https://github.com/codaxy/cxjs-codesandbox-template. It should be enough to detect `cx` package dependency. Please note that `react` will also be listed in dependencies as CxJS uses React for rendering.

nicknisi:
@CompuIves Any ideas on when you might be planning to merge this? Not trying to rush at all, just trying to get an understanding your timeline!

CompuIves:
I've put everything on staging now! I'm planning to merge it today.

If you're interested in helping with testing, send your IP to me via Discord or something, I'll add you to the whitelist on our staging server 😄

CompuIves:
Did a bunch of testing, looks good! One thing I'm worrying a bit about is the amount of cubes on the homepage. I'm going to do a code review now and decide about the cubes after that. But it's definitely certain that we'll have this live today! 🎉

CompuIves:
Okay! Everything is ready to merge in after the checks are green and ready to deploy. I decided to keep the cubes on the homepage, but will probably remove some cubes in 2 days after the initial announcement. The reason for this is that every cube takes quite a lot of performance (5 big box-shadows per cube, with transforms) and I'm afraid that our bounce rate will increase if we make it very hefty.

Planning to use this GIF for the announcement:


cx-dojo-xs-small

But bigger of course :smile:

PR #719, opened 2018-04-11
Reintroduce React module type and optimize module type detection

donysukardi:
Reintroducing module detection for React in a more optimal way 

https://github.com/CompuIves/codesandbox-client/issues/673

**What is the new behavior?**
Module types are determined and persisted to the store during sandbox initialization module creation/rename, and save.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 


CompuIves:
Hmm, I like this approach personally, but there is one aspect to keep in mind.

Until now the sandbox and modules were always left untouched in the state, so what I mean with that is that we don't add custom properties on top of the response from the API. This was introduced with the Cerebral refactor, before that I just added properties as I went 😄.

Maybe it's good to denote that, for example, we could maybe make the `type` property start with a `$` (`$type`) or an `_` (`_type`) to let everyone know that this property is not saved in the API.

I'm also thinking that we could maybe make this a computed property, something called `getType` which will do this:

```js
function getType(moduleId) {
  const module = getModule(moduleId)
  return module.type || getType(module.title, module.code);
}
```

But I don't know if this makes sense, I'm curious what you think on both points!

donysukardi:
I thought about using computed property as well, but then it’d have the same effect as before as it’ll be computed whenever `module.code` changes

donysukardi:
Oopss.. I accidentally hit the close pull request button on my mobile :grin:

lbogdan:
@donysukardi , @CompuIves 
Just wanted to let you know I manually deployed this PR at http://pr719.cs.lbogdan.tk/s/new (as there were some issues with the build server when the PR was created).

donysukardi:
So do you prefer `$` or `_` for the prefix? I’ll update the PR accordingly

CompuIves:
Wow, this is so good. We should've merged this a long time ago! Let's see if we can still fix the merge conflicts.

donysukardi:
I hope it's ok now! :D

CompuIves:
Hmm, it's painful to see my responses on this. This should've been merged when we had the chance. I'm closing this now, as this mostly relates to Cerebral and we nowadays use Overmind. We can still add this here: https://github.com/codesandbox/codesandbox-client/blob/1389461369765c283797354ceb89eb84d6b34ee5/packages/app/src/app/overmind/effects/utils/sandbox.ts#L1, but I think it's better to create this in a new PR, otherwise I'll miss it again. THanks a huge amount for the work you've put into this, and I'm sorry for the wasted effort you've gotten from this.

PR #720, opened 2018-04-11
Indicate turning VIM mode on will enable CodeMirror

frontsideair:
Bug fix

Switching the VIM mode setting on enables CodeMirror without notifying the user.

**What is the new behavior?**
It now indicates the current behavior.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 

I have stumbled upon this when I started with a React Typescript sandbox but didn't see type errors I was expecting. I opened the same sandbox in a different browser to test and the errors were present there. At first, I thought it was a browser problem, but then I noticed the errors were gone after I switched the vim mode on there too. It was confusing.

The wording and detail are up to debate, but it should be indicated. A future improvement can be to make the CodeMirror option disabled at `on` state if VIM mode is selected. I don't know how long until Monaco gains vim support though, it may be too much work for little gain.

### DIFF
--- a/packages/app/src/app/pages/common/Modals/PreferencesModal/EditorPageSettings/EditorSettings/index.js
+++ b/packages/app/src/app/pages/common/Modals/PreferencesModal/EditorPageSettings/EditorSettings/index.js
@@ -80,7 +80,7 @@ function EditorSettings({ store, signals }) {
             {...bindValue('vimMode')}
           />
           <SubDescription>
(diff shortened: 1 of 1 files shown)


CompuIves:
Yep, that does make most sense. Could you fix the lint issue before I merge it? I believe the `'` needs to become `{"'"}` for escaping.

frontsideair:
Done.

CompuIves:
Perfect! Thank you very much, Docker is a known issue (fixed on master 😄). I'll merge this in now!

PR #724, opened 2018-04-12
Fix Select background arrows

donysukardi:
I remember it was fine all along. Can't find any recent changes that might have caused the background image to err either.

Fix Select's arrow background image

image

**What is the new behavior?**

image

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 


CompuIves:
Oh wow I can't believe that Prettify did this, thanks for fixing it!

PR #725, opened 2018-04-12
Omitting homepage value for now deployments

mdugue:
Bug Fix

https://github.com/CompuIves/codesandbox-client/issues/680

**What is the new behavior?**
`homepage` is removed from apiData.package when deploying to `now`

**Checklist**:

- [x] Documentation (kindof)
- [ ] Tests
- [x] Ready to be merged (see comments below)

I'm a bit unsure about the following:

- I wanted it to keep it as simple as possible, so I skipped any check for _is value "https://codesandbox.io/s/new"?_ Any case I missed where you would need to provide the homepage value
- I saw you use `lodash` but not `fp`, still decided for the `fp` way as I thought it would be better readable. Are you fine with that or do you prefer it "non-fp"?
- I could not really e2e-test it, just some console.logging. hope it is still fine 😬


mdugue:
strange, the diffs seem to point to some babel / npm issues… @CompuIves any hint?

CompuIves:
Perfect! The diff looks good, I'll merge this in now. Thank you!

mdugue:
You are welcome! Glad to contribute (even though just a 2 liner 😀), as I personally benefited a lot from your great work… Should we close https://github.com/CompuIves/codesandbox-client/issues/680 ?

PR #766, opened 2018-04-28
Show descriptive error message on file too large upload error

### LINKED ISSUE #759: File Upload "Entity Too Big"
Just wondering about upload filesize limits after this tweet by the all mighty @CompuIves last night.

https://twitter.com/CompuIves/status/988899843656568833

I've confirmed that the filesize limit is larger than the 1mb that I originally thought, but I have tested and found that I can not upload a 4.2mb file, although a 3.9mb file works just fine. I have tested these files with both drag and drop as well as the "upload selection" feature. (I don't know what to call the normal way of uploading files...)

I am assuming that there are variables at play when uploading a file that can... cause the final size to be different? >.> I don't know. I'm not very good with technology. 🤷‍♂️ 

screen shot 2018-04-25 at 1 11 47 pm

lbogdan:
Fixes #759 .

### DIFF
--- a/packages/app/src/app/store/modules/files/actions.js
+++ b/packages/app/src/app/store/modules/files/actions.js
@@ -135,6 +135,9 @@ export async function uploadFiles({ api, props, path }) {
     });
   } catch (error) {
     console.error(error);
+    if (error.message.indexOf('413') !== -1) {
+      return path.discardError();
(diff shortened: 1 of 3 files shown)


CompuIves:
Perfect!

PR #769, opened 2018-04-28
Fix vue-cli @ alias

lbogdan:
This was pretty much broken from the get-go, strangely enough no one noticed. :smile: As seen here https://github.com/vuejs-templates/webpack/blob/develop/template/build/webpack.base.conf.js#L40 , `@` should point to `/src`, not `/`.


CompuIves:
Yep, this is a long standing issue. The scary part is that I don't know which sandboxes will break from this change.

I just checked the most popular sandboxes and they're not using it, so let's just merge this in and fix sandboxes that get reported.

PR #780, opened 2018-05-01
Scripts spring cleanup

lbogdan:
### Changes:
* `app` and `homepage` only depend on `codesandbox-api` and `codesandbox-browserfs`; added `build:deps` to only build those 2 dependencies
* added `dev` - builds deps and starts `app` dev env
* added `dev:fast` - only starts `app` dev env
* added `dev:home` - only starts `homepage` dev env
* hid `(!) 'this' has been rewritten to 'undefined'` rollup warnings
* removed unneeded `extract-text-webpack-plugin` patching from homepage build
* upgraded `uglifyjs-webpack-plugin` to 1.2.5 - fixes build hanging on Windows
* updated `CONTRIBUTING.md`
* updated CircleCI config; fixed error when docker credentials not available


packages/homepage/package.json
```diff
@@ -37,8 +37,7 @@
   "license": "MIT",
   "main": "n/a",
   "scripts": {
-    "patch": "sed -i -e 's/if(c\\.initial) return;/if(!c||c.initial) return;/g' ../../node_modules/extract-text-webpack-plugin/index.js",
```
CompuIves: Hmm, maybe we need this one still.

CONTRIBUTING.md
```diff
@@ -6,10 +6,11 @@ The CodeSandbox client is currently divided in to 5 parts. We use `lerna` to sha
 
 - `app`: The editor, the search, profile page, the embed and the sandbox.
 - `homepage`: The Gatsby website of the homepage.
-- `codesandbox-api`: The npm package that's responsible for communication between the sandbox and the editor.
 - `common`: All common parts between these packages, reusable JS.
+- `codesandbox-api`: The npm package that's responsible for communication between the sandbox and the editor.
+- `codesandbox-browserfs`: An in-browser file system that emulates the Node JS file system API and supports storing and retrieving files from various backends. Forked from [https://github.com/jvilk/BrowserFS](https://github.com/jvilk/BrowserFS), with an additional [CodeSandbox backend](https://github.com/CompuIves/codesandbox-client/blob/master/standalone-packages/codesandbox-browserfs/src/backend/CodeSandboxFS.ts).
```
CompuIves: I love this explanation!

.circleci/config.yml
```diff
@@ -78,10 +78,6 @@ jobs:
           name: Start Test Server
           command: yarn start:test
           background: true
-      - run:
```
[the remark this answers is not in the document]
lbogdan: Were we (supposed to still be) using BrowserStack? Looking though CircleCI logs, this was always failing.

CONTRIBUTING.md
```diff
@@ -6,10 +6,11 @@ The CodeSandbox client is currently divided in to 5 parts. We use `lerna` to sha
...
 
-This version of CodeSandbox is using the production server as source of truth, this is specified by the environment variable `LOCAL_SERVER`. It's not yet possible to sign in using this version, I haven't figured out how to handle this yet.
+This version of CodeSandbox is using the production server as source of truth, this is specified by the environment variable `LOCAL_SERVER`. If you're working on a feature that needs you to be logged in, you can login on [https://codesandbox.io/](https://codesandbox.io/) and copy the contents of the `jwt` local storage key over to your development environment on [http://localhost:3000/](http://localhost:3000/). **Be very careful with how you handle the token**, as anyone who knows it can login as you and have read/write access to all your CodeSandbox content!
```
CompuIves: I'm not sure if we could recommend copying `jwt` token as an official way, since it feels VERY insecure. It could be safe though, I mean the JWT is sent insecure to localhost, but it's localhost, and then sent over https. I'm not sure how to feel about this, what do you think?

lbogdan: I had the same concerns and came to the same conclusions. It's still opt-in, would probably only be used by people that work on features that require to be logged-in, and there's also the warning.

packages/homepage/package.json
lbogdan: From what I've found, this was fixing an old `extract-text-webpack-plugin` issue which got fixed since.

package.json
```diff
@@ -4,13 +4,17 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "build": "yarn build:dependents && yarn build:prod",
+    "build": "yarn build:deps && yarn build:prod",
     "build:prod": "lerna run build --scope app --scope homepage --parallel && gulp",
     "build:dependents": "lerna run build --scope codesandbox-api --scope common --scope codesandbox-browserfs --stream && lerna run build --scope smooshpack --stream && lerna run build --scope react-smooshpack --stream",
+    "build:deps": "lerna run build:dev --scope codesandbox-{api,browserfs}",
     "start": "lerna run build --scope codesandbox-browserfs --stream && lerna run start --parallel",
     "start:fast": "lerna run start --scope app --stream",
```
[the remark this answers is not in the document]
lbogdan: You mean `build:dependents`?

.circleci/config.yml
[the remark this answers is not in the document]
CompuIves: Hmm, this one only works in master I believe, because it uses env variables. They are there to make sure that CodeSandbox works eg. in IE11. I think it'd be good to keep it, but only on master. Very strange that the build runs though, wasn't expecting the tests to work... Weird

package.json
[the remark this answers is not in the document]
CompuIves: Oh yes! And `start:fast` probably.

packages/homepage/package.json
CompuIves: Okay nice! In that case let's remove it!

CONTRIBUTING.md
CompuIves: Agreed, okay let's keep it then. Maybe we can change the dev environment in the near future to work on codesandbox.stream api, feels much safer. Then we should probably add a header in the proxy so nginx knows not to redirect.

package.json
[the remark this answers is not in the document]
lbogdan: Hm, I'm still wondering if we shouldn't keep `start[:something]` instead of `dev[:something]` tbh.

PR #804, opened 2018-05-06
Add support for Dojo theming

nicknisi:
**What kind of change does this PR introduce?** Bug fix

**What is the current behavior?** Theming in Dojo examples does not work properly as the CSS Module JS file is missing a `" _key"` property that the Dojo theming system relies on.

**What is the new behavior?**

For Dojo themes to work properly a special `" _key"` property should exist on the CSS module JS file that's value is the package name plus the name of the css file.

This commit makes the following changes:

- add a custom styles transpiler for Dojo that adds the `" _key"` property
- Update the Dojo preset to use the new styles transpiler
- split out the `classesToDefinition` method into its own utility module and import it into the existing and Dojo style transpilers

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 


CompuIves:
Perfect! 👍

Issue #572, opened 2018-02-20
Sub-ideal first experience with a React typescript app


altano:
I tried working on a new project in code sandbox, which is super awesome in so many ways, but ended up having a bad experience. My sandbox is here: https://codesandbox.io/s/2zp6rqpvp

1. emmet expansion + ctrl+z takes you back further in your edit history than right before the emmet expansion.
1. My anchors were not working and I was going crazy trying to figure out why only to try and reproduce outside codesandbox and it just *didn't* reproduce. For some reason something about codesandbox is interfering with the anchors in the table of contents linking to the headers below. This is a dealbreaker for me because my script I'm writing entirely revolves around this working.
1. I tried to download my sandbox but I couldn't get the downloaded version to work.
    1. A bunch of the @types in the package.json were not valid npm packages.
    1. My package.json was invalid because codesandbox didn't prevent me from using a name with a value of "Name Like This". Converting it to "a-name-like-this" made the package work.
    1. The package didn't include basic dependencies like react-scripts-ts which are used by the scripts
    1. I needed to manually create tslint.json and tsconfig.json as they didn't come with the download
    1. Annnnd then I finally gave up when, after fixing the above, I hit this error:
```
Failed to compile.

./src/index.tsx
Module parse failed: Unexpected token (22:32)
You may need an appropriate loader to handle this file type.
|     textAlign: "center"
| };
| var App = function () { return (<div style={styles}>
|     <TableOfContents_1.default />
|     <Article_1.default />
```

I can tell codesandbox is 99% of the way to something amazing that I would have really enjoyed using, but this stuff got in my way. Just thought I'd leave a note saying as much. Thanks!

CompuIves:
Wow @altano, thank you so much for taking the time to write all this out!

I'll make sure to address your feedback this week, this is really valuable and clear information. Also, thanks for showing the sandbox that's having this!

CompuIves:
Okay, I committed fixes for:

3.i (https://github.com/CompuIves/codesandbox-client/commit/403f82b432b7984c13b87a2b314e1f42ecc58df8)
3.ii (https://github.com/CompuIves/codesandbox-client/commit/f2a01f163cebfe44b70d6e71fa4090431e782f79)
3.iii (changed defaults for the default sandbox `react-ts`)

I will allow configuration of `tsconfig.json` in CodeSandbox tomorrow, this will fix 3.iv and 3.v with that hopefully.

[...]

PR #810, opened 2018-05-08
Fix vue-router <router-link> in hash mode

### LINKED ISSUE #595: Hash routes not working
https://codesandbox.io/s/pk8807zz0

lbogdan:
Fixes #595 .

Test: http://pr810.cs.lbogdan.tk/s/pk8807zz0 .

### DIFF
--- a/packages/app/src/sandbox/url-listeners.js
+++ b/packages/app/src/sandbox/url-listeners.js
@@ -100,6 +100,7 @@ export default function setupHistoryListeners() {
         const el = ev.target;
         if (
           el.nodeName === 'A' &&
+          !el.__vue__ && // workaround for vue-router <router-link>
           el.href.indexOf('#') !== -1 &&
           el.href.substr(-1) !== '#'
         ) {


CompuIves:
Nice! In general it would be nice if we could find a way to make hash links work, I remember having similar issues with react router in hash and I believe it was mentioned here too: #572.

lbogdan:
`react-router` already worked in hash mode, see http://pr810.cs.lbogdan.tk/s/2wow48xvnj . And #572 was about not "scrolling" the page to the element with the id equal to the link hash when clicking on a hash link (which I'll look into shortly).

CompuIves:
Ahh, that makes sense. Thanks for the clarification and the PR 😄

lbogdan:
To go a bit into detail, `react-router` it's fully based on `hashchange` events, while `vue-router` also listens for `click` events on hash links, which we're hijacking, so that's why we only need the workaround for `vue-router`.

PR #812, opened 2018-05-08
Fix update search state when navigating back / forward

### LINKED ISSUE #469: Back and forward on search are not updating the filters or current page
Hi @CompuIves, there appears to be an issue with the codesandbox search.

If I select or modify some filters (say select react, vue, ...) then change the page number the state is getting pushed to the history but on going back or forward (using the browser back/forward navigation) the filters are not updating.

Replicating
1. Go to https://codesandbox.io/search
2. Select any page number
3. Try going back. The history gets updated but the page number doesn't.

- Same can be replicated using any of the filters.

Let me know if I can help further with this issue. 😄

lbogdan:
Fixes #469 .

### DIFF
--- a/packages/app/src/app/pages/Search/index.js
+++ b/packages/app/src/app/pages/Search/index.js
@@ -46,7 +46,7 @@ class Search extends React.PureComponent {
     };
 
     this.unlisten = this.props.history.listen((location, action) => {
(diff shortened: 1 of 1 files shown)


CompuIves:
Great @lbogdan!

Issue #761, opened 2018-04-26
Ctrl + Tab to switch between two files


GingerBear:
Thanks for creating such a great tool that makes programming much easier and accessible.

One of my pain point of writing code on codesandbox is that, switching between two files is not very easy using keyboard shortcuts (like Ctrl + Tab on most editor). I need to use mouse to select the other files, or use cmd + p and type the file name.

CompuIves:
Good one! I like this addition.

grady-lad:
@CompuIves I can take a look at this one if it is still available?

CompuIves:
That'd be great @grady-lad! It's still available.

We store keybindings here: https://github.com/CompuIves/codesandbox-client/blob/master/packages/common/utils/keybindings.js. These keybindings fire signals. The signals that are important are defined here: https://github.com/CompuIves/codesandbox-client/blob/master/packages/app/src/app/store/modules/editor/sequences.js#L29-L36. 

Let me know if you have any questions along the way 😄

lbogdan:
Maybe not Ctrl-Tab, though, as that's usually the native "next tab" browser keyboard shortcut, and it might not be over-writable across browsers.

steppefox:
@SaraVieira does your comment mean that this issue could be closed too?

https://github.com/codesandbox/codesandbox-client/pull/817#issuecomment-478547908

SaraVieira:
Thank you @steppefox 

Yes, we changed to VSCode so all our keybindings now come from VSCode 🗡 
https://code.visualstudio.com/docs/getstarted/keybindings

PR #817, opened 2018-05-09
[Feature] WIP: Keybindings added to navigate between tabs.

grady-lad:
This feature introduces two new keybindings:
`Switch To Next Tab` & `Switch To Previous Tab`

switchtabdemo

See #761 

**What is the new behavior?**

`Switch To Next Tab:` Keybinding allowing users to create a shortcut to switch to the next tab

`Switch To Previous Tab:` Keybinding allowing users to create a shortcut to switch to the previous tab

**Checklist**:

- [ ] Documentation __N/A__
- [ ] Tests __N/A__
- [ ] Ready to be merged __N/A__ 
- [ ] Added myself to contributors table __N/A__ 

**Additional comments**
I have introduced the functionality to switch between tabs in both directions. When testing the PR it seems I ran into a few _snags:_

1) I found it hard to find appropriate default keybinding values that worked, as some were taken by the browser. Open to suggestions here as to what they should be and if they should be able to overwrite browser defaults?

2) Also in terms of usability if a keybinding was created with multiple keys eg. `Control + S` after the first execution I would expect the keybinding to still execute if the user held `Control` and repeatedly pressed `S`, currently this is not the situation.  To switch to the next tab the user has to repeatedly press `Control + S`

**Please Note**:  _The default keybindings in the PR are not final they were just used for testing purposes and can be changed to more appropriate keys_

### DIFF
--- a/packages/app/src/app/store/modules/editor/actions.js
+++ b/packages/app/src/app/store/modules/editor/actions.js
@@ -370,3 +370,24 @@ export function setPreviewBounds({ props, state }) {
     state.set(`editor.previewWindow.height`, props.height);
   }
 }
+
+export function switchToTab({ state, props }) {
+  const tabs = state.get('editor.tabs');
+
+  const currentModuleShortid = state.get(`editor.currentModuleShortid`);
+  const currentTabIndex = tabs.findIndex(
+    t => t.moduleShortid === currentModuleShortid
+  );
+
+  const switchToTabIndex = (tabLength, currentIdx) => {
+    if (props.shouldSwitchNext) {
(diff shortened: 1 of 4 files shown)


bora89:
Yeah, it would be super cool if there will be any shortcuts for tabs navigation. When aprox it will be merged?

bora89:
Consider this shortcuts: CTRL + J, CTRL + K, very handy in my opition

SaraVieira:
Hey!

Sorry for the quiet time and for I am about to say but due to us changing to VSCode this is now handled by VSCode so we don't need that extra logic


Sorry again and thank you for taking the time to contribute

PR #818, opened 2018-05-09
[WIP] Sandbox service worker changes

lbogdan:


CompuIves:
This looks good to me! I'm thinking that we should change it in the future for sandpack, in that case we need to also cache unused transpilers. For example when someone adds a `.scss` project to the compiler it should be able to load that transpiler in offline context. But that's one for the next iteration 😄

PR #853, opened 2018-05-23
Compiler updates

### LINKED ISSUE #467: Alias support?
Is there a way to define alias ? Like webpack config:

image

CompuIves:
1. Better performance for path resolving
2. Clearer error messages for dependency not found
3. We now skip the `browser` field for `babel-core` and `@babel/core`, since they just _work_ in CodeSandbox
4. Put `getAliasedPath` in a common flow, so it doesn't have to be called multiple times
5. Add `main` field compatibility for the parcel template
6. Add `alias` field support for parcel
7. Support for node modules node-like resolving
8. `require.resolve` support

Fixes #467


valtido:
Needs a bug fix for auto imports to use alias instead of `../.../`

bb:
Hi there, is the aliasing issue regarding https://codesandbox.io/s/material-demo-7v6qq mentioned in preactjs/preact#1850 the same as #1374 or a different issue? At least looking at the error message, I think it's different.

Shouldn't that be fixed within this PR? What am I missing?

jyjin:
So? Where is an example for alias setting...

PR #891, opened 2018-06-07
fix(signout): Fixing UI updates after sign out

### LINKED ISSUE #852: After I got out of your profile, then I need to refresh the page to return to the main menu
# 🐛 bug report

## Description of the problem
After I got out of your profile, then I need to refresh the page to return to the main menu. Otherwise I just have a black screen

arthurdenner:
Bug fix. Closes #852.

After signing out, the UI breaks because the `UserMenu` component is connected to the store and reacts to the `user` property being updated. (See #852)

**What is the new behavior?**
Instead of being connected to the store, the `UserMenu` component receives the `user` and `userMenuOpen` props from the `Navigation` component, which also decides if the `UserMenu` should render or not based on the `user` property.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/packages/app/src/app/pages/common/Navigation/index.js
+++ b/packages/app/src/app/pages/common/Navigation/index.js
@@ -14,7 +14,7 @@ import UserMenu from '../UserMenu';
 import { LogoWithBorder, Border, Title, Actions, Action } from './elements';
 
 function Navigation({ signals, store, title }) {
-  const { user, isPatron } = store;
+  const { user, userMenuOpen, isPatron } = store;
 
   return (
     <Row justifyContent="space-between">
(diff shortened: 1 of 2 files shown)


CompuIves:
This is really nice @arthurdenner! Thank you so much for building this 😄. Merging it in now!

arthurdenner:
@CompuIves, sorry for bothering you, but I saw that you reverted this PR and now the sign out is broken again. Also, is there any way to sign in with GitHub during development besides copying the `jwt` token?

arthurdenner:
Sorry, I didn't notice this at all.

arthurdenner:
@CompuIves, I looked at this problem again and noticed that if we use the `isLoggedIn` property instead of the `user` property to compute what to render in the [`Navigation` component](https://github.com/CompuIves/codesandbox-client/blob/master/packages/app/src/app/pages/common/Navigation/index.js#L54), the UI doesn't break (and nothing else needs to be changed).

Can you confirm this?

PR #898, opened 2018-06-12
Recover wave animation after tabs switch

jeezliu:
After switch tabs wave animations will stop. Reset variable lowPerf to false to recover it when tab regain focus.


CompuIves:
Good one! Merging in now, thank you 😄

jeezliu:
😄😄

PR #900, opened 2018-06-13
Fix sandpack updatePreview docs

iamphill:

docs update

Throws an error because the options are passed incorrectly with the example code.

**What is the new behavior?**

Fixes the docs so that the example code is correct.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 


CompuIves:
Ahh, perfect. Thanks for the contribution!

PR #902, opened 2018-06-14
Code search

xtuc:
## Why

We want to allow an user to search across its codebase. We use Algolia because it's fast.

## Usage

Search matches are highlighted in white.

### Searching code

search-code

### Searching a file

search-path

### Keyboard navigation

To open a result you can click on it but also use the directional arrows and press ENTER.

keyboard

### Searchable attributes

Ordered by importance:
  - code
  - path

We prefer searching in the code first. I believe that searching the path won't be as common.

### Faceting

The index is segmented by sandboxId, which is configured as a facet. This could be disabled if we want to search in the code of all the sandboxes at the same time.

### Display & pagination

Hits par page: 10 because more result won't be visible to the user anyway.

Unretrievable attributes: we can omit sandboxId because we already have the info on the client

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 


CompuIves:
Ahh, I'm sorry for not reviewing this properly. I was holding this off for a while and forgot about it, that's my mistake. I do think we need to have search in files for CodeSandbox, but with the current version I would opt to do this using the existing VSCode architecture that we have right now with web workers. I really appereciate the effort you put into this PR, thanks a lot for taking the time to implement this!

vikrantsingh47:
is there any plan to add this feature?

CompuIves:
Yes, nowadays we have this feature. In the sidebar under files you can trigger "Search All".

vikrantsingh47:
awesome.

solarkraft:
For some reason the search tab on the left side only became visible for me (in Firefox) after I joined the same session with another browser (Chrome).

AlexisN94:
I can't see the search feature on Chrome. Am I missing something?

PR #906, opened 2018-06-14
fix(signout): Fixing UI updates after sign out

### LINKED ISSUE #852: After I got out of your profile, then I need to refresh the page to return to the main menu

arthurdenner:
This PR tries to ~really~ fix #852.

After signing out, the UI breaks.

**What is the new behavior?**
After failing in fixing this in #891, I investigated this problem and as weird as it sounds, using the `store.isLoggedIn` instead of `store.user` to compute if the `UserMenu` should render in the `Navigation` component fixes the problem. Please, confirm this before deploying in case I'm wrong again. :see_no_evil: 

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 

### DIFF
--- a/packages/app/src/app/pages/common/Navigation/index.js
+++ b/packages/app/src/app/pages/common/Navigation/index.js
@@ -14,7 +14,7 @@ import UserMenu from '../UserMenu';
 import { LogoWithBorder, Border, Title, Actions, Action } from './elements';
 
 function Navigation({ signals, store, title }) {
-  const { user, isPatron } = store;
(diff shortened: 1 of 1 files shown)


CompuIves:
Ha @arthurdenner, that's perfect, I think you're right with that!

I'll merge this in and will test it on staging before deploy, thanks for the contribution 😄.

PR #913, opened 2018-06-16
fix(EntryTitleInput): in event onblur if force is true it calls reset…

### LINKED ISSUE #912: Rename file or folder is broken when you dont cancel with ESC
# 🐛 bug report

## Description of the problem
Rename file or folder is broken when you dont cancel with ESC

captura de tela de 2018-06-16 18-19-15

### Your Environment
| Browser |  Chrome |
| Operating System | Linux |

emersonlaurentino:
…State

Bux fix. Closes #912.

#912

**What is the new behavior?**
when the user is going to rename a folder or a file and then clicks off the input, onBlur is called with the force parameter true and then resets the edit state.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 


CompuIves:
This is great @emersonlaurentino! Thanks for the fix.

PR #927, opened 2018-06-25
add typescript files for Vue projects

ksnyde:
as discussed on Discord ... this will allow bare `.ts` files to be transpiled instead of just relying on TS embedded in .vue files


CompuIves:
Perfect! Thanks for the contribution, merging in now...

PR #933, opened 2018-06-28
refactor styled-components extend style

franklinjavier:
You should only use Comp.extend if you know that Comp is a styled component. If you're importing from another file or a third party library, prefer to use styled(Comp) as it accomplishes the same thing but works with any React component. 

**Behind the scenes**

The styled() factory generates new component styles with a new class. The classnames are then passed to the React component via the className prop. Calling extend creates new component styles by extending the old one, and thus doesn't generate two classes for a single component. (styled() factory does that)


viankakrisna:
i think `.extend` is on the way to be deprecated on 4.0 
https://github.com/styled-components/styled-components/issues/1546
https://github.com/styled-components/styled-components/issues/1694

franklinjavier:
Sad to hear. I will close the PR then.

PR #972, opened 2018-07-17
Theme fixes & added 2 new themes

lbogdan:
Theme fixes:
- ~~use [strip-json-comments](https://www.npmjs.com/package/strip-json-comments) to... well, strip JSON comments 😃~~
- also strip comments when loading theme from an URL

Themes added:
- Solarized Light
- Cobalt 2

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -205,6 +205,7 @@
     "shortid": "^2.2.8",
     "store": "^2.0.12",
     "string-replace-loader": "^2.1.1",
+    "strip-json-comments": "^2.0.1",
     "styled-components": "^3.2.1",
     "svg-react-loader": "^0.4.4",
     "tern": "^0.21.0",
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -205,7 +205,6 @@
     "shortid": "^2.2.8",
     "store": "^2.0.12",
     "string-replace-loader": "^2.1.1",
-    "strip-json-comments": "^2.0.1",
     "styled-components": "^3.2.1",
     "svg-react-loader": "^0.4.4",
     "tern": "^0.21.0",
--- a/packages/app/src/app/pages/Sandbox/Editor/utils/get-vscode-theme.js
+++ b/packages/app/src/app/pages/Sandbox/Editor/utils/get-vscode-theme.js
@@ -1,4 +1,5 @@
 import JSON from 'json5';
+import stripJsonComments from 'strip-json-comments';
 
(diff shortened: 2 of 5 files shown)


packages/app/package.json
```diff
@@ -205,6 +205,7 @@
     "shortid": "^2.2.8",
     "store": "^2.0.12",
     "string-replace-loader": "^2.1.1",
+    "strip-json-comments": "^2.0.1",
```
CompuIves: I don't think we need this as we already use JSON5 to parse (https://json5.org/). The reason I did that RegEx was because I wanted the default colors which are commented out to be used as well, as we don't have those default colors built in.

So if you have:

```
//"editorActiveLineNumber.foreground": "#0b216f",
```

I want it to become:

```
"editorActiveLineNumber.foreground": "#0b216f",
```

But only for those kind of comments, not for comments that explain something. 

Maybe it works with these colors commented though, is worth testing.

packages/app/src/app/pages/Sandbox/Editor/utils/get-vscode-theme.js
```diff
@@ -43,13 +44,31 @@ function fetchTheme(foundTheme) {
     return foundTheme.get();
   }
 
-  return window.fetch(foundTheme.url).then(x => x.json());
```
CompuIves: Ahhh that makes a lot of sense!

packages/app/package.json
lbogdan: Yeah, seems like I completely misunderstood what that code did. 😃

Still, shouldn't that be
```js
replace(/^\s*\/\/"/gm, '"')
```
or is JSON5 resilient in missing quotes in property names?

CompuIves: You're right! That's what it should be, big mistake on my side 😅.

packages/common/themes/index.js
```diff
@@ -41,4 +41,16 @@ export default [
...
       'https://cdn.rawgit.com/akamud/vscode-theme-onelight/master/themes/OneLight.json',
   },
+  {
+    name: 'Solarized Light',
+    id: 'solarizedLight',
+    url:
+      'https://raw.githubusercontent.com/Microsoft/vscode/d2b6bbb46fbdf535e2c96b3e00ac56ac1d427a69/extensions/theme-solarized-light/themes/solarized-light-color-theme.json',
+  },
+  {
+    name: 'Cobalt 2',
```
CompuIves: Could you move this one above `VSCode Light`? That way we keep the Dark -> Light order 😄.

PR #997, opened 2018-07-21
Fix Contibutors Duplication

satyarohith:
Fixes duplication of contributors in `README.md` file

Extra rows of same contributors are created

**What is the new behavior?**
Fixes duplication.


CompuIves:
Oh thanks! Must've been something from my bad merging efforts hahaha.

PR #999, opened 2018-07-23
[WIP] Connect Emmet to Monaco

### LINKED ISSUE #194: Add emmet support

satyarohith:
Enables emmet support 

Emmet is not connected to the editor

**What is the new behavior?**
We have Emmet support as before :smile: 

closes #194

### DIFF
--- a/packages/app/src/app/components/CodeEditor/Monaco/enable-combi-emmet.js
+++ b/packages/app/src/app/components/CodeEditor/Monaco/enable-combi-emmet.js
@@ -1,7 +1,7 @@
 import extractAbbreviation from '@emmetio/extract-abbreviation';
 import { expand } from '@emmetio/expand-abbreviation';
 
-const field = () => '';
+const field = () => '${0}';
 
 const expandAbbreviation = (source, language) =>
   expand(source.abbreviation, {
@@ -48,7 +48,13 @@ const enableEmmet = (editor, monaco) => {
       }
       if (word) {
         // Get expand text
-        const expandText = expandAbbreviation(word, 'html');
+        let expandText = expandAbbreviation(word, 'html');
+        const posOffsetArr = expandText.split('${0}')[0].split('\n');
+        const lineNumber = pos.lineNumber + posOffsetArr.length - 1;
+        const column =
+          word.location + posOffsetArr[posOffsetArr.length - 1].length + 1;
+
+        expandText = expandText.replace(/\$\{0\}/g, '');
         if (expandText) {
           // replace range content: pos.column , pos.column -word.length;
           const range = new monaco.Range(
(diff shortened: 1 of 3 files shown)


satyarohith:
Oops! Did not notice this error. I learned to open `console`after every modification :smile:, I will get back after I implement `tokenize`.

satyarohith:
Hi @CompuIves, I'm quite confused about how stuff is working right now.
We have two files `enable-combi-emmet.js` and `enable-emmet.js`.
The text expansion and edits in the editor are done via `enableCombiEmmet(editor, monaco);`
but we also have some logic in `enable-emmet.js`.  I couldn't figure out what exactly is it doing?
It seems like it sets the position of the cursor and also attempts to edit the editor according to emmet, I might be wrong here.

Coming to `tokenize` I implemented this way but I was getting error which I was unable to figure out and also cursor positioning does not work.
```js
const token = monaco.editor.tokenize(
   model.getLineContent(lineNumber),
   model._languageIdentifier.id
)[0];
```
loging the token without [0] I got this
screenshot from 2018-07-24 14-46-07

and the error after implementing the above
screenshot from 2018-07-24 14-32-36

If I'm not wrong,The main problem according to what I know is we are unable to place the cursor according to the expansion of elements.
Thanks for taking time to read this Ives, point me where I'm doing wrong.

CompuIves:
Ahh, yes it's quite a mess. We have two implementations of Emmet, one of those was the initial implementation and the other one was added later because the first one broke if I remember correctly. It seems like the error is related to a bad call within Monaco, which possible could be because some services in Monaco are mocked to work in the browser.

I think the main focus is to make `enable-combi-emmet.js` work first, I believe that one works the best. Does that one throw any errors?

CompuIves:
That makes a lot of sense @satyarohith. Great to hear that it works! I think it can be done with something `editor.cursor.setPosition`.

satyarohith:
Hi @CompuIves I got cursor placement working but when expanding multiple elements they are not formatted as expected.
singleelement
multiele
I think the culprit might be `expandText`.
Do you have any Idea about this?

abhinayar:
Hey guys,

Seems to still be an issue! Not sure what the status is on this, is a huge thing for me if I want to move my workflow to CodeSandbox. Let me know if there's anything I can do to help!

Best,
Abhi

SaraVieira:
@CompuIves @satyarohith Same, what is left so this can be merged? If not a lot can we merge into some other branch so other people can work on it ?

satyarohith:
I'm sorry for not being active.
Indentation and cursor placement for multiple expansions don't work
properly.
I'm quite busy now due to exams so would love to know if anyone is
interested to take a look.
Thanks for bearing with me.

CompuIves:
I implemented Emmet directly in our language server now, it should feel very smooth 
screen shot 2018-10-23 at 03 52 42.

Thanks for the work @satyarohith, I appreciate the research behind it! Good luck with the exams!

PR #1013, opened 2018-07-27
fix(genie): allow genie to sort wishes properly

kentcdodds:
One of the cool things about genie is it's sorting algorithm takes into account the user's previous choices. By providing the input value when you make the wish, genie can keep track of what input values lead to which wishes and can optimistically suggest certain wishes as the user's typing.

**What kind of change does this PR introduce?** Bug Fix

**What is the current behavior?** genie isn't able to make proper suggestions based on previous wishes made

**What is the new behavior?** genie will optimistically suggest wishes

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests
- [x] Ready to be merged 
- [ ] Added myself to contributors table  I should probably do this... 😅 But feel free to merge before I get the chance.

### DIFF
--- a/packages/app/src/app/pages/Sandbox/QuickActions/index.js
+++ b/packages/app/src/app/pages/Sandbox/QuickActions/index.js
@@ -78,8 +78,15 @@ class QuickActions extends React.Component {
   }
 
   loadGenie() {
-    const { enteredMagicWords } = JSON.parse(window.localStorage.getItem('genie'))
-    genie.options({ enteredMagicWords })
+    try {    
+      const { enteredMagicWords } = JSON.parse(window.localStorage.getItem('genie'))
+      genie.options({ enteredMagicWords })
+    } catch (error) {
+      // it may not exist in localStorage yet, or the JSON was malformed somehow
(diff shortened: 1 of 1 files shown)


kentcdodds:
Ah, I've worked it out. The `inputValue` is set to the selected item in `onChange`. We'll have to keep track of the input value separately so we know what the inputValue is at the time of the selection.

kentcdodds:
Sweet! It's working!

Next step will be to save genie's options to localstorage whenever a wish is made and then rehydrate genie when the component is mounted...

kentcdodds:
After this the next thing we'll need to do is figure out why the <kbd>⌘</kbd> + <kbd>p</kbd> keyboard shortcut doesn't seem to open the menu most the time...

kentcdodds:
Sweet, this thing's totally working! So now genie will be able to learn the user's preferred wishes as they type, making the menu more helpful.

This is ready to be merged! There's another linting issue in another file.

CompuIves:
Woooooooow this is greaaaat! Thank you so much @kentcdodds! Merging this in now. Such an improvement!

PR #1022, opened 2018-08-02
Fix async issue with fetching typings from meta Fixes #1021

### LINKED ISSUE #1021: Inconsistent Typings Downloads
# 🐛 bug report

## Description of the problem

When trying to use a project, such as a Dojo project, where the types are inlined, the Editor isn't always consistent when it comes to identifying types, even if the code does run.
screen shot 2018-08-01 at 22 07 24

This causes the imports to be underlined in red as if the types are not found

### Your Environment
| Сodesandbox  | latest |
| Browser | any |
| Operating System | macOS |

nicknisi:
Ensure that all fetch Promises have been resolved when returning from the fetchFromMeta method in fetch-dependency-typings.

Currently, it causes inconsistencies when it comes to fetching inline typings as the promise can resolve before all of the types are downloaded.

**What is the new behavior?**
The new fix causes the `fetchFromMeta` to return a `Promise.all` of its fetch calls to ensure its returned Promise doesn't resolve until all of them do.

**Checklist**:

- [ ] Documentation (NA)
- [ ] Tests (skipped as I didn't see existing tests for this file
- [x] Ready to be merged 

### DIFF
--- a/packages/app/src/app/components/CodeEditor/Monaco/workers/fetch-dependency-typings.js
+++ b/packages/app/src/app/components/CodeEditor/Monaco/workers/fetch-dependency-typings.js
@@ -191,13 +191,13 @@ function fetchFromMeta(dependency, version, fetchedPaths) {
         throw new Error('No inline typings found.');
       }
 
-      dtsFiles.forEach(file => {
+      return Promise.all(dtsFiles.map(file =>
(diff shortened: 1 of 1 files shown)


CompuIves:
Ah so that's where that came from!! Thanks for resolving this!

PR #1059, opened 2018-08-28
Support for Vue 3

CompuIves:
This is the initial implementation of Vue 3 support. The main thing to support is the new Vue 3 babel preset. We now support it by executing the preset in the babel worker, which can make initial execution slower (we need to still evaluate and interpret the preset). But vue v3 works!


CompuIves:
Test it here: http://pr1059.cs.lbogdan.tk/s/k243wv9v77 (cc @Akryum)

packages/app/src/sandbox/eval/transpilers/babel/worker/evaluate.js
```diff
@@ -1,13 +1,16 @@
 import resolve from 'browser-resolve';
```
CompuIves: I had to rewrite some parts in the babel evaluator. It turns out that we didn't cache according to node spec! We do now, which makes the babel vue preset work.

Akryum:
Awesome! :heart_eyes_cat: Then wee need a way to install and invoke vue-cli plugins! :smile_cat:

Issue #1075, opened 2018-09-04
 CRA App With Jest / React Testing Library Refreshes Constantly 


stolinski:
# 🐛 bug report

## Description of the problem

A pen that was originally created via CRA and uploaded to Sandbox, refreshes constantly. Refreshes stop when tests are removed.

## How has this issue affected you? What are you trying to accomplish?

This app is to teach Jest & React Testing Library. This works 100% off CodeSandbox as expected. Originally created in CRA, constant refresh stops if tests are removed.

### Your Environment
| Software         | Name/Version|
| ---------------- | ---------- |
| Сodesandbox  | PROD-1536082527-76b9252 |
| Browser | Chrome(68)/Firefox |
| Operating System | MacOS High Sierra |

arthurdenner:
Hi @stolinski! 👋 Could you verify if this is still happening? I tried to reproduce but I couldn't. If it's not, could you close the issue, please?

lbogdan:
Closing this, as it's way old, there were lots of changes in the meanwhile, and can't replicate in the OP sandbox. @stolinski Please re-open if this is still an issue for you.

Issue #1074, opened 2018-09-04
GraphQL Query Never Resolves When Jest Test Is Present


stolinski:
# 🐛 bug report

## Description of the problem

This is a weird one. I have a basic Apollo query that doesn't resolve on page load, it just sits in the loading state. @CompuIves Noticed in an email chain that the query successfully resolves when the default CRA test is removed. 

## How has this issue affected you? What are you trying to accomplish?

This is a CRA project used to show how to write Apollo queries and mutations. The code was imported into CodeSandbox and works 100% as expected outside of CSandbox. I'm cool with removing the CRA test and calling it a day, but I feel like there is probably something weird going on.

### Your Environment
| Software         | Name/Version|
| ---------------- | ---------- |
| Сodesandbox  | PROD-1536082527-76b9252 | 
| Browser | Chrome(68)/Firefox |
| Operating System | Mac OS High Sierra |

CompuIves:
I think this is directly a result of running the tests in the same scope as the application itself. An intermediary fix may be to not run the tests right after the application has been loaded, but only after the user opens the `Tests` tab.

The longer term solution should be to run the tests in another iframe (from the sandbox iframe). I will take a look at this at the start of next week.

christianalfoni:
this should have been resolved, closing now. Please reopen if still an issue

PR #1080, opened 2018-09-07
Make running tests conditional

CompuIves:
Helps with #1075 and #1074.

All the Jest tests are currently executing in the same window as the main application, which can cause some trouble when tests alter the main DOM or do things like `location.reload`. The solution is to execute the tests in a separate iframe, but this will be quite some work before we can get there.

In the meantime I propose to only run the tests when the tests tab is opened, contrary to what we have now where we eagerly run tests as soon as the sandbox is evaluated. This will help with infinite loops cause by eg. `onSubmit` in a test and will not infer with the main evaluation of the sandbox.

There are some drawbacks:

- It will take a bit longer to load the tests when you open the Tests tab
- We will not show eagerly the green/red notification on the Tests tab if it's not opened

There also is the advantage that it should maybe give a small general perf boost since we don't evaluate tests after every run of a sandbox anymore if the tab is not opened.

I'm curious what @lbogdan and @kentcdodds think of this intermediate solution until we have separate iframe evaluation. Suggestions to a solution for the drawbacks are also welcome!

### DIFF
--- a/packages/app/src/app/components/Preview/DevTools/Tests/index.js
+++ b/packages/app/src/app/components/Preview/DevTools/Tests/index.js
@@ -107,6 +107,10 @@ class Tests extends React.Component<Props, State> {
         running: true,
       });
     }
+
+    if (this.props.hidden && !nextProps.hidden) {
+      this.runAllTests();
+    }
   }
 
   selectFile = (file: File) => {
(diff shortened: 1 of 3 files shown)


kentcdodds:
I'm definitely in favor of this! Having the tests run on the same document as the app was a source of frustration.

> We will not show eagerly the green/red notification on the Tests tab if it's not opened

This is unfortunate. I'm happy to lose this, but I'd love to have it come back eventually :)

CompuIves:
Merged with a nice compromise, we do now show the amount of tests in the bottom bar, just not the status.

kentcdodds:
Awesome! Thanks.

If there's any way to get the tests to run in the background as well that would be nice, but I'm happier with this change. Thanks!

Issue #1063, opened 2018-08-29
Unable to delete sandbox tags


Yohanna:
# 🐛 bug report

## Description of the problem

I'm unable to delete any tags. When I press the small `x` button beside the tag, nothing happens. And I get the following error in the console:

```
Uncaught Error: [mobx-state-tree] setParent is not supposed to be called on scalar nodes
    at Ie (mobx-state-tree.module.js:2236)
    at e.setParent (mobx-state-tree.module.js:835)
    at Array.t.willChange (mobx-state-tree.module.js:2929)
    at G (common.3f99b861.chunk.js:1)
    at e.spliceWithArray (common.3f99b861.chunk.js:1)
    at t.splice (common.3f99b861.chunk.js:1)
    at Model.js:225
    at Model.js:94
    at Array.reduce (<anonymous>)
    at Model.js:90
    at runInBatch (Model.js:69)
    at _ (common.3f99b861.chunk.js:1)
    at n (common.3f99b861.chunk.js:1)
    at mobx-state-tree.module.js:1857
    at ie (mobx-state-tree.module.js:1763)
    at Object.r (mobx-state-tree.module.js:1778)
    at t.value (Model.js:89)
    at t.value (Model.js:224)
    at function.t.ProviderConstructor.e.(anonymous function) [as splice] (https://codesandbox.io/static/js/app.317e140b.js:1:198829)
    at Object.Ed [as function] (actions.js:69)
    at e.value (FunctionTree.js:98)
    at executeTree.js:77
    at r.executeBranchWrapper (FunctionTree.js:247)
    at t (executeTree.js:13)
    at n (executeTree.js:15)
    at executeTree.js:43
    at r.executeBranchWrapper (FunctionTree.js:247)
    at t (executeTree.js:13)
    at n (executeTree.js:15)
    at executeTree.js:43
    at r.executeBranchWrapper (FunctionTree.js:247)
    at t (executeTree.js:13)
    at executeTree.js:27
    at e.value (FunctionTree.js:158)
    at executeTree.js:77
    at r.executeBranchWrapper (FunctionTree.js:247)
    at t (executeTree.js:13)
    at executeTree.js:45
    at r.executeBranchWrapper (FunctionTree.js:247)
    at t (executeTree.js:13)
    at executeTree.js:45
    at r.executeBranchWrapper (FunctionTree.js:247)
    at t (executeTree.js:13)
    at executeTree.js:82
    at a (FunctionTree.js:318)
[...]

PR #1081, opened 2018-09-07
Update mobx-state-tree to version 3.3.0

krvajal:
Updated the version of `mobx-state-tree` to version 3.3.0. 

The user is unable to remove a tag from the list of tags by clicking the delete button ( See #1063)

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests 
- [x] Ready to be merged 

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -165,7 +165,7 @@
     "memoize-one": "^4.0.0",
     "mobx": "^4.0.0",
     "mobx-react": "^5.2.3",
-    "mobx-state-tree": "^3.0.0",
+    "mobx-state-tree": "^3.3.0",
     "moment": "^2.18.1",
     "monaco-editor": "^0.13.1",
     "monaco-editor-textmate": "^2.0.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -5,28 +5,24 @@
 "@babel/code-frame@7.0.0-beta.44":
   version "7.0.0-beta.44"
   resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.0.0-beta.44.tgz#2a02643368de80916162be70865c97774f3adbd9"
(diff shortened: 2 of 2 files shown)


krvajal:
I would like to add a test to verify that this works as expected and catch any future regressions.

CompuIves:
This is very nice! I have to admit that our testing system is very bad, we do have Jest tests but nearly not enough, so thanks for adding tests! That would be very helpful.

The part that should be tested is Cerebral, I think this is a good article on testing: https://cerebraljs.com/docs/api/test. This is the sequence that we'd need to test:

https://github.com/CompuIves/codesandbox-client/blob/ea37dbc3b6bb230ac580974f1a2e764a5494f3d3/packages/app/src/app/store/modules/workspace/sequences.js#L100-L107

krvajal:
I am having some trouble with the `ensureOwnedSandbox` action because it is calling `forkSandbox` and this then checks for the `editor.currentId` which is null. Is there a way to create a new sandbox by calling a signal or something in the controller?
```js

it("can add a tag", () => {

  const controller = Controller(rootModule);
  // it this editor.currentSandbox.owned == true is not true
  // the forkSandbox sequence will be executed
  // also this will make all the sequence fail
  expect(controller.getState("editor.currentId")).toBeNull()

  expect(controller.getState('workspace.tags')).toEqual({tagName: ""})
  const tagChanged = controller.getSignal("workspace.tagChanged")

  tagChanged({tagName: "tag_01"})
  expect(controller.getState('workspace.tags')).toEqual({tagName: "tag_01"})
  const tagAdded = controller.getSignal("workspace.tagAdded");
  tagAdded()
})


```

CompuIves:
Hmm yeah I can see that as a problem. We can change `ensureOwnedSandbox` to get another prop like `{ force: true }` and it will skip its behaviour. Something like this:

```
export const ensureOwnedSandbox = sequence('ensureOwnedSandbox', [
  when(state`editor.currentSandbox.owned`, props`force`, (owned, force) => owned && !force),
  {
    true: [],
    false: forkSandbox,
  },
]);
```

This seems like the easiest solution, mostly because we use this sequence so many times.

CompuIves:
Heya! I'm getting a bit more requests for this bugfix, so I think we should merge this one in first and focus on the tests in another PR. Thanks a lot for this contribution 😄

PR #1103, opened 2018-09-23
Add babel7 config

christianalfoni:


packages/app/config/webpack.common.js
```diff
@@ -55,6 +55,7 @@ module.exports = {
   entry: SANDBOX_ONLY
     ? {
         sandbox: [
+          '@babel/polyfill',
```
CompuIves: I'm wondering if this would make the bundler bigger than it needs to be, since with this it's including all babel polyfills out there. Is there a way to make this more fine-grained? I thought that babel-preset-env would do something like this.

packages/app/src/app/utils/Loadable.js
```diff
@@ -0,0 +1,25 @@
...
+      });
+    }
+    render() {
+      const { LoadedComponent } = this.state;
+
+      if (LoadedComponent) {
+        return <LoadedComponent />;
+      }
+
+      return <Loading />;
```
CompuIves: Can we show the spinner after a set time, like 1s? That will make perceived loading speed faster and is what Loadable has been doing.

packages/app/config/babel.prod.js
```diff
@@ -4,33 +4,32 @@ module.exports = {
   // Don't try to find .babelrc because we want to force this configuration.
   babelrc: false,
   presets: [
+    require.resolve('@babel/preset-flow'),
     // Latest stable ECMAScript features
     [
       'env',
```
CompuIves: It seems like this 'env' has been unchanged, but that should work too, right?

packages/app/config/babel.prod.js
```diff
@@ -4,33 +4,32 @@ module.exports = {
...
       {
         targets: {
           ie: 11,
```
CompuIves: I'm thinking that we should replace this target with something sensible for CodeSandbox, like 99%>. I will check the analytics.

packages/app/config/babel.prod.js
```diff
@@ -4,33 +4,32 @@ module.exports = {
...
-          // We currently minify with uglify
-          // Remove after https://github.com/mishoo/UglifyJS2/issues/448
-          uglify: true,
         },
         // Disable polyfill transforms
         useBuiltIns: false,
         modules: false,
+        forceAllTransforms: true,
```
CompuIves: What does this mean?

packages/app/config/webpack.common.js
christianalfoni: Yeah, I think we only need a setImmediate poly... which can basically just be:

```
global.setImmediate = global.setImmediate || global.requestAnimationFrame
```

Will try to remove that and only use custom stuff inside existing polyfill file there

packages/app/src/app/utils/Loadable.js
christianalfoni: Jup, for sure, let me fix :)

packages/app/config/babel.prod.js
christianalfoni: Ah, let me fix that, just to be certain it uses the new stuff

packages/app/config/babel.prod.js
christianalfoni: Great, do that :)

packages/app/config/babel.prod.js
christianalfoni: This is a replacement for the `uglify` option that used to be there. They probably wanted to name the option what it actually does, not what like "I have uglify" type of option.

christianalfoni:
Okay, so I had to do some additional work here. Mostly fixing testing. With Babel 7 there are some adjustments to how it finds and handles test files.

- It is supposed to find a `babel.config.js` file at root of lerna project... which it does. But the pattern matching for node_modules does not work for some reason. I believe this is a bug

- The pattern matching for node_modules has been changed, strengthening my point above, because I had to add a file matcher as well at the end

- I had to install babel 7 bridge version to allow babel 6 and 7 to run

- There was a need to update snapshots, due to returning a SET instead of an ARRAY. This could maybe be related to React update. The content of the collection is the same, so let it through

It really is insane how difficult it is to maintain these large projects, but I think this move it very important. When you get the "insides" on babel 7 as well there is a lot of cleanup that can be done!

Hopefully the last fix is moving in now. Just add the babel 7 bridge thingy as a dep, while it was also a devDep :)

christianalfoni:
Not sure what happens on the integration tests there? Any way to run that locally? Or maybe it is related to PR deployment error there? Maybe I broke some production config. Could you help me find some error message? :)

PR #1105, opened 2018-09-24
💅Add Lucy syntax theme

juliettepretot:
screenshot at sep 24 23-19-03

I think this minimal yet contrasty theme would make an outstanding addition to codesandbox. 
It has already been downloaded by 65K+ users on the VSCode marketplace. I would be honored to see it offered inside of my favorite online code editor.

https://github.com/juliettepretot/lucy-vscode

image


CompuIves:
Wow this looks good, thanks for the addition!

PR #1141, opened 2018-10-02
fix(patron): Fixing crash on slider

### LINKED ISSUE #1138: [bug] Patron slider crashing on chrome
# 🐛 bug report

I was using the slider on the patron page and as soon as I slide, page goes blank and spits some errors in the console

[GIF](http://g.recordit.co/nUuFHWkUhM.gif)
### Your Environment
| Сodesandbox  | PROD-1538418597-247779e
| Browser | Chrome 69.0.3497.100
| Operating System | Windows 7

arthurdenner:
Bug fix. Closes #1138.

The app breaks on Chrome when the slider goes over $14 (and the badge would change) because `Array.forEach` is `undefined` (I had no idea about this).

**What is the new behavior?**
Using lodash-es' `forEach, the app doesn't break.

 An alternative would be using `Array.from().forEach`, but lodash-es' `forEach` syntax is simpler, in my opinion. Let me know if you prefer the native method.

### DIFF
--- a/packages/app/src/app/pages/Patron/PricingModal/Badge/Particles/index.js
+++ b/packages/app/src/app/pages/Patron/PricingModal/Badge/Particles/index.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { forEach } from 'lodash-es';
 
 import badges from 'common/utils/badges/patron-info';
 import { Particle } from './elements';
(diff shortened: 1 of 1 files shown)


CompuIves:
Thank you so much, this is due to our recent switch to Babel 7. I'll merge this in now. I'm wondering if there are any other places where we do Array.forEach().

arthurdenner:
> I'm wondering if there are any other places where we do Array.forEach().

There are not. I searched through the code and `Array.forEach` is only used in this file. Thanks for explaining that Babel 7 was the reason. :+1:

PR #1153, opened 2018-10-03
fix(trash): Updating list of deleted sandboxes after permanent removal

arthurdenner:
Bug fix.

After permanently deleting sandboxes, the grid is not updated because the `oldDeleteCache` is mutated when updating the cache.

**What is the new behavior?**
Creating a `newDeleteCache` with all the previous data and overriding `me.sandboxes`, the grid is updated as expected.


CompuIves:
Ahhh, perfect! Great fix @arthurdenner, thank you!

PR #1154, opened 2018-10-03
add vscode icons

SaraVieira:

Uses https://github.com/PKief/vscode-material-icon-theme for the icons on files and folders on codesandbox so that we have no file extension limits and also removes a lot of regexes.

Uses custom files that every time a new file type is added a new regex must be added.

**What is the new behavior?**
Uses new icons that are the default VSCode ones so recognizes file types even if code sandbox doesn't use them.
if no image is found it falls back to the old icons and if that fails we get the default file icon

Images
screenshot 2018-10-03 at 18 18 56
screenshot 2018-10-03 at 18 23 20
screenshot 2018-10-03 at 18 30 46
screenshot 2018-10-03 at 18 31 09
screenshot 2018-10-03 at 18 31 44
screenshot 2018-10-03 at 18 33 23

Example: /s/7z69vrql5q


ForsakenHarmony:
thx

also stop being so productive 👀

PR #1110, opened 2018-09-27
Add tag names to dependency version list

melanieseltzer:
Didn't see anyone working on this yet and thought it was a pretty good feature to have!

**Corresponding Issue**
#1071

**What kind of change does this PR introduce?**
Adds a function `getTagName` to get the name of a tag (if it has one). The name gets displayed next to its corresponding version.

**What is the current behavior?**
List of versions get spit out without tag names.

**What is the new behavior?**
The versions in the list will have a tag name next to it (if one exists). I just followed what was suggested in the issue for the UI.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 
- [x] Added myself to contributors table 

screen shot 2018-09-26 at 9 36 00 pm

PR #1155, opened 2018-10-03
Remove ternary in favor of short &&

melanieseltzer:
Really small but this was bugging me from my previous PR (#1110) - don't have an else so ternary is unnecessary :)

Use logical `&&` instead.

Ternary.


CompuIves:
Oh great! Thank you for the cleaner code 😄

PR #1158, opened 2018-10-04
Correctly update history for pop and replace actions

### LINKED ISSUE #1102: Calling history.go(-1) breaks fake browser's forward button
# 🐛 bug report

## Description of the problem

When a sandbox calls `history.go(-1)` to move backwards in its session history, the fake browser is then unable to go forwards (the button is disabled) and clicking the fake browser's back button reveals that there are extra entries in the session history.

###  Link to sandbox: [link](https://codesandbox.io/s/vqw1mnnl9y) (optional)

Reproduction (using the above sandbox):

1. Click the <kbd>Push</kbd> button three times. This will result in the history
```
["/", "/1", "/2", "/3"]
```
2. Clicking the fake browser's forward/back buttons, you can jump between pages as expected.
3. While on page `/3`, click the <kbd>Go Back</kbd> button. This results in the current location being `/2` (as expected), but the fake browser's forward button is disabled (unexpected). Clicking the fake browser's back button will act as if the history is:
```js
["/", "/1", "/2", "/3", "/", "/1", "/2", "/3"]
```

### Your Environment
| Сodesandbox  | PROD-1536623422-8794c0e
| Browser | Chrome
| Operating System | Windows 10

pshrmn:

This alters how the fake browser (`<Preview>`) manages its `history` state.

Fixes #1102.

Previously, all `urlchange` dispatches were treated as pushes (append URL to end of `history.state` and increment `state.historyPosition`). The fake browser's back and forward buttons got around this by manually updating the state, but manual `history.go()` and `history.replaceState()` calls resulted in an incorrect history array/position.

This behavior can be seen here:

https://codesandbox.io/s/vqw1mnnl9y

(steps to reproduce are described in #1102)

**What is the new behavior?**

When `history.go()` is called, the `state.history` array is unchanged while the `state.historyPosition` is altered by the go amount.

When `history.replaceState()` is called, the `state.history` array is altered to replace the current entry at `state.historyPosition` with the new URL.

`BasePreview.handleForward()` and `BasePreview.handleBack()` no longer manually set state because they can rely on `commitUrl()` to handle this correctly.

**Checklist**:

- [ ] Documentation (N/A?)
- [ ] Tests (would probably be nice, but I'm not sure if there is a way to test this)
- [x] Ready to be merged 
- [ ] Added myself to contributors table (already in there)

### DIFF
--- a/packages/app/src/app/components/Preview/index.js
+++ b/packages/app/src/app/components/Preview/index.js
@@ -550,60 +550,45 @@ class BasePreview extends React.Component<Props, State> {
     dispatch({
       type: 'urlback',
     });
-
-    const { historyPosition, history } = this.state;
-    this.setState({
-      historyPosition: this.state.historyPosition - 1,
-      urlInAddressBar: history[historyPosition - 1],
-    });
   };
 
   handleForward = () => {
     dispatch({
       type: 'urlforward',
     });
-
-    const { historyPosition, history } = this.state;
-    this.setState({
-      historyPosition: this.state.historyPosition + 1,
-      urlInAddressBar: history[historyPosition + 1],
-    });
   };
 
   commitUrl = (url: string, action: string, diff: number) => {
     const { history, historyPosition } = this.state;
 
-    const currentHistory = history[historyPosition] || '';
-    if (currentHistory !== url) {
-      switch (action) {
-        case 'POP':
-          this.setState(prevState => {
-            const newPosition = prevState.historyPosition + diff;
-            return {
-              historyPosition: newPosition,
-              urlInAddressBar: url,
-            };
-          });
-          break;
-        case 'REPLACE':
-          this.setState(prevState => {
-            return {
-              history: [
-                ...prevState.history.slice(0, historyPosition),
-                url,
-                ...prevState.history.slice(historyPosition + 1),
-              ],
-              urlInAddressBar: url,
-            };
-          });
-          break;
-        default:
-          this.setState({
-            history: [...history, url],
-            historyPosition: historyPosition + 1,
+    switch (action) {
+      case 'POP':
+        this.setState(prevState => {
+          const newPosition = prevState.historyPosition + diff;
+          return {
+            historyPosition: newPosition,
             urlInAddressBar: url,
-          });
-      }
+          };
+        });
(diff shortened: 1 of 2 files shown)


CompuIves:
Heya! Thanks for the PR, this is really helpful!

I'm ready to merge this in, but noticed one issue when testing (http://pr1158.cs.lbogdan.tk/s/vqw1mnnl9y). When I open the sandbox it already has the back button enabled, there is no history though. Do you know what could cause this?

pshrmn:
I fixed the issue by replacing instead of pushing a URL change, but I'm not sure _why_ that `setTimeout()` is there.

CompuIves:
Yeah wuh that setTimeout is weird. Maybe @lbogdan knows why it's there?

I have a new scenario that's a bit different! (Sorry).
If you press `Push` and then `Go Back` (in the preview) it doesn't seem like the back button is disabled anymore.

Other than that it looks good and I think is ready to merge!

pshrmn:
Were you running into that bug immediately or when you had navigated to another page (e.g. `/1`) and then refreshed the preview?

There was a bug in the actual test sandbox (I just fixed it) that always assumed the initial value was 0 on load, so it broke if the pathname was a number (other than 0).

That conflicted with [this line](https://github.com/CompuIves/codesandbox-client/blob/4b74200d035e058d26d92e274abeee8311bb5a00/packages/sandbox-hooks/url-listeners.js#L18), which prevents you from pushing the same URL as the current location. I'm not actually sure that that behavior is ideal. While an app probably shouldn't push the same location as the current one, this introduces a different behavior than a non-sandboxed app would have.

I don't want to touch that here in case I'm missing a reason for that check, but I believe that it can (should?) be removed.

pshrmn:
Hey @CompuIves, I'm curious if you can still replicate that bug.

CompuIves:
Wooow this works perfectly now! I think we can remove the line you mentioned @pshrmn. If it causes the sandbox to behave differently from the real-world scenario we shoudl remove it.

pshrmn:
I added a <kbd>Push Same</kbd> button to test sandbox to verify that calling `pushState()` with the current URL pushes the URL onto the session history.

PR #1163, opened 2018-10-04
fix(profile): Update sandboxes count after delete

arthurdenner:
Bug fix.

If you are on your profile and delete a sandbox, the sandboxes count is not updated.

**What is the new behavior?**
The count is updated if the delete operation is successful.


CompuIves:
Super nice! Also thanks for the clarity on `index`.

PR #1173, opened 2018-10-08
chore(homepage): make sure sandbox count is updated

Haroenv:

The "sandbox count" on the homepage now gets fetched from Algolia

The "sandbox count" was hardcoded

**What is the new behavior?**

The "sandbox count" is now fetched from Algolia with a fallback (of 10x more than the static value!!)

I also made sure the "lite" version of Algolia is used on the home page (it's a significant difference in bundle size)

**Checklist**:

- [x] Documentation N/A
- [X] Tests N/A
- [x] Ready to be merged 

### DIFF
--- a/packages/homepage/src/screens/home/NPMFeature/SandboxCount.js
+++ b/packages/homepage/src/screens/home/NPMFeature/SandboxCount.js
@@ -1,0 +1,32 @@
+import React, { Fragment, Component } from 'react';
+import { search } from '../../../utils/algolia';
+
+class SandboxCount extends Component {
+  state = {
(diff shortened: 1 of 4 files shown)


Haroenv:
I also changed the way the homepage search for facet values is working, by still making sure `preact` is findable, but doing it backend with the typoTolerance setting.

CompuIves:
Hahaha man I love this. 80.000 -> 800.000, what a jump.

Thanks for this update and making it easier for us!

Haroenv:
It's incredible, isn't it!

PR #1194, opened 2018-10-16
Fix New Sandbox Modal

SaraVieira:
Fixes the add create sandbox modal

Overflowing because of server templates

**What is the new behavior?**
screenshot 2018-10-16 at 15 29 15

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 


SaraVieira:
screenshot 2018-10-16 at 15 34 36
 why ? :(

CompuIves:
This is great! We need to merge some other stuff in to make it 'official'. 

These ones:

- https://github.com/codesandbox-app/codesandbox-importers/pull/16
- https://github.com/codesandbox-app/codesandbox-importers/pull/18

Then I have to add them to the server too, but that is almost no work. I was also thinking that we can enable Sapper now (it's just adding the `showOnHomepage: true` for Sapper!

SaraVieira:
Sapper updated 🎉

Merge this when the importers are merged as well then :)

packages/app/src/app/pages/Dashboard/Content/CreateNewSandbox/Modal/index.js
```diff
@@ -93,12 +92,27 @@ export default class Modal extends React.PureComponent {
           ))}
 
           <Title>Server Templates</Title>
+          {mainServerRows.map((ts, i) => (
+            // eslint-disable-next-line react/no-array-index-key
```
franklinjavier: You should not use indexes for keys. This can negatively impact performance and may cause issues with state.

refs
https://reactjs.org/docs/lists-and-keys.html#keys
https://github.com/facebook/react/issues/1342#issuecomment-39230939
https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

franklinjavier: Yeah, I know. But if someone isolate and compose this part of code, may affect it. 
I always do this to ensure it will work regardless of context

PR #1201, opened 2018-10-18
Add minimum value to the PriceInput

jamesplease:
Presently, users can enter invalid values, such as -10, by using the native browser controls associated with inputs with a type of "number".

This prevents the following situation:

image

I went with 5 as the minimum due to the validation that's in place that requires that as the minimum:

image

(p.s. that credit card info is fake)

It's a small update to the Patron page.

Users are able to enter values less than 5 (and also less than 0) into the input.

**What is the new behavior?**

Users are unable to enter numbers less than 5 into the input.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged 


jamesplease:
Thank you @SaraVieira and @CompuIves !

PR #1206, opened 2018-10-18
Pushing wipes "future" history

pshrmn:

Follow up to #1158. I noticed that pushing locations when there is "future" history doesn't work correctly (the "forward" button is still active after pushing).

Pushing adds the new location to the end of the history array. This bug is hidden because the `<Preview>` keeps a `urlInAddressBar` property in state.

`history[historyPosition]` _should_ be usable everywhere that `urlInAddressBar` is. I can take a look at whether that works in another PR, but I'd rather get this fix merged in first.

**What is the new behavior?**

Pushing a new location adds it after the current location, wiping out any "future" locations.

I also updated the `historyList` array kept in `url-listeners` to drop "future" locations when pushing so that `historyList.length` doesn't have to be manually set.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 

### DIFF
--- a/packages/app/src/app/components/Preview/index.js
+++ b/packages/app/src/app/components/Preview/index.js
@@ -567,7 +567,6 @@ class BasePreview extends React.Component<Props, State> {
 
   commitUrl = (url: string, action: string, diff: number) => {
     const { history, historyPosition } = this.state;
-
     switch (action) {
       case 'POP':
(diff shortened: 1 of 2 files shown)


CompuIves:
Makes sense, thanks for the quick followup @pshrmn!

PR #1221, opened 2018-10-23
Now integration

SaraVieira:
Adds better now integration

You can deploy things but no feedback is given on what you deployed

**What is the new behavior?**
* You can see a list of deployments done with that sandbox
* You can delete deployments
* Visit them
* Alias them to what you set up in your now.json

screenshot 2018-10-23 at 10 58 05
screenshot 2018-10-23 at 10 58 18
screenshot 2018-10-23 at 10 58 22
screenshot 2018-10-23 at 10 58 29


CompuIves:
This is insane @SaraVieira!!! This is a huge feature, so well done! I can't believe we have this now! Thank you so much!

Issue #1212, opened 2018-10-21
Prompt on saving frozen sandbox


dance2die:
# 🌈 Feature

## What is this feature?
Prompt a user before forking a `frozen` sandbox.

## How the feature should work?
A user should be notified before forking a frozen sandbox.

## You have examples or an idea how it can be implemented?
Hovering over `?` button shows that saving frozen`

>  ...sandBox will fork on save

![image](https://user-images.githubusercontent.com/8465237/47271167-e7021a80-d543-11e8-8277-e7fb28d2def4.png)

But the transition (when you click `ctrl + s`) is so subtle (and fast on a fast connection), it's hard to tell if you are working on a frozen SandBox or a forked on as you can see below.

![2018-10-21_15-20-09](https://user-images.githubusercontent.com/8465237/47271241-db632380-d544-11e8-983c-eec1c8aa8b3e.gif)

We can prompt a user with a modal dialog to confirm if s/he wants to fork the frozen sandbox.

SaraVieira:
Hey !

You are completely right ! It was an oversight to not include an alert for that .

Is the alert that shows up when you click to fork your own sandbox something that would work here ?

dance2die:
> Is the alert that shows up when you click to fork your own sandbox something that would work here ?

The easiest way to implement this is to **disable** `ctrl+s` keyboard shortcut.
But yes, an alert to prevent you from forking accidentally would work better 
_(as the intention is more explicit and leave people hanging why it's not saving/forking)_

JonShort:
Hi both 👋 

I've submitted a PR adding in this functionality, like Sara suggested above I added a browser alert when the user is trying to save edits to a frozen sandbox which they own.

I'd be interested in getting opinions on the message we should display to the user, or other approaches to solving this problem 👍

PR #1232, opened 2018-10-24
Added alert when saving frozen sandbox

JonShort:

This PR adds a browser alert prior to saving a frozen sandbox. The alert describes that the sandbox will be forked.

Also included is a minor code spelling fix (FreezeConatainer > FreezeContainer)

Currently saving a frozen sandbox will silently fork the sandbox, similar to saving a sandbox your don't own.

Issue #1212 mentioned that a prompt identifying that the sandbox will be forked could be useful.

**What is the new behavior?**

When saving edits to a frozen sandbox which you own, an alert will appear describing that the sandbox will be forked.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 

**Comments / Questions**

Q: Could this be a `confirm()` rather than an `alert()`?

A: This would be my ideal implementation, allowing the user to cancel and 'unfreeze' their sandbox if they wish. Unfortunately I don't know a way to cancel the proceeding cerebral `sequence` (which would need to be approached generically as save is triggered in many places). I'd be interested in seeing other approaches to this problem which allow this functionality.

Q: Does this negatively affect UX?

A: As mentioned in the issue prompting this work, the application silently forking a frozen sandbox felt like a bit of a dark pattern to the individual raising the issue. I'm happy to offer up this PR, and get feedback around the UX impact of it.

If anyone can think of a message clearer than "This sandbox is frozen, and will be forked." please let me know, I feel the message might not be clear to someone who doesn't know the concept of 'frozen' and 'forked' sandboxes 😃


packages/app/src/app/store/sequences.js
```diff
@@ -146,15 +147,25 @@ export const forkSandbox = sequence('forkSandbox', [
   },
 ]);
 
+export const forkFrozenSandbox = sequence('forkFrozenSandbox', [
+  when(state`editor.currentSandbox.isFrozen`),
+  {
+    true: [alertForkingFrozenSandbox, forkSandbox],
```
CompuIves: To make this work with a `confirm` we can probably use the `path` directive of Cerebral. Here is an example: https://github.com/CompuIves/codesandbox-client/blob/master/packages/app/src/app/store/actions.js#L8-L19

The idea is that you can return different paths from an action, and in the sequence you can define what to do for every returned path. So here you could do:

```
{
	confirm: [forkSandbox],
    cancel: []
}
```

I think that would solve your question for how to use `confirm`.

packages/app/src/app/store/modules/editor/actions.js
```diff
@@ -301,6 +301,10 @@ export function confirmForkingOwnSandbox({ browser, path }) {
     : path.cancelled();
 }
 
+export function alertForkingFrozenSandbox({ browser }) {
```
CompuIves: So here you could add `path` and return either `path.confirm`/`path.cancel`.

packages/app/src/app/pages/Sandbox/Editor/Workspace/Project/elements.js
```diff
@@ -35,7 +35,7 @@ export const PrivacyContainer = styled.span`
   margin-bottom: 1rem;
 `;
 
-export const FreezeConatainer = styled.span`
+export const FreezeContainer = styled.span`
```
CompuIves: Nice find!

packages/app/src/app/store/sequences.js
JonShort: I've given this a try locally, and the fork is cancelled as we'd expect when user cancels the confirm 🎉

However the issue then becomes cancelling the remaining actions e.g:
image

Here when `ensureOwnedEditable` returns without forking, the actions below are still dispatched (e.g. save is performed on frozen sandbox).

To be honest I feel as though it'd be too disruptive to the architecture to add in `if` `else` type logic, especially in this PR 😆

As this PR is a UX enhancement, i'll leave it as-is, feel free to reject/approve as you see fit 👍

karlhorky:
This seems like a cool change, just wanted this exact feature.

Is there anything still open before this can be merged?

JonShort:
I think this PR is perhaps a bit too old now, I'll close it so others can tackle this issue if they wish 👍

SaraVieira:
For bug week I will finish your work!

Thank you so much! I will make sure to add you as a co-author and contributor

karlhorky:
Work continuing in #1716 for those who want to follow along :)

PR #1234, opened 2018-10-25
Fix showing "Connection Trouble" warning after fork

lbogdan:


packages/app/src/app/components/Preview/index.js
```diff
@@ -50,6 +50,7 @@ type State = {
   overlayMessage: ?string,
   hibernated: boolean,
   sseError: boolean,
+  localClose: boolean,
```
CompuIves: What does `localClose` mean? Maybe we could add a comment describing what it does.

packages/app/src/app/components/Preview/index.js
```diff
@@ -217,15 +220,12 @@ class BasePreview extends React.Component<Props, State> {
         frameInitialized: false,
       });
       if (this.$socket) {
+        this.state.localClose = true;
```
CompuIves: I'm not sure if this is a good way of altering state. If there's a direct assignment I think it makes sense to not put it on state but as an instance field on the class.

lbogdan:
> You got the magic number 1234!! 😄

Now reaching for 12345! 🙂

Issue #1211, opened 2018-10-21
Initialize CSB on nested routes (not only when accessing index)


manniL:
# 🌈 Feature

If you currently access a nested route of an **inactive CSB**, a 404 will be thrown (example: https://1rmqv4xkx3.sse.codesandbox.io/reactivity/wrong).

It'd be great if the sandbox would start initializing (similar like when you access the root/index page).

lbogdan:
This has been added. Please test and close this if it works as expected.

manniL:
Whoop, that was fast! 
Works as expected, thank you ☺️ 👍

PR #1237, opened 2018-10-25
Container loading screen improvements

lbogdan:
Related issue: #1211 .


packages/common/.babelrc
```diff
@@ -0,0 +1,16 @@
...
+        "polyfill": false,
+        "regenerator": true
+      }
+    ],
+    "transform-object-rest-spread",
+    "transform-class-properties",
+    "babel-plugin-emotion",
+    "transform-flow-strip-types"
+  ]
+}
```
lbogdan: Added this so that we can build `sse-loading-screen`.

PR #1306, opened 2018-11-17
Default to vertical layout when height > width

CompuIves:
This is especially useful when looking with a phone. We automatically change the layout of embeds to vertical when width < height.

Curious what you think @SaraVieira, @lbogdan.

image

image


SaraVieira:
This makes sense to me :)

PR #1313, opened 2018-11-18
Redirect to sandbox wizard when deleting sandbox

### LINKED ISSUE #1312: Redirect to template selection when deleting sandbox
# 🌈 Feature

Currently, when you delete a sandbox, the application redirects to `/s/new`, which is the template for a React application.

I think that it would be nice to redirect to `/s` (the "template selection" route) instead.

Alternatively, deleting could redirect to the new template page for the type of sandbox that was deleted.

I prefer redirecting to the template selection route. If I delete a sandbox, I don't necessarily want to create a new one right away, so it feels a bit incongruous to be taken to a new sandbox.

pshrmn:

This PR changes where the user is redirected after deleting a sandbox.

Fixes #1312 

The user is redirected to the new React template (`/s/new`).

**What is the new behavior?**

The user is redirected to the sandbox wizard (`/s`).

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 

### DIFF
--- a/packages/app/src/app/store/providers/Router.js
+++ b/packages/app/src/app/store/providers/Router.js
@@ -10,6 +10,9 @@ export default Provider({
   redirectToNewSandbox() {
     history.push('/s/new');
   },
+  redirectToSandboxWizard() {
+    history.push('/s');
+  },
   getSandboxOptions() {
     return getSandboxOptions(decodeURIComponent(document.location.href));
   },
--- a/packages/app/src/app/store/providers/Router.js
+++ b/packages/app/src/app/store/providers/Router.js
@@ -11,7 +11,7 @@ export default Provider({
     history.push('/s/new');
   },
   redirectToSandboxWizard() {
-    history.push('/s');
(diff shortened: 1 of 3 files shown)


packages/app/src/app/store/providers/Router.js
```diff
@@ -10,6 +10,9 @@ export default Provider({
   redirectToNewSandbox() {
     history.push('/s/new');
   },
+  redirectToSandboxWizard() {
+    history.replace('/s');
```
lbogdan: Hey @pshrmn , can you please make it `/s/`, to prevent the redirect?

pshrmn:
```suggestion
    history.replace('/s/');
```

CompuIves:
Thanks a lot @pshrmn!

lbogdan:
There were still lint errors... 😟

CompuIves:
Fixed!

PR #1325, opened 2018-11-22
Provide contact address in code of conduct

Noviny:
In part just raising that this is missing. This is probably not the ideal email. If you know what a better email would be, probably reject this and update with that instead.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged 


SaraVieira:
Thank you so much 🙌

PR #1328, opened 2018-11-24
Missing styles Ives :(

SaraVieira:
Adds import of the styles


CompuIves:
Hah, merci @SaraVieira!

Issue #1286, opened 2018-11-07
Babel env preset browser target is not respected in vanilla template


blikblum:
# 🐛 bug report

## Description of the problem
In a project forked from vanilla template with a custom .babelrc, the browser target of env preset is not respected.

For example, setting the browser target to "chrome 68" the ES classes are still compiled to ES5, while in a local setup no transpilation is done at all:

```
"presets": [
    ["env", {"targets": {"browsers": ["chrome 68"]}}]
  ] 
```

## How has this issue affected you? What are you trying to accomplish?
I'm trying to use LitElement package which is distributed as unstranspiled ES classes. Mixing transpiled with unstranspiled classes does not work for Custom Elements. Gives a "Class constructor LitElement cannot be invoked without 'new'" error

### Your Environment
| Software         | Name/Version|
| ---------------- | ---------- |
| Сodesandbox  | 
| Browser |  Chrome 70
| Operating System | Windows 10

CompuIves:
We now respect it! Sorry for the inconvenience. 

Note that I enabled it for babel 7 only, you can force babel7 by adding `@babel/core` to the `devDependencies`.

blikblum:
Hi, it fixed partially and example with lit-element still not working.

See https://codesandbox.io/s/47l21wjlx

I added babel/core to devDependencies and updated babelrc syntax. Set preset to chrome 68 i.e. should not transpile ES classes or async functions 

The sandbox own files (src/index.js) are not transpiled (ES class syntax is kept) but dependency (@polymer/lit-element/lib/updating-element.js) is transpiled both classes and async (sandbox fails with "regeneratorRuntime is not defined" error)

yugandhar-pathi:
Me too face the same issue, irrespective of target it always moves the code to ES5.

ntucker:
seems to ignore browserslist configurations set in package.json

Issue #1266, opened 2018-11-02
Error with BrowserFS while rendering a React app


feelextra:
# 🐛 bug report
## Description of the problem
This React project works for me on my local environment. Tried to upload it and this issue happens on the provided URL below.

Might be related to [this open issue at BrowserFS](https://github.com/jvilk/BrowserFS/issues/210#issuecomment-435501601)

Attached image showing what it looks like.
```
Error: EIO: Initialize BrowserFS with a file system using BrowserFS.initialize(filesystem)

11 stack frames were expanded.
C
https://codesandbox.io/static/js/sandbox.3fc888ce.js:1:220188
Worker.<anonymous>
https://codesandbox.io/static/js/sandbox.3fc888ce.js:1:223203
g
https://codesandbox.io/static/js/common-sandbox.730613df.chunk.js:1:207992
Generator._invoke
https://codesandbox.io/static/js/common-sandbox.730613df.chunk.js:1:207780
Generator.e.(anonymous function) [as next]
https://codesandbox.io/static/js/common-sandbox.730613df.chunk.js:1:208171
u
https://codesandbox.io/static/js/common-sandbox.730613df.chunk.js:1:2891
a
https://codesandbox.io/static/js/common-sandbox.730613df.chunk.js:1:2980
(anonymous function)
https://codesandbox.io/static/js/common-sandbox.730613df.chunk.js:1:3019
new r
https://codesandbox.io/static/js/common-sandbox.730613df.chunk.js:1:71362
Worker.<anonymous>
https://codesandbox.io/static/js/common-sandbox.730613df.chunk.js:1:2822
Worker.<anonymous>
https://codesandbox.io/static/js/sandbox.3fc888ce.js:1:224364
```
## How has this issue affected you? What are you trying to accomplish?
Trying to demonstrate an issue that I raised on [emotion's repo](https://github.com/emotion-js/emotion/issues/985)
### Your Environment
| Software         | Name/Version|
| ---------------- | ---------- |
| Сodesandbox  |  PROD-1541005726-0dba21e |
| Browser | Chromium 70.0.3538.77 (Official Build) |
| Operating System | Built on Ubuntu , running on LinuxMint 18.3 (64-bit) |

![codesandbox-issue](https://user-images.githubusercontent.com/43674238/47940918-dd3dc700-def4-11e8-8401-1b91e190d808.png)

feelextra:
Update: previous error message is gone, here is the new error message:

[...]

Issue #1179, opened 2018-10-09
Babel error on Vue CLI 3 template


Haroenv:
# 🐛 bug report

## Description of the problem

This is likely because some mismatch of Babel versions?

## How has this issue affected you? What are you trying to accomplish?

<img width="562" alt="screen shot 2018-10-09 at 14 05 50" src="https://user-images.githubusercontent.com/6270048/46668406-9f3ad500-cbcc-11e8-9d46-00d1e3109644.png">

```
[BABEL] /src/main.js: The 'useBuiltIns' option has been removed. The @babel/runtime module now uses builtins by default. (While processing: "base$0$7")
```

### Your Environment
| Software         | Name/Version|
| ---------------- | ---------- |
| Сodesandbox  | PROD-1534913825-2619708 |
| Browser | Safari (tech preview Release 63 (Safari 12.1, WebKit 13607.1.2.1)) |
| Operating System | MacOS High Sierra 10.13.6 (17G65) |

Haroenv:
I think it has something to do with one of these dependencies: 

```
    "@vue/cli-plugin-babel": "^3.0.5",
    "@vue/cli-plugin-eslint": "^3.0.3",
    "@vue/cli-service": "^3.0.3",
    "vue-template-compiler": "^2.5.17"
```

Haroenv:
Maybe it's because the cli-plugin-babel has babel 7 alpha, and you might have the released version?

lbogdan:
@Haroenv Meanwhile you can play around with (most probably) the prototype to the Vue CLI 3 server / container template here: https://codesandbox.io/s/github/lbogdan/cs-vue-cli-3 . 🙂

Haroenv:
thanks @lbogdan, did you fix it by adding `"vue-template-compiler": "^2.5.17"`?

Haroenv:
Seems like this is still happening, but now a different error message:

```
Invalid preset specified in Babel options: "env"
▼ 9 stack frames were expanded.
C
https://codesandbox.io/static/js/sandbox.616011f0.js:1:228358
(anonymous function)
https://codesandbox.io/static/js/sandbox.616011f0.js:1:231365
g
https://codesandbox.io/static/js/common-sandbox.730613df.chunk.js:1:207996
(anonymous function)
https://codesandbox.io/static/js/common-sandbox.730613df.chunk.js:1:207781
u
https://codesandbox.io/static/js/common-sandbox.730613df.chunk.js:1:2891
[...]

PR #1349, opened 2018-12-03
Add support for dynamic macros

### LINKED ISSUE #981: [Question] How to use babel-plugin-macros in create-react-app template?
# ❔ Question

hi! I am trying to use @kentcdodds ' `babel-plugin-macros` inside the create-react-app template in CSB. I have upgraded `react-scripts` to a v2.0 beta (where babel-plugin-macros is supported) but it seems to still not be running. I feel like there is some key piece of knowledge I am missing, and am unable to debug this myself. any hints please?

CompuIves:
This change will detect if an error is thrown for a missing module, if this
error is detected we will try to traverse the file and download all remaining
dependencies recursively. We then sync this back to the manager so new babel
workers will have these files as well.

This comes down to having a mini bundler and file fetcher for every babel
worker to execute babel macros.

Fixes #981 #1266 #1179 #1286

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -115,7 +115,7 @@
     "@cerebral/mobx-state-tree": "^3.0.0",
     "@emmetio/codemirror-plugin": "^0.3.5",
     "@svgr/core": "^2.4.1",
-    "@vue/babel-preset-app": "^3.0.1",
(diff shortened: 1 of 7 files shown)


sw-yx:
woohoo @kentcdodds

kentcdodds:
YES!

PR #1367, opened 2018-12-10
[WIP] Remove inline styles

SaraVieira:
Remove most inline styles

Count:
Start: 404
Current: 285


SaraVieira:
240

SaraVieira:
146

CompuIves:
NICe!

SaraVieira:
80

SaraVieira:
Done!!

Website is breaking because of 

```
styled-components.browser.esm.js:226 Uncaught Error: It seems you are interpolating a keyframe declaration (bcCCNc) into an untagged string. This was supported in styled-components v3, but is not longer supported in v4 as keyframes are now injected on-demand. Please wrap your string in the css\`\` helper (see https://www.styled-components.com/docs/api#css), which ensures the styles are injected correctly.
```

Things like this:

```
export const Container = styled.div`
  position: absolute;
  display: flex;
  flex-direction: column;
  justify-content: space-between;
  background-color: ${props =>
    props.theme['sideBar.background'] || props.theme.background};
  color: ${props => props.theme['sideBar.foreground'] || 'inherit'};
  height: 100%;
  width: 100%;
  overflow-y: overlay;
  overflow-x: auto;

  > div {
    ${fadeIn(0)};
  }
`;
```

SaraVieira:
So the whole website is broken now so I will close this at least for now

PR #1386, opened 2018-12-16
Submit Feedback

CompuIves:


CompuIves:
I moved the submit logic to the modal itself, this way we can lazy load the modal with code splitting and split the airtable dependency to the modal.

PR #1388, opened 2018-12-17
Tag focus

### LINKED ISSUE #1385: "add a tag" loses focus on every key press
# 🐛 bug report

When adding a tag, the input loses focus on every key press.

SaraVieira:
Fixes loose focus on keydown

Removed focus on keypress

**What is the new behavior?**
Works now

closes #1385

### DIFF
--- a/packages/app/src/app/components/EditableTags/index.js
+++ b/packages/app/src/app/components/EditableTags/index.js
@@ -2,8 +2,8 @@ import React, { Fragment } from 'react';
 import { createGlobalStyle } from 'styled-components';
 import TagsInput from 'react-tagsinput';
 
-export default function(color) {
-  const GlobalStyle = createGlobalStyle`
+const GlobalStyle = createGlobalStyle`
+${console.log(props => props)}
   .react-tagsinput {
     display: flex;
     flex-wrap: row;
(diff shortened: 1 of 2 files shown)


packages/app/src/app/components/EditableTags/index.js
```diff
@@ -58,10 +67,9 @@ export default function(color) {
...
-  return props => (
-    <Fragment>
-      <GlobalStyle />
-      <TagsInput {...props} />
-    </Fragment>
-  );
-}
+export default props => (
+  <Fragment>
+    <GlobalStyle color={props.template.color()} />
```
CompuIves: I think you can put this in, with this you don't have to do `hexToRgb` of `color` in the tag.

```suggestion
    <GlobalStyle color={props.template.color} />
```

packages/app/src/app/components/EditableTags/index.js
```diff
@@ -43,7 +52,7 @@ export default function(color) {
     width: 80px;
 
     &:focus {
-      border-color: ${color.clearer(0.3)()};
+      border-color: ${props => hexToRgb(props.color, 0.3)};
```
CompuIves: This should work too:

```suggestion
      border-color: ${props => props.color.clearer(0.3)};
```

Because the template color also has a decorated change.

SaraVieira:
That fixed it yes 💯

PR #1391, opened 2018-12-19
add zeit warning

SaraVieira:
Feature 

No warning saying we only support version 1 of zeit

**What is the new behavior?**
Adds a new warning for version 1 and also a way for people to say they want version 2

zeit2

It also shows a notification saying your feedback was saved but that's not on the gif

### DIFF
--- a/packages/app/src/app/store/modules/deployment/actions.js
+++ b/packages/app/src/app/store/modules/deployment/actions.js
@@ -44,20 +44,20 @@ export function getPeopleWhoWant2() {
 export function addPersonWhoWant2({ path, props }) {
   const base = Airtable.base('apppgSmcJWwuXac6t');
 
-  const getPeople = () =>
-    new Promise(res => {
-      base('zeit2').create({ username: props.username }, (err, record) => {
+  const AddPerson = () =>
+    new Promise((res, rej) => {
+      base('zeit2').create({ username: props.username }, err => {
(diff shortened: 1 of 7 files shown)


CompuIves:
Omg I *love* this!

packages/app/src/app/store/modules/deployment/actions.js
```diff
@@ -14,6 +15,51 @@ export function loadZip({ props, jsZip }) {
   return jsZip.loadAsync(file).then(result => ({ contents: result }));
 }
 
+export function getPeopleWhoWant2() {
```
CompuIves: Could we perhaps make this function async and dynamically import `airtable`? That should shave us some bundle size 😄

SaraVieira: You want a fancy require inside that ? 😄

packages/app/src/app/store/modules/deployment/actions.js
```diff
@@ -1,5 +1,5 @@
 import { omit } from 'lodash-es';
-
+import Loadable from 'app/utils/Loadable';
```
CompuIves: I think that this one can be removed.

SaraVieira: 🤦‍♂️

PR #1396, opened 2018-12-20
Fix bug where deploying to Now breaks binary files

lbogdan:
Unfortunately in `createApiData()` (I think) we lose the `module.isBinary` flag, so I ended up sending all files to the Now API with `base64` encoding, which has a bit of overhead for text files.

@CompuIves Is there any easy way to determine if a file is text or binary in `createApiData()`?

### DIFF
--- a/packages/app/src/app/store/modules/deployment/actions.js
+++ b/packages/app/src/app/store/modules/deployment/actions.js
@@ -132,9 +132,9 @@ export async function createApiData({ props, state }) {
     const file = contents.files[filePath];
 
     if (!file.dir && filePath !== 'package.json') {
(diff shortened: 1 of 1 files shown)


CompuIves:
Hmm, yes, we lose that data when we go from files -> zip -> API. I think the way to fix this would be to separate the zip file step to make it files -> API. Or maybe we can determine from the start of the base64 whether it's a text file?

Nevertheless, I think the overhead is not super big. So if you feel happy with these changes we can merge it in!

lbogdan:
Well, I at least think it's better to have a bit of overhead than binary files not working at all. :-)

lbogdan:
@CompuIves , can we merge this?

PR #1417, opened 2019-01-01
Add zeit 2 support

SaraVieira:
Adds support for zeit now version 2

Currently we have a counter and it had like 140 people so I guess it's time

**What is the new behavior?**
If the now.json has `version:2` specified the new now platform will be used

Here is an example: 
https://codesandbox-v2-test.now.sh/graphql

Also all the other endpoints remained the same so it was easier then I thought o.o

cc @timothyis @rauchg

### DIFF
--- a/packages/app/src/app/store/modules/deployment/actions.js
+++ b/packages/app/src/app/store/modules/deployment/actions.js
@@ -13,61 +13,6 @@ export function loadZip({ props, jsZip }) {
   return jsZip.loadAsync(file).then(result => ({ contents: result }));
 }
 
-export async function getPeopleWhoWant2() {
-  const Airtable = await import(/* webpackChunkName: 'airtable' */ '../../utils/setAirtable');
-
-  const base = await Airtable.default.base('apppgSmcJWwuXac6t');
-  const params = base('zeit2').select({
-    view: 'Grid view',
-    maxRecords: 100000,
-  });
-  const people = [];
-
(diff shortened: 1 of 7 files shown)


packages/common/templates/template.js
```diff
@@ -126,8 +126,8 @@ export default class Template {
         serve: '^10.1.1',
       },
       scripts: {
-        ...parsedFile.scripts,
         'now-start': `cd ${this.distDir} && serve -s ./`,
+        ...parsedFile.scripts,
```
SaraVieira: This was moved because I may want to override your now-start

packages/app/src/app/store/modules/deployment/actions.js
```diff
@@ -167,10 +121,11 @@ export async function aliasDeployment({ http, path, props, state }) {
 export async function postToZeit({ http, path, props, state }) {
   const { apiData } = props;
   const token = state.get('user.integrations.zeit.token');
+  const deploymentVersion = apiData.version === 2 ? 'v6' : 'v3';
```
CompuIves: Very smart!

SaraVieira: That's what my momma says

SaraVieira:
Kind of amazing that adding Zeit 2 actually removed more code then it added 😂

darielnoel:
+1

PR #1418, opened 2019-01-01
add private gh to patron features

SaraVieira:
Docs update.

**What is the new behavior?**
Adds the fact that private github import is from patrons only

screenshot 2019-01-01 at 22 59 18


SaraVieira:
Maybe we should also add the server sandboxes

CompuIves:
Yep! Let's add that too.

SaraVieira:
@lbogdan @CompuIves Done :D

CompuIves:
Cooool! Merging it in now 😄

PR #1423, opened 2019-01-03
Make tabs on new sandbox modal

SaraVieira:
Adds tabs to the new sandbox modal since we are getting too many templates

Currently ut's a long modal

**What is the new behavior?**
Tabs for popular, client, server and preset templates

screenshot 2019-01-03 at 22 54 02

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Content/CreateNewSandbox/Modal/index.js
+++ b/packages/app/src/app/pages/Dashboard/Content/CreateNewSandbox/Modal/index.js
@@ -77,7 +77,7 @@ export default class Modal extends React.PureComponent {
       <Fragment>
         <TabContainer>
           {availableTemplates.map((t, i) => (
-            <Button selected={selectedTab === i} onClick={() => this.setTab(i)}>
+            <Button key={i} selected={selectedTab === i} onClick={() => this.setTab(i)}>
               {t.name}
             </Button>
           ))}
--- a/packages/app/src/app/pages/Dashboard/Content/CreateNewSandbox/Modal/index.js
+++ b/packages/app/src/app/pages/Dashboard/Content/CreateNewSandbox/Modal/index.js
@@ -76,9 +76,13 @@ export default class Modal extends React.PureComponent {
     return (
       <Fragment>
         <TabContainer>
-          {availableTemplates.map((t, i) => (
-            <Button key={i} selected={selectedTab === i} onClick={() => this.setTab(i)}>
-              {t.name}
+          {availableTemplates.map(({ name }, i) => (
+            <Button
+              key={name}
+              selected={selectedTab === i}
+              onClick={() => this.setTab(i)}
+            >
+              {name}
             </Button>
           ))}
         </TabContainer>
@@ -91,10 +95,10 @@ export default class Modal extends React.PureComponent {
             {availableTemplates.map((tab, i) => (
               <Tab visible={selectedTab === i}>
                 <Templates>
-                  {tab.templates.map(t => (
+                  {tab.templates.map(template => (
                     <Template
-                      key={t.name}
-                      template={t}
+                      key={template.name}
+                      template={template}
                       selectTemplate={this.selectTemplate}
                     />
                   ))}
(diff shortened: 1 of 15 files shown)


packages/app/src/app/pages/Dashboard/Content/CreateNewSandbox/Modal/index.js
```diff
@@ -55,123 +54,73 @@ export default class Modal extends React.PureComponent {
...
-                  subtitle={`Using ${t.variantName} template`}
-                />
-              ))}
-            </Templates>
+    const { selectedTab } = this.state;
+    return (
+      <Fragment>
+        <TabContainer>
+          {availableTemplates.map((t, i) => (
+            <Button selected={selectedTab === i} onClick={() => this.setTab(i)}>
```
SaraVieira:
```suggestion
            <Button key={i} selected={selectedTab === i} onClick={() => this.setTab(i)}>
```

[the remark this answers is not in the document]
SaraVieira: Added! How this passed liniting is beyond me :/

Also I added `i` but it's okay because they will never move in the array

packages/app/src/app/pages/Dashboard/Content/CreateNewSandbox/Modal/index.js
```diff
@@ -55,123 +54,73 @@ export default class Modal extends React.PureComponent {
...
+        <Container
+          closing={closing}
+          forking={forking}
+          onMouseDown={e => e.preventDefault()}
+        >
+          <InnerContainer forking={forking} closing={closing}>
+            {availableTemplates.map((tab, i) => (
+              <Tab visible={selectedTab === i}>
+                <Templates>
+                  {tab.templates.map(t => (
```
donavon: kinda seems like cheating to call this `t`. could we please name it `template` like it probably should be?

packages/app/src/app/pages/Dashboard/Content/CreateNewSandbox/Modal/index.js
```diff
@@ -55,123 +54,73 @@ export default class Modal extends React.PureComponent {
...
-                  selectTemplate={this.selectTemplate}
-                  subtitle={`Using ${t.variantName} template`}
-                />
-              ))}
-            </Templates>
+    const { selectedTab } = this.state;
+    return (
+      <Fragment>
+        <TabContainer>
+          {availableTemplates.map((t, i) => (
```
donavon: you could deconstruct `name` inline as you don't use the entire keys

```js
{availableTemplates.map(({ name }, i) => (
```

packages/app/src/app/pages/Dashboard/Content/CreateNewSandbox/Modal/index.js
SaraVieira: I fixed this and forgot to comment :(

packages/app/src/app/pages/Dashboard/Content/CreateNewSandbox/Modal/index.js
SaraVieira: Same :)

SaraVieira:
le promised ping

CompuIves:
This one is tricky, it will directly affect how templates are discovered by people. I like the distinction between Popular/Client/Server/Preset, we could even make it personalized in the future so we can eg. show the most used template for a user in "Popular". 

I think the most "organized" way to handle this is according to this:

The "Popular" tab stays as it is now, but we'll remove all "small" buttons. So we'd have:

### Client

Angular - React - Vanilla - Vue

### Server

Gatsby - Next.js - Node - Nuxt

### Presets

React + TS - Vanilla + TS

We'd keep the same layout though (so the headers then stay). Then we have server and client tab, I think we should just put all server templates under server and client templates under client. Like you have now.

It feels like that's the most organized and flexible way to handle this. In the future we can change this once we extend the template behaviour.

PR #1512, opened 2019-01-31
add markojs server template support

nm123github:
Adding markoJS server template to list of options.

https://github.com/CompuIves/codesandbox-client/issues/847

**Checklist**:

- [ N/A ] Documentation
- [ N/A ] Tests
- [ x ] Ready to be merged (i think)
- [ N/A ] Added myself to contributors table (.all-contributorsrc has malformed JSON)

codesandbox-client/packages/homepage/src/screens/home/Frameworks/index.js
^ Do we need to add an entry to TEMPLATE_SUPPORT? Where exactly does that appear?

image


nm123github:
@CompuIves not sure if you got a chance to look at this one. Attempting to add markoJS template support to the list of supported templates.

CompuIves:
Hey! Sorry for not responding, I missed this PR. We can merge this within a week! Could you fix the merge conflicts? Then we're ready to go!

nm123github:
@CompuIves fixed conflicts 👍

SaraVieira:
Checking out why the deploy failed so we can get this merged :)

SaraVieira:
Hey!

It doesn't seem to update when saved in any way: https://codesandbox.io/s/github/nm123github/marko-codesandbox :(

nm123github:
@SaraVieira Could you try again? Updated the default template.

packages/common/src/utils/url-generator.ts
```diff
@@ -32,6 +32,7 @@ export const parcelSandboxUrl = () => `/s/vanilla`;
 export const newReactTypeScriptSandboxUrl = () => `/s/react-ts`;
 export const newDojoSandboxUrl = () =>
   `/s/github/dojo/dojo-codesandbox-template`;
+export const newMarkoSandboxUrl = () => `s/marko`;
```
CompuIves: I don't think you need this part of the code, since it's used nowhere. 
```suggestion
```

CompuIves:
Great! Let's merge this in together with #1753 when ready.

nm123github:
This is ready from my end. Just few things
- Would be nice to have some description of Template variables (e.g. main, mainFiles, showOnHomePage etc) and how/where they are used.
- Right now i've set template to 'node' in my project's [sandbox.config.json](https://github.com/nm123github/marko-codesandbox/blob/master/sandbox.config.json
) (and it works). However, when i set the template to marko it says 'invalid template specified'. I am guessing this should work after code is merged.

packages/common/src/utils/url-generator.ts
nm123github: deleted.

CompuIves:
Great! 

> Would be nice to have some description of Template variables (e.g. main, mainFiles, showOnHomePage etc) and how/where they are used.

Yes, I agree, we should document this better.

> Right now i've set template to 'node' in my project's sandbox.config.json (and it works). However, when i set the template to marko it says 'invalid template specified'. I am guessing this should work after code is merged.

Yep! This will work as soon as I make a new deploy with the server.

I'll merge this in now, thanks a lot for this contribution!

nm123github:
Quick question @CompuIves , whats the release cadence like? When would we expect to see this live?

PR #1524, opened 2019-02-08
Update Import from Github section

fraindz:
Docs update

The current import wizard at https://codesandbox.io/s/github expects actual github URL and automatically converts it to sandbox URL. User is not required(and not allowed) to enter sandbox URL anymore.

**What is the new behavior?**
Now, unnecessary part is removed from doc and it is updated with proper example URL

**Checklist**:

- [ x] Documentation
- [- ] Tests
- [x ] Ready to be merged 

Just a doc jira so no need to verify tests. Cannot add to contributors as there appears to be some issue with all-configurators-cli tool.

### DIFF
--- a/packages/homepage/content/docs/2-importing.md
+++ b/packages/homepage/content/docs/2-importing.md
@@ -14,7 +14,7 @@ The Create Wizard shows you all public templates that are currently available, a
 
 ## Import from GitHub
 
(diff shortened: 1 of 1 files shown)


CompuIves:
Hey! I'm not sure if I fully understand the removal of that part of the docs. It still holds, if you copy everything after github.com/ and past it after codesandbox.io/s/github/ it will import that GitHub repo.

fraindz:
The current import wizard at https://codesandbox.io/s/github expects actual github URL and automatically converts it to sandbox URL. User is not required(and not allowed) to enter sandbox URL anymore.

Updated the PR with same information. I was naive to expect that all maintainers would be aware of all functionality changes.

CompuIves:
> User is not required(and not allowed) to enter sandbox URL anymore.

That's right, it's not required, but it's still allowed. I think it makes sense to note it because it shows how simple the workings behind it actually is. The import util does the exact same thing, it uses a regex to get the last part and puts it after codesandbox.io/s/github.

I'm going to close this PR as I think that it's fine as it is. I'll reopen it if my mind changes though.

PR #1529, opened 2019-02-09
Move Now Logic to Service

SaraVieira:
This is a clean up so we can call a service instead of talking directly to now moving the responsibility to node for faster deploys and better maintability

All the logic and endpoints are defined in the FE

**What is the new behavior?**
Created a node service here: https://github.com/codesandbox/now-deploy

This will handle most of the things we need to do in order to deploy to now


SaraVieira:
All working!!

SaraVieira:
Closed due to innactivity

PR #1573, opened 2019-03-01
Build pipeline for common

CompuIves:
This is a build pipeline for common so we don't need custom webpack configurations for other packages.

- [x] Convert existing files to TypeScript
- [x] Add pipeline for JavaScript files
- [x] Update existing paths from `common/*` to `common/lib/*`


packages/common/src/components/Preference/PreferenceKeybinding/KeybindingInput.tsx
```diff
@@ -55,7 +68,7 @@ export default class KeybindingInput extends React.Component {
 
     if (
       this.state.recordedKeys.indexOf(upperCaseKey) === -1 &&
-      IGNORED_KEYS.indexOf(e.key === -1)
+      IGNORED_KEYS.indexOf(e.key) === -1
```
CompuIves: Bug found while converting to TS 😄

SaraVieira: TS says hi

packages/common/src/components/logos/Vue.tsx
```diff
@@ -1,13 +1,13 @@
-import React from 'react';
+import * as React from 'react';
```
SaraVieira: You have the pragma set up si `import React from 'react'`should work no?

CompuIves: Yeah as alternative I could've enabled `esModuleInterop` for TS, but I felt like I was on a roll so I changed everything 😅. No big opinion on this.

CompuIves: I caved 😂 https://github.com/CompuIves/codesandbox-client/pull/1573/commits/33f64407d8396a1ac3fa7bc7510504634f31f4a6

SaraVieira: 😍

PR #1584, opened 2019-03-07
Switch from react-tippy to @tippy.js/react for tooltips

### LINKED ISSUE #1582: Switch from react-tippy to @tippy.js/react for tooltips
# 🌈 Feature

CodeSandbox is using `react-tippy` for tooltips. I made the underlying `tippy.js` library which it uses, but not `react-tippy`.  I think it's using a version of tippy.js from 2017 – since then, many bugs have been fixed and features released. It's been iterated on/improved a lot and is much more stable and mature now in 2019.

[`@tippy.js/react`](https://github.com/atomiks/tippy.js-react) is the official React wrapper I released and is using the latest version (v4).

Some benefits out of the box:

- Smaller size compared to react-tippy
- Better appearance (animations are quicker, tooltips are smaller). The current ones feel quite large and slugglish at the moment. Technically you can fix this with react-tippy, but still
- Guaranteed updates (react-tippy is essentially unmaintained now)

Obviously there are quite a few breaking changes (tooltip DOM is different, React API is different), but I believe the switch should be worth it

SaraVieira:
This PR removes react-tippy to  @tippy.js/react

closes #1582 

@atomiks


atomiks:
Nice work 🎉 

I think the manually controlled tooltips (e.g. update tooltip) might be missing the conversion over to the new API?

atomiks:
https://github.com/CompuIves/codesandbox-client/blob/500b18c99e67f872748feaa54139af32251eb970/packages/app/src/app/pages/Sandbox/Editor/Header/UpdateFound/index.js

This will need to be updated to the new API. I'm not familiar with CodeSandbox codebase so maybe there are others too. Are there tests for this?

SaraVieira:
@atomiks You mean this file: packages/app/src/app/pages/Sandbox/Editor/Header/UpdateFound/index.js?

SaraVieira:
This is the only file that did contain new things that wern't known to me :/

What is the new API for open (isEnabled?) and trigger?

atomiks:
- `open` => `isVisible`

That's it 😅

SaraVieira:
Oh! Nice!

Done!

Also is this file okay?

https://github.com/CompuIves/codesandbox-client/blob/69b0360bed301f50327bd780138e7fa6eb5eb42f/packages/common/src/components/Tooltip.tsx

atomiks:
I think you should be spreading the props into `<Tippy />` rather than the `<span>`. That would remove the need to repeat the props for every option you need

SaraVieira:
That doesn't work because when I do that and have some option that is actually for styled components the app breaks saying that the option isn't valid :(

Example:

```
      <ActionTooltip
        disabledAction={!onClick}
        title={placeholder || tooltip}
      >
        <Icon />
      </ActionTooltip>
```

That disabled action is something for styled components but bad things happen


Event the style tags was breaking it :(

atomiks:
You can destructure the invalid props out and only pass valid ones to tippy (assuming the invalid ones can be known beforehand)

SaraVieira:
I don't because since we are using styled components you can pass anything.

I could know the valid ones and do it that way by going to the tippy website and destructure those

atomiks:
There are almost 50 options so that would suck 😱 

But you could import the underlying library `tippy.js`.

It exposes the default options object on `tippy.defaults`, so you can loop through those keys to verify!

SaraVieira:
Is there not a way to tell Tippy to ignore those things? And I pass style to the span?

atomiks:
See my edit(s), I made like 100 of them 😰 

Does that answer your question?

SaraVieira:
Sorry, edits where? :(

atomiks:
https://github.com/CompuIves/codesandbox-client/pull/1584#issuecomment-470465638

SaraVieira:
I think I am done, just need the new api for hideOnClick

SaraVieira:
Ignore that ahah

atomiks:
I think the `placement` destructure is not necessary, apart from that looks good

SaraVieira:
It should all 100%

Fixed all the props because those props attached were not the best idea so they have been moved to a container and I passed style and css to the span :)

SaraVieira:
> I think the placement destructure is not necessary, apart from that looks good

Removed 😄 

One last look?

Thank you :)

atomiks:
I don't know how to test it but from my cursory glance it looks OK, I guess we should wait for @CompuIves  to approve ✅

CompuIves:
Yes! I've been wanting to upgrade Tippy for a long time. 

This is quite a big change, so it will require some more testing. One thing I noticed is that for icons we only get a tooltip when we hover over the icon instead of over the container with this change. I assume this happens because all elements are now wrapped in a `span`:

screenshot 2019-03-07 at 11 53 12

Is there a way to retain the simplicity without a `span` wrap?

atomiks:
I think the tooltips should be wrapping the buttons instead?

`react-tippy` used to use a wrapper `<div>` element which is what the tooltip was applied to

SaraVieira:
That is probably a css issue, since react tippy used a div and a div is block but a span is not. 

Adding a div and setting always to 100% of the size fixes that @CompuIves :)

CompuIves:
Nice! The icons work!

I found some layout problem here:

image

Other than that, works!

SaraVieira:
Fixed 

screenshot 2019-03-07 at 14 00 27

SaraVieira:
To be merged after editor redisign

atomiks:
Great work on the redesign! I notice you're using tooltips with delays for buttons next to each other in the new design. The `<TippyGroup />` wrapper component is perfect for this use case

https://atomiks.github.io/tippyjs/misc/#groups

CompuIves:
This can be merged after the conflicts are done! Good tip @atomiks, seems like exactly what we need!

SaraVieira:
Duuude, I have no idea how to even start on these ahah

lbogdan:
Ouch! 😄

SaraVieira:
SHIP IT

CompuIves:
YES! Thank you @atomiks and @SaraVieira!!

SaraVieira:
Omg yay! One down! 6 to go

Issue #1280, opened 2018-11-06
Feature: "Empty trash" button


manniL:
Heya! :wave: 

As the trash becomes more important with server-side CSBs, as the limit is fairly lower in comparison to client-side ones and the trashed ones count as well, I'd love to see a "delete all from trash" button (with an extra confirmation dialog or so).

zixuan75:
Trash?

SaraVieira:
ohhhhh

This has actually been fixed https://github.com/CompuIves/codesandbox-client/pull/1592

PR #1592, opened 2019-03-12
Dashboard improvements

SaraVieira:
When there are folders in a folder it will show the folders inside that folder in the folder.

/giphy magic

Also adds an empty trash button


packages/app/src/app/pages/Dashboard/utils/getChildCollections.js
```diff
@@ -0,0 +1,15 @@
+import getDirectChildren from './get-direct-children';
+
+export default (folders, path = '/') => {
```
CompuIves: Could this become a named function? That way we have nicer auto-completions (eg. `getChildCollections` would autocomplete and auto import).

packages/app/src/app/pages/Dashboard/utils/getChildCollections.js
```diff
@@ -0,0 +1,15 @@
+import getDirectChildren from './get-direct-children';
```
CompuIves: It'd be best if this file would be renamed to `get-child-collections` for consistency.

packages/app/src/app/pages/Dashboard/Content/routes/PathedSandboxes/Folders.js
```diff
@@ -0,0 +1,55 @@
...
+  let folderInfo = {};
+  if (!loading) {
+    folderInfo = getChildCollections(
+      me.collections,
+      (me.collection || {}).path
+    );
+  }
+
+  const { children, foldersByPath, folders } = folderInfo;
+  return !loading && (children || []).size ? (
```
CompuIves: I think it would be more readable to do an early return here:

```
if (loading) {
  return null;
}

const folderInfo = getChildCollections(
  me.collections,
  me.collection.path
);

if (folderInfo.children.size === 0) {
  return null;
}
```

This saves a lot of null checks.

packages/app/src/app/pages/Dashboard/Content/routes/PathedSandboxes/Folders.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import { withRouter } from 'react-router-dom';
+import FolderEntry from '../../../Sidebar/SandboxesItem/FolderEntry';
+import getChildCollections from '../../../utils/getChildCollections';
+
+import { Folder, FoldersWrapper } from './elements';
+
+const getPath = name => {
```
CompuIves: It's probably better to use `react-router`  for this (`props.match.params.path`). This may break if someone has a folder  called `sandboxes`.

packages/app/src/app/pages/Dashboard/Content/routes/PathedSandboxes/index.js
```diff
@@ -67,6 +65,9 @@ const PathedSandboxes = props => {
                 isLoading={loading}
                 possibleTemplates={possibleTemplates}
                 Header={<Navigation teamId={teamId} path={path} />}
+                Folders={
```
CompuIves: This could be renamed to something more generic like `SubHeader`. This way we can maybe use it in the future for the `Recent Sandboxes` or `Search` page as well.

CompuIves:
I realized that making the folder card behave like a sandbox card is suuuper complicated (like dragging and handling moving dirs AND sandboxes at the same time). This needs some thinking.

packages/app/src/app/pages/Dashboard/utils/getChildCollections.js
SaraVieira: Changed and also changed getMostUsedTemplate to get-most-used-template

packages/app/src/app/pages/Dashboard/Content/routes/PathedSandboxes/Folders.js
SaraVieira: Good point! Changed :)

packages/app/src/app/pages/Dashboard/Content/routes/PathedSandboxes/index.js
SaraVieira: Changed :)

packages/app/src/app/pages/Dashboard/Content/routes/PathedSandboxes/Folders.js
SaraVieira: omg so much clean

```js
  const getPath = name => (params.path ? `${params.path}/${name}` : '/' + name);
```

SaraVieira:
Folders are now wrapped to and I think the selecting should be left for later because of complexity

CompuIves:
Very nice! Another thing I saw is that the scroll behaves a bit weird, there is a permanent  scrollbar and scrolling down always shows the sandboxes in view. I think this can be "fixed" by including the folders in `react-virtualized`, but that requires it to be added to `Sandboxes` I think.

I'm thinking the best way is to have an extra argument to `Sandboxes` with `folders` and let `SandboxGrid` do the rendering of the folders, this way it renders correctly with scroll and automatically works with selection. But this is quite complicated behaviour, and will probably take a couple days to implement.

I'm thinking whether there is a different way to fix scroll as first version.

SaraVieira:
Dude I have no idea what you mean 😅

SaraVieira:
closes #1280

SaraVieira:
Commented out

CompuIves:
Nice!

PR #1605, opened 2019-03-16
Make CodeSandbox logo icon clickable again

tzhelyazkova:
### Steps to reproduce
- Create a new sandbox
- Give it a name (preferably something long)
- Try to click the codesanbox logo in the left top corner
### New Changes
The navigation container's centered style overrides the position of the logo's <a> element.
Removing `position: absolute` makes it clickable again. The title of the sandbox is not positioned the same way as before though.


SaraVieira:
Looks 💯 

We need to add a max width to this that will also overflow hidden so that people don't put huge names but this works great and doesn't seem to break anything!

Thank you!

You win a macdonalds burguer! Not a choice! You have to eat it

PR #1607, opened 2019-03-18
Fix small user testing issues

SaraVieira:
Fixes three small things that confused a user

- Carousel now goes around when you click next in the last sandbox
- The text in custom vscodeTheme is now a placeholder
- Text at the top for renaming a sandbox is a placeholder if the name is not set yet. otherwhise it's the value


CompuIves:
Very nice!

PR #1609, opened 2019-03-18
Search Improvements

SaraVieira:
In a user test it was pointed we can't order the search and it makes sense to do so

This PR adds order by likes and views

This PR also:

* Changes the cards in search for the new cards we have in the explore page
* Removes how long the search took as that is not important information for the user
* Changes the API of wide sandbox to have the prop `selectSandbox` instead of `pickSandbox` to make it more generic
* Adds common script
* Moves some .css files to gobal js styles so we can use the theme

Design
https://www.figma.com/file/akq4c3BMhkW5EHipeIBEbKmf/Search-Page?node-id=0%3A1


SaraVieira:
cc @Haroenv

Haroenv:
The sorting of "relevance" and "views" seems very similar to me, was a different sort required for this? 

I'm seeing some flashes when I change index, but I don't see any particular reason why in the code by a quick glance, maybe it already was there before



Also, it seems like the search is pretty hidden with the new design, I can't seem to find any link towards it anymore, I had to guess the URL.

Finally there's a small other thing (maybe only on Safari), and that's the search icon isn't correctly positioned: 

Screenshot 2019-03-21 at 15 56 36

SaraVieira:
Good points!

I deleted the views one as that is basically the default one and also added a link to the search in the header

The other two:

The flickering seems to only happen on safari :/

I'm looking at the CSS thing now

CompuIves:
The bottom part of the card (with description) seems a bit big to me, does it have different styles compared to `SandboxCard` in `common`?

SaraVieira:
@CompuIves Nop, itmay be because most of them don't have a description

packages/app/src/app/pages/Search/Results/index.js
```diff
@@ -2,17 +2,30 @@ import React from 'react';
...
 import { Container } from './elements';
 
 function Results() {
   return (
     <Container>
       <ResultInfo />
-      <Hits hitComponent={SandboxCard} />
+      <Hits
+        hitComponent={({ hit }) => (
+          <WideSandbox
```
CompuIves: It does seem a bit weird, after checking it seems like the description of the sandbox in search is ~150px and in explore ~130px height

packages/common/src/components/WideSandbox/index.tsx
```diff
@@ -34,7 +34,7 @@ export type Props = {
   small?: boolean;
   defaultHeight?: number;
   noMargin?: boolean;
-  pickSandbox: (
+  selectSandbox: (
```
CompuIves: Nice!!

packages/app/src/app/pages/Search/Filters/index.js
```diff
@@ -1,12 +1,22 @@
...
+import { ALGOLIA_DEFAULT_INDEX } from 'common/lib/utils/config';
 
 import Filter from './Filter';
+import Sort from './Filter/Sort';
 import { Container } from './elements';
 
 function Filters() {
   return (
     <Container>
+      <Sort
```
CompuIves: This is nice! Maybe we can also re-use 

image

in a later version.

package.json
```diff
@@ -75,7 +76,8 @@
     "gulp": "^3.9.1",
     "lerna": "^2.5.1",
     "lint-staged": "^5.0.0",
-    "prettier": "^1.8.2"
+    "prettier": "^1.8.2",
+    "typescript": "^3.3.4000"
```
CompuIves: Could this one be `^3.3.1`? That way we ensure that we have a single typescript version in the whole repo.

SaraVieira: Fixed

PR #1611, opened 2019-03-19
Blog Page

SaraVieira:
👋

This PR adds a blog page that gets the articles from medium and the local directory

It combines them and orders them to show by date


packages/homepage/gatsby-node.js
```diff
@@ -49,6 +50,69 @@ exports.createPages = async ({ graphql, actions }) => {
...
+    }
+  `);
+
+  blogsFromMedium.data.allFeedMediumBlog.edges.forEach(edge => {
+    const slug = edge.node.title
+      .toLowerCase()
+      .replace(/[^\w ]+/g, '')
+      .replace(/ +/g, '-');
+    const id = edge.node.id;
+    if (edge.node.categories) {
```
SaraVieira: If no categories its a comment

packages/homepage/src/utils/makePosts.js
```diff
@@ -0,0 +1,65 @@
+// Im so sorry about this whole file
```
SaraVieira: 🙃

packages/homepage/gatsby-config.js
```diff
@@ -71,10 +71,21 @@ module.exports = {
...
+      },
+    },
     {
       resolve: `gatsby-plugin-google-fonts`,
       options: {
-        fonts: [`Poppins:600,700,800`, `source sans pro:300,400,500,600,700`],
+        fonts: [
+          `Poppins:600,700,800`,
+          `source sans pro:300,400,500,600,700`,
+          'open sans:400',
```
CompuIves: What do we use open sans for?

packages/homepage/gatsby-node.js
CompuIves: Would be good to add this as a comment (hah!) so future readers know this too.

packages/homepage/gatsby-node.js
```diff
@@ -49,7 +50,72 @@ exports.createPages = async ({ graphql, actions }) => {
...
+    `
+  );
+
+  if (allMarkdownArticles.data) {
+    allMarkdownArticles.data.allMarkdownRemark.edges.forEach(edge => {
+      const slug = edge.node.frontmatter.slug;
+      const id = edge.node.id;
+
+      createPage({
+        path: 'post/' + slug,
```
CompuIves: Would it make sense to do `/blog/` prefix?

packages/homepage/src/components/PostElements.js
```diff
@@ -0,0 +1,40 @@
...
+  font-family: 'Poppins';
+  font-weight: 600;
+  font-size: 30px;
+  line-height: 1.5;
+
+  color: #f2f2f2;
+`;
+
+export const PostDate = styled.span`
+  font-family: 'Poppins';
```
CompuIves: I know that I've been doing this wrong as well, but would be better to have this as a fallback.

```suggestion
  font-family: 'Poppins', sans-serif;
```

packages/homepage/src/pages/blog.js
```diff
@@ -0,0 +1,122 @@
...
+  Aside,
+  Header,
+  PageTitle,
+  PageSubtitle,
+} from './_blog.elements';
+
+import Layout from '../components/layout';
+import { makeFeed } from '../utils/makePosts';
+
+// UNCOMMENT AT THE BOTTOM IF IT BREAKS
```
CompuIves: Will this be uncommented?

packages/homepage/src/templates/docs.js
```diff
@@ -274,7 +274,6 @@ export const pageQuery = graphql`
             title
           }
           fields {
-            slug
```
CompuIves: Don't we need slug for docs?

packages/homepage/src/utils/makePosts.js
CompuIves: I like it!

packages/homepage/gatsby-config.js
SaraVieira: The text in the post single page

packages/homepage/src/pages/blog.js
SaraVieira: Only when someone adds a markdown blog post

packages/homepage/src/templates/docs.js
SaraVieira: They don't even have them in the code ahah

PR #1633, opened 2019-03-21
Update manifest.json

lifeiscontent:
Fix Top-bar color when in standalone mode.

changing the top bar color from blue to the black theme color

Screen Shot 2019-03-20 at 10 45 09 PM


lifeiscontent:
@CompuIves I thought the black color would look better, what do you think?

CompuIves:
Agree! Nice work!

PR #1667, opened 2019-03-28
Disable fork button while forking

### LINKED ISSUE #844: Fork button doesn't show feedback while forking
# 🐛 bug report

## Description of the problem
Today at React-Europe, we used CodeSandbox (again) and had to fork a sandbox. However, the wifi was pretty slow and thus the fork was not immediate. Thus, it feels like it wasn't working so I clicked again (many times ^^). Then the sandbox reloaded several times with each time a new fork.

## Suggestion
Display a loader while forking or disable the button.

ValentinH:
Fixes #844.

This is a first improvement:
fork-btn

After this, I'd like to work on a better feedback, maybe a circular loader inside the button. What do you think?

**Checklist**:
- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged 

### DIFF
--- a/packages/app/src/app/pages/Sandbox/Editor/Header/index.tsx
+++ b/packages/app/src/app/pages/Sandbox/Editor/Header/index.tsx
@@ -44,7 +44,10 @@ type ButtonProps = {
   signals: any;
   style: React.CSSProperties;
   secondary?: boolean;
-  disabled?: boolean;
+};
+
+type ForkButtonProps = ButtonProps & {
+  isForking: boolean;
 };
 
 const LikeButton = ({
@@ -65,19 +68,24 @@ const LikeButton = ({
   />
 );
 
-const ForkButton = ({ signals, secondary, disabled, style }: ButtonProps) => (
+const ForkButton = ({
+  signals,
+  secondary,
+  isForking,
+  style,
+}: ForkButtonProps) => (
   <Button
     onClick={() => {
       signals.editor.forkSandboxClicked();
     }}
     style={style}
     secondary={secondary}
-    disabled={disabled}
+    disabled={isForking}
     small
   >
     <>
       <Fork style={{ marginRight: '.5rem' }} />
-      Fork
+      {isForking ? 'Forking' : 'Fork'}
     </>
   </Button>
 );
@@ -212,7 +220,7 @@ const Header = ({ store, signals, zenMode }: Props) => {
         />
         <ForkButton
           secondary={sandbox.owned}
-          disabled={store.editor.isForkingSandbox}
+          isForking={store.editor.isForkingSandbox}
           style={{ fontSize: '.75rem' }}
           signals={signals}
           store={store}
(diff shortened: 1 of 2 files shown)


lbogdan:
I think it would also be nice to have a *positive* feedback, like maybe changing the text to "Forking...".

ValentinH:
Totally agreed, I've just updated the PR.

SaraVieira:
Hey!!

I was also just working on this! Bless you!! 

There is also a couple more places we have that button, it would be awesome to have that there too 😁

ValentinH:
@SaraVieira I've just checked and found this place where there is also a fork button:
image

Do you see any other places?

SaraVieira:
Awesome!! 

I'm at home and my laptop is in the car but the code looks great and I will check the UI Tomorrow morning and most likely merge it 😁

Can you add to the contributors table ?

ValentinH:
I'm already in the contributors table 🙂

ValentinH:
What do you guys think about this animation?

animated-fork

SaraVieira:
I really like it! Nice css loading 🎉

packages/app/src/app/pages/Sandbox/Editor/Header/index.tsx
```diff
@@ -64,18 +68,24 @@ const LikeButton = ({
...
     }}
     style={style}
     secondary={secondary}
+    disabled={isForking}
     small
   >
     <>
       <Fork style={{ marginRight: '.5rem' }} />
-      Fork
+      {isForking ? 'Forking' : 'Fork'}
```
CompuIves: I think it would be best to name it "Forking..." with the three dots to indicate an action.

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/More/index.tsx
```diff
@@ -27,15 +27,21 @@ class More extends React.PureComponent<Props> {
...
           {!owned ? (
-            <Button small block onClick={this.forkSandbox}>
-              Fork Sandbox
+            <Button
+              small
+              block
+              disabled={isForkingSandbox}
+              onClick={this.forkSandbox}
+            >
+              {isForkingSandbox ? 'Forking Sandbox' : 'Fork Sandbox'}
```
CompuIves: Same here, adding 3 dots `...` would be nice.

packages/app/src/app/pages/Sandbox/Editor/Header/index.tsx
ValentinH: Yes, i was hesitating between both because the dots makes the button really larger while forking. Now you convinced me :)

ValentinH:
I'll keep the animation for a follow-up PR. I'll update following requested changes once I have access to my computer :)

lbogdan:
Also, maybe make the button a bit wider, so that when you change the text to "Forking" it doesn't do that annoying wiggle.

ValentinH:
I think we need to decide between both solutions, because having "Forking..."  will force to have the Fork button pretty large. This will look weird compare to siblings buttons. 
What do you think?

ValentinH:
For the animation, being fancy depends on whether we want to change the button markup or not. In my screenshot, I only used one before pseudo element.

CompuIves:
Just played with it, works really well! Thanks for this contribution @ValentinH!

PR #1681, opened 2019-04-01
Start Preview's history at index of 0

pshrmn:

Bug fix

The browser preview starts with an empty `history` array in state and a `historyPosition` of `-1`. After initializing, the URL listener dispatches a `REPLACE`, which adds an entry to the preview's `history` array, but the `historyPosition` is still `-1`. This causes two incorrect behaviors: 1) the preview thinks that it can go forward immediately and 2) the preview can never return to the initial location.

(I made the `REPLACE` change in #1158, so I suppose that this bug is my fault. I'm not sure how I didn't notice the issue until now, but I didn't spot any other changes, so it must have existed since that PR.)

**What is the new behavior?**

By starting with the `historyPosition` of `0`, the preview 1) cannot go forward immediately and 2) can return to the initial location.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 

### DIFF
--- a/packages/common/src/components/Preview/index.tsx
+++ b/packages/common/src/components/Preview/index.tsx
@@ -170,7 +170,7 @@ class BasePreview extends React.Component<Props, State> {
     this.state = {
       frameInitialized: false,
       history: [],
-      historyPosition: -1,
+      historyPosition: 0,
       urlInAddressBar: this.serverPreview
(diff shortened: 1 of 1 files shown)


CompuIves:
This is great! Thanks a lot for fixing this.

PR #1682, opened 2019-04-01
add min width 100 to both panes

### LINKED ISSUE #1615: 🐛 Preview Pane can be Resized Off-Screen
Please refer to the attached gif for a visual explanation:

Imgur

When the user clicks and drags the separator between the code and preview areas, they can effectively re-size the preview area off-screen, at which point they can no longer drag it back into view. Refreshing the page seems to be the only way to revert this change.

Instead, perhaps have a minimum size for the preview pane to prevent this?

SaraVieira:
Adds a min-width of 100px to both panes so you can't drag anything out of view

Screenshot 2019-04-01 at 11 13 08

closes #1615

### DIFF
--- a/packages/app/src/app/pages/Sandbox/Editor/Content/index.js
+++ b/packages/app/src/app/pages/Sandbox/Editor/Content/index.js
@@ -515,6 +515,7 @@ class EditorPreview extends React.Component<Props, State> {
             }
           />
           <SplitPane
+            maxSize={-100}
             onDragFinished={() => {
               this.props.signals.editor.resizingStopped();
             }}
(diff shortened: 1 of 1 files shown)


packages/app/src/app/pages/Sandbox/Editor/Content/index.js
```diff
@@ -515,6 +515,7 @@ class EditorPreview extends React.Component<Props, State> {
             }
           />
           <SplitPane
+            maxSize={-100}
```
SaraVieira: https://github.com/tomkp/react-split-pane#maxsize

CompuIves:
Great!!

PR #1687, opened 2019-04-01
add cmd+click for opening in new tab in dashboard

### LINKED ISSUE #1024: [UX issue] Open link in New Tab when CMD or CTRL Clicked on Projects
# 🌈 Feature

I think this should be common in every website while clicking link with CMD or CTRL it should open in a new tab. I'm on https://codesandbox.io/dashboard/recent page & can't open multiple projects bcz of that.

SaraVieira:
Adds code to check if the user has click cmd(or ctrl)+ double click, if so opens the sandbox in a new tab

closes #1024


CompuIves:
Niiice!

PR #1690, opened 2019-04-01
Sandbox name resonance between sidebar and header

samrith-s:
Update `CollectionInfo` in `Header`'s render function.

This issue fixes a bug and maintains resonance of the sandbox name between the sidebar and the header.

https://github.com/CompuIves/codesandbox-client/issues/1680

**Checklist**:
- [ ] Documentation (*NA*)
- [ ] Tests (*NA*)
- [x] Ready to be merged

The issue was `CollectionInfo` was being passed a reference to a sub-sect of observable `store`, from `Header`. But MobX states that for the observable to trigger a reaction, it needs to be called in `render`. 

The quick fix is to pass a new reference of `sandbox` to`CollectionInfo`, rather than passing the observable.


CompuIves:
Great! Thanks a lot @samrith-s

packages/app/src/app/pages/Sandbox/Editor/Header/index.tsx
```diff
@@ -154,7 +154,10 @@ const Header = ({ store, signals, zenMode }: Props) => {
 
       {sandbox.owned && (
         <Centered style={{ margin: '0 3rem' }}>
-          <CollectionInfo isLoggedIn={store.isLoggedIn} sandbox={sandbox} />
+          <CollectionInfo
+            isLoggedIn={store.isLoggedIn}
+            sandbox={{ ...sandbox }}
```
CompuIves: Could you leave a comment as to why you do this? Also, maybe `sandbox.toJS()` works too.

samrith-s: Will check that out, and shall leave a comment too!

samrith-s:
@CompuIves done! Do have a look.

CompuIves:
Great! Thank you a lot, really nice to have these contributions! Merging now.

PR #1692, opened 2019-04-02
Adds search to the docs

### LINKED ISSUE #1100: Search on the docs
# 🌈 Feature

## What is this feature?
adding [DocSearch](https://community.algolia.com/docsearch) to [/docs](https://codesandbox.io/docs).

If you type in the search bar, you will find answers in the docs.

Adding an input to the sidebar, an input and a similar lazy loading strategy as on the react docs: 

https://github.com/reactjs/reactjs.org/blob/master/src/components/LayoutHeader/DocSearch.js

You have to apply (as site owner) on the DocSearch website first :)

SaraVieira:
Adds DocSearch by algolia to the docs

We didn't have any search in the docs

**What is the new behavior?**
There is search now

Screenshot 2019-04-02 at 18 44 52

closes #1100


packages/homepage/content/docs/3-configuration.md
```diff
@@ -6,7 +6,7 @@ description: "You can configure sandboxes and other options with configuration f
 
 ## Configuration Files
 
-We think that configuration is used when a program is too stupid to figure out what the user really wants, and try to steer away from it as much as possible. However, there are always advanced use cases where the user wants to control every aspect of their program, and thus configure everything. CodeSandbox is a clear example of a program that requires configuration, it doesn't fit the 95%.
+There are always advanced use cases where the user wants to control every aspect of their program, and thus configure everything. CodeSandbox is a clear example of a program that requires configuration, it doesn't fit the 95%.
```
CompuIves: Hmm, I think this line is actually a valid line. It tells something about the program, not a person. I have it from here:

https://fishshell.com/docs/current/design.html

Under "Configurability is the root of all evil". We can rewrite it to remove the word, but I think that this sentence is the most important sentence of that paragraph.

SaraVieira: Changed to: We think that configuration is used when a program can't understand what the user really wants,

Is that okay?

SaraVieira: Because even if you got it from somewhere it's still a thing where without meaning you are calling all software that requires config stupid

Haroenv:
a light on dark color scheme would fit better here, but not that obvious to make

SaraVieira:
@Haroenv I tried that but It didn't look at good and it wasn't very readable that why I stuck with the light one and changed the colors to be ours :/

CompuIves:
I noticed that the search is not there on other pages (eg. check here https://codesandbox.io/docs/git).

packages/homepage/content/docs/3-configuration.md
CompuIves: > "We think that configuration is used when a program can't understand what the user really wants"

Yes, that's okay! I understand that the `stupid` can offend people, good suggestion to change it to this.

SaraVieira: It's what happens when you live in berlin for long 😂

SaraVieira:
@Haroenv bas made me do a dark theme

CompuIves:
Some laaaast nitpicks, then we're ready to roll!

image

I would remove the header "Search our docs" and put it as the placeholder, looks cleaner. I would make the placeholder a bit more gray in color, to make it look more like nothing is entered yet. I would also make the `border-color` dark when not focused, otherwise it can look like the search is focused but in reality it isn't (I think codesandbox.io/search is a good example).

Other than that it looks really great! Can't wait to have this live, the docs were getting quite big and unmanageable 😅.

SaraVieira:
@CompuIves All done :)

packages/homepage/src/components/DocSearch.js
```diff
@@ -0,0 +1,62 @@
...
+  };
+  componentDidMount() {
+    // Initialize Algolia search.
+    // eslint-disable-next-line no-undef
+    if (window.docsearch) {
+      window.docsearch({
+        apiKey: '45db7de01ac97a7c4c673846830c4117',
+        indexName: 'codesandbox',
+        inputSelector: '#algolia-doc-search',
+        debug: true,
```
s-pace: You can set this param to `false` to close the dropdown once the user click outside of it.

s-pace:
```suggestion
        debug: false,
```
You can set this param to false to close the dropdown once the user click outside of it.

SaraVieira: Oh yeah, I forgot to turn that back to false :/

Done!

Haroenv:
The PR deploy seems to be empty on the / and /docs

SaraVieira:
I can see them here 
http://pr1692.cs.lbogdan.tk/docs

Haroenv:
Looks like with the dark theme, you overrode the "highlighted" item with a dark background, so you can no longer use a keyboard to use it reliably :)

packages/homepage/src/components/DocSearch.js
```diff
@@ -0,0 +1,62 @@
...
+      // eslint-disable-next-line
+      this.setState({ enabled: false });
+    }
+  }
+
+  render() {
+    const { enabled } = this.state;
+
+    return enabled ? (
+      <form>
```
CompuIves:
```suggestion
      <form style={{ margin: 0 }}>
```

CompuIves:
Left one last suggestion, you can merge it in after doing that one!

PR #1694, opened 2019-04-02
fix(gridsome): Update logo + color

tomtev:
The current logo does not show on my Chrome. It's a problem with the SVG markup.
Looks like fill="url(#a)" / id="a" does not work.  (But fill="url(#aa)" / id="aa" works :) ).

Anyway, I have updated the logo to a new one that doesn't have this problem.


SaraVieira:
Awesome! 

I will never fully understand this chrome fill bug 

Thank you!

PR #1700, opened 2019-04-03
add arrowParens to prettier config

SaraVieira:
Also adds a max height to make it scrollable

### DIFF
--- a/packages/app/src/app/components/CodeEditor/Configuration/elements.ts
+++ b/packages/app/src/app/components/CodeEditor/Configuration/elements.ts
@@ -4,8 +4,8 @@ export const Container = styled.div`
   padding: 1rem;
   color: rgba(255, 255, 255, 0.8);
   box-sizing: border-box;
-
-  overflow: auto;
+  max-height: 100%;
+  background: red;
 `;
 
 export const Title = styled.h1`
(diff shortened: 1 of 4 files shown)


packages/app/src/app/components/CodeEditor/Configuration/elements.ts
```diff
@@ -4,8 +4,8 @@ export const Container = styled.div`
   padding: 1rem;
   color: rgba(255, 255, 255, 0.8);
   box-sizing: border-box;
-
-  overflow: auto;
+  max-height: 100%;
+  background: red;
```
CompuIves: 😅?

packages/common/src/templates/configuration/prettierRC/ui.tsx
```diff
@@ -152,6 +152,19 @@ export class ConfigWizard extends React.Component<ConfigurationUIProps> {
             line instead of being alone on the next line.
           </ConfigDescription>
         </PaddedConfig>
+        <PaddedConfig>
```
CompuIves: Nice! I think we also need to add this to the preferences: `packages/app/src/app/pages/common/Modals/PreferencesModal/CodeFormatting/Prettier/index.js`

packages/app/src/app/components/CodeEditor/Configuration/elements.ts
SaraVieira: LOL! Removed, it was the one that didn't work ahah

packages/common/src/templates/configuration/prettierRC/ui.tsx
SaraVieira: Added there too :)

CompuIves:
Great! Looks good!

PR #1711, opened 2019-04-03
Bugfix/cross origin error

### LINKED ISSUE #667: how to get rid of cross-origin error?
**IN FIREFOX**: When I execute my code the typical error I should get is: "TypeError: Cannot read property 'setState' of undefined", instead I received a very weird cross-origin error.

Here is a screenshot of the error:
http://prntscr.com/iwipnb
image

> Error
> A cross-origin error was thrown. React doesn't have access to the actual error object in development. See https://fb.me/react-crossorigin-error for more information.

here is my code: https://codesandbox.io/s/4885l37xrw

How can I avoid the cross-origin error in Codesandbox?

AndyOGo:
fixes #667 

Currently `eval` is used to execute scripts.

**What is the new behavior?**
Now `new Function` constructor is used to execute scripts.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 


AndyOGo:
@CompuIves React recommends to use `cheap-module-source-map` for sourcemaps and webpack.
> If you use Webpack, we recommend using the `cheap-module-source-map` setting in development to avoid this problem.
https://reactjs.org/docs/cross-origin-errors.html#source-maps

So I would change that too:
https://github.com/CompuIves/codesandbox-client/blob/3bbf7128bc9712ac6ff121908dc34c625052a6af/packages/app/config/webpack.dev.js#L21

packages/app/config/webpack.dev.js
```diff
@@ -18,7 +18,7 @@ module.exports = merge(
   },
   commonConfig,
   {
-    devtool: 'eval',
+    devtool: 'cheap-module-source-map',
```
AndyOGo: Is React the only known affected library?
https://reactjs.org/docs/cross-origin-errors.html#source-maps

AndyOGo:
@CompuIves 
hmh, running `yarn test` locally works 🤔 
Above test-integrations errors are all related to `*.png` snapshots, what are these about?

AndyOGo:
@CompuIves Just made a quick test run locally and it still works:)

Screenshot 2019-04-03 at 21 56 56

packages/app/src/sandbox/eval/loaders/eval.js
```diff
@@ -41,8 +41,8 @@ export default function(
       `) {` +
       code +
       `\n})`;
-    // eslint-disable-next-line no-eval
-    (0, eval)(newCode).apply(
+    // eslint-disable-next-line no-new-func
+    new Function(`return ${newCode}`)().apply(
```
AndyOGo: @CompuIves 
I couldn't find details on collaboration and doclets.
Do your prefer complementary inline comments?
E.g. with a link to https://reactjs.org/docs/cross-origin-errors.html#source-maps and/or the issue #667

CompuIves:
Wow, thanks for this contribution! I'm wondering, does using `Function` over `(0)(eval)` fix the cross-origin error from React?

lbogdan:
Looks like this breaks source-maps, though?

AndyOGo:
Seems not, as the react docs says:
> Some JavaScript bundlers may wrap the application code with `eval` statements in development. (For example Webpack will do this if `devtool` is set to any value containing the word “eval”.) This may cause errors to be treated as cross-origin.
https://reactjs.org/docs/cross-origin-errors.html#source-maps

AndyOGo:
But to be 100% sure we should double check by testing a currently broken example

AndyOGo:
In that case I will undo the sourcemap config of webpack

packages/app/config/webpack.dev.js
AndyOGo: @lbogdan 
I revert the commit changing the webpack config:
https://github.com/CompuIves/codesandbox-client/pull/1711/commits/61f828e282cc07102063a0c5d97339f6d0a75104

AndyOGo:
Here is a reproduction script:
https://codesandbox.io/embed/ly0q02y9q9

AndyOGo:
@CompuIves 
@lbogdan 
I can approve, by running my [reproduction test](https://codesandbox.io/embed/ly0q02y9q9) locally I don't have the cross origin issue now:)

Instead I get the expected error message of:
> TypeError
> Cannot assign to read only property 'children' of object '#<CECrossOriginError>'

Screenshot 2019-04-10 at 20 17 59

AndyOGo:
Thanks @CompuIves 
I'm keen to see it in production, if the browser treats `new Function` same as `eval`ed code with special origin rules for those dynamic scripts 🤔 

If that fails in prod, other solutions whiich come to my mind are:
 - CSP header for [`script-src` of `'unsafe-eval'`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src)
 - some CORS header
 - last option would be dynamically inserted `<script>` tags with inline code (but could a DOM performance issue).

BTW:
The only CI test that fails is:
https://circleci.com/gh/CompuIves/codesandbox-client/20207?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
> Your job ran 38 tests with 3 failures
>
>    sandboxes loads the sandbox with id 'github/reactjs/redux/tree/master/examples/real-world' - sandboxes loads the sandbox with id 'github/reactjs/redux/tree/master/examples/real-world'
>
>Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
>
>sandboxes loads the sandbox with id 'ymjwwrw2rj' - sandboxes loads the sandbox with id 'ymjwwrw2rj'
>
>Error: Expected image to match or be a close match to snapshot.
>
>sandboxes loads the sandbox with id 'zw9zjy0683' - sandboxes loads the sandbox with id 'zw9zjy0683'
>
>Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.

lbogdan:
PR build server is up again!

AndyOGo:
Thanks @lbogdan .
Unfortunately the `new Function` constructor has the same **cross origin error* issue:
http://pr1711.cs.lbogdan.tk/s/jzq5rlmr0w
Screenshot 2019-04-18 at 21 42 38

packages/app/scripts/start.js
```diff
@@ -170,6 +171,14 @@ function openBrowser(port, protocol) {
 }
 
 function addMiddleware(devServer, index) {
+  devServer.use(
```
AndyOGo: @CompuIves @lbogdan 
I don't know how many servers you are using.
From a first glimpse I think you have this webpackDevServer, an iframe with the running app, some proxies and it seems it's hosted in the cloud.

Any other CSP settings you deem necessary or any other servers?

AndyOGo: Am I right, that nginx is also used at prod servers?

AndyOGo:
Closing, because `new Function` Constructor doesn't help.

I suggest to go with proper **CSP** HTTP headers or `<meta>` tags.

> According to the **Content Security Policy** spec, a [`<meta>` tag](https://w3c.github.io/webappsec-csp/#meta-element) can also be used to set [`script-src`](https://w3c.github.io/webappsec-csp/#script-src) 🎉 Like:
>
>```html
><meta http-equiv="Content-Security-Policy" content="script-src 'self' 'unsafe-eval'">
>```
>
>That doesn't even need a server change and would cover all cases, not only React.

PR #1716, opened 2019-04-04
Alert frozen

### LINKED ISSUE #1212: Prompt on saving frozen sandbox

SaraVieira:
Alert when sandbox is frozen

Closes  #1212


packages/app/src/app/store/sequences.js
```diff
@@ -172,15 +178,35 @@ export const forkSandbox = sequence('forkSandbox', [
...
+  },
+]);
+
 export const ensureOwnedEditable = sequence('ensureOwnedEditable', [
-  when(
-    state`editor.currentSandbox.owned`,
-    state`editor.currentSandbox.isFrozen`,
-    (owned, frozen) => !owned || frozen
-  ),
+  when(state`editor.currentSandbox.owned`, owned => !owned),
```
CompuIves: Can we reverse this one? It's a bit confusing for me to see a negation here 😅

SaraVieira: We can't reverse because of the old one checked if it was frozen too

What do you suggest?

CompuIves: I'm thinking of this:

```
  when(state`editor.currentSandbox.owned`, owned => owned),
  {
  	true: [
      when(state`editor.currentSandbox.isFrozen`, isFrozen => isFrozen),
      {
        true: forkFrozenSandbox,
        false: [],
      },
    ],
    false: forkSandbox,
  },
```

CompuIves: So flipping the true/false statements and setting `owned => owned`

SaraVieira: Done :)

CompuIves:
Great!

PR #1720, opened 2019-04-04
fix UX confusion with Github

### LINKED ISSUE #1209: Import from Github UX confusion.
# 🐛 bug report

## Description of the problem

The UX for importing a repo from Github is not great. It took me ~10 minutes to figure out how to do it.

## How has this issue affected you?

It has made me 10 minutes older.

## What are you trying to accomplish?

I am trying to get younger.

😆 When I clicked "create sandbox" and then "import from Github", I really expected to land on the form where you provide the repo URL, but I landed on the docs view, instead, which confused me.

I felt like I arrived at the wrong place, so instead of reading docs, I naturally went back to see if there was another option for importing from Github.

I believe the user should be provided a link to the import form, as well as a link to the importing docs, both from the "create a sandbox" view. 👍

SaraVieira:
Adds a link from the import from Github to the actual link to import from GitHub.

In there is a link to the docs

closes #1209

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Content/CreateNewSandbox/Modal/index.js
+++ b/packages/app/src/app/pages/Dashboard/Content/CreateNewSandbox/Modal/index.js
@@ -82,10 +82,7 @@ export default class Modal extends React.PureComponent {
             </Tab>
           ))}
           <ImportChoices>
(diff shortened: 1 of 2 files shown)


CompuIves:
Nice!!

PR #1733, opened 2019-04-05
Show a loader on the button while forking

ValentinH:
This is a follow-up of #1667.

animated-fork

The effect is simple because for now I didn't want to modify the markup of the button.

With my limited CSS skills, I think @CompuIves suggestion regarding nzbin.github.io/three-dots would require to have a div wrapper around the button and a div to display the 3 dots. Do we want to do this?

### DIFF
--- a/packages/app/src/app/pages/Sandbox/Editor/Header/index.tsx
+++ b/packages/app/src/app/pages/Sandbox/Editor/Header/index.tsx
@@ -82,6 +82,7 @@ const ForkButton = ({
     style={style}
     secondary={secondary}
     disabled={isForking}
+    loading={isForking}
     small
   >
     <>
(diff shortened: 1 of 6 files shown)


ValentinH:
I'll provide another solution by creating a <ProgressButton> component that would wrap the Button itself. It will keep the same API (and pass most props to the Button itself)  and support a `loading` prop (maybe also a `success` one).
Like this we are not limited and can be creative. Also we are not bloating the default Button component.

ValentinH:
I have updated the PR to have a better loader. It introduces a `<ProgressButton>` that wrap the button and add a loader on top of it.

fork-loading

CompuIves:
I like that! Could you make the dots a bit smaller? That would make the effect a bit more subtle.

ValentinH:
Do you like this better? I also made it a bit less transparent.

6px was feeling too small to me.

loading-button

CompuIves:
Yes, great! Thanks a lot!

CompuIves:
This implementation is super clean, I love it!

ValentinH:
I'd love to continue contributing to codesandbox, maybe on something bigger. Do you have a good issue for me?

PR #1734, opened 2019-04-05
allow to change version in the dep name

### LINKED ISSUE #1730: Allow choosing a dependency version without removing it
# 🌈 Feature

When clicking the version of a dependency it should show a dropdown and the user should be able to switch a version without removing the dependency and adding the new version.

SaraVieira:
I took a different route and show the dropdown when you hover in the version 

Let me know your thoughts

https://www.dropbox.com/s/p5uqiwr4zk5cnve/1.mov?dl=0

closes #1730

cc @kitze


CompuIves:
This is beautiful!

kitze:
@SaraVieira nice!

PR #1736, opened 2019-04-05
Add a tooltip for each VersionEntry button

ValentinH:
A tooltip was added for each version entry action: 
- Show/Hide sizes
- Update to latest
- Remove

image

There is no tooltip and the Refresh button is not 100% obvious.

**What is the new behavior?**
There is now a tooltip for each action

I had to force `outline: none` to avoid having a blue square around the icon:
image


SaraVieira:
This is awesome!! 
Thank you so much 😍

Issue #1654, opened 2019-03-25
🐛 Terminal does not resize with it's container


Saeris:
As the title says, the terminal window's contents does not resize along with it's container.

Please see attached gif for example:

![terminal resize](https://user-images.githubusercontent.com/3144549/54961363-87a69d00-4f1d-11e9-86a4-81ea80efa782.gif)

PS: This was observed in Firefox 67 on Windows 10. This doesn't appear to occur in Chrome 73. The scrollbars being the system default also appears to be a Firefox specific bug.

Saeris:
Just double-checked in Chrome 73, it appears to occur there too, though it has some weird jumping behavior too. See attached gif:

![terminal resize 2](https://user-images.githubusercontent.com/3144549/54962731-38636b00-4f23-11e9-9399-06561467e958.gif)

lbogdan:
Can confirm, noticed this too the other day, but was too caught up in stuff to file an issue. Thanks!

lbogdan:
This should be fixed by #1740 , @Saeris can you please confirm?

Saeris:
Appears to be fixed!

CompuIves:
Great!

lbogdan:
Hey @Saeris ,

Can you give the fix a go, see https://github.com/codesandbox/codesandbox-client/issues/2103#issuecomment-506626841 ?

armujahid:
@lbogdan It's appeared to be fixed.

PR #1740, opened 2019-04-06
Terminal does resize with it's container

rabingaire:
Fixed bug #1654 where Terminal does not resize with its container

Terminal does resize with its container

**What is the new behavior?**

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 

The bug occurred when `ResizeObserver` was not called even though the HTML element `ResizeObserver` was listening to changed, I gave `height` and `width` to this element to `100%` to fix the issue.


CompuIves:
This is awesome!! Thank you so much!

PR #1793, opened 2019-04-21
Add theme support to tests tab

SaraVieira:
Screenshot 2019-04-21 at 12 38 06
Screenshot 2019-04-21 at 12 38 00
Screenshot 2019-04-21 at 12 37 54
Screenshot 2019-04-21 at 12 37 48
Screenshot 2019-04-21 at 12 37 43


CompuIves:
Ohh wow this is great!

Is it possible to keep a darker background on these?

image

CompuIves:
The main thing that I'd like to see before we can merge this are the blocks set out: 
image

I think the blocks give a better overview and look cleaner. This is pretty hard to do with themes, but maybe we can actually use `Color` to create some variations in the theme, alternatively, we can invent some new names for the theme, like `props => props.theme["tests.blockBackground"]`. Almost no theme has these values, but we can add them to the CodeSandbox theme and if they don't exist we can default to `sideBar.background` like we do now. Hopefully some VSCode themes will then add these keys in the future to customize how CodeSandbox looks 😄.

SaraVieira:
All ideas implemented

Screenshot 2019-04-30 at 12 12 23
Screenshot 2019-04-30 at 12 11 19
Screenshot 2019-04-30 at 12 10 55
Screenshot 2019-04-30 at 12 10 47
Screenshot 2019-04-30 at 12 10 37
Screenshot 2019-04-30 at 12 10 05

PR #1807, opened 2019-04-23
Expo support

### LINKED ISSUE #1806: Expo support
# 🌈 Feature

## What

Now with Expo for web it would be great to get it working in code sandbox.

## How

Need to add/edit babel aliases.

https://github.com/facebook/create-react-app/pull/6813

EvanBacon:

fixes #1806

**What is the new behavior?**

You can now use a wide variety of [Expo](https://github.com/expo) modules.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged 


EvanBacon:
Can someone recommend a good way to test this?

CompuIves:
Wooow, this is awesome!! Thanks a lot @EvanBacon! We deployed the PR to here: http://pr1807.cs.lbogdan.tk/s/new. There you can also test if you can add expo modules. I'd be glad to help get this through!

EvanBacon:
@CompuIves thanks for the deploy! The aliases seem to work as expected. It looks like there is a bug where `.web` extensions aren't being used in node modules. 

```js
// This works: `expo-asset/build/AssetRegistry.web.js`
import * as AssetRegistry from 'expo-asset/build/AssetRegistry.web'

// This doesn't, it resolves to: `expo-asset/build/AssetRegistry.js`
// which also exists but is being used for the native functionality
// essentially returning an empty object as it's looking for a React Native `NativeModule`. 
import * as AssetRegistry from 'expo-asset/build/AssetRegistry'
```

CompuIves:
Ahh, I see! I pushed a fix to master, if you merge it should work (on staging at least).

EvanBacon:
This works really well! Still some bugs that I can work out in the individual modules. One more thing that would be nice is the environment variable `__DEV__` that Metro provides. Not sure if this will conflict with anything on your end though, we do this in the `expo/webpack-config` [here](https://github.com/expo/expo-cli/blob/d59cbe334dd55231ee9c68c66041a018c48aba5a/packages/webpack-config/webpack/createClientEnvironment.js#L41) to better support libraries that aren't fully optimized for web.

packages/common/src/config/env.ts
```diff
@@ -19,5 +20,6 @@ export default Object.keys(process.env)
       'process.env.LOCAL_SERVER': !!LOCAL_SERVER,
       'process.env.STAGING': 'STAGING_BRANCH' in process.env,
       'process.env.VSCODE': !!JSON.stringify(process.env.VSCODE),
+      __DEV__,
```
EvanBacon: I'm not sure if this actually worked.

CompuIves: I think we need to set it here: https://github.com/CompuIves/codesandbox-client/blob/6508607312aec5c34cfa0aeec480214834ee5169/packages/app/src/sandbox/eval/transpiled-module.ts#L964-L967. It's currently hardcoded but I can make it an option for templates.

One thing we need to make sure of is that the code written in CodeSandbox should also work when you download it to your local environment. Do you know if CRA also has the __DEV__ global enabled?

EvanBacon: it doesn't. I opened a PR to get it added here https://github.com/facebook/create-react-app/pull/6951

EvanBacon: After talking to @ide it may seem more logical to make `react-native-web` polyfill this value. Going to roll the change back. After I publish a new version of expo with some better web support, the aliases should be enough to get Expo running in code sandbox.

EvanBacon:
Given that `__DEV__` is not an actual blocker, I think it would make sense to merge this PR as is unless @CompuIves has any other concerns.

CompuIves:
That's great!! I'll merge this in tomorrow morning!

PR #1812, opened 2019-04-24
Use jsdom instead of window in jest test

### LINKED ISSUE #1800: Separate the app DOM from the tests DOM
# 🌈 Feature

Right now the tests run in the same JavaScript/DOM environment as the app. This can lead to unexpected and pretty confusing results which I (and other maintainers/contributors) have to explain over and over again to people reporting issues to react-testing-library.

Here's a simple example: https://codesandbox.io/s/n3j6r28z10

We have to tell people that they need to NOT render anything in their `index.js` file to avoid the issues which is annoying, inconvenient, and confusing.

I think it would make a lot of sense to run the tests in an iframe. And maybe provide a codesandbox option to run the tests with jsdom rather than the browser DOM (this would simulate local reality better for create-react-app based codesandboxes).

CompuIves:
Hacky implementation, I need to improve the performance and probably code split + find a better way to require jsdom inside the browser. But it works!

This replaces `document` in tests with a version provided by  `jsdom`, same way  as how it happens in Jest. This means that the DOM  of the application and the tests are not conflicting in any way.

It's not as good as running the tests in a separate context, but  building that  requires us to extract the `listen`/`dispatch` first to a version that's not global.

Fixes #1800.


CompuIves:
Testable here: http://pr1812.cs.lbogdan.tk/s/n3j6r28z10

packages/app/src/sandbox/eval/tests/jest-lite.js
```diff
@@ -102,12 +116,17 @@ export default class TestRunner {
...
+    const { window } = new JSDOM('');
+    const { document } = window;
+
     return {
       ...jestTestHooks,
       expect,
       jest: jestMock,
       test,
       it,
+      document,
```
kentcdodds: I'm guessing this is the "globals" object. This could be pretty problematic if it's not overriding _all_ browser globals (you'd end up with `window !== document.window` in tests for example)...

packages/app/src/sandbox/eval/tests/jest-lite.js
```diff
@@ -102,12 +116,17 @@ export default class TestRunner {
       });
     };
 
+    const { JSDOM } = JSDOM2;
+    const { window } = new JSDOM('');
```
kentcdodds: This should maybe be initialized the same way it is in jest?

https://github.com/facebook/jest/blob/4d3c1a187bd429fd8611f6b0f19e4aa486fa2a85/packages/jest-environment-jsdom/src/index.ts#L33-L39

CompuIves:
Probably, I just wanted to take a look if something would work before going to sleep. I can come up with a list of things that need to happen before we can merge this tomorrow.

packages/app/src/sandbox/eval/tests/jest-lite.js
CompuIves: Yeah, I need to take a closer look at how Jest implements this. I initially tried to replace window, but this broke all tests for some reason. Will need to do some more testing

kentcdodds:
Thanks for looking into it!!

packages/app/src/sandbox/eval/tests/jest-lite.js
CompuIves: Ah nice, thanks for the pointer! I'll take a look, maybe we can have a 1:1 implementation!

CompuIves:
My test sandbox: https://codesandbox.io/s/vvy844w25

CompuIves:
I think we're now very close in terms of API with Jest @kentcdodds. I also improved the integration with the problem view and the UI, with some scrolling fixes.

Looks like this now: 
image

kentcdodds:
Awesome! Thanks! Any chance we can fix the eslint errors for the test globals?

CompuIves:
Those errors in the screenshot are actual errors because the tests failed there 😄

kentcdodds:
Oooohhhhhh 😅 nice!

PR #1822, opened 2019-04-29
expose common types on types package

christianalfoni:
This is a preparation for Typescript refactor.

- The `common/src/types.ts` file is moved to `common/src/types/index.ts`
- The `tsconfig.json` of **app** now has a shortcut path named *"types"* to this moved file

This allows ut to put all typings in a single folder and expose them through `../types/index.ts`. It also allows to import any type by:

```js
import { Sandbox, Module } from 'types'
```

I think this will clean up a lot and make things more consistent, also preventing duplicate typing. All application wide types should be defined in this folder so you always know where to look :-)


CompuIves:
Beautiful!

PR #1824, opened 2019-04-29
add overmind and refactor utils to typescript overmind effect

christianalfoni:
Prepared the `overmind` folder and effects (previously providers) structure. `Utils` is the first provider that has now been refactored. 

Safe to pull in as this code is not imported or used anywhere.


christianalfoni:
This PR has been updated with copying all providers and utils of `store` over to `overmind`. Still nothing changed in the app itself, no imports to running code.

A couple of notes:

```ts
  const primaryBindingHits = state.pendingPrimaryBindings
    .filter(
      binding => binding.bindings[0][binding.bindings[0].length - 1] === key
    )
    .sort((a, b) => (a.bindings.length <= b.bindings.length ? -1 : 1));
```

The `sort` here returned a boolean, which is invalid. Now it it prioritizes the `a` binding if it has less or same length as `b`. I do not know if this is correct. Great comment there, but I think you have to understand how this works :)

```ts
  getRecoverList(currentId, modules) {
    const localKeys = Object.keys(localStorage).filter(key =>
      key.startsWith(`recover:${currentId}`)
    );

    return modules
      .filter(m => localKeys.indexOf(getKey(currentId, m.shortid)) > -1)
```

This actually had invalid syntax `localKeys.indexOf(getKey(currentId, m.shortid) > -1)`, so fixed that. Unsure what needs to be tested though.

CompuIves:
Hah, the first bugs have been found with TypeScript! I'm merging this in now, thanks for this part 😄

PR #1830, opened 2019-04-30
fix babel version compatibility

loveky:

change dependency in `react-sandpack` from `babel-core` to `@babel/core`

**What is the new behavior?**

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 

### DIFF
--- a/packages/react-sandpack/package.json
+++ b/packages/react-sandpack/package.json
@@ -106,7 +106,7 @@
     "@types/react": "16.8.12",
     "@types/react-broadcast": "^0.6.1",
     "awesome-typescript-loader": "^3.5.0",
-    "babel-core": "7.0.0-bridge.0",
+    "@babel/core": "^7.0.0",
     "colors": "^1.1.2",
     "commitizen": "^2.9.6",
     "coveralls": "^3.0.0",


SaraVieira:
Hey!

Thank you!

Did you do yarn after doing this? I don't see any yarn.lock changes

loveky:
You are right, I forgot that 🤣. Updated PR with yarn.lock change.

packages/react-sandpack/package.json
```diff
@@ -106,7 +106,7 @@
     "@types/react": "16.8.12",
     "@types/react-broadcast": "^0.6.1",
     "awesome-typescript-loader": "^3.5.0",
-    "babel-core": "7.0.0-bridge.0",
+    "@babel/core": "7.0.0",
```
CompuIves: I think it would be best to add a caret and redo `yarn`. That way we have the same version of `@babel/core` for all dependencies 😄 
```suggestion
    "@babel/core": "^7.0.0",
```

loveky: merged latest commits and updated PR

CompuIves:
Did you test running `react-sandpack`? I'm not sure if this breaks existing build tools.

CompuIves:
We need to test `react-sandpack` before merging this, once this is tested I'll re-open the PR!

PR #1834, opened 2019-05-02
add ability to open template in a new tab

SaraVieira:
Allows to open templates in a new tab because it has been bothering me for a while

Also fixes some svg complaints by react


packages/app/src/app/pages/NewSandbox/index.js
```diff
@@ -16,8 +16,14 @@ class NewSandboxComponent extends React.PureComponent {
     this.props.signals.sandboxPageMounted();
   }
 
-  createSandbox = template => {
-    history.push(sandboxUrl({ id: template.shortid }));
+  createSandbox = (template, e) => {
```
CompuIves: I think it'd be nicer to have the second argument be something like `options: { openInNewTab: boolean }`.

packages/app/src/app/pages/Dashboard/Content/CreateNewSandbox/index.js
```diff
@@ -35,11 +35,16 @@ class CreateNewSandbox extends React.PureComponent {
     document.removeEventListener('keydown', this.keydownListener);
   }
 
-  createSandbox = template => {
-    this.setState({ forking: true }, () => {
+  createSandbox = (template, e) => {
```
CompuIves: I think it'd be nicer to have the second argument be something like options: `{ openInNewTab: boolean }` together with a default.

packages/app/src/app/pages/Dashboard/Content/CreateNewSandbox/index.js
```diff
@@ -35,11 +35,16 @@ class CreateNewSandbox extends React.PureComponent {
...
+    const cmd = e.ctrlKey || e.metaKey;
+    const url = sandboxUrl({ id: template.shortid });
+    if (cmd === true) {
+      return window.open(url, '_blank');
```
CompuIves: This will break creating sandboxes from a subdirectory, the sandbox will then not be forked/put in the directory. I think this one should be moved to where `history.push`is as well (which we do if the sandbox doesn't have a subdirectory).

The logic currently is:

If a sandbox is created without a subdirectory we just redirect you to the sandbox. If the sandbox is created from a subdirectory we actually create a new sandbox with the subdirectory properly set.

packages/app/src/app/pages/Dashboard/Content/CreateNewSandbox/Modal/Template/index.js
```diff
@@ -7,7 +7,7 @@ import { Button, IconContainer, Title, SubTitle } from './elements';
 export default ({ template, subtitle, selectTemplate, small }) => {
   const Icon = getIcon(template.name);
 
-  const select = () => selectTemplate(template);
+  const select = e => selectTemplate(template, e);
```
CompuIves: `e` is undefined here (on line 20 we only call `select()`)

SaraVieira:
Closed due to innactivity

Issue #1839, opened 2019-05-02
[Expo] Supporting the expo-cli start command.


EvanBacon:
When using the [expo-cli with code sandbox](https://codesandbox.io/s/pkkj56zrr7), the start command doesn't seem to work as expected. The preview doesn't seem to update with the project, if you scroll down you'll find a white box with "an unexpected error ".

How can we make the expo-cli start command more compliant with code sandbox? Are there some environment variables that should be parsed in order to serve as expected? Is there a way to determine if the process is running in the code sandbox environment, it'd be nice to disable everything related to native development like the QR code.

The majority of relevant expo-cli startup code can be found here btw: [`XDL.Webpack`](https://github.com/expo/expo-cli/blob/master/packages/xdl/src/Webpack.js)

lbogdan:
Hey @EvanBacon ,

It looks like it's trying to connect to a websocket over http, can you somehow force https (wss)?

![image](https://user-images.githubusercontent.com/2083930/57331445-45c66480-7121-11e9-9687-524c8225bc4d.png)

lbogdan:
> Is there a way to determine if the process is running in the code sandbox environment

There's two environment variables explicitly set inside a sandbox:
- `CODESANDBOX_SSE`, always set to the string "true";
- `SANDBOX_URL`, set to the sandbox public URL, e.g. `https://node.sse.codesandbox.io/` (notice the trailing `/`).

EvanBacon:
Sorry for the delay, thanks for this. Will ensure that https can be passed through as a prop or set as an ENV variable like CRA.

lbogdan:
Hey @EvanBacon ,

Any updates on this? I saw you closed the `expo-cli` PR. We're doing a bit of issue cleaning, let us know if we can assist you further, otherwise we'd appreciate if you could close this. Thanks!

EvanBacon:
Seems stale at this point

PR #1840, opened 2019-05-03
[Expo] Added Expo CLI template

EvanBacon:

Better support for Expo web.

**What is the new behavior?**

Added a template for creating an Expo (web) project.

* [Template](https://github.com/expo/web-examples/tree/master/code-sandbox-template). May change to create a more concise URL.
* Currently using a node setup pending better expo-cli support #1839

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 

Screen Shot 2019-05-02 at 7 03 37 PM


packages/common/src/templates/configuration/expo/index.ts
```diff
@@ -0,0 +1,13 @@
+import { ConfigurationFile } from '../types';
+
+const config: ConfigurationFile = {
+  title: 'app.json',
+  type: 'expo',
+  description: 'Configuration for your universal app.',
+  moreInfoUrl: 'https://docs.expo.io/versions/latest/workflow/configuration/',
+  partialSupportDisclaimer: `Native app development is not supported in Code Sandbox yet. You can download your project and run it locally, or copy your code to https://snack.expo.io to run your project in the Expo client.`,
```
EvanBacon: I'm thinking about decoupling Metro from React Native to run with Webpack, it would then be pretty straightforward to run Expo Client with Code Sandbox. Until then I could investigate snack-sdk integration, but that also could be kinda hard. 😅

CompuIves:
Sorry for the delay on this! So if I understand correctly Expo Web will work in the browser? That would be absolutely awesome!

Is there anything else you need to do before we can merge? I see that the status is on Draft

SaraVieira:
@EvanBacon Are you still working on this? If so can you please make a new PR?

This one is very old and our configuration of templates as changed a lot since when you made so I feel like if you make it work just send me or @CompuIves a message on twitter and we will add that sandbox as a template :)

Thank you

PR #1843, opened 2019-05-04
Support dynamic typescript version from package.json

### LINKED ISSUE #942: Configuring which TypeScript version is used?
Can we make it so that I can configure which version of TypeScript to use? Can I (for instance) install `typescript@next` and use that?

### LINKED ISSUE #1683: Upgrade to TypeScript version 3.4
TS 3.4 adds new syntax ([const contexts](https://github.com/Microsoft/TypeScript/pull/29510) and [readonly tuples](https://github.com/Microsoft/TypeScript/pull/29435)) which isn't supported now. [Repro on CodeSandbox](https://codesandbox.io/s/7w0pwz6lj6?fontsize=14&module=%2Fsrc%2Findex.ts)

Could be solved with #942 but this is smaller scope, so I've created separate issue.

CompuIves:
Fixes #942
Fixes #1683

This doesn't change which version is used by the editor yet, I'm meaning to change this later by introducing a BrowserFS backend that can handle UNPKG.


aleclarson:
Wooo! 👏

PR #1845, opened 2019-05-04
fix(ember): display ember in the client framework

sarbbottam:
... instead of the server framework

Bug fix (?)

Ember is primarily a client web framework thus it would be easy to list it under the client framework for better discoverability

**What is the new behavior?**
Display Ember in the client framework section

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/packages/common/src/templates/ember.ts
+++ b/packages/common/src/templates/ember.ts
@@ -24,7 +24,7 @@ export default new Template(
   'github/mike-north/ember-new-output',
   decorateSelector(() => '#E04E39'),
   {
-    isServer: true,
+    isServer: false,
     showOnHomePage: true,
     main: false,
     netlify: false,


NullVoxPopuli:
I'm a fan of this change. 

Ember not being on the overview or "Client Templates" page has been endlessly confusing for ember users. 

Many thing, "does codesandbox have ember" 

:(

Much sadness :cry:

CompuIves:
This change would make the ember template break I think, because we won't execute the code on a server anymore. We did change the name from "Server" to "Container". Does that sound better? Otherwise we should look into different naming.

NullVoxPopuli:
Why would moving it break things? Sounds like opportunity for a refactor

SaraVieira:
Hey!

The names server and client are mostly for us and that's why we changed to container, Ember needs a container to run.

We are working on making everyone create their own templates so this will be easier on everyone.

So the client runs in the browser and container runs in a server and ember is the latter and needs a server to run.

We have other examples like gatsby that are right now in container but are mostly front end frameworks.

I'm sorry for your trouble and I hope this makes sense.

I will close in favor of the new template system

NullVoxPopuli:
Look, this in a huge visibility problem for those incorrectly categorized.

This is an implementation detail being exposed that doesn't match up with how people think of these tools.

The tabs neeeeeed to be decoupled from implementation. 

I guess I'll submit a PR and fix all this :)

mansona:
@SaraVieira there is a somewhat simple fix that would prevent the need to do a major refactor to "fix" this issue 🤔 we could add Ember to the "Overview" tab under "Popular Templates" 👍 Seeing as done is available there it doesn't care for the separation between "client" and "server" 

Would this be something that you would consider? It would go a long way to "fix" this for Ember and will buy us some time while we wait for the fix to move it into the client tab 👍

sarbbottam:
Thank you, everyone, for your sharing your thoughts, suggestions, and concerns.

>The names server and client are mostly for us and that's why we changed to container, Ember needs a container to run.
Hi @SaraVieira thank you for sharing your thoughts. I feel naming rather a UI representation, i.e. a CTA classification should be aligned with the purpose it is serving, rather how it the purpose is being served. Ember thus belongs to the client-side framework classification IMHO.

>container
Is it due to the execution of `ember serve` command? If so, I am wondering how `create-react-app` based sandboxes are being managed, I guess they are started in a similar fashion. Please excuse my ignorance if I sound so.

>we could add Ember to the "Overview" tab under "Popular Templates" 
Hi @mansona thank you for sharing it, this seems to be the best feasible solution at the moment.

PR #1859, opened 2019-05-07
Reset isForkingSandbox on error

### LINKED ISSUE #1803: "Fork" button doesn't reset its state on fork error
When there's an error forking a (server) sandbox (for example because the server sandboxes limit was reached), the "Fork" button doesn't reset its state and remains in the "forking" state:

image

ValentinH:
Close #1803.

I'm not sure what's the best way to reset some state on error with Cerebral though.
Basically, I'd like to be able to always reset this state at the end of the `forkSandbox` sequence (https://github.com/codesandbox/codesandbox-client/blob/master/packages/app/src/app/store/sequences.js#L162) instead of doing this in every single action called by this sequence.


CompuIves:
This is a good approach! I think a bit cleaner approach would be to use paths, like we do here: https://github.com/codesandbox/codesandbox-client/blob/master/packages/app/src/app/store/sequences.js#L492-L496. In that example in the action we import `path` and in a successful case we return `path.success()`, otherwise `path.error()`.

ValentinH:
Hey @CompuIves, I've updated to paths. Let me know if this matches what you had in mind 🙂

SaraVieira:
Looks good!

Thank you!

PR #1865, opened 2019-05-08
New Cards

SaraVieira:

**What is the new behavior?**

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 

### DIFF
--- a/packages/common/src/components/SandboxCard/index.tsx
+++ b/packages/common/src/components/SandboxCard/index.tsx
@@ -53,12 +53,12 @@ export type Props = {
 };
 
 const kFormatter = (num: number): number | string => {
+  if (num > 9999) {
+    return (num / 10000).toFixed(1) + 'M';
+  }
+  
   if (num > 999) {
     return (num / 1000).toFixed(1) + 'K';
-  }
-
-  if (num > 9999) {
-    return (num / 1000).toFixed(1) + 'M';
   }
 
   return num;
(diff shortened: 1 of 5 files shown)


CompuIves:
Niiiice, looks really good already! I have some very small nitpicks:

- On hover I would show the dark overlay anyway, no matter if there's no description or not. This lets the user know that there is the possibility of the description.
- I would also show the dark overlay when you hover the card itself (under). This might need some JS magic though, less nice but better UX.
- The icons of the stats can maybe be a bit darker, inspiration: https://twitter.com/steveschoger/status/872865304719892480
- There is a jump when switching pages, that's probably because the loading of the images cause the card height to jump

After having these things resolved we're ready to rumble!

SaraVieira:
Everything fixed, maybe?

CompuIves:
Nice!!! One last tiiiiiiny thing I forgot to add! Could you add `M` (million) support? 

image

packages/common/src/components/SandboxCard/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+  small?: boolean;
+  noHeight?: boolean;
+  defaultHeight?: number;
+  noMargin?: boolean;
+  selectSandbox: (params: Sandbox) => void;
+};
+
+const kFormatter = (num: number): number | string => {
+  if (num > 9999) {
+    return (num / 10000).toFixed(1) + 'M';
```
CompuIves:
```suggestion
    return (num / 1000000).toFixed(1) + 'M';
```

PR #1867, opened 2019-05-08
fix(getters): Do not use store.jwt to compute hasLogIn

arthurdenner:

This PR aims to fix two bugs I noticed related to sign out that are described below.

- When you're on the dashboard and signs out, the dashboard doesn't update;
- When you're on a sandbox page and signs out, the sidebar doesn't update;

I believe the problem it's in how `hasLogIn` is computed, because it uses `store.jwt`, which doesn't get updated in the `signOut` sequence. I'm not sure if removing it from the computation is the proper fix to these problems, but since in [this commit](https://github.com/codesandbox/codesandbox-client/commit/705679b88d6057583587b8ab32539f3062a4a479#diff-fcc68a385e98ecb8fed8d41e68d69c39L14) the logic was changed to add `!!this.jwt` and not only use `!!store.get('jwt')` I thought that maybe it is.

**What is the new behavior?**

When you sign out in the situations described above, the UI is updated as expected.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 

### DIFF
--- a/packages/app/src/app/store/getters.js
+++ b/packages/app/src/app/store/getters.js
@@ -1,5 +1,3 @@
-import store from 'store/dist/store.modern';
-
 export function isPatron() {
   return Boolean(
     this.user && this.user.subscription && this.user.subscription.since
@@ -11,5 +9,5 @@ export function isLoggedIn() {
 }
 
 export function hasLogIn() {
-  return !!this.jwt || !!store.get('jwt');
+  return !!this.jwt;
 }


arthurdenner:
@SaraVieira @CompuIves pinging you to see if this PR can get a review

SaraVieira:
Hey!

Sorry for the delay

Can you please explain the change?

Im also pinging @CompuIves

arthurdenner:
Hi Sara! :wave:  No worries.

Try to reproduce what I described in the current behavior section and you'll see that the UI doesn't update as expected.

When I was debugging it, I found that removing the `store.jwt` fix the problem and (if I'm not mistaken) wherever `hasLogin` is being used continues to behave as before.

Sorry I'm on the phone right now and can't provide more information, but if you need any other details I can provide later today.

arthurdenner:
Current problems:

csb-logout-dashboard
csb-logout-sandbox

lbogdan:
@arthurdenner Can you please rebase this to `master`, so that it gets built by the new PR build system? Thanks!

arthurdenner:
@lbogdan, rebased.

I also updated my previous comment to add the new behaviour after the PR.

To reinforce: I'm not sure if this PR contains the proper fix to these problems because I didn't tried to logout during a live session, but the remaining scenarios where `hasLogin` is being used are working as before.

arthurdenner:
@lbogdan, is there any way I see exactly why the integration tests are failing and maybe fix them? :thinking:

CompuIves:
This is great @arthurdenner!! Just tested this and the logic is far better now. Thank you!

PR #1871, opened 2019-05-09
Refactor SearchInput to functional component with hooks

MichaelDeBoey:
Update components to functional components with Hooks

Class components

**What is the new behavior?**
Functional components with Hooks

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/packages/homepage/src/screens/home/NPMFeature/SearchInput/Hit.js
+++ b/packages/homepage/src/screens/home/NPMFeature/SearchInput/Hit.js
@@ -1,11 +1,9 @@
-import React from 'react';
-import styled, { css } from 'styled-components';
-
 import { TimelineMax } from 'gsap';
+import React, { useEffect, useRef } from 'react';
+import styled, { css } from 'styled-components';
 
 const Container = styled.a`
   transition: 0.3s ease all;
-  display: block;
   margin-bottom: 0.5rem;
   padding: 1rem;
   display: flex;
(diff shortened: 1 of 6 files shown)


packages/homepage/src/screens/home/NPMFeature/SearchInput/Hit.js
```diff
@@ -38,37 +37,31 @@ const Count = styled.div`
...
-      this.el,
-      0.1,
-      { opacity: 0 },
-      { opacity: 1 }
-    );
-  }
+const Hit = ({ hit: { count, value } }) => {
+  const el = useRef(null);
+  const tl = useRef(
+    new TimelineMax().fromTo(el.current, 0.1, { opacity: 0 }, { opacity: 1 })
```
SaraVieira:
```suggestion
  let tl;

  useEffect(() => {
    tl = new TimelineMax().fromTo(
      el.current,
      0.1,
      { opacity: 0 },
      { opacity: 1 }
    );
  }, []);
```

SaraVieira:
Hey!

I get an error on running it because sometimes the current target does not exist and I added a comment that fixes it by only checking the current target on component mount :)

MichaelDeBoey:
@SaraVieira Needed to create a `ref` for `tl`, otherwise I got the following warning from ESLint:
```sh
44:10  warning  Assignments to the 'tl' variable from inside React Hook useEffect will be lost after each render. To preserve the value over time, store it in a useRef Hook and keep the mutable value in the '.current' property. Otherwise, you can move this variable directly inside useEffect  react-hooks/exhaustive-deps
```

Also added `eslint-plugin-react-hooks` dependency and `react-hooks/rules-of-hooks` and `react-hooks/exhaustive-deps` rules, so we can take advantage of them.

SaraVieira:
Interesting error but it looks good, works well in your machine?

MichaelDeBoey:
@SaraVieira Works just like before on my machine

SaraVieira:
Awesome!! Thank you again!

MichaelDeBoey:
With pleasure 🙂

PR #1890, opened 2019-05-13
Refactor GraphQL Result Param Reassignment

Saeris:
Instead of reassigning values on the result object, which previously required disabling eslint, this refactor uses ES6 object destructuring  and rest/spread to create a new result object.

**What is the new behavior?**

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged 

### DIFF
--- a/packages/app/src/app/graphql/client.js
+++ b/packages/app/src/app/graphql/client.js
@@ -24,14 +24,11 @@ const authLink = setContext((_, { headers }) => {
 });
 
 const absintheAfterware = new ApolloLink((operation, forward) =>
-  forward(operation).map(result => {
-    /* eslint-disable no-param-reassign */
-    result.errors = result.payload.errors;
-    result.data = result.payload.data;
-    /* eslint-enable */
-
-    return result;
-  })
+  forward(operation).map(({ payload, ...result }) => ({
+    ...result,
+    errors: payload.errors,
+    data: payload.data
+  }))
 );
 
 const errorHandler = onError(({ graphQLErrors, networkError }) => {


SaraVieira:
Cleany! Niiice

packages/app/src/app/graphql/client.js
```diff
@@ -24,14 +24,11 @@ const authLink = setContext((_, { headers }) => {
...
 const absintheAfterware = new ApolloLink((operation, forward) =>
-  forward(operation).map(result => {
-    /* eslint-disable no-param-reassign */
-    result.errors = result.payload.errors;
-    result.data = result.payload.data;
-    /* eslint-enable */
-
-    return result;
-  })
+  forward(operation).map(({ payload, ...result }) => ({
```
MichaelDeBoey: @Saeris Why didn't you further destructur `payload` here?

Saeris: I didn't run this locally, I made the edit quickly here on GitHub, so I don't know what else is included in the payload object from Absinthe, or if anything else from payload is actually getting used somehow (I don't believe Apollo Client uses it).

If there is a specific reason to forward it, I'd be interested in hearing it. Otherwise this should not be a breaking change.

MichaelDeBoey: You only use `data` & `errors ` from `payload`, so you could just have done it like this

```js
.map(({ payload: { data, errors }, ...result }) => ({
  ...result,
  data,
  errors,
})
```

PR #1918, opened 2019-05-16
Add condition for primary action in toast

### LINKED ISSUE #1917: When saving a project, a toast might be undefined
# 🐛 bug report

## Description of the problem

When saving (cmd+s) an error is thrown which causes the page to unmount.

page becomes white, error thrown on save, but last save was still taken in account.

There should probably be an error boundary around the toast, and other non-essential parts of the DOM.

Based on the stack trace, I think something in this is undefined:

https://github.com/codesandbox/codesandbox-client/blob/cabd5f9761ecec35cc31266cc8e753d1edeec034/packages/notifications/src/component/Toast.tsx#L149

### Your Environment

| Сodesandbox      | PROD-1557437664-2c65ca12f
| Browser          | Safari Release 81 (Safari 12.2, WebKit 14608.1.19.5)
| Operating System | MacOS 10.14.2 (18C54)

CompuIves:
Fixes #1917

### DIFF
--- a/packages/notifications/src/component/Toast.tsx
+++ b/packages/notifications/src/component/Toast.tsx
@@ -134,20 +134,23 @@ export function Toast({ toast, removeToast, getRef }: Props) {
                   fontSize: '.875rem',
                 }}
               >
-                <Button
-                  small
-                  onClick={() => {
-                    removeToast(toast.id);
-                    toast.notification.actions.primary[0].run();
-                  }}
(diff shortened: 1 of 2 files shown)


SaraVieira:
@CompuIves ? Is this mergy mergy?

SaraVieira:
Ives said mergy mergy :D

PR #1919, opened 2019-05-16
refactored sandboxPageMounted with related stuff

christianalfoni:
"sandboxPageMounted" signal is refactored to Overmind, with anything related to that.

One thing to notice here is "joining the notification channel". Neither Cerebral or Overmind resolves this effect. It is supposed to resolve the number of unread notifications. I am not sure if this is a local thing or maybe it is a bug in production? 

But yeah, nothing here affecting production, but just something I noticed.


packages/app/src/app/overmind/effects/api.ts
```diff
@@ -1,43 +1,29 @@
...
   const response = error.response;
 
   if (!response || response.status >= 500) {
     logError(error);
   }
 
-  if (response && response.result) {
-    if (response.result.errors) {
-      const errors = values(response.result.errors)[0];
+  if (response && response.data) {
```
batiskafff: This if - else hell looks overcomplicated. Is it any way that you can simplify it?

christianalfoni: Ah, yeah, there are tons of places we want to improve logic. Though currently I have to focus on moving over to Typescript or it will never be done :) So first over to Typescript and Overmind, then when things are running we can do another round of refactor to improve stuff like this. 

I totally agree with you, I just know myself that if I start doing these tiny improvements I get totally lost in the main goal of getting over to Typescript and new state management :)

PR #1925, opened 2019-05-17
🔨 Hooks Refactor

Saeris:
Removed unused container styles and reorganized code to follow updated coding conventions (to be documented at a later date). Added a new useScript hook (based on https://usehooks.com/useScript/) and a hooks directory.


Saeris:
Build failure appears to be unrelated to this commit.

Appears to run fine locally. If I'm missing a step here let me know!

packages/app/src/app/index.js
```diff
@@ -120,15 +121,17 @@ function boot() {
...
-        <Provider {...controller.provide()}>
-          <ApolloProvider client={client}>
-            <ThemeProvider theme={theme}>
-              <Router history={history}>
-                <App />
-              </Router>
-            </ThemeProvider>
-          </ApolloProvider>
-        </Provider>,
+        <Cerebral.Provider value={controller.provide()}>
```
CompuIves: Interesting, why do we do this? Aren't we setting the context twice now?

packages/app/src/app/pages/Sandbox/Editor/Workspace/Advertisement/Advertisement.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { useScript } from 'app/hooks';
+import { Container } from './elements';
+
+export const Advertisement = () => {
+  useScript(`https://codefund.app/properties/24/funder.js`);
```
CompuIves: Nice and clean!

packages/app/src/app/pages/Sandbox/Editor/Workspace/ConnectionNotice/ConnectionNotice.js
```diff
@@ -0,0 +1,18 @@
+import React, { useContext } from 'react';
```
CompuIves: Would it be better to call this file `index.js` for consistency with the other files?

packages/app/src/app/pages/Sandbox/Editor/Workspace/index.js
```diff
@@ -16,8 +16,8 @@ import Deployment from './items/Deployment';
 import ConfigurationFiles from './items/ConfigurationFiles';
 import NotOwnedSandboxInfo from './items/NotOwnedSandboxInfo';
 
-import ConnectionNotice from './ConnectionNotice';
-import Advertisement from './Advertisement';
+import { ConnectionNotice } from './ConnectionNotice';
```
CompuIves: This is really nice!

packages/app/src/app/index.js
Saeris: Technically yes, we are using two providers here. This seems to be the recommended approach to migrating to mobx-react-lite, as this new provider uses the react context api, and the old way does it different somehow. Basically to use `useContext` we have to do it this way. Doesn't appear to be a problem at the moment, but I'm happy to do further tests before we commit to this.

packages/app/src/app/pages/Sandbox/Editor/Workspace/Advertisement/Advertisement.js
Saeris: Yup! Now we can reuse useScript anywhere we need to load an outside script on component mount.

packages/app/src/app/pages/Sandbox/Editor/Workspace/ConnectionNotice/ConnectionNotice.js
Saeris: No, the reason why we always split `index.js` and `Component.js` is to deliberately improve readability in a contributor's code editor. It quickly becomes confusing which file you're looking at when every component is defined inside of an `index.js`. Additionally, we may want to export multiple components, functions, etc from a single directory using named exports. A good example of this is a component that comes paired with say a GraphQL query fragment. It makes sense to keep those files separate for readability's sake, much like how we split component definitions from css-in-js with `elements.js`. The goal is to break files down into small, highly readable chunks. My general philosophy is that if I have to scroll to see all the contents of a single file, it's getting too big. Sometimes that's unavoidable, but you can use your best judgement there.

packages/app/src/app/pages/Sandbox/Editor/Workspace/index.js
Saeris: So keep in mind that with my previous comment, I'm in the process of re-organizing files one by one to keep these reviews easy on you. Every component will be refactored to conform to the conventions I'm looking to establish here.

Saeris:
Will need some additional help testing the Live components refactor, as you can't start a live session when running the client locally.

packages/app/src/app/index.js
Saeris: This has been re-addressed in 519b214, as now it's only calling `controller.provide()` once and passing it's values to our providers. Once the whole refactor is complete, we can remove the old provider entirely.

packages/app/src/app/hooks/useInterval.ts
```diff
@@ -0,0 +1,21 @@
+// Based on https://overreacted.io/making-setinterval-declarative-with-react-hooks/
+import { useEffect, useRef } from 'react';
+
+export const useInterval = (callback = () => {}, delay: number) => {
+  const savedCallback: { current: any } = useRef();
```
MichaelDeBoey: No need to type `savedCallback` here.

packages/app/src/app/hooks/useInterval.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  useEffect(() => {
+    savedCallback.current = callback;
+  }, [callback]);
+
+  useEffect(() => {
+    function tick() {
```
MichaelDeBoey: This can be an arrow function

```js
const tick = () => savedCallback.current();
```

packages/app/src/app/hooks/useInterval.ts
```diff
@@ -0,0 +1,21 @@
...
+      savedCallback.current();
+    }
+    if (delay !== null) {
+      const id = setInterval(tick, delay);
+      return () => clearInterval(id);
+    }
+    return undefined;
```
MichaelDeBoey: No need to return `undefined`, since no return is the same

packages/app/src/app/hooks/useScript.ts
```diff
@@ -0,0 +1,60 @@
...
+  useEffect(() => {
+    if (cachedScripts.includes(src)) {
+      setState({
+        loaded: true,
+        error: false,
+      });
+    } else {
+      cachedScripts.push(src);
+
+      const script = document.createElement('script');
```
MichaelDeBoey: This can  be simplified with the spread operator

```js
const script = {
  ...document.createElement('script'),
  async: true,
  src,
};
```

packages/app/src/app/hooks/useScript.ts
```diff
@@ -0,0 +1,60 @@
...
+        setState({
+          loaded: true,
+          error: false,
+        });
+      };
+
+      const onScriptError = () => {
+        const index = cachedScripts.indexOf(src);
+
+        if (index >= 0) cachedScripts.splice(index, 1);
```
MichaelDeBoey: It will make this piece of code a bit clearer if you wrap it in curly braces (`{}`).

```js
if (index >= 0) {
  cachedScripts.splice(index, 1);
}
```

packages/app/src/app/hooks/useScript.ts
```diff
@@ -0,0 +1,60 @@
...
+
+      script.addEventListener('load', onScriptLoad);
+      script.addEventListener('error', onScriptError);
+
+      document.body.appendChild(script);
+
+      return () => {
+        script.removeEventListener('load', onScriptLoad);
+        script.removeEventListener('error', onScriptError);
+      };
```
MichaelDeBoey: Shouldn't the script be detached from `document.body` too? 🤔

packages/app/src/app/hooks/useScript.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    return undefined;
```
MichaelDeBoey: No need to return `undefined`, since no return is the same

packages/app/src/app/pages/Dashboard/index.js
```diff
@@ -52,7 +52,11 @@ class Dashboard extends React.Component {
     } = this.props;
     const { showSidebar } = this.state;
 
-    history.listen(() => {
+    history.listen(({ state }) => {
+      if (!!state && state.from === 'sandboxSearchFocused') {
```
MichaelDeBoey: Try using the `Boolean` function instead of double negation

```suggestion
      if (Boolean(state) && state.from === 'sandboxSearchFocused') {
```

packages/app/src/app/pages/Sandbox/Editor/Workspace/Advertisement/Advertisement.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { useScript } from 'app/hooks';
+import { Container } from './elements';
+
+export const Advertisement = () => {
+  useScript(`https://codefund.app/properties/24/funder.js`);
```
MichaelDeBoey: You can just use normal quotes here

```suggestion
  useScript('https://codefund.app/properties/24/funder.js');
```

packages/app/src/app/hooks/useInterval.ts
Saeris: That return is simply there to satisfy a linting rule. We could use an ignore instead, though it wouldn't make a difference in how this hook works.

Saeris:
@MichaelDeBoey Thanks for the suggestions! I'll consider a few of those. For reference, both of these were taken from the blog posts linked at the top of each file and slightly modified from their respective original author's source primarily to fit our linting rules.

packages/app/src/app/pages/Sandbox/Editor/Workspace/Advertisement/Advertisement.tsx
Saeris: Actually going to keep this one as-is. While yes in this case we're not making use of template literals, there's no cost to using backticks as opposed to quotes and that being the case it's just a good practice to prefer backticks for the simple reason that you can at any time drop in expressions and mixed quotes at any time.

This article sold me on template literals being strictly better a while back and so I've been using them everywhere since:

https://ponyfoo.com/articles/template-literals-strictly-better-strings

packages/app/src/app/hooks/useScript.ts
Saeris: Hmm, maybe, I'm not entire sure what the possible side-effects would be of removing a script from a page. My intuition tells me that it could cause unwanted behavior.

packages/app/src/app/hooks/useScript.ts
Saeris: I would have to test this. To my understanding the script tag needs to be present in the DOM before you can set a src attribute on it, which triggers the network request.

packages/app/src/app/hooks/useInterval.ts
Saeris: Not destructuring here, though we possibly could. Just adding a type to satisfy the linter. Probably clean this up later as we make another pass to flesh out types across the code base.

MichaelDeBoey:
Really great job in refactoring all these components @Saeris! 🙂

I think we should close this PR tho.
This one is only getting bigger and bigger and I think it would be better to have separate PR's for each refactored component instead of this one big PR which is getting to big right now imo.

Like I said: really great job so keep up the good work! 🙂

You can just cherry-pick your commit from this branch in new branches so this work doesn't need  to be done multiple times 🙂

CC @CompuIves @SaraVieira

packages/app/src/app/pages/Sandbox/Editor/Workspace/ConnectionNotice/ConnectionNotice.js
CompuIves: Right, the only thing I'm afraid of is that we're now following two different naming schemes inside the repo (alongside two languages). For new contributors we need to be clear what the preferred way is.

packages/app/src/app/hooks/useScript.ts
CompuIves: Not much would happen if you remove it, on the other hand, I don't see a reason to remove it.

packages/app/src/app/hooks/useScript.ts
CompuIves: Hmm yes, the more "conventional" way for me is to not spread anything that isn't a plain object.

.all-contributorsrc
```diff
@@ -1198,6 +1198,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "twhite96",
```
CompuIves: Isn't this related to another commit?

CompuIves:
This is a huge improvement over the existing code! I agree with @MichaelDeBoey, this PR is huge, and that it's a big risk merging this all in at once. We did this once before and it caused some small bugs of which some of those are still not fixed today. I prefer to split this up into smaller testable commits, but I understand that that might be a lot of work.

I'm okay with merging this PR in, but only if we do **a lot** of testing on staging before we're going to put this on prod. Here's a list of things that we'll need to test:

1. Change title/subtitle of sandbox
2. Sandbox forking
3. Saving files
4. Deploying to ZEIT
5. Deploying to Now
6. Generating a diff for a commit
7. Changing a file and making sure that the diff is regenerated properly
8. Committing & Opening a PR
9. Which components update when changing code (this is a performance test and to make sure that the proxy mechanism still works with hooks)
10. Adding/Removing Dependencies (without mouse)
11. CodeSandbox Live (will work with staging :))
12. CodeSandbox Teams Live
13. Adding/removing people to live session, toggling classroom, toggling all settings
14. Make sure that the More tab works, that analytics is tracked and that the buttons there work

I'm also curious what @christianalfoni thinks about the state management, the main thing I want to know if this refactor will be as performant as our current solution, because that's one of the main reasons that we switched. I'm not sure how `shouldComponentUpdate` works in React Hooks and if it can be overridden.

MichaelDeBoey:
@CompuIves Another option would be to first merge #1956 (which is an extracted part from this PR).
#1956 is the basis on which all other changes are based on in this PR.

So if that one's merged, this PR can be extracted (via `git cherry-pick`) in multiple smaller PRs (each component it's own PR) to make it easier to review and reason about

.all-contributorsrc
MichaelDeBoey: I think this is because of merging `master` back into this branch

CompuIves:
That'd be best, also leads to a cleaner git history. My points about testing still stand, but that way it will still be clean.

> Another option would be to first merge #1956 (which is an extracted part from this PR).

Makes sense, I want to know how this affects `shouldComponentUpdate` before we can merge this. We had a substantial performance increase when switching to cerebral and its custom `shouldComponentUpdate` and it would be a shame if we would lose that benefit.

Saeris:
Yeah apologies for the huge PR! After that mixup the other day where all my work in progress got wiped out, I figured it would be safer for me to make multiple commits for each section I ended up having to re-write.

So far in all of my manual testing, everything appears to work. Here's some of what I've tried successfully:

1. Advertisements still load
2. Can create new GitHub Repos
3. Changes correctly appear in the GitHub tab
4. More correctly shows the right button when Logged in/out, opens new window to login (can't fully test on local), and forks when logged in.
5. NotOwnedSandboxInfo properly displays it's data (tested on `~/s/new`
6. Tested Netlify deployment. All buttons still work as expected, states update when you kick off a build, can launch build info modal.
7. And I'm in the middle of refactoring the Server workspace right now. All the scripts work, restart/reboot works, can change/edit/delete environment variables. Just needs some more styling moved from the components to `elements.ts`

Forking and saving still work. Haven't made changes to ProjectInfo, so changing the Sandbox Title/Subtitle should be unaffected. No changes to File Explorer, so adding/removing dependencies should be unaffected.

That leaves testing Live, Now Deployment, Commits/PRs, Offline, and performance.

As far as `shouldComponentUpdate` goes, I don't know if there's anything special going on under the hood that would be affected by these changes, but in the specific case of WorkspaceItem (which is untouched), before my progress got reset, I did refactor that using `React.memo()`, which can only compare props.

christianalfoni:
Hi guys!

The refactor cleans up a lot. I have not had time to go through everything, but conceptually it works exactly the same as before. The only thing to watch out for is that hooks does not handle `shouldComponentUpdate`, which the HOC does.

So basically most components will probably benefit from being `React.memo` components, as that checks the props.... and also avoids reconciliation when there are no props. That said, it is extremely difficult to evaluate. Generally it should not be an issue as components will render individually based on state changes of the state they actually look at. But if a component high up in the tree needs to render, all the children will reconcile until it meets a `React.memo` or similar.

Anyways... I would drop going crazy with optimizations and if one pops up, just add `React.memo`. 

I hope React will implement "no props, no reconcile", so when a parent causes child to reconcile it skips it when no props are passed. Cause with external state stores a surprising amount of components has no props. Anyways... this should also make the transition to Overmind simpler, as we are already in hooks world.

MichaelDeBoey:
@CompuIves I've extracted all @Saeris' changes into (smaller) separate branches/PRs:

-`Advertisement`:  #1962
- `ConnectionNotice`: #1963
- `Live`: #1964 (#1962 needs to be merged first)
- `SSEDownNotice`: #1965
- `GitHub`: #1968
- `Deployment`: #1969
- `More`: #1970
- `NotOwnedSandboxInfo`: #1971

MichaelDeBoey:
@CompuIves I thought a decision was made to not continue with major refactors just for the sake of refactoring, so I guess this one (and #1964) can be closed? 🤔

christianalfoni:
hm, has this PR moved to far away from current `master`. Been there many times with Codesandbox and refactoring. Might be an idea to create a new PR and more incrementally introduce the concepts from this PR? @CompuIves @Saeris ?

SaraVieira:
yeah, this is impossible to merge now :/

I am gonna close this and explain to @Saeris

PR #1927, opened 2019-05-18
Fix Like icon alignment in Project Info tab

ValentinH:
It fixes the alignment of the Like button on the Project Info tab

image

**What is the new behavior?**

image


SaraVieira:
Thank you! :D

Issue #1930, opened 2019-05-19
Vanilla template does not support dynamic import


pshrmn:
# 🐛 bug report

Parcel supports dynamic import without configuration by internally configuring Babel ([v7](https://github.com/parcel-bundler/parcel/blob/75310e15f6fa3c8a0ce6593ada1bcca00240ea54/packages/core/parcel-bundler/src/transforms/babel/babel7.js#L24) sets a parser option while [v6](https://github.com/parcel-bundler/parcel/blob/75310e15f6fa3c8a0ce6593ada1bcca00240ea54/packages/core/parcel-bundler/src/transforms/babel/babel6.js#L21-L24) uses a custom plugin); the vanilla template does not.

## Description of the problem

Using dynamic `import()` in the vanilla template results in the syntax error:

> Support for the experimental syntax 'dynamicImport' isn't currently enabled (1:1):

## How has this issue affected you? What are you trying to accomplish?

Unable to use dynamic `import()` in the vanilla template.

### Link to sandbox:

[Babel v7](https://codesandbox.io/s/jovial-grass-b2wwg)
[Babel v6](https://codesandbox.io/s/muddy-field-v05k2)

### Your Environment

| Software         | Name/Version |
| ---------------- | ------------ |
| Сodesandbox      | PROD-1557935916-f212746b9 |
| Browser          | Chromium 73 |
| Operating System | Linux Mint |

PR #1931, opened 2019-05-19
Support dynamic import in vanilla template for Babel v7

pshrmn:

This PR enables use of dynamic `import()` in the vanilla template.

Fixes https://github.com/codesandbox/codesandbox-client/issues/1930

A syntax error occurs.

**What is the new behavior?**

`import()` works as expected.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged 

I split the way that the change is added based on how Parcel does it internally.

v7 should be ready, it simply uses `parserOpts`.

v6 could use some work. My "fix" uses a plugin, which is what Parcel does for Babel v6. Parcel creates a custom plugin that is essentially the same as `babel-plugin-syntax-dynamic-import`. That wasn't playing nicely when I tried to do the same, so I updated the `babel.6.26.js` module to support `babel-plugin-syntax-dynamic-import`. I don't like this solution, particularly because the `babel.6.26.min.js` module wasn't updated, but it wasn't immediately obvious where these static files get generated.

**Edit:** I realized why the custom plugin wasn't working. `getDefaultCode` serializes the return object, so the function is transformed into `null`. I think the best solution will be to use `babel-plugin-syntax-dynamic-import` provided I figure out how to generate the minified Babel 6 module.


lbogdan:
@pshrmn Very nice!

Doesn't seem to be working with v6, though? http://pr1931.cs.lbogdan.tk/s/muddy-field-v05k2

pshrmn:
@lbogdan I think that is because it is using the minified version, which I didn't fix.

pshrmn:
Ideally I would prefer to get Babel 6 working without touching those static files, but if that is the best approach, I would rather not try manually editing the minified version. Are those modules generated by some external script?

pshrmn:
I removed the Babel 6 support for now. Both static Babel files were added in https://github.com/codesandbox/codesandbox-client/pull/446, so I think that the builds must have happened externally.

CompuIves:
Ahh, good point! I think we can merge this one in the meantime. I think I got the other ones by building `babel-standalone` with webpack.

pshrmn:
Does the PR build through `cs.lbogdan.tk` no longer work? I wanted to do a sanity check that this still works before pinging about getting it merged but I'm getting a 503 error.

lbogdan:
Hey @pshrmn ,

We recently made some improvements to the PR building workflow, please push a new (empty) commit so that it will pick up and build this PR.

lbogdan:
Or you can merge `master`, as there were a few changes in the meanwhile. 🙂

pshrmn:
I think that this is fine to merge. I'm not sure why the integration test is failing; the failing sandbox (https://codesandbox.io/s/6w66jzw3mn) is a Preact starter while this PR only affects Parcel templates.

PR #1946, opened 2019-05-21
Refactor Patron to functional components with hooks

MichaelDeBoey:
Update Patron to functional components with Hooks

Class components

**What is the new behavior?**
Functional components with Hooks

**Checklist**:
- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged 

### DIFF
--- a/packages/app/src/app/pages/Patron/PricingModal/Badge/Particles/index.js
+++ b/packages/app/src/app/pages/Patron/PricingModal/Badge/Particles/index.js
@@ -1,26 +1,25 @@
-import React from 'react';
-import { forEach } from 'lodash-es';
-
 import badges from '@codesandbox/common/lib/utils/badges/patron-info';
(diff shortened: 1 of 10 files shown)


SaraVieira:
Hey 

So I ran this on my computer and now there is a huge delay when you scroll for a point where the animation needs to trigger :(

CompuIves:
I'm thinking that we should maybe skip this PR. In some cases it doesn't make sense to move to hooks IMO, in this case we have a highly optimized class component with hacks to make it more performant. Moving this to hooks would be so much effort that doesn't seem to outweigh the cost.

I'm curious what others think of this, otherwise I'll close the PR.

MichaelDeBoey:
Needs to be rebased anyways, so it's probably better to start from scratch anyways if we want  to have hooks here.

PR #1954, opened 2019-05-22
feat(version): more efficient query

Haroenv:
Note that I blind coded this, so definitely check out in the preview if it works.

reference: https://www.algolia.com/doc/api-reference/api-methods/get-objects/?language=javascript

I also changed the require out for the way smaller `algoliasearch/lite`


SaraVieira:
Hey man!

This completely breaks so I tried to fix but screwed up git and made my own branch by accident :(

Sorry!

This does make it faster tho! Thank you

Im gonna close and merge my fixed branch

PR #1962, opened 2019-05-22
Refactor Advertisement to use useScript

MichaelDeBoey:
I've extracted @Saeris' changes to `Advertisement` out of #1925.

> Removed unused container styles and reorganized code to follow updated coding conventions (to be documented at a later date). Added a new useScript hook (based on https://usehooks.com/useScript) and a hooks directory.

- Class component `CodeFund` 

**What is the new behavior?**
- Rename `CodeFund` component to `Advertisement`
- Created `useScript` hook
- Use `useScript` hook in `Advertisement`
- Refactor code to TS

**Checklist**:
- [ ] Documentation N/A
- [ ] Tests
- [ ] Ready to be merged 


packages/app/src/app/pages/Sandbox/Editor/Workspace/Advertisement/index.ts
```diff
@@ -0,0 +1 @@
+export { Advertisement } from './Advertisement';
```
MichaelDeBoey: @Saeris Why cant we just delete this file and rename `Advertisement.jsx` to `index.tsx`? 🤔

Right now we're creating every time an extra file for no reason at all imo.

packages/app/src/app/hooks/useScript.ts
```diff
@@ -0,0 +1,60 @@
...
+        setState({
+          loaded: true,
+          error: false,
+        });
+      };
+
+      const onScriptError = () => {
+        const index = cachedScripts.indexOf(src);
+
+        if (index >= 0) cachedScripts.splice(index, 1);
```
MichaelDeBoey: It will make this piece of code a bit clearer if you wrap it in curly braces (`{}`).

```js
if (index >= 0) {
  cachedScripts.splice(index, 1);
}
```

CompuIves: Sounds good!

MichaelDeBoey:
@CompuIves @Saeris Came across [`useScript` from @palmerhq's `the platform`](https://github.com/palmerhq/the-platform/blob/e46f715b04f040fb1a7e8ad1d692c80c2184fa78/src/Script.tsx#L28-L30)

Maybe we should use this one (will add a dependency tho) or copy it over, since that's a more modern way of this hook and is `Suspense` compatible? 🤔

CompuIves:
This is nice!! I'll merge this in now, thank you!

> Maybe we should use this one (will add a dependency tho) or copy it over, since that's a more modern way of this hook and is Suspense compatible? 🤔

Let's evaluate it when we are going to support suspense. I want to keep it as simple as possible now.

PR #1963, opened 2019-05-22
🔨 Refactor ConnectionNotice to use hooks

MichaelDeBoey:
I've extracted @Saeris' changes to `ConnectionNotice` out of #1925.

> Small example refactor showing how to use Cerebral with React Hooks

- Use `mobx-react`
- Use `inject` HOC

**What is the new behavior?**
- Use `mobx-react-lite`
- Use new `useStore ` hook

**Checklist**:
- [ ] Documentation N/A
- [ ] Tests
- [ ] Ready to be merged 

### DIFF
--- a/packages/app/src/app/pages/Sandbox/Editor/Workspace/ConnectionNotice/index.ts
+++ b/packages/app/src/app/pages/Sandbox/Editor/Workspace/ConnectionNotice/index.ts
@@ -1,0 +1,1 @@
+export { ConnectionNotice } from './ConnectionNotice';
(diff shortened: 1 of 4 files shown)


packages/app/src/app/pages/Sandbox/Editor/Workspace/ConnectionNotice/index.ts
```diff
@@ -0,0 +1 @@
+export { ConnectionNotice } from './ConnectionNotice';
```
MichaelDeBoey: @Saeris Why cant we just delete this file and rename `ConnectionNotice.jsx` to `index.tsx`? 🤔

Right now we're creating every time an extra file for no reason at all imo.

Saeris: Two reasons: readability in the editor (every open tab becomes index.ts otherwise), and having a quick reference for what a directory exports.

MichaelDeBoey: Sounds reasonable to me

We're not doing this anywhere else in the code, so I think @CompuIves is the best man to decide on this one

CompuIves: As long as we're going to do this consistently in the future I'm okay with this!

PR #1969, opened 2019-05-23
🔨 Refactor Deployment to use hooks

MichaelDeBoey:
I've extracted @Saeris' changes to `Deployment` out of #1925.

- Use `mobx-react`
- Use `inject` HOC

## What is the new behavior?
- Use `mobx-react-lite`
- Use new `useSignals` & `useStore` hooks
- Refactor code to TS

### ZEIT Deploys
- [ ] Verify that deployment modal of ZEIT says you need to sign in
- [ ] Sign into ZEIT
- [ ] Deploy to ZEIT
- [ ] Make change to sandbox
- [ ] Deploy again, verify new URL

### Netlify Deploys
- [x] Deploy to Netlify
- [ ] Check that logs show
- [x] Check that deployment is added
- [x] Make change
- [x] Deploy again, see if deployment is added
- [x] Claim deployment, see if it works

## Checklist
- [ ] Documentation N/A
- [ ] Testing 
- [x] Ready to be merged 


packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import { Netlify } from './Netlify';
+import { Zeit } from './Zeit';
+
+export const Deployment = () => {
+  const {
+    deployment: { getDeploys },
+  } = useSignals();
+
+  useEffect(() => getDeploys(), []); // eslint-disable-line
```
MichaelDeBoey: @Saeris It's a best practice to add the `getDeploys ` function to the `useEffect` dependencies even though you only want the effect to run on mount

```suggestion
  useEffect(() => getDeploys(), [getDeploys]);
```

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/Netlify.tsx
```diff
@@ -0,0 +1,200 @@
...
+    return [];
+  }
+};
+
+export const Netlify = observer(() => {
+  const {
+    deployment: { getNetlifyDeploys, deployWithNetlify },
+    modalOpened,
+  } = useSignals();
+  const { deployment, editor } = useStore();
```
MichaelDeBoey: @Saeris I think you should destructur `deployment` & `editor`  just like you did in the other files

```suggestion
  const {
    deployment: { building, deploying, netlifyClaimUrl, netlifyLogs, netlifySite },
    editor: { currentSandbox },
  } = useStore();
```

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/Netlify.tsx
```diff
@@ -0,0 +1,200 @@
...
+  );
+
+  return (
+    template.netlify !== false && (
+      <Wrapper loading={deployment.deploying}>
+        <WorkspaceInputContainer style={{ marginTop: '1rem', marginBottom: 0 }}>
+          <DeploymentIntegration
+            beta
+            color="#FFFFFF"
+            deploy={() => deployWithNetlify()}
```
MichaelDeBoey: @Saeris This can be simplified

```suggestion
            deploy={deployWithNetlify}
```

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/Zeit.tsx
```diff
@@ -0,0 +1,144 @@
...
+  const { deployment, user } = useStore();
+
+  const [isVisible, setVisibility] = useState(false);
+
+  return user.integrations.zeit ? (
+    <Wrapper loading={deployment.deploying}>
+      <WorkspaceInputContainer style={{ marginTop: '1rem', marginBottom: 0 }}>
+        <DeploymentIntegration
+          color="#000000"
+          deploy={() => deploySandboxClicked()}
```
MichaelDeBoey: @Saeris This can be simplified

```suggestion
          deploy={deploySandboxClicked}
```

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/elements.ts
```diff
@@ -0,0 +1,133 @@
+import styled, { css } from 'styled-components';
+
+const mapColorToState = (state: string, theme: any) => {
+  const STARTING = ['DEPLOYING', 'BUILDING', 'INITIALIZING'];
+  const ERROR = ['DEPLOYMENT_ERROR', 'BUILD_ERROR', 'ERROR'];
+  const STARTED = ['BOOTED', 'READY'];
+
+  if (STARTING.includes(state)) return '#fccb7e';
```
MichaelDeBoey: @Saeris Add curly braces (`{}`) to make this more clear

```suggestion
  if (STARTING.includes(state)) {
    return '#FCCB7E';
  }
```

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/elements.ts
```diff
@@ -0,0 +1,133 @@
...
+  if (ERROR.includes(state)) return theme.red;
```
MichaelDeBoey: @Saeris Add curly braces (`{}`) to make this more clear

```suggestion
  if (ERROR.includes(state)) {
    return theme.red;
  }
```

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/elements.ts
```diff
@@ -0,0 +1,133 @@
...
+  if (STARTED.includes(state)) return theme.green;
```
MichaelDeBoey: @Saeris Add curly braces (`{}`) to make this more clear

```suggestion
  if (STARTED.includes(state)) {
    return theme.green;
  }
```

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/elements.ts
```diff
@@ -0,0 +1,133 @@
...
+  if (state === 'FROZEN') return theme.blue;
```
MichaelDeBoey: @Saeris Add curly braces (`{}`) to make this more clear

```suggestion
  if (state === 'FROZEN') {
    return theme.blue;
  }
```

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/Netlify.tsx
```diff
@@ -0,0 +1,200 @@
...
+export const Netlify = observer(() => {
+  const {
+    deployment: { getNetlifyDeploys, deployWithNetlify },
+    modalOpened,
+  } = useSignals();
+  const { deployment, editor } = useStore();
+
+  const [isVisible, setVisibility] = useState(false);
+
+  useEffect(() => getNetlifyDeploys(), []); // eslint-disable-line
```
MichaelDeBoey: @Saeris It's a best practice to add the `getNetlifyDeploys ` function to the `useEffect` dependencies even though you only want the effect to run on mount

```suggestion
  useEffect(() => getNetlifyDeploys(), [getNetlifyDeploys]);
```

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/Zeit.tsx
```diff
@@ -0,0 +1,144 @@
...
+export const Zeit = observer(() => {
+  const {
+    deployment: {
+      aliasDeployment,
+      deploySandboxClicked,
+      setDeploymentToDelete,
+    },
+    modalOpened,
+  } = useSignals();
+  const { deployment, user } = useStore();
```
MichaelDeBoey: @Saeris I think you should destructur `deployment` & `user`  just like you did in the other files

```suggestion
  const {
    deployment: { deploying, hasAlias, sandboxDeploys },
    editor: { integrations: { zeit } },
  } = useStore();
```

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/Deployment.tsx
CompuIves: Agree, we can add it as commit!

CompuIves: (I don't have access to accept the suggestion)

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/Netlify.tsx
CompuIves: Hmm, for me it looks cleaner to not destructure everything, makes it harder to read.

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/Netlify.tsx
CompuIves: Make sure to test whether it works, sometimes there's some iffy `this` or given params.

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/Zeit.tsx
CompuIves: Same!

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/elements.ts
CompuIves: Agree!

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/elements.ts
CompuIves: Agree!

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/Zeit.tsx
CompuIves: Same as last time, I would not destructure it too much. Makes it header to read for me.

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/Netlify.tsx
CompuIves: Agree!

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/Netlify.tsx
MichaelDeBoey: It's  because @Saeris did it in all his other commits too, so to keep the code consistent.

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/Deployment.tsx
Saeris: Also can't commit the suggested change. GitHub is telling me I don't have permission to push.

But yeah this change is fine. I haven't included signals as dependencies because they should never change.

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/Netlify.tsx
Saeris: I found during development of these refactors that sometimes errors are thrown when you overly destructure like this, that's why it's inconsistent where I drill down this deeply. Right now I'm not sure what the rules should be around destructuring.

Personally I think it's good to declare all of the state properties you'll be using in a component up at the top because you can see at a glance what a component depends on to render. I'll have to talk to Ives on this one.

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Deployment/Netlify.tsx
Saeris: Yeah a lot of the time where I'm doing this, there used to be a class method or something similar where there was a named function/method that wrapped a signal call. This is probably fine but it should be tested of course.

MichaelDeBoey:
@CompuIves @Saeris I've rebased onto `master` and pushed the suggested changes except for the  destructuring.
When decided, I'll push those changes too. 🙂

MichaelDeBoey:
Fixed some TypeScript errors (that weren't in #1986 yet) too

MichaelDeBoey:
@CompuIves Since all checks pass, I think this one can be merged after a last review 🙂

packages/app/src/app/pages/Sandbox/Editor/Workspace/index.tsx
```diff
@@ -1,20 +1,22 @@
 import VERSION from '@codesandbox/common/lib/version';
 import { observer } from 'mobx-react-lite';
 import React from 'react';
+//  Fix css prop types in styled-components (see https://github.com/DefinitelyTyped/DefinitelyTyped/issues/31245#issuecomment-463640878)
+import * as CSSProps from 'styled-components/cssprop'; // eslint-disable-line
```
CompuIves: Is there a way to not do this? I also am not sure what the advantage of the `css` prop is over `style`.

Saeris: I believe the core difference between the `css` prop and the `style` prop is that `style` is in-lined on the element, and `css` compiles to a styled component (ie: a uniquely named class). That would mean they have different levels of specificity as well as in-lining the styles into the markup.

However, just as we don't want to bloat the rendered HTML with styles, we shouldn't be bloating our JSX with style definitions either, as it makes it more difficult to read the structure of a component. It's generally better if we avoid inline styles entirely and use `styled` instead to definite presentation attributes in an `elements.ts` file.

The only reason why we would need to import the cssprop type definition is if we wanted a component to accept inline styles from a parent.  The better option would be to wrap the component in a call to `styled` instead.

MichaelDeBoey: I haven't touched @Saeris in this PR, but if you guys want I can convert the components with a `css` props to be a new styled component.

Saeris: Yeah I think when I went over this originally, I kept some things as they were because I wasn't confident it didn't have some other purpose. My vote is to do away with inline styles, so if you'd like to make the change go right ahead.

MichaelDeBoey: I'm on it! 😉

MichaelDeBoey: @CompuIves @Saeris Converted all `css` prop call to new components in a `elements.ts` file

MichaelDeBoey:
@CompuIves I tested the Netlify build flow and that one works, except for the "show logs" button that's not showing.
I saw there was a problem with the WebSocket connection too, so maybe it has something to do with that? 🤔
```sh
WebSocket connection to 'wss://pr1969.build.csb.dev/socket/websocket?guardian_token=eyJhbGciOiJIUzUxMiIsInR5cCI6IkpXVCJ9.eyJhdWQiOiJDb2RlU2FuZGJveCIsImV4cCI6MTU2NTI1ODU3OCwiaWF0IjoxNTYyNjY2NTc4LCJpc3MiOiJDb2RlU2FuZGJveCIsImp0aSI6Ijk2YzhlOTYwLWQ0MzctNGZhMy1iOTljLTUzMDM0YWYwNGJmMCIsIm5iZiI6MTU2MjY2NjU3Nywic3ViIjoiVXNlcjpmZDE2ZmI3Mi1jNDllLTQ5NDktYjZlZi1mZTVjZWUwOWU0MjgiLCJ0eXAiOiJhY2Nlc3MifQ.xuUQwYx3ArzROl9wQQPlzohfrPr9uoFmn4apmu_-xjoCdfIxxkDJA3e2evpGR-2Im0q8qzjOeU5E8AUyUtGZJw&vsn=2.0.0'
failed: Error during WebSocket handshake: Unexpected response code: 200
```

I couldn't test the ZEIT build flow because of #2169

CompuIves:
Thanks a lot @MichaelDeBoey, this is really good. So two things left for testing are:

1. View logs (why doesn't it show), seems like the webhook isn't related
2. ZEIT

I can do that now!

MichaelDeBoey:
@CompuIves I won't have time until late this evening to take a look at it unfortunately, so be my guest if you have 🙂

CompuIves:
After doing some testing on Netlify I also noticed that the status doesn't update. It should show this: 
image

But it shows this:
image


I'm not sure if this is related to me testing on `localhost`? cc @SaraVieira

CompuIves:
Oh wait now it suddenly shows "Building..." but much later, maybe an observability issue?

CompuIves:
An observer was missing.

packages/app/src/app/components/DeploymentIntegration/DetailInfo/DetailInfo.tsx
```diff
@@ -10,7 +10,7 @@ export const DetailInfo = ({ info, deploy, bgColor, light, loading }) => (
       <Info light={light}>{info}</Info>
     </Margin>
 
-    <Button small disabled={loading} onClick={deploy}>
+    <Button small disabled={loading} onClick={() => deploy()}>
```
CompuIves: This is quite important, otherwise you get this:

image

or other subtle bugs. This is why I prefer an arrow func whenever we don't want to explicitly pass the event to the function.

MichaelDeBoey: Will keep that in mind next time 🙂

MichaelDeBoey:
I just figured out I _always_ need the observer when I use the `useStore` hook, so will take that in mind too from now on. 🙂

PR #1786, opened 2019-04-18
Rewrote circleci/config.yml to 2.1 schema

thecodechef:
**What kind of change does this PR introduce?**

**What is the current behavior?**

**What is the new behavior?**

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged 
- [ ] Added myself to contributors table

PR #1989, opened 2019-05-28
CircleCI: fix check for environment variable not set

lbogdan:
When using `[ -n "$DOCKER_USER" ]` to check if the environment variable is set, the quotes are mandatory. PR https://github.com/codesandbox/codesandbox-client/pull/1786 removed the quotes, this PR adds them back.


lbogdan:
The check not working:

image

PR #1990, opened 2019-05-28
allow webcam and audio when possible

SaraVieira:
Screenshot 2019-05-28 at 17 58 17


CompuIves:
Great - awesome that you also took the share model into account!

PR #1993, opened 2019-05-28
Fix .eslintrc

MichaelDeBoey:
`overrides` [should be an array](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/eslintrc.json#L547)

`overrides` isn't an array

**What is the new behavior?**
`overrides` is an array

**Checklist**:
- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged 

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -45,40 +45,42 @@
     "react-hooks/rules-of-hooks": "error",
     "react-hooks/exhaustive-deps": "warn"
   },
-  "overrides": {
-    "files": ["**/*.ts", "**/*.tsx"],
-    "parser": "@typescript-eslint/parser",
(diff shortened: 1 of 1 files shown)


CompuIves:
Hmm, interesting. I got it from here: https://github.com/facebook/create-react-app/blob/master/packages/eslint-config-react-app/index.js#L55-L103. Would be good to make it an array anyway, but I don't think it works yet (at least in my editor I don't see eslint warnings in ts files).

MichaelDeBoey:
Didn't see them either, but thought it was something with my IDE...

CompuIves:
Oh well, let's merge this in the meantime!

PR #1998, opened 2019-05-30
🔨 Replace Parcel and VuePress Icons

Saeris:
Dropping the old icons for the new SVGs published in v0.3.0 of @codesandbox/template-icons


Saeris:
Please note, if there's an error with linting regarding missing type definitions, I think that's the result of a build error I need to investigate on the template-icons package. Pika should be running it's full build during publish, but it appears to have skipped creating new types for some reason. It shouldn't affect the icons rendering however.

CompuIves:
Great!

PR #2013, opened 2019-06-02
fix filter options in deleted sandboxes

jyash97:
Looks like a bug, but I am not sure if this was intentional 

Updating Filter option message when there are no deleted sandboxes

We currently see all the templates even when we don't have a deleted sandbox and when we deselect a template it shows message `Showing 0 templates` and in `OverlayContainer`, select all option is always checked. 
image

**What is the new behavior?**
Now if we don't have sandboxes it shows `Showing 0 templates` with message `No templates found`.Check the below gif :
CSB

**What steps did you take to test this?**

**Checklist**:

- [ ] Documentation - N/A
- [x] Testing - Have added a gif to show the working 
- [ ] Ready to be merged 
- [ ] Added myself to contributors table  - Already Contributor for Code


CompuIves:
Very good and clean implementation! Thanks a lot for this contribution. Merging it in now!

PR #2023, opened 2019-06-05
Adds prisma and netlify icons

SaraVieira:
Adds some more icons and deletes ones not needed because they were added to vscode icon theme we use

Screenshot 2019-06-05 at 12 26 48


CompuIves:
Beautiful!

PR #2025, opened 2019-06-05
Add a check to make sure that versions are strings in version UI

### LINKED ISSUE #1908: Can't access sandbox after adding alias in package.json
# 🐛 bug report

## Description of the problem
Can't access my sandbox 
Black page when loading the sandbox. I added an alias in the package.json (probably at the wrong place in the package.json)

### Your Environment
OSX chrome

| Сodesandbox      | 88zo93kxz0
| Browser          | Chrome
| Operating System |OSX

CompuIves:
Bug fix, fixes #1908

We check if a version is a string, if it's not a string we don't render the dependency in the dependency overview.

There is no check for a string, which causes React to error since we try to render an object.

**What is the new behavior?**

No crashing editor on object for version :)

**What steps did you take to test this?**

Loaded https://codesandbox.io/s/88zo93kxz0

**Checklist**:

- [ ] Documentation
- [x] Testing 
- [x] Ready to be merged 

### DIFF
--- a/packages/app/src/app/pages/Sandbox/Editor/Workspace/Dependencies/VersionEntry/index.js
+++ b/packages/app/src/app/pages/Sandbox/Editor/Workspace/Dependencies/VersionEntry/index.js
@@ -126,6 +126,10 @@ export default class VersionEntry extends React.PureComponent {
   render() {
     const { dependencies, dependency } = this.props;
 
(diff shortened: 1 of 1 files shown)


SaraVieira:
das very gud

danke u

PR #2028, opened 2019-06-06
[Refactor] Remove modules from editor

CompuIves:
This is a big refactor. The goal of this PR is to get rid of `module` (so `module.id` and `module.shortid`) from the codebase so we have a better groundwork for an asynchronous file system. We will use `fs` for all loading/saving for files in CodeSandbox. The main functionality enabled from this will be that we can edit any file, even if they're not stored in a sandbox.

There's a lot going on, here's an incomplete list of things that are refactored:

- [ ] Live
	- [x] User selections
	- [x] Current module
	- [ ] Code operations
	- [ ] Module moving/creating/deleting
- [ ] Unsaved file storage

The question here is where we want to have the source of truth for unsaved files (so edited code that isn't saved). Do we want it in the Cerebral store and sync it to VSCode, or do we want to use VSCode as the source of truth? The latter option would make us very dependent on VSCode, which I don't prefer. On the other hand, how dependent are we already?

Right now I introduced a new field on the store called `changedFiles`, this is an array with the changed modules (`{ code: string, path: string }`). This is the first way I mentioned, alternatively we can use the vscode effect (discussed with @christianalfoni) as our source of truth for the models. I actually like that idea a lot.

cc @christianalfoni this is what I've been thinking of. But now that I'm doing all these changes I'm thinking it might be safer to wait for Overmind.


packages/app/src/app/store/providers/ModelManager/ModelManager.ts
```diff
@@ -0,0 +1,24 @@
+interface IModel {
```
CompuIves: This was my initial idea, but putting something like this in an effect would be ideal.

PR #2033, opened 2019-06-06
fix(dashboard): Properly highlight the trash menu when the route is active

arthurdenner:
Bug fix.

When you are in a dashboard route that's not `dashboard/trash` (e.g `dashboard/recent`) and switch to the trash route, the menu doesn't update properly because the `TrashItem` component doesn't rerender.

And if you start in the `dashboard/trash` route and switch to another menu, the menu stays highlighted for the same reason above.

## What is the new behavior?
The menu updates properly in both cases described.

Note: Only passing down the `currentPath` prop would solve the problem, but I decided to add the `active` computation to prevent someone from thinking that the prop is useless and remove it, bringing back the bug.

Follow the use cases described in the current behaviour section.

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged 

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Sidebar/TrashItem/index.js
+++ b/packages/app/src/app/pages/Dashboard/Sidebar/TrashItem/index.js
@@ -7,11 +7,12 @@ import { inject, observer } from 'mobx-react';
 
 import Item from '../Item';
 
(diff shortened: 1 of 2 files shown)


CompuIves:
Man you're on fire!! This has been bugging me for a while, thank you so much for building this!!

PR #2040, opened 2019-06-07
fix(devtools): Fix terminal output when it's positioned along with the browser

### LINKED ISSUE #1759: 🐛 Terminal doesn't display anything when moved
See attached gif:

codesadbox

Appears that unless the terminal panel is in it's original position, it won't display any output when scripts are run from the server control panel. Steps to reproduce should be as simple as opening any container sandbox, dragging the terminal up next to the browser tab, and running any script from the server control panel to verify.

arthurdenner:
Bug fix.

Closes #1759.


CompuIves:
Nice!!! I was meaning to fix this, thank  you!

PR #2042, opened 2019-06-09
fix(git): Properly update Git integration UI after a commit

### LINKED ISSUE #1076: after adding a commit redirect fails
# 🐛 bug report

- connect to github repository
- add something to the code and create a commit
- UI will say "creating commit", and then on the left "fetching changes"
- changes will never be fetched, nothing happens after, you need to manually go back to the `/s/github/...` url

### Your Environment
| Browser | Chrome
| Operating System | Ubuntu 16

arthurdenner:
Bug fix. Closes #1076.

After finishing a commit, the state `originalGitChanges` is set to `null` and because of [this ternary](https://github.com/codesandbox/codesandbox-client/blob/master/packages/app/src/app/pages/Sandbox/Editor/Workspace/items/GitHub/Git/Git.tsx#L80) the UI shows `Fetching changes...` but it's not doing anything.

## What is the new behavior?
By setting `originalGitChanges` to `undefined` instead of `null` when the state is "empty", we can use `defaultProps` to handle `undefined` values and remove the `&& gitChanges` condition in the ternary mentioned above, properly updating the UI to show `There are no changes`.

**Notes:**

- This approach raises a warning in `mobx-state-tree` about `trying to read or write to an object that is no longer part of a state tree`, so it may not be the best solution;
- One solution I thought is to always use an object with the expected keys (`added`, `deleted`, `modified` and `rights`) in the `originalGitChanges` state instead of setting it to `null` but I decided to propose this change because `undefined` is preferred over `null` in some TypeScript styleguides;
- Another solution would be to fire `fetchGitChanges` after the commit and everything would work as expected, but we would be doing an unnecessary request;

Let me know if you disagree with the solutions I provided and mentioned, and if possible, please suggest another approach to solve the problem. :pray: 

I reproduced what is described in the issue.

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged 

### DIFF
--- a/packages/app/src/app/pages/Sandbox/Editor/Workspace/items/GitHub/Git/Changes/Changes.tsx
+++ b/packages/app/src/app/pages/Sandbox/Editor/Workspace/items/GitHub/Git/Changes/Changes.tsx
@@ -18,3 +18,7 @@ export const Changes = ({ changes, color, hideColor, Icon, title }) => (
     ))}
   </>
 );
+
+Changes.defaultProps = {
+  changes: [],
+};
--- a/packages/app/src/app/pages/Sandbox/Editor/Workspace/items/GitHub/Git/Git.tsx
+++ b/packages/app/src/app/pages/Sandbox/Editor/Workspace/items/GitHub/Git/Git.tsx
@@ -79,7 +79,7 @@ export const Git = observer(() => {
 
         {!isFetching ? (
           <Margin top={1}>
-            <TotalChanges gitChanges={gitChanges} />
+            <TotalChanges gitChanges={gitChanges || {}} />
 
             {changeCount > 0 ? (
               <Margin top={1}>
--- a/packages/app/src/app/pages/Sandbox/Editor/Workspace/items/GitHub/Git/Git.tsx
+++ b/packages/app/src/app/pages/Sandbox/Editor/Workspace/items/GitHub/Git/Git.tsx
@@ -74,10 +74,10 @@ export const Git = observer(() => {
 
       <Margin bottom={0}>
         <WorkspaceSubtitle>
-          Changes ({gitChanges ? changeCount : '...'})
+          Changes ({isFetching ? '...' : changeCount})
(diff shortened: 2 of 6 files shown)


CompuIves:
Interesting!

> This approach raises a warning in mobx-state-tree about trying to read or write to an object that is no longer part of a state tree, so it may not be the best solution;

I wonder where this error comes from, is it a warning from the component? 

I always liked putting `gitChanges` explicitly to `null` (or `undefined`, doesn't really matter to me) because it communicates that the changes haven't been loaded yet. That's also why I prefer it over an object with `added`/`modified`/`deleted`. I think that the update with `isFetching` is a really good change that we should've done from the start. The only thing that I don't understand is the warning, since I would expect that it uses the `defaultProps`.

I think the best solution would be to do an additional request. It's a cheap request (only few bytes of payload) and it will make sure that also after the commit we can show if something changed between the source and the current sandbox (I'm not sure if that's even possible, but with that we can be sure 😄).

arthurdenner:
> I wonder where this error comes from, is it a warning from the component?

I think so. However, it's not outputting the warning anymore. Weird. :thinking:

I'll revert some changes and update the branch to make an additional request.

Thanks for the suggestion!

arthurdenner:
@CompuIves, I reverted the changes related to `undefined` instead of `null` and added the request as you suggested. I had to put a fallback value to the `gitChanges` prop on `TotalChanges`, otherwise the code would break because it expects an object.

CompuIves:
Ah, that makes sense. Great! The fix looks good to me, did you test the last version? If it's tested it's ready for merge!

arthurdenner:
> Did you test the last version?

I did. It's working as expected since we fire the `fetchGitChanges` action instead of set `originalGitChanges` to `null`, but you may check please? I don't want to risk it. :sweat_smile:

CompuIves:
Hah, I'll do a last test and merge it in. Thanks for this fix! Was nagging me for a whule.

PR #2051, opened 2019-06-11
Keep history in one place

### LINKED ISSUE #2050: Symbol % in the address bar crashes the app
# 🐛 bug report

## Description of the problem

I was trying to check how react-router works with encoded url params but faced codesandbox crash when i typed the % symbol in the address bar.

## Steps to reproduce
Open sandbox: https://codesandbox.io/s/vanilla
Type % symbol in the address bar of browser window
Codesandbox will crash with errors in the console

I captured this on a video: https://streamable.com/sbjp0

### Your Environment

| Сodesandbox      | https://codesandbox.io/s/vanilla |
| Browser          | Yandex Browser 19.6 (like Chrome 74) |
| Operating System | Windows 10 Pro 17763.503 |

pshrmn:

Bug fix-ish

Currently the fake browser's history is stored in two places: the `Preview` component's state and in the `url-listeners` module. 

## What is the new behavior?

The history is only stored in the `url-listeners` module. Instead of having the `Preview` store an array of locations and an index, it keeps track of the current url and knows whether or not it can go backwards and forwards.

This PR also removes the automatic decoding of the URL displayed in the `Preview`'s address bar, which fixes #2050.

I have verified that it works as I expect it to with this sandbox: https://codesandbox.io/s/suspicious-river-unnjk This includes:

1. On load, neither forward nor back are usable
2. On navigation (push), the user can go back.
3. After going back, the user can go forward.
4. If you push while you can go forward, you can no longer go forward.
5. If you refresh while typing a pending URL in the address bar, the current (not pending) URL is the one that gets used.

I wasn't entirely sure what the intended behavior for refreshing/hitting enter in the address bar are. Both of those currently reset the history, which I assume is the correct behavior, but I haven't looked closely enough to see if that

I am confident in these changes, but I also wouldn't want to break the `Preview`'s navigation for anyone, so I would appreciate any double checking.

## Checklist

- [ ] Documentation
- [x] Testing 
- [x] Ready to be merged 

### DIFF
--- a/packages/common/src/components/Preview/index.tsx
+++ b/packages/common/src/components/Preview/index.tsx
@@ -52,10 +52,10 @@ export type Props = {
 
 type State = {
   frameInitialized: boolean;
-  history: string[];
-  historyPosition: number;
+  url: string;
   urlInAddressBar: string;
-  url: string | undefined;
+  back: boolean;
+  forward: boolean;
   showScreenshot: boolean;
   useFallbackDomain: boolean;
 };
@@ -81,12 +81,14 @@ class BasePreview extends React.Component<Props, State> {
     // templates that are executed in a docker container.
     this.serverPreview = getTemplate(props.sandbox.template).isServer;
 
+    const initialUrl = this.currentUrl();
+
     this.state = {
       frameInitialized: false,
-      history: [],
-      historyPosition: 0,
-      urlInAddressBar: this.currentUrl(),
-      url: null,
+      urlInAddressBar: initialUrl,
+      url: initialUrl,
+      forward: false,
+      back: false,
       showScreenshot: true,
       useFallbackDomain: false,
     };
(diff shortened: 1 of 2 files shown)


lbogdan:
Doing a quick test on https://pr2051.build.csb.dev/s/unnjk , it looks like when manually entering an URL or clicking the refresh button, the back button remains active.

pshrmn:
Interesting. I wonder if there is a race condition because it always works when I am debugging. I'll try to narrow down the issue to figure out where it is going wrong.

pshrmn:
I think that the issue is that when the `Preview` is refreshed/a new URL is submitted, the frame is not registered when the `urlchange` event is emitted, so the app's listeners to do not receive the dispatch. I'm piecing things together, so maybe I'm missing some implementation details, but that appears to be what the problem is.

As far as a solution goes, the quick fix would be to hard-code the backwards/forwards buttons to be turned off by a refresh/submit. Ideally the actual dispatch would make it through, but that feels too far outside the scope of this PR.

pshrmn:
I hard coded resetting the buttons on refresh/url submit.

lbogdan:
@pshrmn LGTM. Before a final test, can you please do a merge with `master` and fix the conflicts? Thanks!

pshrmn:
Okay, I rebased and I think that this is in a good place now. Sorry about the delay!

CompuIves:
Perfect! Thank you so much for this change @pshrmn, big improvement that required a lot of thinking.

Issue #2062, opened 2019-06-14
Cannot Open Another terminal (Ember)


jrock2004:
# 🐛 bug report

## Description of the problem
I have created an ember app and I want to open a tab so I can generate new ember files `ember g route home`. All I get it is this

<img width="959" alt="image" src="https://user-images.githubusercontent.com/655716/59521948-7bd2d300-8e9b-11e9-9bee-41b46da0a6f7.png">

## How has this issue affected you? What are you trying to accomplish?
I want to use the ember cli tool to generate new files

### Link to sandbox: [link]() (optional)
https://codesandbox.io/s/ember-international-rmjdz

### Your Environment

| Software         | Name/Version |
| ---------------- | ------------ |
| Сodesandbox      | Not sure|
| Browser          | Firefox 67.0.2 | 
| Operating System | Mac Mojave |

lbogdan:
Hey @jrock2004 , we're sorry for this, looks like it's a regression introduced by #2040, which I reverted in #2063, and already deployed. Can you please check again and let us know it works as expected? Thanks!

jrock2004:
No worries, its now bring up the terminal. Thank you guys

PR #2063, opened 2019-06-14
Revert #2040

lbogdan:
Looks like #2040 is causing #2062 , reverting to check that's actually the case.


lbogdan:
Confirmed, will merge and deploy ASAP.

PR #2065, opened 2019-06-15
📦 Upgrade dependency styled-components to v5.x

Saeris:
Check out this Medium post for more info on v5:

https://medium.com/styled-components/announcing-styled-components-v5-beast-mode-389747abd987


Saeris:
Appears to break Gatsby SSR, going to roll back the changes to the homepage dependencies.

Saeris:
Looks like we can't mix versions. Maybe all the common/shared packages can use an older version, but we can't have v5 for common and v4 for homepage.

I'll take a look into what can be done about the homepage for now. Worst case scenario we'll just have to wait to push this out to whenever this comes out of beta and there' Gatsby support for it.

MichaelDeBoey:
@Saeris Would be nice to have `styled-components@5.0`, but I think we should wait until it's out of beta before updating.

This will otherwise set a path of having more beta dependencies, which is not prefered.

Saeris:
We're probably going to have to wait. This was brought up in Slack and it was requested to see if we can get it working. If I can't find a fix for the homepage, then we'll just wait and I'll finish off this PR once all the pieces are in place.

CompuIves:
Ah, I didn't know it was still in beta. Let's wait until it's stable!

MichaelDeBoey:
Latest version at this moment is `4.3.1`

https://www.npmjs.com/package/styled-components

SaraVieira:
Closing this because we are waiting and then when its not beta anymore we need to review gatsby anyway

PR #2003, opened 2019-05-31
TEMPLATES YAAAY

SaraVieira:
We add templates

YAY????

PR #2066, opened 2019-06-16
🔨 Refactor app/components to TypeScript

Saeris:
This PR refactors (nearly) everything under app/components to TypeScript, adding new types and applying coding conventions.

It's being doing in multiple stages:

- [x] Change file extensions of all components to be refactored by this PR
- [x] Apply coding conventions and add types to each component in a separate commit and request review
- [x] Make requested changes to touched files
- [x] Reorganize files and update import/export references in each component and component instance throughout the code base to use named exports as per coding conventions and update props where necessary
- [x] Test each component to ensure they are still working as they did previously
- [x] Request final approval and merge

This is being done as one large PR instead of several smaller ones as each component is relatively small (most are single files and require very few changes besides adding types and minor coding style changes). Changes to each component have been done in their own commit to make it easier to review changes on a per-component basis.

The why is simple: we're in the process of transitioning the entire code base to TypeScript so it makes sense to get the easy wins out of the way first so that when we refactor more complex components, we'll have the updated prop types already in place. Along the way it makes sense to apply some consistency and best practices to existing code as well.

Here are some basic rules regarding coding conventions (to be more thoroughly documented at a later date):

- Always use Named Exports (here's some good reasons why this is a good idea: https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/)
- Write functional components with hooks (makes it much easier to detect unused methods, among numerous other good reasons)
- Prefer arrow functions (death to `this`)
- Prefer destructuring to property accessors
- Components should always be their own file with a Capitalized name matching the component name
- Multiple file components should be in a Capitalized folder matching the component name, with an `index.ts` file handling all exports to be consumed by outside components
- Prefer `interface` over `type` for typing Props
- In general, prop interfaces should follow a IComponentNameProps naming scheme, though this is largely stylistic
- Prefer `const SomeComponent: React.FC<ISomeComponentProps> = ({ . . . }) => { }` over `const SomeComponent = ({ . . . }: ISomeComponentProps) => { }`  to avoid needing to define common React component props
- For styled components, Inline types are fine where their footprint is small, ie: 
```typescript
const SomeElement = styled.div<{ large: boolean }>` 
  ${({ large }) => css`
    /* 
    Always wrap style declarations in a single function call and don't forget to
    use "css" for syntax highlighting!
    */
  `}
`
```
- All type interface definitions should live in a `types.ts` file adjacent to the component to reduce clutter in the component definition
- If a component uses GraphQL queries/mutations, all GraphQL documents should live in their files adjacent to the component that uses them, ie: `queries.gql` and `mutations.gql` to reduce clutter in the component definition
- All styles should be applied with styled-components and all styled elements should be defined in elements.js to reduce clutter in the component definition
- Global styles should be defined using styled-components in a `GlobalStyles.ts` file adjacent to the root component that requires it
- When using props in styled components, always wrap all of the styles in a single arrow function to make it easy to destructure props all in one place and always wrap nested template literals with the `css` tagged template to enable style linting/syntax highlighting.
- Style rules should be well ordered (to be enforced by stylelint at a later date). In general this means rules should be grouped by category, starting with Position > Box Model > Fonts > Animation > Misc > pseudo-selectors. For now this is low-priority.

The general rule of thumb here is: avoid clutter and if you have to scroll a component file, it probably needs to be broken up into smaller pieces. By separating concerns into small, well-organized files, code will be easier to read and thus easier to maintain.

That should cover more than what's implemented by this PR. These rules will be applied progressively so as to make the initial code review for refactors easier (as for the most part it will just require some components to be broken up into multiple files).


packages/app/src/app/components/Stat/index.tsx
```diff
@@ -1,14 +1,24 @@
...
-function StatComponent({ count, name }) {
-  return (
-    <Stat>
-      <Property>{name}</Property>
-      <Number>{formatNumber(count)}</Number>
-    </Stat>
-  );
+import { Property, Number, Container } from './elements';
+
+interface IStatProps {
```
MichaelDeBoey: @Saeris I think you should use a normal `type` instead of an `interface`

```suggestion
type StatProps = {
```

packages/app/src/app/components/SubscribeForm/CheckoutForm/index.tsx
```diff
@@ -1,111 +1,104 @@
...
+import {
+  CardContainer,
+  Card,
+  NameInput,
+  ErrorText,
+  Label,
+  Submit,
+} from './elements';
+
+interface ICheckoutFormProps {
```
MichaelDeBoey: @Saeris I think you should use a normal `type` instead of an `interface`

```suggestion
type CheckoutFormProps = {
```

packages/app/src/app/components/SubscribeForm/CheckoutForm/index.tsx
```diff
@@ -1,111 +1,104 @@
...
-      this.setState({ errors: {}, name: nextProps.name });
-    }
-  }
-
-  setName = e => {
-    if (e) {
-      this.setState({ errors: {}, name: e.target.value });
-    }
-  };
+interface IErrorsState {
```
MichaelDeBoey: @Saeris I think you should use a normal `type` instead of an `interface`

```suggestion
type ErrorsState = {
```

packages/app/src/app/components/SubscribeForm/index.tsx
```diff
@@ -2,33 +2,39 @@ import React from 'react';
 import { StripeProvider, Elements } from 'react-stripe-elements';
 import { STRIPE_API_KEY } from '@codesandbox/common/lib/utils/config';
 import CheckoutForm from './CheckoutForm';
-
 import { Container } from './elements';
 
-function SubscribeForm({
+interface ISubscribeFormProps {
```
MichaelDeBoey: @Saeris I think you should use a normal `type` instead of an `interface`

```suggestion
type SubscribeFormProps = {
```

packages/app/src/app/components/SubTitle/index.ts
```diff
@@ -1,15 +1,20 @@
...
-  width: 100%;
-  font-size: 1.75rem;
-  color: ${props =>
-    props.theme.light ? 'rgba(0, 0, 0, 0.7)' : 'rgba(255, 255, 255, 0.7)'};
-  font-weight: 300;
-  margin-top: 0;
-  margin-bottom: 1.5rem;
-  line-height: 1.4;
+const SubTitle = styled.h2`
+  ${({ delay = 0, theme }: { delay: number | null; theme: any }) => css`
```
MichaelDeBoey: @Saeris I think it's better to extract the `SubTitleProps` here, just to be consistent

packages/app/src/app/components/Title/index.tsx
```diff
@@ -1,14 +1,18 @@
...
-  color: white;
-  font-size: 2.5rem;
-  font-weight: 300;
-  background-color: transparent;
-  margin-top: 0;
-  border: none;
-  outline: none;
-  text-align: center;
+const Title = styled.h1`
+  ${({ delay = 0 }: { delay: number | null }) => css`
```
MichaelDeBoey: @Saeris I think it's better to extract the `TitleProps` here, just to be consistent

packages/app/src/app/components/UploadProgress/Cube/elements.ts
```diff
@@ -9,35 +9,37 @@ const rotateAnimation = keyframes`
...
 `;
 
-export const Cube = styled.div`
-  position: absolute;
-  width: ${props => props.size + 1}px;
-  height: ${props => props.size + 1}px;
-
-  * {
+export const Container = styled.div`
+  ${({ size }: { size: number }) => css`
```
MichaelDeBoey: @Saeris I think it's better to extract the `ContainerProps` here, just to be consistent

packages/app/src/app/components/UploadProgress/Cube/elements.ts
```diff
@@ -9,35 +9,37 @@ const rotateAnimation = keyframes`
...
 
 export const Sides = styled.div`
-  ${({ noAnimation }) =>
-    !noAnimation &&
-    css`
-      animation: ${rotateAnimation} 1s linear infinite;
-    `};
-  transform-style: preserve-3d;
-  transform: rotateX(-37.5deg) rotateY(45deg);
+  ${({ size }: { size: number }) => css`
```
MichaelDeBoey: @Saeris I think it's better to extract the `SidesProps` here, just to be consistent

packages/app/src/app/components/UploadProgress/Cube/elements.ts
```diff
@@ -9,35 +9,37 @@ const rotateAnimation = keyframes`
...
+      background-color: rgba(242, 119, 119, 0.5);
+      border: ${size / 10}px solid white;
+    }
+  `}
 `;
 
 export const Side = styled.div`
-  transform-origin: 50% 50%;
-  transform: ${({ rotate }) => rotate} translateZ(${({ size }) => size / 2}px);
+  ${({ rotate, size }: { rotate: string; size: number }) => css`
```
MichaelDeBoey: @Saeris I think it's better to extract the `SideProps` here, just to be consistent

packages/app/src/app/components/UploadProgress/elements.ts
```diff
@@ -1,38 +1,40 @@
...
 `;
 
 export const DeployAnimationContainer = styled.div`
-  display: flex;
-  justify-content: center;
-  align-items: center;
-  padding: 3rem;
-
-  ${({ deploying }) => deploying && delayInEffect(0, false)};
+  ${({ deploying }: { deploying: boolean }) => css`
```
MichaelDeBoey: @Saeris I think it's better to extract the `DeployAnimationContainerProps` here, just to be consistent

packages/app/src/app/components/UploadProgress/elements.ts
```diff
@@ -54,10 +56,12 @@ const cubeAnimation = keyframes`
 `;
 
 export const StyledCube = styled(Cube)`
-  position: absolute;
-  animation: ${cubeAnimation} 2s ease-in infinite;
-  animation-delay: ${({ i }) => i * 0.5}s;
-  transform: translateY(20px) translateX(-100px) scale(0, 0);
+  ${({ delay }: { delay: number }) => css`
```
MichaelDeBoey: @Saeris I think it's better to extract the `StyledCubeProps` here, just to be consistent

packages/app/src/app/components/UploadProgress/index.tsx
```diff
@@ -10,25 +9,28 @@ import {
...
-              <StyledCube key={i} i={i} size={20} />
-            ))}
-            <StyledLogo width={70} height={70} />
-          </DeployAnimationContainer>
-          <DeployText>{message}</DeployText>
-        </div>
-      )}
-    </Container>
-  );
+interface IUploadProgressProps {
```
MichaelDeBoey: @Saeris I think you should use a normal `type` instead of an `interface`

```suggestion
type UploadProgressProps = {
```

packages/app/src/app/components/Alert/index.tsx
```diff
@@ -1,23 +1,35 @@
...
-        <Button small block secondary onClick={onCancel}>
-          Cancel
-        </Button>
-        <Button small block danger onClick={onDelete}>
-          Delete
-        </Button>
-      </Buttons>
-    </Container>
-  );
+interface IAlertProps {
```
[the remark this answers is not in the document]
Saeris: We're going to stick with `interface` over `type` because interfaces are what are used in the TypeScript language docs for JSX, in Microsoft's React TypeScript starter, as well as various other tutorials. There doesn't appear to be a big advantage to one over the other, but if that's what the language docs use, better to stay consistent.

packages/app/src/app/components/DelayedAnimation/index.ts
```diff
@@ -2,5 +2,5 @@ import styled from 'styled-components';
 import delayEffect from '@codesandbox/common/lib/utils/animation/delay-effect';
 
 export default styled.div`
-  ${props => delayEffect(props.delay || 0)};
+  ${({ delay }: { delay: number }) => delayEffect(delay || 0)};
```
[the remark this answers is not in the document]
Saeris: Yeah we can probably extract these out. I haven't made a hard decision whether we should extract out all props to interface definitions, because if we do that for all styled elements then we should also move those interfaces to `types.ts` to reduce clutter.

packages/app/src/app/components/DeploymentIntegration/index.tsx
```diff
@@ -1,67 +1,59 @@
...
+  Up,
+  Down,
+} from './elements';
 
-const Integration = ({
+interface IDeploymentIntegrationProps {
+  light?: boolean;
+  loading: boolean;
+  bgColor: string;
+  Icon: React.ComponentClass<any> | React.StatelessComponent<any>;
```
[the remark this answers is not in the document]
Saeris: Sure, we can do this. Documentation on this was sparse so what I put her was just the first good answer I could find.

packages/app/src/app/components/EditableTags/index.tsx
```diff
@@ -53,13 +53,15 @@ const GlobalStyle = createGlobalStyle`
   }
 
   .react-tagsinput-tag a::before {
-    content: " ×";
+    content: ' ×';
   }
-`;
+`}`;
```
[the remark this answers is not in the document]
Saeris: No this closes the css tagged template:

```jsx
const GlobalStyle = createGlobalStyle`${({ color }: { color: any }) => css`
  // ... styles
`}`;
```

packages/app/src/app/components/Integration/DetailInfo/index.tsx
```diff
@@ -3,30 +3,39 @@ import CrossIcon from 'react-icons/lib/md/clear';
...
+  heading: string;
+  info: string;
+  onSignOut?: () => void;
+  onSignIn?: () => void;
 }
 
+const DetailInfo = ({
+  heading,
+  info,
+  onSignOut,
```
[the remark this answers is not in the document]
Saeris: Sure, I'll add that in a resolve conversations commit

packages/app/src/app/components/IntegrationModal/elements.ts
```diff
@@ -1,55 +1,54 @@
 import styled from 'styled-components';
 
 export const Container = styled.div`
-  background-color: ${props => props.theme.background};
   margin: 0;
+  background-color: ${props => props.theme.background};
```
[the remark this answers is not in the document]
Saeris: Whoops, guess I missed one. I'll fix this in a resolve conversations commit.

packages/app/src/app/components/Skeleton/index.tsx
```diff
@@ -1,23 +1,29 @@
...
-        {titles.slice(1).map((title, index) => (
-          <SubTitle key={String(index)} delay={title.delay}>
-            {title.content}
-          </SubTitle>
-        ))}
-      </Centered>
-    </Fullscreen>
-  );
+interface ISkeletonProps {
+  titles: [
```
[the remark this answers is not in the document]
Saeris: Sure, will fix in a resolve-conversations commit

Saeris:
@MichaelDeBoey Thanks for giving everything a thorough review! In the future I don't think it's necessary to comment on every instance of a proposed change. It just adds noise to the review. Suggesting a change once for all instances will suffice.

MichaelDeBoey:
@Saeris Yeah I just wanted to be sure everything was handled that's all 🙂

packages/app/src/app/components/Alert/index.tsx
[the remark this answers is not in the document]
MichaelDeBoey: I really think we should use `type` for the reasons explained in [this Medium](https://medium.com/@martin_hotell/interface-vs-type-alias-in-typescript-2-7-2a8f1777af4c) post tho.

It basically comes down to: "Use `type` for internal things, use `interface` if you're exporting them"

packages/app/src/app/components/DeploymentIntegration/index.tsx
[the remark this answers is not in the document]
MichaelDeBoey: Yeah sure it's correct what you did, just a bit shorter to use `ComponentType` and will always stay up to date too.
For example when Class components won't be supported anymore in the future.

packages/app/src/app/components/HoverMenu/index.tsx
```diff
@@ -13,7 +13,7 @@ const HoverMenu = ({ onClose, children }: IHoverMenuProps) => {
 
     document.addEventListener('click', handleDocumentClick);
 
-    return () => {
+    return () => document.removeEventListener('click', handleDocumentClick);
       document.removeEventListener('click', handleDocumentClick);
```
MichaelDeBoey: @Saeris This line can be deleted now

packages/app/src/app/components/Alert/index.tsx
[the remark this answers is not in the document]
Saeris: I read that too and I don't find that to be a compelling reason to use `type` over `interface`. At the time that was written in 2018 I believe his suggestion of using HOCs for type extensions was a good point, but now with hooks I believe we'll be using less HOCs overall. So when comparing the arguments of public versus private types against having code be consistent with official documentation examples, I think prefer `interface` wins out here as this is an open source project and we want it to be as approachable as possible for contributors. As stated by the author of the post, use whatever suits your team best.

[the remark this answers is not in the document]
MichaelDeBoey: The "use what the official docs do" argument doesn't really fit imho, since the article makes it clear that the official docs about `type` vs. `interface` (see [here](http://www.typescriptlang.org/docs/handbook/advanced-types.html#interfaces-vs-type-aliases)) is outdated too

@CompuIves @SaraVieira What're your opinions about this one?

packages/app/src/app/components/DeploymentIntegration/index.tsx
```diff
@@ -1,67 +1,59 @@
...
+  name: string;
+  beta: boolean;
+  open: boolean;
+  onToggle: (event: React.MouseEvent<HTMLDivElement>) => void;
+  onDeploy: (event: React.MouseEvent<HTMLButtonElement>) => void;
+  children: React.ReactChildren;
```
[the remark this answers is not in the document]
MichaelDeBoey: @Saeris Any reason you've changed this back to `ReactChildren`? 🤔

shiftyp:
Hi! Is this PR still a WIP? What's the best way to contribute?

Saeris:
@shiftyp Hey! Currently it's on pause as we haven't had time to properly review and test the changes. That majority of the work is already done, so I'm not sure this would be the best place to contribute.

However, you can use the guidelines given above and take a look at the code here and in #2003 for reference to get started on converting other parts of the app that are still written in JavaScript. I'd recommend picking components with a small footprint so that any PRs you open don't necessitate reviewing more than say a dozen files, just so that we can quickly review and test it. This PR is an example of what not to do, as it touches more than we've had bandwidth to work through right now. It is however a good example of the steps you should take: convert file extensions first, then refactor, test locally, submit for review, integrate requested changes, all as separate commits.

Thanks for reaching out and feel free to ping me in any PRs you open if you have any questions!

PR #2071, opened 2019-06-17
Use official babel-eslint dependency instead of own fork

MichaelDeBoey:
babel/babel-eslint#772 is fixed by babel/babel-eslint#773, so we can use the official `babel-eslint` dependency again instead of [our own fork](https://github.com/codesandbox/babel-eslint).

The fork can be deleted too now.


CompuIves:
Great! Thanks for the PR 😄

MichaelDeBoey:
With pleasure! 🙂

PR #2074, opened 2019-06-19
client test performance improvement, reduce screen freezing

batiskafff:
Bug fix.

Preconditions:
- 15 files in the project
- 50 tests in 8 different files
- Tests tab is open
- Watching for tests is **on**

Editor freeze, sometimes quite often, when you write code because it runs tests after each editing

## What is the new behavior?

Editor freeze less often (at least twice less)

Test methodology described here with few test output:
https://gist.github.com/batiskafff/925f07637485ffc88940d47b36872bd1

Brief description:
- use sandbox https://codesandbox.io/s/boring-dawn-fzvzd 
- include performance meter component in UI
- edit code

## Checklist

- [ ] Documentation N/A
- [x] Testing 
- [x] Ready to be merged 
- [x] Added me to contributors table 

It is proposal fix. It solves some issues, however, it is not the best possible solution for this issue.
The best possible solution from my opinion is added webworkers as test runner env, however, quite a big amount of dependency's and not simple implementation could unreasonable increase complexity of this part of the project.

### DIFF
--- a/packages/app/src/app/components/Preview/DevTools/Tests/index.tsx
+++ b/packages/app/src/app/components/Preview/DevTools/Tests/index.tsx
@@ -4,6 +4,7 @@ import { actions, dispatch, listen } from 'codesandbox-api';
 import SplitPane from 'react-split-pane';
 
 import immer from 'immer';
+import { debounce } from 'lodash-es';
 
 import { Container, TestDetails, TestContainer } from './elements';
 
@@ -68,6 +69,7 @@ type State = {
 
 type SandboxMessage = { type: 'test' | 'done' } & (
   | InitializedTestsMessage
+  | CompilationDoneMessage
   | TestCountMessage
   | TotalTestStartMessage
   | TotalTestEndMessage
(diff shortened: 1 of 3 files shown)


CompuIves:
This is veery interesting, I wonder if the performance hit is from the editor UI updating or from running the tests in the sandbox. If it's the latter, we're fixing that for Chrome by moving the url of the sandbox to a new domain. We're going to start using `csb.dev` instead of `codesandbox.io` for the preview soon, the way Chrome works is that it will run the iframe in a separate process if it's from another domain (https://www.chromium.org/developers/design-documents/oop-iframes), so if the biggest hit is from executing tests that would already fix it for Chrome.

We're not only building CodeSandbox for Chrome though, so this PR is greatly appreciated! Reviewing now.

CompuIves:
> The best possible solution from my opinion is added webworkers as test runner env

I agree with this. For this we could repurpose the bundler to generate a bundle that can be sent to the web worker to use.

packages/app/src/app/components/Preview/DevTools/Tests/index.tsx
```diff
@@ -201,7 +207,19 @@ class Tests extends React.Component<DevToolProps, State> {
 
   handleMessage = (data: SandboxMessage) => {
     if (data.type === 'done' && this.state.watching && !this.props.hidden) {
-      this.runAllTests();
+      let delay = 1000;
```
CompuIves: This is good, I agree with the change. It would be good to add a comment explaining why you did this for the next readers 😄.

batiskafff: done

CompuIves:
Great!! Thanks for this!

PR #2083, opened 2019-06-22
Add separate sandbox host

lbogdan:


packages/app/src/sandbox/index.js
```diff
@@ -18,6 +18,11 @@ const debug = _debug('cs:sandbox');
 export const SCRIPT_VERSION =
   document.currentScript && document.currentScript.src;
 
+const sandboxHost = {
```
CompuIves: Very clean!

packages/app/src/sandbox/index.js
```diff
@@ -32,9 +37,21 @@ function getId() {
     return document.location.host.match(re)[1];
   }
 
-  const hostRegex = host.replace(/https?:\/\//, '').replace(/\./g, '\\.');
-  const sandboxRegex = new RegExp(`(.*)\\.${hostRegex}`);
-  return document.location.host.match(sandboxRegex)[1];
+  let result;
+  [host, sandboxHost[host]].filter(Boolean).forEach(tryHost => {
```
MichaelDeBoey: @lbogdan I think you can make this a bit cleaner and remove the extra `let result` line by doing
- a map to the matches
- add an extra filter (if there are matches or not)
- reverse the mapped array
- Take the first one's second value

instead of the forEach an repopulating the result

```js
const [, result] = [host, sandboxHost[host]]
  .filter(Boolean)
  .map(tryHost => {
    const hostRegex = tryHost.replace(/https?:\/\//, '').replace(/\./g, '\\.');
    const sandboxRegex = new RegExp(`(.*)\\.${hostRegex}`);

    return document.location.host.match(sandboxRegex);
  })
  .filter(Boolean)
  .reverse()[0]
```

lbogdan: @MichaelDeBoey We're currently just testing... 🙂 

But TBH, seeing a few other of your suggestions, you sometimes tend to forget that readability is code's most important quality (after correctness, of course). Taking current suggestion as an example, I had to read the whole chain a few times while executing it in my head, and I'm still not sure it's equivalent to the initial code. Also, the need of explaining is usually a good indicator that it's probably over-complicated. Just my 2¢. 🙂

MichaelDeBoey: After reading @adamwathan's ["Refactoring to Collections"](https://adamwathan.me/refactoring-to-collections) I always try to use chained array methods.
In his book, Adam explains how you can refactor loops (`for`, `while`, `forEach`) to `map`, `filter`, `reduce`, ... and why that's in most cases a cleaner solution than using `if` statements inside a loop.

Once you're used to this method, you'll love it and see it's (mostly) a cleaner solution.

What I've done is just refactor your `forEach`, since having a loop (`forEach` in this case) where you (conditionally) populate a variable outside that loop is mostly a good indicator you can refactor it with other chainable array methods.

MichaelDeBoey: Using `const [, result] = ...` is the same as using `const result = ....reverse()[0][1]` if that wasn't directly clear. I just like using array destructuring a bit more in this case since `[0][1]` is uglier imho

lbogdan: I partly agree, but it's not about you or me here, it's about anyone that could stumble over the code and try to understand what it does at a quick glance. And more often than not "smart" code is less readable and comprehensible.

MichaelDeBoey: Yeah I agree you don't have to make it utterly complex and give things a good name to make the code understandible, readable and comprehensible

Maybe this is a bit more clear? 🤔
```js
const matches = [host, sandboxHost[host]]
  .filter(Boolean)
  .map(tryHost => {
    const hostRegex = tryHost.replace(/https?:\/\//, '').replace(/\./g, '\\.');
    const sandboxRegex = new RegExp(`(.*)\\.${hostRegex}`);

    return document.location.host.match(sandboxRegex);
  })
  .filter(Boolean)
const lastMatch = matches.reverse()[0]
const result = lastMatch[1]
```

PR #2084, opened 2019-06-22
codesandbox-deps: use rollup instead of parcel

lbogdan:


probablyup:
I think this broke things. After the latest SW update I'm getting this error:

browserfs.rollup.js:3932 Uncaught Error: WARNING: File system has desynchronized. Received following error: Error: ENOTSUP: Operation is not supported.
$
    at n (browserfs.rollup.js:3932)
    at browserfs.rollup.js:6631

https://codesandbox.io/embed/thirsty-river-nlgn5

lbogdan:
@probablyup This PR is supposed to be related to only the `tern` dep of CodeMirror, so I don't think it's related to your issue.

lbogdan:
Can reproduce it on an older PR build as well: https://pr2071.build.csb.dev/s/nlgn5 .

probablyup:
Hmm.. what do you suggest? I cleared all site data including the SW and it's still broken.

PR #2060, opened 2019-06-13
Executor Refactor

CompuIves:
Will give more info tomorrow!

PR #2105, opened 2019-06-27
DevTools / Terminal UI fixes

### LINKED ISSUE #2103: Double Scrollbars with Terminal Tabs and Full-Height Terminal (in Container)
# 🐛 bug report

## Description of the problem

In a CodeSandbox Container, the terminal receives two scrollbars when it's been expanded to the full height of a browser window.

It's kind of hard to see what I'm doing here, but I'm typing in the console while it appears that nothing is happening. You cannot see the input because it is scrolled off of the end of the screen 😞:

ezgif com-resize (1)

As soon as the user types something, the terminal is scrolled up, causing the typing to be obscured.

## Reproduction Steps

- open a CodeSandbox Container
- resize the terminal pane to the full height of the window (obscuring the browser preview)
- open a new tab in the terminal
- type in a lot of text so that a scrollbar appears
- notice that any new text entered is not visible on the screen
- to make the effect worse: resize the terminal pane again

Note: I have successfully reproduced this on a 13 inch Retina MacBook Pro on default resolution.

## Potentially Related

https://github.com/codesandbox/codesandbox-client/issues/1654 (and the pull request from that: https://github.com/codesandbox/codesandbox-client/pull/1740)

I would like to type in the terminal and have it show up.

### Your Environment

| Сodesandbox      | `PROD-1561557066-4a5dea6e4 ` |
| Browser          | Chrome 75.0.3770.100 |
| Operating System | macOS 10.14.5 |

lbogdan:
Should fix #2103, #1654 .

This mainly uses Flexbox to shrink / grow containers instead of manually managing widths / heights.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,21 +6,21 @@ version: 2.1
 
 aliases:
   - &restore_repo_cache
-    key: v18-repo-{{ .Environment.CIRCLE_SHA1 }}
+    key: v19-repo-{{ .Environment.CIRCLE_SHA1 }}
 
   - &save_repo_cache
-    key: v18-repo-{{ .Environment.CIRCLE_SHA1 }}
+    key: v19-repo-{{ .Environment.CIRCLE_SHA1 }}
     paths:
       - ~/codesandbox-client
 
   - &restore_deps_cache
     keys:
-      - v18-dependency-cache-{{ .Branch }}-{{ checksum "yarn.lock" }}
-      - v18-dependency-cache-{{ .Branch }}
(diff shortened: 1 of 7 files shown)


karlhorky:
@CompuIves Oh no, reverted? Did I miss testing something? 🙈

lbogdan:
There were some (collateral, unrelated) things breaking, like the Console tab. And as it turns out, @CompuIves also fixed those quirks in #2060 , which we'll try hard to release early next week!

karlhorky:
Ah ok cool! Great, no worries then :)

PR #2115, opened 2019-07-03
fix(RollingText): inherit width from parent

### LINKED ISSUE #2114: bug: Homepage Hero Animation broken
# 🐛 bug report

## Description of the problem
The animation in the `Title__Container` is broken for page width below 1280px.
This can be best seen at a page width of 1279px.
When a "slide" is fading out it's container width is "jumping". I find it a little bit hard to describe, so I took a video:
https://youtu.be/GIjVP2fI-Vg

Timestamps where you can see the bug:
0:04
0:10
0:16

RafaelKr:
Fixes #2114

The container for the old children should also inherit it's width from parent.
RollingText is used in the Title- and the Frameworks-component and they both seem to behave correctly now.

Jumping animation

See https://youtu.be/GIjVP2fI-Vg

Timestamps where you can see the bug:
0:04
0:10
0:16

## What is the new behavior?
Animation working as intended.

## Checklist

- [ ] Documentation
- [x] Testing 
- [x] Ready to be merged 


CompuIves:
Great PR! Thank you for the fix!

PR #2132, opened 2019-07-04
Show Template Banner and name for not owned sandboxes

SaraVieira:
Screenshot 2019-07-04 at 15 26 57

### DIFF
--- a/packages/app/src/app/pages/Sandbox/Editor/Header/SandboxName/SandboxName.tsx
+++ b/packages/app/src/app/pages/Sandbox/Editor/Header/SandboxName/SandboxName.tsx
@@ -118,12 +118,7 @@ export const SandboxName = observer(() => {
             />
           </Form>
         ) : (
-          <Name
-            css={`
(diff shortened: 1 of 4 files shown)


packages/app/src/app/pages/Sandbox/Editor/Header/SandboxName/SandboxName.tsx
```diff
@@ -117,7 +118,14 @@ export const SandboxName = observer(() => {
             />
           </Form>
         ) : (
-          <Name onClick={handleNameClick}>{sandboxName}</Name>
+          <Name
+            css={`
```
Saeris: Name is imported from elements.js next to this file, right? how aobut we pass a prop to change the cursor state instead of using inlined css here?

SaraVieira: done

CompuIves:
This is great! Really good PR, thank you!

PR #2137, opened 2019-07-06
Overmind refactor: Round 2

christianalfoni:
## STRATEGY UPDATE

So I have found a strategy that will save us a lot of time. This is the suggested approach:

- By adding `?overmind=true` Codesandbox will fire up with Overmind as the state store and logic. The components are completely unaware of what state management system is running

- To make this possible it was necessary to revert some of the "hooks" work. It is impossible to make Overmind work with the `useStore` and `useSignals` hooks. This revert is only temporary

Now we are able to move on running the actual UI to test if stuff works. Can run one browser tab with default Codesandbox and Cerebral devtools, and an other browser tab with `?overmind=true` and the Overmind devtools. We can even get help from our users by asking them to try the new version (which just adds ?overmind=true) to the url, localStorage or whatever.

When we are confident it works we can continue the work on the components. That means:

```ts
const SomeComponent = inject('store', 'signals')(observer(({ store, signals }) => {

})
```

becomes:

```ts
const SomeComponent = () => {
  const { state, actions } = useOvermind()
}
```

Everything typed and ready :)

Really sorry about reverting some work here, but it is a big win to not having to refactor components with the state management refactor :)

Here is a video where I explain some more: https://www.youtube.com/watch?v=LNKxAVhnPZo&feature=youtu.be

## Initial description

The **first round** of refactoring moved all the logic of Codesandbox over to Typescript and Overmind **actions** and **effects**. This was a blunt "moving of code" where nothing is tested yet, the code has not even been run. In **round 2** we have two goals:

### Goal 1
Clean up up the existing code to take advantage of best practices in Overmind. The current logic is conceptually split up by:

Screenshot 2019-07-06 at 14 27 51

The **signals** is where we expressed state mutations and referenced pieces of logic, called **actions**, that managed talking to **providers** and performing other logic. The **providers** was small low level wrappers allowing Codesandbox to "talk to the outside world". 

With **Overmind** we are moving to a concept looking more like this:

Screenshot 2019-07-06 at 14 27 56

First of all we are leaving the **declarative** world of signals in favour of the imperative straight forward **actions** of Overmind. This has excellent Typescript support and will hopefully lower the threshold of contributions as there is no functional concept to learn.

The **internal actions** are currently the old **actions** of Cerebral. In this refactor we want to move this logic to the front of **actions** or further back into **effects**. Basically we want to avoid having **internal actions**.

The **effects** are the old **providers** of Cerebral, but they are getting A LOT more responsibility and becoming domain specific. An example of this is the old **api** provider, which now moves from being an API surface of **get**, **post** etc. to **getSandbox**, **deleteSandbox** etc. This moves a lot of unnecessary complexity in the **actions** and also ensures that we do not have to insert types manually everywhere, reducing the developer experience. An example of this is the old **api** where you would have to`api.get<Sandbox>(...)`, where you now just do `api.getSandbox(...)`. We are also adding new effects to move even more of this code out of the **actions**. Effects like **zeit** and **netlify**. 

A good rule of thumb is that **actions** should manage state and grab data through **effects**. Any logic that does not directly relate to managing state should ideally live in  **effects**. 

The only reason to keep an **internal action** is where we have some shared logic that relates to managing state. **Actions** are able to call these internal actions to execute this shared logic.

This is one example of how this approach looks like:

```ts
export const deployClicked: AsyncAction = async ({
  state,
  effects,
  actions,
}) => {
  state.deployment.deploying = true;
  const zip = await effects.utils.getZip(state.editor.currentSandbox);
  const contents = await effects.jsZip.loadAsync(zip.file);

  try {
    state.deployment.url = await effects.zeit.deploy(
      contents,
      state.editor.currentSandbox
    );
  } catch (error) {
      'An unknown error occurred when connecting to Zite'
    );
  }

  state.deployment.deploying = false;

  actions.deployment.getDeploys();
};
```

As you can see, compared to the old code, we have put a lot of complexity into the effect. Also notice that we have only defined a single type here, `AsyncAction`, but everything is completely type safe. 

### Goal 2
Test the execution of all actions listed below and compare them to the execution of existing Cerebral signals. They have the same name. We do this by using the Overmind and Cerebral devtools. They tell us all the mutations and effects executed. We simply just have to compare.

**How to test actions coming soon**

### Root

**NOTE!** Unable to test notifications as it connects to `wss`

- [x] appUnmounted
- [x] searchMounted
- [x] sandboxPageMounted
- [x] cliMounted
- [x] cliInstructionsMounted
- [x] githubPageMounted
- [x] connectionChanged
- [x] modalOpened
- [x] modalClosed
- [ ] **signInClicked** - *Unable to test locally*
- [ ] **signInCliClicked** - *Unable to test locally*
- [x] userMenuOpened
- [x] userMenuClosed
- [ ] **notificationAdded** - *Need to clear up why we have two notification systems*
- [ ] **notificationRemoved** - *Need to clear up why we have two notification systems*
- [ ] **signInZeitClicked** - *Unable to test locally*
- [ ] **signOutZeitClicked** - *Unable to test locally*
- [x] authTokenRequested
- [x] requestAuthorisation
- [ ] **signInGithubClicked** - *Unable to test locally*
- [ ] **signOutClicked** - *Unable to test locally*
- [ ] **signOutGithubIntegration** - *Unable to test locally*
- [x] setUpdateStatus
- [ ] **refetchSandboxInfo** - *Is now resetting a lot of state that seems to be counter intuitive? Need to discuss this*
- [x] track

### dashboard

- [x] dashboardMounted
- [x] sandboxesSelected
- [x] setTrashSandboxes
- [x] dragChanged
- [x] orderByChanged
- [x] blacklistedTemplateAdded
- [x] blacklistedTemplateRemoved
- [x] blacklistedTemplatesCleared
- [x] blacklistedTemplatesChanged
- [x] searchChanged
- [ ] **createSandboxClicked** - *Can not figure out when this is possible?*

### deployment

- [ ] **deployWithNetlify** - *Unable to test locally*
- [ ] **getNetlifyDeploys** - *Unable to test locally*
- [ ] **getDeploys** - *Unable to test locally*
- [ ] **deployClicked** - *Unable to test locally*
- [ ] **deploySandboxClicked** - *Unable to test locally*
- [ ] **setDeploymentToDelete** - *Unable to test locally*
- [ ] **deleteDeployment** - *Unable to test locally*
- [ ] **aliasDeployment** - *Unable to test locally*

### editor

- [x] addNpmDependency - *Needs more testing, several states here*
- [x] npmDependencyRemoved - *Needs more testing, seems to be messing around with dirty and changed state, but should it?*
- [x] sandboxChanged - *Needs to be tested with different recover states etc.*
- [x] contentMounted
- [x] resizingStarted
- [x] resizingStopped
- [x] codeSaved - *Not sure about adding to changed modules and dirty, is that not only when changing code?*
- [x] codeChanged
- [ ] saveClicked
- [ ] createZipClicked
- [ ] forkSandboxClicked
- [ ] likeSandboxToggled
- [ ] moduleSelected
- [ ] clearModuleSelected
- [ ] moduleDoubleClicked
- [ ] tabClosed
- [ ] tabMoved
- [ ] prettifyClicked
- [ ] errorsCleared
- [ ] projectViewToggled
- [ ] previewActionReceived
- [ ] privacyUpdated
- [ ] frozenUpdated
- [ ] quickActionsOpened
- [ ] quickActionsClosed
- [ ] setPreviewContent
- [ ] togglePreviewContent
- [ ] currentTabChanged
- [ ] discardModuleChanges
- [ ] fetchEnvironmentVariables
- [ ] updateEnvironmentVariables
- [ ] deleteEnvironmentVariable
- [ ] toggleEditorPreviewLayout
- [ ] onNavigateAway

### explore

- [ ] popularSandboxesMounted
- [ ] pickSandbox
- [ ] pickSandboxModal
- [ ] getSandbox
- [ ] pickedSandboxesMounted

### files

- [ ] addedFileToSandbox
- [ ] gotUploadedFiles
- [ ] deletedUploadedFile
- [ ] filesUploaded
- [ ] moduleCreated
- [ ] moduleRenamed
- [ ] directoryCreated
- [ ] directoryRenamed
- [ ] moduleMovedToDirectory
- [ ] directoryMovedToDirectory
- [ ] directoryDeleted
- [ ] moduleDeleted
- [ ] createModulesByPath
- [ ] syncSandbox
- [ ] removeModule
- [ ] removeDirectory

### git

- [ ] repoTitleChanged
- [ ] createRepoClicked
- [ ] gitMounted
- [ ] createCommitClicked
- [ ] subjectChanged
- [ ] descriptionChanged
- [ ] createPrClicked

### live

- [ ] roomJoined
- [ ] createLiveClicked
- [ ] liveMessageReceived
- [ ] onTransformMade
- [ ] applyTransformation
- [ ] onCodeReceived
- [ ] onOperationApplied
- [ ] onSelectionChanged
- [ ] onSelectionDecorationsApplied
- [ ] onModeChanged
- [ ] onAddEditorClicked
- [ ] onRemoveEditorClicked
- [ ] onSessionCloseClicked
- [ ] onNavigateAway
- [ ] onToggleNotificationsHidden
- [ ] onSendChat
- [ ] onChatEnabledChange
- [ ] onFollow
- [ ] onModuleStateMismatch

### patron

- [ ] patronMounted
- [ ] priceChanged
- [ ] createSubscriptionClicked
- [ ] updateSubscriptionClicked
- [ ] cancelSubscriptionClicked
- [ ] tryAgainClicked

### preferences

- [ ] viewModeChanged
- [ ] devtoolsToggled
- [ ] setDevtoolsOpen
- [ ] itemIdChanged
- [ ] settingChanged
- [ ] setBadgeVisibility
- [ ] paymentDetailsRequested
- [ ] paymentDetailsUpdated
- [ ] keybindingChanged
- [ ] zenModeToggled
- [ ] codeMirrorForced

### profile

- [ ] profileMounted
- [ ] sandboxesPageChanged
- [ ] likedSandboxesPageChanged
- [ ] selectSandboxClicked
- [ ] newSandboxShowcaseSelected
- [ ] deleteSandboxClicked
- [ ] sandboxDeleted

### server

- [ ] statusChanged
- [ ] containerStatusChanged

### user-notifications

- [ ] notificationsOpened
- [ ] notificationsClosed
- [ ] messageReceived

### workspace

- [ ] valueChanged
- [ ] tagChanged
- [ ] tagAdded
- [ ] tagRemoved
- [ ] sandboxInfoUpdated
- [ ] externalResourceAdded
- [ ] externalResourceRemoved
- [ ] integrationsOpened
- [ ] sandboxDeleted
- [ ] sandboxPrivacyChanged
- [ ] setWorkspaceItem
- [ ] toggleCurrentWorkspaceItem
- [ ] setWorkspaceHidden


packages/app/src/app/overmind/effects/netlify.ts
```diff
@@ -0,0 +1,112 @@
...
+        url: `${NetlifyBaseURL}/${sandboxId}`,
+      });
+
+      return response.data;
+    },
+    async deploy(file: string, sandbox: Sandbox): Promise<string> {
+      const userId = _options.getUserId;
+      const template = getTemplate(sandbox.template);
+      const buildCommand = (name: string) => {
+        if (name === 'styleguidist') return 'styleguide:build';
```
MichaelDeBoey: @christianalfoni Put curly braces around this one to make it less error-prone

```suggestion
        if (name === 'styleguidist') {
          return 'styleguide:build';
        }
```

packages/app/src/app/overmind/effects/netlify.ts
```diff
@@ -0,0 +1,112 @@
...
+        if (name === 'nuxt') return 'generate';
```
MichaelDeBoey: @christianalfoni Put curly braces around this one to make it less error-prone

```suggestion
        if (name === 'nuxt') {
          return 'generate';
        }
```

packages/app/src/app/overmind/effects/zeit.ts
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  if (!nowJSON.files) {
+    apiData.files.push({
+      file: 'package.json',
+      data: JSON.stringify(packageJSON, null, 2),
+    });
+  }
+
+  for (let i = 0; i < filePaths.length; i += 1) {
```
MichaelDeBoey: @christianalfoni This  can just be a `forEach`

```js
filePaths.forEach(filePath => {
  // ...
});
```

packages/app/src/app/overmind/namespaces/preferences/internalActions.ts
```diff
@@ -0,0 +1,36 @@
...
+  { state },
+  sandboxOptions
+) => {
+  state.preferences.showPreview =
+    sandboxOptions.isPreviewScreen || sandboxOptions.isSplitScreen;
+
+  state.preferences.showEditor =
+    sandboxOptions.isEditorScreen || sandboxOptions.isSplitScreen;
+
+  if (sandboxOptions.initialPath)
```
MichaelDeBoey: @christianalfoni I think it's better to put curly braces around each `if`-block to make it less error-prone.

I know this will make this file a lot bigger, but (especially) for new(er) developers this will make the code a bit clearer.

packages/app/src/app/overmind/effects/api/apiFactory.ts
```diff
@@ -1,85 +1,59 @@
...
-    camelizedData.data.npmDependencies = response.data.data.npm_dependencies;
-  }
+export type Options = {
+  shouldCamelize: boolean;
+};
 
-  return camelizedData.data ? camelizedData.data : camelizedData;
-}
+export type Api = {
+  get<T>(path: string, params?: Params, options?: Options): Promise<T>;
```
MichaelDeBoey: @christianalfoni I think function types are more commonly declared (using the arrow) like

```ts
export type Api = {
  get: <T>(path: string, params?: Params, options?: Options) => Promise<T>;
}
```

packages/app/src/app/overmind/effects/api/apiFactory.ts
```diff
@@ -93,67 +67,110 @@ export default (() => {
...
+  if (!response || response.status >= 500) {
+    logError(error);
+  }
+
+  if (response && response.data) {
+    if (response.data.errors) {
+      const errors = values(response.data.errors)[0];
+      if (Array.isArray(errors)) {
+        if (errors[0]) {
+          error.message = errors[0]; // eslint-disable-line no-param-reassign
```
MichaelDeBoey: @christianalfoni Why don't you just return the new value instead of reassigning `error.message` and returning it at the end of the function?  🤔

packages/app/src/app/overmind/createConnect.ts
```diff
@@ -0,0 +1,248 @@
...
+import { IConfiguration, Overmind, EventType } from 'overmind';
+import {
+  StatelessComponent,
+  ComponentClass,
+  ClassicComponentClass,
+  Component,
+  createElement,
+} from 'react';
+
+export type IReactComponent<P = any> =
```
MichaelDeBoey: @christianalfoni You could just use `ComponentType` instead

packages/app/src/app/overmind/createConnect.ts
```diff
@@ -0,0 +1,248 @@
...
+  | StatelessComponent<P>
+  | ComponentClass<P>
+  | ClassicComponentClass<P>;
+
+// Diff / Omit taken from https://github.com/Microsoft/TypeScript/issues/12215#issuecomment-311923766
+type Omit<T, K extends keyof T> = Pick<
```
MichaelDeBoey: @christianalfoni `Omit` is included in the core of TypeScript since 3.5
https://devblogs.microsoft.com/typescript/announcing-typescript-3-5

packages/app/src/app/pages/Sandbox/Editor/Header/Buttons/PickButton/PickButton.tsx
```diff
@@ -1,33 +1,36 @@
...
-export const PickButton = observer(() => {
-  const {
-    explore: { pickSandboxModal },
-  } = useSignals();
-  const {
-    editor: {
-      currentSandbox: { id, title, description, owned },
-    },
-  } = useStore();
+export const PickButton = inject('store', 'signals')(
```
MichaelDeBoey: I don't thinks if we want this back actually. 🤔
The hooks are a real improvement I think.
Using the hooks will provide us with code completion and can give use typechecking too btw.

CC @CompuIves @SaraVieira @Saeris

christianalfoni: @MichaelDeBoey Check out this youtube clip: https://www.youtube.com/watch?v=LNKxAVhnPZo&feature=youtu.be :) 

We have to move back to this approach before we can move to hooks. The reason is that we need the components to work with both Cerebral and Overmind. After we get that running we can use the Overmind hook instead:

```ts
const { state, actions } = useOvermind()
```

It does not require any HOC, so no observer HOC or anything :)

But yeah... using `useStore` and `useSignals` is definitely an improvement, but it was a bit premature I realize now. Have to get over to Typescript and Overmind first :)

packages/app/src/app/overmind/actions.ts
```diff
@@ -103,21 +108,13 @@ export const signInZeitClicked: AsyncAction = async ({
...
-      });
-      await actions.internal.getZeitUserDetails();
+      state.user = await api.createZeitIntegration(data.code);
+      await actions.deployment.internal.getZeitUserDetails();
     } catch (error) {
-      notificationToast.add({
-        message: 'Could not authorize with ZEIT',
-        status: notificationToast.convertTypeToStatus('error'),
-      });
+      notificationToast.error('Could not authorize with ZEIT');
```
SaraVieira: So much cleaner 😍

PR #2143, opened 2019-07-08
Support .ts(x) eslint linting

CompuIves:
This demos that our linter doesn't work with TS files. What could be wrong with our `.eslintrc` configuration?


Saeris:
Appears to be working in Atom:

image

MichaelDeBoey:
WebStorm seems to work too.

Schermafbeelding 2019-07-08 om 16 41 28
Schermafbeelding 2019-07-08 om 16 42 40

lbogdan:
I think @CompuIves is referring to the [linting running in CI](https://circleci.com/gh/codesandbox/codesandbox-client/26560) . 🙂

MichaelDeBoey:
CI is bit bloated now with all of `overmind`'s type errors, but I indeed can't see the errors there at first look.

lbogdan:
The type errors are in a separate `typecheck` step, which fails, but the `lint` one succeeds.

MichaelDeBoey:
Ah my bad, I’m sorry... 😓

Locally it passes too.
I thought it could be because we copied over the config from CRA and they mostly throw warnings instead of errors, but can't see any warnings either...

lbogdan:
@CompuIves There, I fixed it! Happy now, with all the errors? 😃

packages/app/package.json
```diff
@@ -17,8 +17,8 @@
     "test:watch": "jest --watch --env=jsdom",
     "test:integrations": "jest --config integration-tests/jest.config.json --maxWorkers=2",
     "lint": "npm run lint:app && npm run lint:embed",
-    "lint:app": "cross-env LINT=1 eslint src/app",
-    "lint:embed": "cross-env LINT=1 eslint src/embed",
+    "lint:app": "cross-env LINT=1 eslint --ext .js,.ts,.tsx src/app",
```
MichaelDeBoey: It's weird that we explicitly have to set the extensions.
I would assume ESLint will automatically run on `.js` files and with the `@typescript-eslint/parser` that it would automatically run on `.ts(x)` files too, but happy if it works like this 🙂

lbogdan: Not that weird, from [command line docs](https://eslint.org/docs/user-guide/command-line-interface):
```
--ext [String]                 Specify JavaScript file extensions - default: .js
```

MichaelDeBoey: Maybe not that weird then indeed.

packages/app/src/app/pages/Dashboard/Content/routes/Templates/Navigation/Navigation.tsx
```diff
@@ -15,6 +15,7 @@ export const Navigation = ({ teamId, number }: INavigationProps) => (
     >
       {teamId ? 'Team Templates' : 'My Templates'}
     </NavigationLink>
-    {number != undefined && <Number>{number}</Number>}
+
+    {number && <Number>{number}</Number>}
```
CompuIves: Oh, small thing I found here @MichaelDeBoey. This check will cause `<Number>` not to show when value is `0`, because `0` is falsy.

CompuIves:
I also love that you updated VSCode settings, it now works in my editor too @MichaelDeBoey 	@lbogdan 	<3.

lbogdan:
Those were actually added by @Saeris . 🙂

CompuIves:
Ohhhh thank you @Saeris!!

CompuIves:
Merge this on Monday, when we're going to deploy executors.

CompuIves:
HAH! Merged already!

PR #2149, opened 2019-07-09
✨ Feature: Add CodeSadbox Crash Reporter Error Boundary

Saeris:
This PR adds a new re-usable Error Boundary component (modified from react-error-boundary by @bvaughn) with a default CodeSadbox fallback component that acts as an error page that provides an easy means for users to file a new crash report issue with a pre-filled issue template.

While users of CodeSandbox should never experience a crash, it's inevitably going to happen and the current behavior of a silent error and a blank screen isn't very helpful. So with this new addition, users will know that they've managed to break the app and as a bonus we're making it easy for them to tell us about it.

This will be the best new feature we hope our users never have to use.

## Demo

Here is a demo sandbox that exhibits the core functionality:

https://codesandbox.io/s/codesadbox-6vgjh

## Checklist
- [x] Create Error Boundary Component
- [x] Create CodeSadbox Fallback Component
- [x] Wrap Routes in Error Boundary
- [ ] ~Add Recent Actions to Crash Report~
- [x] Add Navigation Options to CodeSadbox
- [x] Remove Test Route
- [x] Documented
- [x] Tested
- [x] Ready to be merged


CompuIves:
@Saeris I fixed the mentioned issue with the modal. The modal component itself was inside the error boundary, so if the error boundary is rendered the modal component doesn't get rendered anymore (as it's replaced with the boundary).

The second thing with the not signed in state is because we manually call a signal for every page in CodeSandbox that initializes the state (including signing in), if you get the error boundary from another page (like the editor page) you should be signed in already. We don't have this initialization for the codesadbox page, which makes sense since it's only a debugging page right now.

packages/app/package.json
```diff
@@ -211,10 +212,10 @@
     "lru-cache": "^4.1.3",
     "match-sorter": "^1.8.1",
     "memoize-one": "^4.0.0",
-    "mobx": "^4.0.0",
-    "mobx-react": "^5.2.3",
-    "mobx-react-lite": "^1.3.2",
-    "mobx-state-tree": "^3.3.0",
+    "mobx": "^5.11.0",
```
CompuIves: Is there anything breaking in these major version upgrades? I'm quite scared if Cerebral likes this since it uses mobx extensively.

Saeris: https://github.com/mobxjs/mobx#mobx-4-vs-mobx-5

Doesn't appear that there are any breaking changes besides dropped IE11 support

https://github.com/mobxjs/mobx-react#choosing-your-version

mobx-react 6 adds hooks support by integrating mobx-react-lite into the main package. AFAIK this means there's no big difference between the two, and the long-term goal will be to replace mobx-react with mobx-react-lite entirely. Updating this dependency to latest is just to keep up to date with the latest bug fixes.

A quick look around the app doesn't seem to reveal anything broken. If there is, it's likely an extreme edge case. We can always roll back these deps if necessary.

Saeris:
Alright, I think we can call this one done! Going to remove the test route and get to writing up an announcement!

packages/app/package.json
MichaelDeBoey: @Saeris Dropping IE11 support is quite a major one, since we still support IE11. @lbogdan has done some extra work for IE11 in #2176 recently too.

Saeris: What he's doing with that PR is specific to the embed. AFAIK we don't actually support IE11 for the full app. @CompuIves can correct me if I'm wrong. We already had a discussion on this.

lbogdan:
Yeah, the plan is to only support IE11 in the sandbox (so `https://sandboxId.codesandbox.io/`), and, maybe, the embed. (and even so, I'm kinda struggling to make most of worker-loaders work)

Saeris:
Just waiting on a banner image and a final review then this is ready to be merged.

packages/homepage/content/articles/codesadbox-introducing-improved-crash-reporting/index.md
```diff
@@ -0,0 +1,69 @@
+---
+featuredImage: ./images/banner.png
+slug: CodeSadbox
+authors: ['Drake Costa']
+photo: 'https://avatars0.githubusercontent.com/u/3144549?s=460&v=4'
```
MichaelDeBoey:
```suggestion
photo: https://avatars0.githubusercontent.com/u/3144549?s=460&v=4
```

packages/app/src/app/pages/common/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,63 @@
...
+    const { children, FallbackComponent } = this.props;
+    const { error, info } = this.state;
+
+    return error !== null ? (
+      <FallbackComponent
+        error={error}
+        trace={info ? info.componentStack : ''}
+      />
+    ) : (
+      children || null
```
MichaelDeBoey: No need for the extra `|| null` here

packages/app/src/app/pages/common/ErrorBoundary/types.ts
```diff
@@ -0,0 +1,24 @@
...
+  componentStack: string;
+};
+
+export interface IFallbackComponentProps {
+  error?: Error;
+  trace?: string;
+}
+
+export interface IErrorBoundaryProps {
+  children?: React.ReactNode;
```
MichaelDeBoey: @Saeris I think it would be a bit less verbose when you just delete the `React.` part here

```suggestion
  children?: ReactNode;
```

packages/app/src/app/pages/common/ErrorBoundary/types.ts
```diff
@@ -0,0 +1,24 @@
...
+  FallbackComponent: React.ComponentType<IFallbackComponentProps>;
```
MichaelDeBoey: Same

```suggestion
  FallbackComponent: ComponentType<IFallbackComponentProps>;
```

packages/app/src/app/pages/index.tsx
```diff
@@ -54,35 +49,37 @@ const Curator = Loadable(() =>
...
-  render() {
-    return (
-      <Container>
-        <Route
-          path="/"
-          render={({ location }) => {
-            if (process.env.NODE_ENV === 'production') {
-              routeDebugger(
-                `Sending '${location.pathname + location.search}' to ga.`
+  useEffect(() => () => appUnmounted(), [appUnmounted]);
```
MichaelDeBoey: @Saeris This can be

```suggestion
  useEffect(() => {
    appUnmounted();
  }, [appUnmounted]);
```

MichaelDeBoey:
> Yeah, the plan is to only support IE11 in the sandbox (so https://sandbox-id.codesandbox.io)

So @Saeris' `mobx` update needs to be reverted since it's not IE11 compatible? 🤔

Saeris:
> > Yeah, the plan is to only support IE11 in the sandbox (so https://sandbox-id.codesandbox.io)
> 
> So @Saeris' `mobx` update needs to be reverted since it's not IE11 compatible? 🤔

No the two shouldn't affect each other.

packages/app/src/app/pages/common/ErrorBoundary/types.ts
Saeris: We're already using the long syntax elsewhere in the codebase and I'd prefer to stay consistent with other TypeScript examples

packages/app/src/app/pages/common/ErrorBoundary/ErrorBoundary.tsx
Saeris: Going to keep this as-is, since the fallback `null` was part of the original code I adapted this from. It may catch an edge case I'm not aware of.

CompuIves:
Hmm, only a parsing error in the md file 🤔

CompuIves:
Great!

PR #2152, opened 2019-07-09
Stackbit Import Support

CompuIves:
image


packages/common/src/components/Button/elements.ts
```diff
@@ -70,7 +70,8 @@ const getColor = ({
   return 'white';
 };
 
-const getHoverColor = ({ secondary }: OptionProps) => {
+const getHoverColor = ({ secondary, disabled }: OptionProps) => {
+  if (disabled) return '';
```
MichaelDeBoey: @CompuIves I think it's better to use curly braces around the returns

```suggestion
  if (disabled) {
    return '';
  }
```

packages/app/src/app/pages/Dashboard/Content/CreateNewSandbox/NewSandboxModal/Imports/StackbitImport.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { ImportHeader, ImportDescription, Section } from './elements';
+import { StackbitButton } from './StackbitButton';
+
+interface Props {
+  username: string;
+}
+
+export const StackbitImport = ({ username }: Props) => {
+  return (
```
MichaelDeBoey: @CompuIves This can be an implicit return

packages/app/src/app/pages/Dashboard/Content/CreateNewSandbox/NewSandboxModal/Imports/StackbitButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+  );
+}
+
+interface Props {
+  username: string;
+  style?: React.CSSProperties;
+}
+
+export const StackbitButton = ({ username, style }: Props) => {
+  return (
```
MichaelDeBoey: @CompuIves This can be an implicit return

packages/common/src/components/Button/elements.ts
CompuIves: I prefer to keep consistency here, I either change this here and all the other lines as well, or I do that in a separate PR. I prefer to hold it off this PR.

PR #2154, opened 2019-07-10
Fix ESLint errors/warnings on linting-issues-repro branch

MichaelDeBoey:
Fix ESLint errors/warnings on `linting-issues-repro` branch (#2143)

ESLint errors/warnings
```sh
✖ 119 problems (98 errors, 21 warnings)
```

## What is the new behavior?
Less ESLint errors/warnings
```sh
✖ 11 problems (0 errors, 11 warnings)
```

Run `yarn lint`

## Checklist
- [ ] Documentation N/A
- [x] Testing 
- [x] Ready to be merged 


lbogdan:
@MichaelDeBoey Can we limit to **only** eslint errors / warnings, please?

MichaelDeBoey:
@lbogdan Reverted the find/replace stuff

MichaelDeBoey:
@lbogdan I still have a lot of these errors, but  can't figure out how to fix them

```sh
codesandbox-client/packages/app/src/app/components/Preview/DevTools/Problems/index.tsx
  5:8  error  Unable to resolve path to module 'codesandbox-api/dist/types/actions/correction'  import/no-unresolved
  5:8  error  Missing file extension for "codesandbox-api/dist/types/actions/correction"        import/extensions
  9:8  error  Unable to resolve path to module 'codesandbox-api/dist/types/actions/error'       import/no-unresolved
  9:8  error  Missing file extension for "codesandbox-api/dist/types/actions/error"             import/extensions
```

packages/app/src/app/components/Preview/DevTools/Tabs/index.tsx
```diff
@@ -69,7 +69,7 @@ const DevToolTabs = ({
               devToolIndex={devToolIndex}
               moveTab={moveTab}
               index={i}
-              key={i}
+              key={pane.id}
```
CompuIves: This should be `i` (with an eslint ignore), since `pane.id` is not unique. Leads to small bugs.

packages/app/src/app/components/Preview/DevTools/Tests/TestElement/elements.ts
```diff
@@ -22,8 +22,7 @@ export const Actions = styled.div`
 `;
 
 export const TestName = styled.div`
-  padding: 0.25rem;
-  padding-left: 20px;
+  padding: 0.25rem 0.25rem 0.25rem 20px;
```
CompuIves: In general I'd like the explicit separate rule for `padding-left`. I don't mind this change, but I don't think future usages of `padding-left` etc need to be refactored.

packages/app/src/app/components/Preview/DevTools/Tests/TestElement/elements.ts
```diff
@@ -87,88 +86,84 @@ export const FileData = styled.div`
...
-    background-color: ${props =>
-      !props.theme.light ? 'rgba(0, 0, 0, 0.2)' : 'rgba(255, 255, 255, 0.2)'};
-    color: ${props =>
-      props.theme.light ? 'rgba(0, 0, 0, 1)' : 'rgba(255, 255, 255, 1)'};
-    border-left-color: ${props => props.theme.secondary.clearer(0.5)};
-
-    ${Actions} {
-      opacity: 1;
-    }
+  ${({ selected, theme }) => css`
```
CompuIves: Are these just style changes? I'm not sure whether I'm comfortable merging this since it can fundamentally change style behaviour and we're not working on this part of the codebase right now. I'd prefer to revert the style changes here.

packages/app/src/app/components/Preview/DevTools/Tests/index.tsx
```diff
@@ -25,24 +23,26 @@ export type Status = 'idle' | 'running' | 'pass' | 'fail';
...
+        }[]
+      | null;
+  }[];
 };
 
 export type Test = {
-  testName: Array<string>;
-  duration: number | undefined;
+  testName: string[];
+  duration?: number;
```
CompuIves: Let's keep this `duration: number | undefined`. Like I said before, it changes whether the key is set on the object or not (in this case it's always set, but can be `undefined`);

```js
const a = { test: undefined }; // `('test' in a) === true
const a = { }; // `('test' in a) === false
```

packages/app/src/app/components/Preview/DevTools/Tests/TestElement/elements.ts
MichaelDeBoey: These are fixes for having a double use of the `props` name as a variable.

packages/app/src/app/components/Preview/DevTools/Tests/index.tsx
```diff
@@ -25,24 +23,26 @@ export type Status = 'idle' | 'running' | 'pass' | 'fail';
...
     _originalFunctionName: string;
     _originalColumnNumber: number;
     _originalLineNumber: number;
-    _originalScriptCode: Array<{
-      lineNumber: number;
-      content: string;
-      highlight: boolean;
-    }> | null;
-  }>;
+    _originalScriptCode:
```
CompuIves: In this case I prefer using `Array`, so `Array` when using object literals, `[]` for types/interfaces.

packages/app/src/app/vscode/extensionHostWorker/services/searchService.ts
```diff
@@ -11,15 +11,6 @@ initializeGlobals();
 
 async function initialize() {
   await initializeBrowserFS({ syncSandbox: true, syncTypes: true });
-
-  self.addEventListener('message', e => {
```
CompuIves: Why is this removed?

MichaelDeBoey: The listener wasn't doing anything

```ts
self.addEventListener('message', e => {	
    if (e.data.$type === 'input-write') {	
      const { $type, $data } = e.data.$data;	
      if ($type === 'file-search') {	
        const { query } = $data;	
      }	
    }	
  });
```

packages/app/src/app/vscode/index.ts
```diff
@@ -74,9 +75,9 @@ class VSCodeManager {
         this.controller.getSignal('editor.toggleEditorPreviewLayout')();
       },
       keybindings: {
-        primary: KeyMod.Shift | KeyMod.Alt | KeyCode.KEY_9,
+        primary: KeyMod.Shift || KeyMod.Alt || KeyCode.KEY_9,
```
CompuIves: This has to be a `|`, keyboard shortcuts use binary shifting to determine the shortcut. If we use `||` it will break the shortcut. We should change our eslint rules for this.

packages/app/src/app/vscode/initializers.ts
```diff
@@ -106,7 +106,7 @@ export function setVimExtensionEnabled(vimEnabled: boolean) {
   } else {
     // Auto disable vim extension
     if (
-      localStorage.getItem('vs-global://extensionsIdentifiers/disabled') ==
+      localStorage.getItem('vs-global://extensionsIdentifiers/disabled') ===
```
CompuIves: This has to be `==`, to also account for `null` values.

packages/node-services/src/child_process.ts
```diff
@@ -34,15 +34,15 @@ class Stream extends EventEmitter {
     super();
   }
 
-  setEncoding(encoding: string) {}
+  setEncoding() {}
 
-  write(message: string, encoding: string) {
+  write(message: string) {
```
CompuIves: Are we required to remove these parameters? I think they have a documentation purpose. If linting requires it it's okay to remove them.

packages/node-services/src/module.ts
```diff
@@ -21,12 +21,12 @@ function updateChildren(
 export default class Module {
   id: string;
   exports: any;
-  parent: Module | undefined;
+  parent?: Module;
```
CompuIves:
```suggestion
  parent: Module | undefined;
```

packages/node-services/src/module.ts
```diff
@@ -84,7 +84,7 @@ export default class Module {
   }
 
   _compile(content: string, filename: string) {
-    var _self = this;
+    const _self = this;
```
CompuIves:
```suggestion
    var _self = this;
```

This is pretty important. This should not be a block variable, this is copied from the node code.

packages/app/src/app/store/providers/FSSync.ts
```diff
@@ -90,9 +90,9 @@ async function syncDependencyTypings(
           sendTypes();
         } catch (e) {
           if (process.env.NODE_ENV === 'development') {
+            // eslint-disable-next-line no-console
             console.warn('Trouble fetching types for ' + depName);
           }
-          return {};
```
CompuIves: We should return an empty object here.

packages/node-services/src/child_process.ts
MichaelDeBoey: They trigger the `no-undefined` rule

packages/app/src/app/vscode/initializers.ts
MichaelDeBoey: Will make it an
```ts
[null, undefined].includes( /* ... */)
```
instead so we don't have to disable the linting

CompuIves: I think the rule has an exception for `== null`, I'm not 100% sure though!

packages/app/src/app/vscode/extensionHostWorker/services/searchService.ts
CompuIves: That's true, I'd prefer to keep it there since I was planning on using this for the search implementation.

MichaelDeBoey: OK I'll revert it and just disable the following line then 👍
```ts
const { query } = $data;
```

packages/app/src/app/components/Preview/DevTools/Tests/index.tsx
MichaelDeBoey: I understand the difference when you test those 2, but if I put  `duration?: number`, WebStorm gives me `duration:  number | undefined`, so these 2 are the same I think.

packages/app/src/app/vscode/initializers.ts
MichaelDeBoey: But you're checking on `undefined` here so the rule errors out.

Doing both a check to `null` and `undefined` is the same as doing the
```ts
[null, undefined].includes(/* ... */)
```
Which is a bit cleaner than the double check I think, no? 🤔

MichaelDeBoey:
@CompuIves I resolved all discussions 🙂

MichaelDeBoey:
Always a pleasure 🙂

PR #2159, opened 2019-07-14
Rename unsafe React lifecycle methods

skovy:

Refactor. Rename unsafe lifecycle methods to use the `UNSAFE_` prefix.

- [Update on Async Rendering](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html): referenced here in the update post
- [React Component: Legacy Lifecycle Methods](https://reactjs.org/docs/react-component.html#legacy-lifecycle-methods): referenced here in the `React.Component` documentation

These methods are considered unsafe. I wanted to test changes made in https://github.com/reactjs/react-codemod/pull/228 to add support for TypeScript in the `react-codemod` package against a public repository.

If this is something you don't want feel free to close this PR. 

## What is the new behavior?

Should be identical _(but these methods will not be removed in React 17 and make it clearer these methods should be changed if refactoring)_

Relying on automated testing and relying on React's internal tests.

Locally, I ran (with the above branch `yarn link`-ed locally):

- `./node_modules/react-codemod/bin/react-codemod.js rename-unsafe-lifecycles packages --parser flow --extensions js`
- `./node_modules/react-codemod/bin/react-codemod.js rename-unsafe-lifecycles packages --parser tsx --extensions tsx`

## Checklist

- [x] Documentation 
- [x] Testing 
- [x] Ready to be merged 

### DIFF
--- a/packages/app/src/app/components/CodeEditor/FuzzySearch/index.js
+++ b/packages/app/src/app/components/CodeEditor/FuzzySearch/index.js
@@ -23,7 +23,7 @@ export default class FuzzySearch extends React.PureComponent {
   // This is a precached map of paths to module
   paths = {};
 
-  componentWillMount() {
+  UNSAFE_componentWillMount() {
     const { modules, directories } = this.props;
     const modulePathData = modules.map(m => {
       const path = getModulePath(modules, directories, m.id);
--- a/packages/app/src/app/components/CodeEditor/MonacoDiff/index.js
+++ b/packages/app/src/app/components/CodeEditor/MonacoDiff/index.js
@@ -30,7 +30,7 @@ export default class MonacoDiff extends React.Component<Props>
     this.settings = props.settings;
   }
 
-  componentWillUpdate(nextProps: Props) {
+  UNSAFE_componentWillUpdate(nextProps: Props) {
     if (
       this.props.originalCode !== nextProps.originalCode ||
       this.props.modifiedCode !== nextProps.modifiedCode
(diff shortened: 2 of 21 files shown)


MichaelDeBoey:
Hey @skovy!

Thanks for having interest in contributing to the repo and putting in your time in this PR!

However I think this change is unnecessary, since all these deprecated lifecycle methods will be gone once we’ve fully refactored to using hooks.

Also this won’t throw any warnings anymore and will eventually slow down the codebase.

Feel free to keep contributing to the repo tho! 👊

CompuIves:
I'm wondering, will this change become required in the next release of React?

skovy:
@CompuIves in the [Update on Async Rendering](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#gradual-migration-path) I linked to it sounds like these `UNSAFE_` methods will exist in 17 but the non-unsafe methods will be removed. However, if everything is refactored to hooks before 17 like @MichaelDeBoey mentioned then this change probably isn't necessary.

image

Other potential motivations for this change are to make it immediately obvious to readers that if they're touching one of these components it would be ideal to refactor the component to remove this unsafe lifecycle method. You could also consider adding something in CI to error if `componentWillMount`, `componentWillReceiveProps` or `componentWillUpdate` is used to put some friction between adding new usages of these methods (but still allow `UNSAFE_` as an escape hatch but makes it a lot more obvious to the author and reviewers that they're opting into the unsafe lifecycle methods and would ideally find a different approach).

Your call! I already ran this while testing that command so figured I'd open the PR. If you want to pass on this change, feel free to close the PR 👍

CompuIves:
Hmm, I'm not sure whether we will be 100% hooks soon. I have the mentality of "if it ain't broke, don't fix it", and thus I expect that we still might have some class components in a year or even two (Facebook has the same approach here). If this will make it possible for us to move to React 17 without being hooks 100% yet I would prefer to do this change first!

PR #2164, opened 2019-07-14
Show placeholder icons if not signed in or not owned sandbox

CompuIves:
People often don't know what you can do with CodeSandbox because they're not signed in or they don't own the sandbox. This change is an experiment that should increase the discoverability of these features by showing a "disabled" version of the button if necessary:

image

When clicking on it they see this:

image

image

image


lbogdan:
Nice!

If you go to https://pr2164.build.csb.dev/s/new and click on the "Deployment" icon, you get an "An unknown error occurred when connecting to ZEIT" error.

CompuIves:
Good catch! Fixed it by refactoring to hooks and making the check use the same condition.

MichaelDeBoey:
Nice one @CompuIves! 👊

Will interfere with #1969 (which is here already some time) tho, so maybe that one should be merged first? 🤔
That one can be rebased too tho, the choice is up to you 🙂

CompuIves:
Good point! Let's merge the other one first. Did you test all the deployment flows in the PR?

packages/app/src/app/pages/Sandbox/Editor/Navigation/Navigation.tsx
```diff
@@ -31,6 +34,44 @@ const IDS_TO_ICONS = {
...
+interface IconProps {
+  item: INavigationItem;
+  store: any;
+  isDisabled?: boolean;
+  setWorkspaceHidden: (params: { hidden: boolean }) => void;
+  setWorkspaceItem: (params: { item: string }) => void;
+}
+
+const IconComponent = observer(
+  ({ item, isDisabled, setWorkspaceHidden, setWorkspaceItem }: IconProps) => {
```
MichaelDeBoey:
```suggestion
  ({ item, isDisabled }: IconProps) => {
```

packages/app/src/app/pages/Sandbox/Editor/Navigation/Navigation.tsx
```diff
@@ -31,6 +34,44 @@ const IDS_TO_ICONS = {
...
+    const { id, name } = item;
+    const store = useStore();
+
```
MichaelDeBoey:
```suggestion
    const {
      workspace: { setWorkspaceHidden, setWorkspaceItem },
    } = useSignals();

```

packages/app/src/app/store/modules/workspace/items.ts
```diff
@@ -1,59 +1,79 @@
...
-};
-
-const SERVER = {
+const SERVER: INavigationItem = {
   id: 'server',
   name: 'Server Control Panel',
 };
 
-export default function getItems(store) {
+export function getDisabledItems(store: any): INavigationItem[] {
```
MichaelDeBoey: Maybe create a custom hook instead of passing the store? 🤔

packages/app/src/app/store/modules/workspace/items.ts
```diff
@@ -1,59 +1,79 @@
...
+  const { currentSandbox } = store.editor;
+
+  if (!currentSandbox.owned || !store.isLoggedIn) {
+    return [GITHUB, DEPLOYMENT, LIVE];
+  }
+
+  return [];
+}
+
+export default function getItems(store: any): INavigationItem[] {
```
MichaelDeBoey: Maybe create a custom hook instead of passing the store? 🤔

packages/app/src/app/pages/Sandbox/Editor/Navigation/Navigation.tsx
```diff
@@ -44,34 +85,33 @@ export const Navigation = observer(
...
-                  <Icon />
-                </IconContainer>
-              </Tooltip>
-            );
-          })}
+        {shownItems.map(item => (
+          <IconComponent
+            key={item.id}
+            item={item}
+            store={store}
```
MichaelDeBoey: `store`, `setWorkspaceHidden` & `setWorkspaceItem` can be deleted here (see above).

packages/app/src/app/store/modules/workspace/items.ts
CompuIves: This one is used in multiple places, to make it reusable outside of React I think the store given is a nice solution.

packages/app/src/app/pages/Sandbox/Editor/Navigation/Navigation.tsx
CompuIves: Thank you! Much cleaner!

PR #2174, opened 2019-07-16
Use script tag loading for react

CompuIves:
Very hacky test to see if this fixes https://github.com/codesandbox/codesandbox-client/issues/667

### DIFF
--- a/packages/app/src/sandbox/eval/transpiled-module.ts
+++ b/packages/app/src/sandbox/eval/transpiled-module.ts
@@ -972,7 +972,8 @@ export default class TranspiledModule {
       globals.__dirname = pathUtils.dirname(this.module.path);
       globals.__filename = this.module.path;
 
-      const exports = (this.module.path.indexOf('/node_modules/react/') === 0
+      const exports = (this.module.path.indexOf('/node_modules/react/') === 0 ||
+        this.module.path.indexOf('/node_modules/react-dom/') === 0
         ? evaluateScript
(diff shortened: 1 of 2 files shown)


packages/app/src/sandbox/eval/transpiled-module.ts
```diff
@@ -971,7 +972,10 @@ export default class TranspiledModule {
       globals.__dirname = pathUtils.dirname(this.module.path);
       globals.__filename = this.module.path;
 
-      const exports = evaluate(
+      const exports = (this.module.path.indexOf('/node_modules/react/') === 0 ||
```
MichaelDeBoey: @CompuIves A regex can be used instead, no? 🤔

```ts
const exports = this.module.path.match(/\/node_modules\/react(-dom)?\//)[0]
  ? evaluateScript
  : evaluate)(
  // ...
);
```

CompuIves: Yeah this is super hacky, I don't even want to do that check there. It was purely for testing because it doesn't exist in dev.

In the final version this will probably be a different check, I'm not sure if I would opt for regex since this is a reaaaally hot code path, I'm not sure if I would do a check here 😂.

MichaelDeBoey: No problem, just found it a bit verbose like this so wanted to help you out 🙂

PR #2203, opened 2019-07-23
Stencil Compiler POC

CompuIves:
This uses the Stencil transpiler (https://github.com/ionic-team/stencil/tree/master/src/compiler/browser#stencil-compiler) to support stencil components. Note that right now I just override the compiler of every environment, just as a test. In the final version we want to create a new environment/template for this.

When we make a environment for this we can also define that changing a file requires a refresh (same for html and js).

Example: https://pr2203.build.csb.dev/s/stencil-compiler-poc-v2qbw


SaraVieira:
Merged because the source code is working, when we want we can now add a stencil template that will read from this compiler

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/2203/commits/17d2e8c76d39721d6262b645ca8ca57ac266ae82 is at https://pr2203.build.csb.dev/s/new.

PR #2204, opened 2019-07-24
Update setup instructions

jyash97:
Docs

`yarn start:home` gave this error `Cannot find module '@codesandbox/common/lib/config/env'`

## What is the new behavior?
Have updated the setup instructions to include this command

After running `yarn build:deps` I was able to run the homepage locally.

## Checklist

- [x] Documentation
- [ ] Testing 
- [x] Ready to be merged 


CompuIves:
Hmm, in this case `yarn start` automatically runs `yarn build:deps`. Maybe it would be better to add `yarn build:deps` as an extra script to `yarn start:home` instead.

jyash97:
Right this happens only when we do `start:home` I can revert this change and add the following scripts.

jyash97:
If we add this script in `start:home` than it would build this dependencies everytime we run `yarn start:home` and thus slowing down the process. Thoughts @CompuIves  ?

lbogdan:
Maybe we can do the same as with `start` and `start:fast` - have `start:home` and `start:home:fast`?

lbogdan:
Hey @jyash97 ,

Can you please revert all those formatting changes in `CONTRIBUTING.md`? Thanks!

jyash97:
Hey @lbogdan  You can check the #2215 . I messed up with the formatting here. Thanks

PR #2206, opened 2019-07-24
fix: add max 50 for number input in Patron Page

jyash97:
Bug Fix

Currently when we press Arrow up price gets over 50 but range slider is limited to 50 only.

## What is the new behavior?
After setting `max={50}` it limits the input to 50 when increasing from Arrow up Key.

Tested it locally on `localhost:3000/patron`

## Checklist

- [ ] Documentation
- [x] Testing 
- [x] Ready to be merged 

### DIFF
--- a/packages/app/src/app/pages/Patron/PricingModal/PricingChoice/index.tsx
+++ b/packages/app/src/app/pages/Patron/PricingModal/PricingChoice/index.tsx
@@ -52,6 +52,7 @@ const PricingChoice = observer<Props>(({ badge }) => {
             onChange={e => priceChanged({ price: Number(e.target.value) })}
             value={patron.price}
             min={5}
+            max={50}
             type="number"
           />
           <Month>/month</Month>


CompuIves:
Thanks for the contribution! It's a good catch.

In our case this bug is actually a feature ™️! We have had one patron who wanted to donate more than 50$, so they were able to do this by using the arrow keys. If we would disable the arrow key functionality as well, we would disable paying more than 50$ month altogether. That's why I prefer to keep this the way it is.

jyash97:
Right! This won't allow increasing more than 50$ using Arrow Up key but I think can be done by manually entering the amount. 

If this is the expected behavior we can close this PR. 
Thanks :)

MichaelDeBoey:
I think the whole Patron section needs to be rethought eventually so people can pay whatever they want (even if they want to pay $100+/month) in an easy way.

jyash97:
True, I think having a slider limited to $50 creates a confusion.

christianalfoni:
Thanks for bringing this to our attention @jyash97! I have added this to our user experience improvement docs and will close the PR for now. Thanks again :)

PR #2215, opened 2019-07-25
fix: add build scripts for homepage

jyash97:
Fix #2204


CompuIves:
This is great fix! Thanks a lot @jyash97!

PR #2225, opened 2019-07-27
✨ Follow Templates

Saeris:
This update adds a new MultiAction Button to be used for Following Templates, a prototype can be found here:

https://codesandbox.io/s/codesadbox-6vgjh

Additionally, this includes some refactoring and additions to common components. Button has been updated with a new Link component. Link can be used to both create an internal or external Link using a single API, intelligently using either a React Router link or an anchor tag with the appropriate attributes.

The MultiAction Button extends Button, sharing it's styles and base props, and uses Reakit under the hood to provide full keyboard accessibility.

Theme has been reorganized and extended with a set of fonts taken from the design system files in Figma. Additionally a set of breakpoint utilities was added, based on those found in the React documentation site.

A new `withoutProps` has been added that can be used to exclude props from being passed down to a component. This is useful for styled components, which sometimes pass down non-HTML attributes which cause console errors. Example usage can be observed in the MultiAction button elements.ts.

An error was fixed in Dependencies/VersionEntry for incorrect use of selected on an option element.

NotOwnedSandboxInfo now has a Follow Template button.

TODO:
- [ ]: Add interactivity to the Follow Template button
- [ ]: Test functionality
- [ ]: Incorporate requested changes


packages/common/src/components/Button/elements.ts
```diff
@@ -1,164 +1,177 @@
...
+  width: ${block ? '100%' : 'inherit'};
+  padding: ${small ? `2px 6px` : `0.65rem 2.25rem`};
+  border: ${getBorder};
   border-radius: 4px;
-
+  ${getBackgroundColor};
   box-sizing: border-box;
-  font-size: 1.125em;
+  color: ${getColor};
+  ${small ? theme.fonts.secondary.normal : theme.fonts.secondary.medium};
```
CompuIves: The buttons styles have changed with this update and some paddings are uneven now (especially in notifications). Let's change the button styles back to how they were before before we're going to implement the design system.

packages/app/src/app/components/CreateNewSandbox/NewSandboxModal/Explore/Explore.tsx
```diff
@@ -35,20 +36,23 @@ export const Explore = () => {
...
+        .search({
+          facetFilters: ['custom_template.published: true', category],
+          hitsPerPage: 50,
+          query,
+          page,
+        })
+        .then(rsp => {
+          setAllPages(rsp.nbPages);
+          const newTemplates = makeTemplates(rsp.hits);
+          if (page === 0) return setTemplates(newTemplates);
```
MichaelDeBoey: Put curly braces around body, to make this less  error-prone

```suggestion
          if (page === 0) {
            return setTemplates(newTemplates);
          }
```

packages/app/src/app/components/CreateNewSandbox/NewSandboxModal/ScrollableContent/index.tsx
```diff
@@ -9,33 +9,53 @@ import {
...
-        onBottomReached();
-      }
-      setScrolled(e.currentTarget.scrollTop > 0);
-    });
-  }, [onBottomReached, scrollableElRef]);
+    if (scrollbar && keepCalling) {
+      scrollbar.addListener(status => {
+        setScrolled(status.offset.y > 10);
+        if (status.offset.y === status.limit.y) {
+          if (lastCalled !== status.offset.y) {
```
MichaelDeBoey: These 2 conditions can be merged into 1 `if`-statement.

```ts
if(status.offset.y === status.limit.y && lastCalled !== status.offset.y) {
  // ...
}
```

packages/app/src/app/components/CreateNewSandbox/NewSandboxModal/elements.ts
```diff
@@ -24,6 +24,10 @@ export const Container = styled.div<{
   border: 1px solid #242424;
   box-shadow: 0px 8px 8px rgba(0, 0, 0, 0.24), 0px 2px 4px rgba(0, 0, 0, 0.12);
   font-family: Inter;
+
+  * {
+    box-sizing: border-box;
```
MichaelDeBoey: @tailwindcss is using the following their preflight (which extends Suit CSS):

```css
html {
  box-sizing: border-box; /* 1 */
}

*,
*::before,
*::after {
  box-sizing: inherit;
}
```

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/2225/commits/fe79e9610f7a541521f4d04c7227c9e654532cf0 failed.

packages/app/src/app/vscode/initializers.ts
```diff
@@ -3,9 +3,12 @@ import { vscode } from 'app/vscode';
 export function initializeThemeCache() {
   try {
     if (!localStorage.getItem('vs-global://colorThemeData')) {
+      localStorage.setItem('newUser', 'true');
```
CompuIves: I think we shouldn't put the `newUser` code in the `vscode` initializers, we don't use this code in some cases and new user is unrelated to whether VSCode initialized or not. Let's find a different property that we can use for this, I think there are easier ways to check this, by eg. checking the `signed_in` cookie. cc @SaraVieira

PR #2252, opened 2019-08-05
React DevTools

### LINKED ISSUE #2226: 💥 Crash Report: <Short Description of Crash Circumstances>
<h1>💥 Crash Report</h1>

<h2>What were you trying to accomplish when the crash occurred?</h2>

<h3>Link to sandbox: [link]() (optional)</h3>

<h3>Crash Details</h3>

<details>
<summary>Environment</summary>

| Browser |  Version  | Operating System |
| ------- | --------- | ---------------- |
| chrome | 75.0.3770 | OS X 10.14.5           |

</details>

<details>
<summary>Error Message</summary>

```bash
TypeError: Cannot read property 'editorService' of undefined

    in n
    in div
    in t
    in elements__Container
    in n
    in div
    in n
    in div
    in div
    in t
    in div
    in t
    in div
    in t
    in n
    in r
    in inject-with-signals-store(n)
    in div
    in t
    in div
    in t
    in div
    in div
    in t
    in styled.div
    in div
    in t
    in elements__Container
    in t
    in n
    in r
    in inject-with-signals-store(n)
    in n
    in r
    in inject-with-signals-store(n)
    in t
    in t
    in div
    in t
    in elements__Content
    in n
    in t
    in withRouter(n)
    in div
    in t
    in elements__Container
    in Unknown
    in t
    in withRouter()
    in DragDropContext(withRouter())
    in t
    in t
    in u
    in t
    in t
```

</details>

SaraVieira:
Fixes #2226


packages/app/src/app/components/Preview/DevTools/React-Devtools/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+  return (
+    <Container>
+      {devToolsReady && ReactDevTools.current && <ReactDevTools.current />}
+    </Container>
+  );
+};
+
+export default {
+  id: 'codesandbox.devtools',
+  title: 'React DevTools',
```
bvaughn: Btw instead of "React DevTools" I wonder if we should show "⚛️ Components" (to better align with the browser extension)? I wonder if that would be more or less confusing? :smile:

packages/app/src/app/components/Preview/DevTools/React-Devtools/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+    ReactDevTools.current = initialize(contentWindow);
+    iframe.onload = () => {
+      contentWindow.postMessage(
+        {
+          type: 'activate',
+        },
+        '*'
+      );
+    };
+    setDtReady(true);
```
bvaughn: You could just store the `DevTools` component directly in state and you'd only need one hook 😁 

```js
const [DevTools, setDevTools] = useState(null);

// in the effect
setDevTools(initialize(contentWindow));
```

bvaughn:
By the way, I'm seeing a bug when you select an element for the first time and DevTools tries to suspend to read its props- where the right hand panel isn't updating. I noticed this when working on my lightweight Code Sandbox app as well.

There's a new API we are adding to React that sits between the legacy render (`ReactDOM.render`) and the upcoming concurrent mode (`ReactDOM.unstable_createRoot`) that's a batched bust mostly synchronous API. That's the one I used in my Code Sandbox mimic app:
https://github.com/bvaughn/faux-codesandbox-client/blob/4a6da405cab3fc0fbc082a86e74be664042285f0/frontend/src/index.js#L8-L9

Note that I'm also using the latest canary (what will soon likely be 16.9.0) to pull in bugfixes and stuff. This helps (but may not entirely eliminate) the issue above. Going to dig into it more on my end but I would suggest trying to replace your current legacy render with the technique mentioned above:
https://github.com/codesandbox/codesandbox-client/blob/4f295f72cef58f02f3e32499a6df1fda5f67b7ba/packages/app/src/app/index.js#L136

And see if that helps? :smile:

packages/app/src/app/index.js
```diff
@@ -133,7 +133,8 @@ function boot() {
 
     try {
       const { signals, store } = controller.provide();
-      render(
+      const root = createSyncRoot(rootEl);
+      root.render(
```
bvaughn: You forgot to remove `rootEl` as the second param to `root.render()` 😀

packages/app/src/app/components/Preview/DevTools/React-Devtools/index.tsx
bvaughn: You can remove the `devToolsReady` state entirely now 😁 If `ReactDevTools` isn't `null`, you know you're ready to roll.

packages/app/src/app/index.js
CompuIves: Ohhh boy!

SaraVieira: I reverted this because the function does not exist, I think react dom needs to be upgraded

packages/app/src/app/components/Preview/DevTools/React-Devtools/index.tsx
SaraVieira: YES! All done and updated!

Thanks!

packages/app/src/app/components/Preview/DevTools/React-Devtools/index.tsx
SaraVieira: Naa, that makes sense!

Done :)

packages/app/src/app/index.js
bvaughn: That makes sense. This change (render to sync root) is really important though, so maybe we can bump to the 16.9 rc?

CompuIves: Hmm, I did bump the versions, let me try again...

packages/app/src/app/components/Preview/DevTools/React-Devtools/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+const DevTools = (props: DevToolProps) => {
+  const [ReactDevTools, setDevTools] = useState(null);
+  const theme = useContext(ThemeContext);
+
+  useEffect(() => {
+    const iframe = document.getElementById(
+      'sandbox-preview'
+    ) as HTMLIFrameElement;
+    const contentWindow = iframe.contentWindow;
+    // in the effect
```
bvaughn: 😆 Very meta

packages/app/src/sandbox/index.js
```diff
@@ -65,6 +68,15 @@ requirePolyfills().then(() => {
       // Means we're in the editor
       setupHistoryListeners();
       setupConsole();
+      window.addEventListener('message', ({ data }) => {
+        switch (data.type) {
+          case 'activate':
```
bvaughn: I just picked "activate" off the top of my head by the way. I'm not very familiar with the CS client so I don't know if it's guaranteed to be unique or what else might be posting messages to this `iframe`.

bvaughn:
By the way, I plan to do one more NPM publish fo the inline package this morning that fixes the sometimes blank panel issue.

packages/app/src/sandbox/index.js
CompuIves: Yap! This is fine! We can otherwise do a check on `data.codesandbox == null`.

packages/app/src/app/components/Preview/DevTools/React-Devtools/index.tsx
CompuIves: Whoooops hahaha

bvaughn:
Regarding the "open component source file" feature- the `DevTools` UI component accepts a prop `viewElementSourceFunction` which should be a function:
```js
function viewElementSourceFunction(id, data) {
  const { source } = data;
  // source.fileName, source.lineNumber
}
```

bvaughn:
Bump to `react-devtools-inline@4.0.0-alpha.8` for the CSS/line-height bugfix I mentioned above.

bvaughn:
Exciting!

PR #2266, opened 2019-08-09
Add stacktrace to error report from codesadbox

CompuIves:


Saeris:
Looks good. I'm curious whether the stack traces here will be any useful, because as far as React errors go, I think the stack trace would show the framework internals not the actual point in our code where the error occurred. On top of that, won't it still be mangled and thus very difficult to read?

CompuIves:
Some function names are not mangled, so in those cases I think it would be nice to deduct which file it's from 😄.

PR #2274, opened 2019-08-10
Fix duplicate theme-color meta tag

vxna:

Removing duplicate `theme-color` meta tag from head.

When running CodeSandbox as desktop pwa, [theme-color](https://github.com/codesandbox/codesandbox-client/blob/master/packages/app/src/app/index.html#L22) meta tag from page head takes precedence over [manifest.json](https://github.com/codesandbox/codesandbox-client/blob/master/packages/app/public/manifest.json#L56) value. In this case, `theme-color` is being used to style the app window [title bar](https://developers.google.com/web/progressive-web-apps/desktop#app-window) and that vivid blue color introduces huge visual inconsistence.

## What is the new behavior?

According to Google web app manifest [article](https://developers.google.com/web/updates/2015/08/using-manifest-to-set-sitewide-theme-color),
> If you page already has a theme-color meta tag — for example `<meta name="theme-color" content="#2196F3">` — then the page level configuration will be used instead of the value in the manifest.

Visual inspection, no breaking changes were introduced.

## Checklist

- [ ] Documentation
- [x] Testing 
- [x] Ready to be merged 

### DIFF
--- a/packages/app/src/app/index.html
+++ b/packages/app/src/app/index.html
@@ -19,7 +19,6 @@
 	<meta property="twitter:image:src" content="https://codesandbox.io/static/img/banner.png">
 	<meta property="twitter:image:width" content="1200">
 	<meta property="twitter:image:height" content="630">
-	<meta name="theme-color" content="#40A9F3">
 	<meta name="description" content="CodeSandbox is an online editor tailored for web applications.">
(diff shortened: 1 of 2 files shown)


CompuIves:
Yup, this makes a lot of sense. Thanks for the PR!

PR #2287, opened 2019-08-14
[Build] Fix SourceMaps for prod builds

CompuIves:
The main goal is to make source maps work again, which will improve our error reporting. It also adds some changes to our build system to improve build output, since I noticed that there is a big performance regression by adding sourcemaps.

This adds:
- New pipeline for css/scss (new version of autoprefixer with current list of browsers, removal of custom/old prefixers, mark css as side effects to not drop css)
- New settings for the (upgraded) minifier, bit more aggressive and we have sourcemaps. Also set parallel option to 2. For some weird reason I see that that's faster than `require('os').cpus().length - 1` for `thread-loader` so also applied it here.
- Caching for our babel output between production builds. Will hopefully improve the build times.
- Enable `futureEmitAssets`, which should make builds more memory-efficient
- Add `pathinfo` to dev builds, will add comments to require statements so we know where the file is
- Upgraded babel dependencies to also get new browsers-list
- Removed unused babel dependencies to lower our dependency complexity

Now let's see how fast (or slow) (subsequent) builds will be!


lbogdan:
@CompuIves That's quite a big change, can you please add a (short) description of the logic behind?

CompuIves:
Yap, on it!

CompuIves:
Added a description.

PR #2290, opened 2019-08-15
Add .nvmrc

baremetalfreak:

Adds `.nvmrc` config file for nvm / fnm to indicate what node version is required and help avoid frustration from a failing build

one has to read the docs and remember to set node to v10

## What is the new behavior?

The presence of `.nvmrc` indicates that a particular node version is required. Running `nvm use` or `fnm use` sets it properly 

- [x] Documentation
- [ ] Testing 
- [x] Ready to be merged 


baremetalfreak:
Because of the commit hook, the CONTRIBUTING.md gets reformatted. I split my change and the reformatting into separate commits to make it easier to review what was actually changed.

MichaelDeBoey:
Can’t we just set the `engine` field in `package.json` instead? 🤔

lbogdan:
I also think @MichaelDeBoey's suggestion is a bit better, considering that not everybody uses `nvm` / is used to look at `.nvmrc`, whereas when setting the `engine` in `package.json` running `yarn` will hard fail with a descriptive error message. That doesn't of course mean we can't have both. 🙂

baremetalfreak:
That would be cool. As far as I know `nvm` and `fnm` do not support `package.json`. I think the reason is that `nvm` was written as a bash script and the author did not want to parse json. And `fnm` is intended to be plug compatible replacement of `nvm`.

I think it would be a good idea for an improvement in fnm. Since it is written in reason, I do not see why it would not parse package.json

baremetalfreak:
@lbogdan that is what I have in some of my projects + a test that verifies they are in sync

baremetalfreak:
Someone already asked about `fnm` support of `engines` here https://github.com/Schniz/fnm/issues/39 but they did not implement it because it can be ambiguous

baremetalfreak:
@lbogdan @MichaelDeBoey I will add the `engines` part. Do you want me to keep `.nvmrc` and add a test to verify the two are in sync or just the `engines` part?

CompuIves:
I think we should do both, set the `package.json#engines` and `.nvmrc`. Best experience on both accounts (automatic switching and fail on install). I also saw that there were changes in `CONTRIBUTING.md`, though I think those were accidental. Could you revert the changes for that file?

baremetalfreak:
@CompuIves I added myself to the contributor list according to your guidelines. But I can do that in a separate PR later.

baremetalfreak:
Oops, mixed two things together. The changes in `CONTRIBUTING.md` added a note about using `nvm` or `fnm` https://github.com/codesandbox/codesandbox-client/pull/2290/commits/8c8d4e8d2282df0e79c1db47e0899455af35e951#diff-6a3371457528722a734f3c51d9238c13R44 do you want me to remove that?

CompuIves:
Oh no that makes sense! The CONTRIBUTING.md file doesn't normally change when you add yourself to the table, just README.md and all-contributors.json.

baremetalfreak:
@CompuIves you are right, I just noticed see my comment above https://github.com/codesandbox/codesandbox-client/pull/2290#issuecomment-521583813

CompuIves:
Great! Let me know if you want to add the test, I'm already happy with these changes but having a test would be super nice to have.

baremetalfreak:
Not sure where to put the test since this is a monorepo. If you tell me the right place, I'll be happy to add it.

lbogdan:
https://github.com/codesandbox/codesandbox-client/blob/64e62e6f7f6e4c89d5cfb3cb8f5bbf0c24cefd17/package.json#L98
We should probably not make it **that** restrictive (the PR build is currently failing, because that uses `10.15`).

baremetalfreak:
@lbogdan `v10.15.0` ?

lbogdan:
Frankly, I'd probably test if it works on `10.0`, and use that instead. I don't know how many people keep their node version up to date (I usually don't 🙂 ), and it's annoying to have to upgrade just to be able to run a project you want to contribute to.

CompuIves:
> If you tell me the right place,

Haha, I don't know either! Let's just merge this change in already, I think this change will be really good for new contributors.

baremetalfreak:
Changed to 10.0..

CompuIves:
This is perfect, thanks a lot for this change @baremetalfreak!!

PR #2295, opened 2019-08-16
Design system yay

SaraVieira:
http://csb-storyboook.surge.sh/?path=/story/*


SaraVieira:
Weird ts error :(

packages/common/src/__snapshots__/theme.test.js.snap
```diff
@@ -1,1767 +0,0 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
```
CompuIves: Don't we need these anymore? 🤔

packages/common/src/templates/configuration/sandbox/ui.tsx
```diff
@@ -1,7 +1,7 @@
-import * as React from 'react';
+import React from 'react';
 import sortBy from 'lodash/sortBy';
 
-import * as templates from '../../../templates';
+import templates from '../../../templates';
```
CompuIves:
```suggestion
import * as templates from '../../../templates';
```

I think this one should stay the same, since we want all exports.

CompuIves: I think this is also the cause of the build error.

packages/common/src/__snapshots__/theme.test.js.snap
SaraVieira: yeah, readded

Issue #2293, opened 2019-08-15
Add dependency: "TypeError: Cannot read property 'toString' of undefined"


alexengrig:
# 🐛 bug report

## Description of the problem

The crash application after a dependency search by name that starts with "end".

## How has this issue affected you? What are you trying to accomplish?

Steps:

1.  Create React Sandbox.
2.  Add Dependency.
3.  Input "end".
4.  Get the crash.

Console output:
```javascript
breadcrumbs.js:58 TypeError: Cannot read property 'toString' of undefined
    at n.value (formatDownloads.js:12)
    at To (react-dom.production.min.js:173)
    at Eo (react-dom.production.min.js:172)
    at Po (react-dom.production.min.js:180)
    at Vs (react-dom.production.min.js:232)
    at Gs (react-dom.production.min.js:233)
    at Ta (react-dom.production.min.js:249)
    at Ca (react-dom.production.min.js:248)
    at xa (react-dom.production.min.js:245)
    at Qs (react-dom.production.min.js:243)
    at Object.enqueueSetState (react-dom.production.min.js:130)
    at y.w.setState (react.production.min.js:13)
    at createConnector.js:83
    at createStore.js:6
    at Array.forEach (<anonymous>)
    at Object.setState (createStore.js:5)
    at m.D (createInstantSearchManager.js:186)
    at m.u.emit (events.js:146)
    at algoliasearch.helper.js:1264
    at e.exports (_arrayEach.js:15)
    at e.exports (forEach.js:37)
    at m._dispatchAlgoliaResponse (algoliasearch.helper.js:1257)
```

### Your Environment

| Software         | Name/Version |
| ---------------- | ------------ |
| Сodesandbox      | Online
| Browser          |    Chromium
| Operating System | Linux Mint

JayBee007:
Can I be assigned this issue? It will be a **good first issue** for me. Thanks

CompuIves:
Yes absolutely @JayBee007, thanks for taking the time!

PR #2297, opened 2019-08-16
Pr/issue 2293

JayBee007:
It is a fix for issue [2293](https://github.com/codesandbox/codesandbox-client/issues/2293).

The response for dependency search was not containing `downloadsLast30Days` property when searching for something like `end` leading to crash of the function formatDownloads.js and eventually the whole app.

## What is the new behavior?
Add zero as a default value when calling the function formatDownloads.

Wrote test to check if the function works properly when it is not supplied an argument.

## Checklist

- [ ] Documentation N/A
- [x] Testing 
- [x] Ready to be merged 


CompuIves:
Clean solution and even has tests! Thanks a lot 😄.

Issue #2301, opened 2019-08-18
I have a question about EntryTitleInput


R4M80MrX:
# ❔ Question

### Your Environment

| Software         | Name/Version |
| ---------------- | ------------ |
| Сodesandbox      |local dev
| Browser          |chrome 73
| Operating System |mac OS && win10

local dev
![屏幕快照 2019-08-18 下午4 50 04](https://user-images.githubusercontent.com/29365693/63222273-45d20800-c1d8-11e9-8eb1-c25836ad5396.png)

codesandbox.io
![屏幕快照 2019-08-18 下午4 50 44](https://user-images.githubusercontent.com/29365693/63222275-55515100-c1d8-11e9-8eae-faf7fdd8ecb9.png)

R4M80MrX:
![屏幕快照 2019-08-18 下午5 29 28](https://user-images.githubusercontent.com/29365693/63222710-c8a99180-c1dd-11e9-8b92-8ef7ee374b28.png)
// top: 0.1rem;
// bottom: 0.1rem;
// margin: 0.2rem;

CompuIves:
Ahh, so that's it! Would you be interested in opening a PR to fix this? This is clearly unwanted behavior 😅 .

R4M80MrX:
> Ahh, so that's it! Would you be interested in opening a PR to fix this? This is clearly unwanted behavior 😅 .

Of course, it's my pleasure.

arthurdenner:
@R4M80MrX, since you've fixed the error (thank you for that), can you close this issue, please?

PR #2307, opened 2019-08-19
fix:EntryTitleInput overflow problem in workspace

R4M80MrX:
Bug fix

https://github.com/codesandbox/codesandbox-client/issues/2301

## What is the new behavior?
Behave like codesandbox.io

`yarn test` and `yarn lint`

## Checklist

- [ ] Documentation N/A
- [X] Testing 
- [X] Ready to be merged 
- [X] Added myself to contributors table


R4M80MrX:
And I found some problems when using Monaco Editor
批注 2019-08-19 180932
批注 2019-08-19 181155
Is it necessary to fix these problems?

CompuIves:
So interesting that this didn't happen on production before, well, the mystery has been solved! Thanks for helping out on this!

CompuIves:
> Is it necessary to fix these problems?

I don't think it's needed right now. Can you reproduce this consistently?

R4M80MrX:
> > Is it necessary to fix these problems?
> 
> I don't think it's needed right now. Can you reproduce this consistently?

Thank you for your timely reply. I forgot to give the details above. 

The first problem is because the \<Cofiguration> component doesn't have `this.sandbox` on it. So when changing module from a supported config to something else, `editor.sandbox`  in `disposeModuleHandler` at `Content\index.js` Ln298 will be `undefined`. 

The second maybe because that when changing module from a no-\<Monaco> component to \<Monaco>. The editor tries to re-add the existing model. But I noticed that all models and editor itself were disposed when componentWillUnmount. So I'm a little confused and trying to figure out.

Neither of them affects the app now. I'm just learning the code.☺️

PR #2315, opened 2019-08-20
remove moment

SaraVieira:
We have both moment and date-fns and we don't really need moment since all we do can be done with date-fns and smaller bundle size yay

The deleted code was not being shown, no date was coming back from the server


packages/app/src/app/pages/common/Modals/StorageManagementModal/FilesList/index.js
```diff
@@ -71,10 +70,10 @@ class FilesList extends Component {
                 `}
               />
               <HeaderTitle>File</HeaderTitle>
-              <HeaderTitle>Created</HeaderTitle>
```
CompuIves: What's happening here?

packages/app/src/app/components/SandboxList/index.js
```diff
@@ -56,8 +56,8 @@ export default ({ sandboxes, isCurrentUser, onDelete }: Props) => (
               <Link to={sandboxUrl(s)}>{s.title || s.id}</Link>
               <PrivacyStatus privacy={s.privacy} asIcon />
             </td>
-            <td>{moment(s.insertedAt).format('ll')}</td>
-            <td>{moment(s.updatedAt).format('ll')}</td>
+            <td>{format(s.insertedAt, 'DD/MM/YYYY')}</td>
```
CompuIves: This is a different format than `ll` (which is `5 Jan 2017` for example), is there a way to get that with `date-fns` as well? It would clear up a lot of confusion between US/EU.

CompuIves: You need this one `MMM DD, YYYY`

CompuIves:
Big improvement!! Thanks!

I want to make sure that nothing changes for the user, so we need to make sure that all dates are formatted the same way as before. I left some comments for that.

packages/app/src/app/pages/common/Modals/StorageManagementModal/FilesList/index.js
SaraVieira: The date was not coming back from the server :(

it was always undefined

packages/app/src/app/pages/common/Modals/SelectSandboxModal/Sandbox/index.js
```diff
@@ -17,7 +17,7 @@ export default class Sandbox extends React.PureComponent {
           {getSandboxName(sandbox)}
           {active && ' (Selected)'}
         </div>
-        <Date>{moment(sandbox.insertedAt).format('ll')}</Date>
+        <Date>{format(sandbox.insertedAt, 'DD/MM/YYYY')}</Date>
```
MichaelDeBoey:
```suggestion
        <Date>{format(sandbox.insertedAt, 'MMM DD, YYYY')}</Date>
```

SaraVieira: Danke

CompuIves:
Love this!

PR #2322, opened 2019-08-22
start tests

SaraVieira:


packages/common/src/components/AutosizeInput/index.test.tsx
```diff
@@ -0,0 +1,16 @@
+import 'jest-styled-components';
+import React from 'react';
+import mountWithTheme from '../../test/themeMount';
+import AutosizeInput from '.';
+
+let wrapper;
+
+beforeEach(() => {
+  wrapper = mountWithTheme(<AutosizeInput />);
```
MichaelDeBoey: I think it would be more consistent with the other tests if you do this inside the `it` callback.

packages/common/src/components/AutosizeTextArea/index.test.tsx
```diff
@@ -0,0 +1,16 @@
+import 'jest-styled-components';
+import React from 'react';
+import mountWithTheme from '../../test/themeMount';
+import AutosizeTextArea from './';
+
+let wrapper;
+
+beforeEach(() => {
+  wrapper = mountWithTheme(<AutosizeTextArea />);
```
MichaelDeBoey: Same

packages/common/src/components/Preference/index.stories.tsx
```diff
@@ -7,18 +7,39 @@ const stories = storiesOf('components/Preference', module);
 const keyBindingKeys = Object.keys(KEYBINDINGS);
 
 stories
-  .add('Boolean Perference', () => <Preference type="boolean" />)
-  .add('String Perference', () => <Preference type="string" value="Test" />)
-  .add('Keybinding Perference', () =>
+  .add('Boolean Preference', () => (
+    <Preference
+      setValue={() => {}}
```
MichaelDeBoey: You could use `noop` from `lodash` or just use `() => undefined` instead

packages/common/src/components/Preference/index.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+import Preference from './';
+import { KEYBINDINGS } from '../../utils/keybindings';
+
+const keyBindingKeys = Object.keys(KEYBINDINGS);
+
+describe('<Preference /> rendering', () => {
+  it('boolean', () => {
+    const wrapper = mountWithTheme(
+      <Preference
+        setValue={() => {}}
```
MichaelDeBoey: Same

packages/common/src/components/SandboxCard/index.test.tsx
```diff
@@ -0,0 +1,30 @@
+import 'jest-styled-components';
+import React from 'react';
+import mountWithTheme from '../../test/themeMount';
+import SandboxCard, { Props } from './';
+import * as fixtures from './fixtures';
+
+const createSandboxStory = ({
+  sandbox = fixtures.sandbox(),
+  selectSandbox = () => {},
```
MichaelDeBoey: Same

packages/common/src/components/RunOnClick/index.test.tsx
```diff
@@ -0,0 +1,11 @@
+import 'jest-styled-components';
+import React from 'react';
+import mountWithTheme from '../../test/themeMount';
+import RunOnClick from './';
+
+describe('<RunOnClick /> rendering', () => {
+  it('basic', () => {
+    const wrapper = mountWithTheme(<RunOnClick onClick={() => {}} />);
```
MichaelDeBoey: Same

packages/common/src/test/setupTests.ts
```diff
@@ -0,0 +1,34 @@
+import { configure } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+
+configure({ adapter: new Adapter() });
+
+type LocalStorageMockProps = {
```
MichaelDeBoey: This type isn't used atm

packages/common/src/test/setupTests.ts
```diff
@@ -0,0 +1,34 @@
...
+  store: any;
+};
+
+class LocalStorageMock {
+  store: any;
```
MichaelDeBoey: `store` could be of type `{ [key: string]: string }`

packages/common/src/test/setupTests.ts
```diff
@@ -0,0 +1,34 @@
...
+  clear() {
+    this.store = {};
+  }
+
+  getItem(key: string) {
+    return this.store[key] || null;
+  }
+
+  setItem(key: string, value: string) {
+    this.store[key] = value.toString();
```
MichaelDeBoey: `value` is already of type `string`, so why doing an extra `toString`? 🤔

packages/common/src/test/themeMount.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { ReactElement } from 'react';
+import { mount } from 'enzyme';
+import { ThemeProvider } from 'styled-components';
+import theme from '../theme';
+
+const mountWithTheme = (tree: ReactElement<any>) => {
+  const WrappingThemeProvider: React.FC<{
```
MichaelDeBoey: Using `React.FunctionComponent` instead of `React.FC` will make it a bit more clear I think

packages/common/src/test/themeMount.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { ReactElement } from 'react';
+import { mount } from 'enzyme';
+import { ThemeProvider } from 'styled-components';
+import theme from '../theme';
+
+const mountWithTheme = (tree: ReactElement<any>) => {
```
MichaelDeBoey: I think we should try to avoid using `any` and use `unknown` instead.
https://mariusschulz.com/blog/the-unknown-type-in-typescript

packages/common/src/test/setupTests.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  removeItem(key: string) {
+    delete this.store[key];
```
MichaelDeBoey: Maybe reassign instead of using `delete` so we don't mutate the `store` directly? 🤔

```ts
const { [key]: deletedKey, ...storeRest } = store;

this.store = storeRest;
```

PR #2355, opened 2019-09-02
show graphiql port on the server tab

SaraVieira:
The idea of this PR is to open the graphiql endpoint when start when its a gastby template


packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Server/elements.tsx
```diff
@@ -0,0 +1,51 @@
...
+  padding-left: 1rem;
+  font-size: 0.875rem;
+`;
+
+export const TasksContainer = styled(WorkspaceInputContainer)<{
+  disconnected: boolean;
+}>`
+  display: flex;
+  flex-direction: column;
+  pointer-events: ${props => (props.disconnected ? 'none' : 'initial')};
```
MichaelDeBoey: I think it's better to use @Saeris' method of only having 1 time the props (destructur).

```ts
export const TasksContainer = styled(WorkspaceInputContainer)<{
  disconnected: boolean;
}>`
  ${({ disconnected }) => css`
    // ...
    pointer-events: `${disconnected ? 'none' : 'initial'}`;
    // ... 
  `};
`;
```

SaraVieira: Oh man, I don't like that syntax I think it makes waaay more confusing

MichaelDeBoey: But that's the syntax we agreed on starting to use, no? 🤔

SaraVieira: I don't think so 🤔

SaraVieira: I will ask in the chat

Saeris: So there isn't really a consensus here I don't think. IMO it's preferable to do one destructure so that it's easier to extend a stylesheet with more variables later on and see in one place all the individual properties being used. By accessing them off of `props`, eslint isn't able to tell you when a property is unused, and in the case where you use a property multiple times in a single stylesheet, you can quickly lose track of the usage. Additionally it cuts down on overall verbosity in the code, because you're not repeatedly accessing off of `props`.

MichaelDeBoey: It was in @Saeris' PR where he did the big TypeScript refactor of a lot of components (#2066).

> - For styled components, Inline types are fine where their footprint is small, ie:
```ts
const SomeElement = styled.div<{ large: boolean }>`
  ${({ large }) => css`
    /* 
      Always wrap style declarations in a single function call and don't forget to
      use "css" for syntax highlighting!
    */
  `}
`
```

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Server/index.tsx
```diff
@@ -58,7 +58,13 @@ export const Server = inject('store', 'signals')(
           })
         );
       }
-    }, [editor.currentSandbox.id, ports, sandbox.template, server.ports]);
+    }, [
+      editor.currentSandbox.id,
```
MichaelDeBoey: This line can be removed I think, since you're not using it inside the effect. 🤔

SaraVieira: Good catch! DOne :)

Saeris:
Clicking the Open Ports button in the Server tab does open a new browser tab, but the URL in it added an extra `https://`, ex: `https://https://rdj2t.sse.codesandbox.io/___graphql`

Also is there any way we could:

1) Rename the browser tab to be `GraphiQL` instead of `Browser (8080)`
2) Don't open a new tab if one is already open (maybe focus the tab instead)

The sandbox does seem to open with the graphiql tab already open in addition to the regular preview. Just once you close that one and try to open it again from the server tab, the URL is messed up. Clicking that button will open a second graphiql tab in addition to the one opened by default.

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Server/elements.tsx
```diff
@@ -0,0 +1,53 @@
+import styled, { css } from 'styled-components';
+import { Button } from '@codesandbox/common/lib/components/Button';
+import PowerIcon from 'react-icons/lib/md/power-settings-new';
+import { WorkspaceInputContainer } from '../../elements';
+
+export const SubTitle = styled.div`
+  text-transform: uppercase;
+  font-weight: 700;
+  color: ${props =>
```
MichaelDeBoey: Don't forget to use @Saeris' convention here 🙂

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Server/EnvVars/elements.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export const InputContainer = styled.div`
+  display: flex;
+  width: 100%;
+  flex-direction: column;
+  margin-bottom: 0.5rem;
+`;
+
+export const ErrorMessage = styled.div`
+  color: ${props => props.theme.red};
```
MichaelDeBoey: Same

SaraVieira:
@Saeris All fixed! It won't focus on the tab but no more opening 10 of them :D

Also changed the name

packages/app/src/app/pages/Sandbox/Editor/Content/index.js
```diff
@@ -417,8 +417,14 @@ class EditorPreview extends React.Component {
 
     const browserConfig = {
       id: 'codesandbox.browser',
-      title: options =>
-        options.port ? `Browser (:${options.port})` : `Browser`,
+      title: options => {
+        const graphiql =
```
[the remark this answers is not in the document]
SaraVieira: Added :)

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Server/EnvVars/elements.tsx
SaraVieira: In the theme I'm not gonna add it because we don't type it anyway

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Server/elements.tsx
SaraVieira: In the theme I'm not gonna add it because we don't type it anyway

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Server/index.tsx
```diff
@@ -18,172 +13,183 @@ import {
...
+    const [graphiQLTabOpen, setGraphiQLTabOpen] = useState(false);
+    const disconnected = server.status !== 'connected';
+    const sandbox = editor.currentSandbox;
+
+    const openPort = (port: Port) => {
+      signals.server.onBrowserFromPortOpened({ port });
+    };
+
+    useEffect(() => {
+      if (sandbox.template === 'gatsby') {
```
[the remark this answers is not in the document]
SaraVieira: Done :)

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Server/index.tsx
```diff
@@ -18,172 +13,183 @@ import {
...
+                No ports are opened. Maybe the server is still starting or it
+                doesn't open any ports.
+              </Description>
+            )}
+            {graphqlPort && server.ports.length ? (
               <EntryContainer
                 style={{ position: 'relative' }}
-                onClick={() => openPort(port)}
+                onClick={() => {
+                  if (!graphiQLTabOpen) {
```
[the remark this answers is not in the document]
SaraVieira: All wired up and works without the if

packages/app/src/app/overmind/namespaces/server/actions.ts
```diff
@@ -103,6 +104,39 @@ export const onSSEMessage: Action<{
           }
         });
       });
+      const browserTabs = editor.devToolTabs
```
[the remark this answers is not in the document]
SaraVieira: All done!

packages/app/src/app/pages/Sandbox/Editor/Content/index.js
```diff
@@ -424,7 +424,9 @@ class EditorPreview extends React.Component {
     const browserConfig = {
       id: 'codesandbox.browser',
       title: options =>
-        options.port ? `Browser (:${options.port})` : `Browser`,
+        options.port
```
CompuIves: Need to have a check here for title, since now it doesn't work if only title is given.

packages/common/src/components/Preview/index.tsx
```diff
@@ -182,15 +182,31 @@ class BasePreview extends React.Component<Props, State> {
...
-  currentUrl = () =>
-    this.props.url ||
-    (this.serverPreview
-      ? getSSEUrl(this.props.sandbox, this.props.initialPath)
-      : frameUrl(
-          this.props.sandbox,
-          this.props.initialPath || '',
-          this.state && this.state.useFallbackDomain
-        ));
+  currentUrl = () => {
```
CompuIves: I think this function needs to be a bit different, there are two issues I see:

* It now returns `this.props.url` by default, even if it isn't set (last statement)
* It tries to append the url to sseURL or frameURL, but sometimes these urls already have something appended (thanks to `initialPath`).
* There is some double logic (the check on this.serverPreview is twice)

I think it can be simplified to:

```ts
currentUrl = () => {
  const { url, sandbox } = this.props;
  if (url && !url.startsWith('/')) {
    // An absolute url is given, just return that
    return url;
  }

  // url may be a relative path (/test), so start with that
  const initialPath = url || this.props.initialPath || '';

  return this.serverPreview
      ? getSSEUrl(sandbox, initialPath)
      : frameUrl(sandbox, initialPath, this.state && this.state.useFallbackDomain);
}
```

packages/common/src/templates/gatsby.ts
```diff
@@ -1,8 +1,39 @@
...
+  getViews(): ViewConfig[] {
+    const GATSBY_VIEWS: ViewConfig[] = [
+      {
+        views: [
+          { id: 'codesandbox.browser' },
+          {
+            id: 'codesandbox.browser',
+            closeable: true,
+            options: {
+              port: 8080,
```
CompuIves: Is the devtools always on port `8080`? If someone would choose to host the main Gatsby app on port 3000, would it still work?

CompuIves: If the devtools are always on the same port as the main app, it may be smart to just remove the `port` option, so it will default to being hosted on the main port.

packages/app/src/app/pages/Sandbox/Editor/Content/index.js
SaraVieira: pushed

packages/common/src/templates/gatsby.ts
SaraVieira: ohhh, I didnt know you could use without the port option

done

packages/common/src/components/Preview/index.tsx
SaraVieira: Seems to work great and changed

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Server/EnvVars/elements.tsx
```diff
@@ -0,0 +1,41 @@
...
+  font-size: 12px;
+  margin: 0.5rem 0;
+  font-style: italic;
+`;
+
+export const EnvName = styled(Input)`
+  margin-left: 0;
+  margin-right: 0;
+  margin-bottom: 0.25rem;
```
CompuIves: I don't know why, but it seems this margin got lost somewhere?

image

SaraVieira: It needs to overwrite another one so its back to inline

PR #2370, opened 2019-09-03
embed: make folders collapsable

### LINKED ISSUE #2342: Collapsible file tree in embeds
# 🌈 Collapsible file tree in embeds

File trees in the embeds are fully expanded and cannot be collapsed. If you have more than 20 or so file it can be a bit noisy. This is an issue for me since I'm using the embeds in a tutorial setting, and I want them to be a clean as possible. 

image

Add some embed options like "startcollapsed=1" and make the file tree work like a normal file tree.

siddharthkp:

Enhancement in `Embed` sidebar

Fixes #2342 

All folders are open by default

## What is the new behavior?

Makes folders collapsable in the sidebar.

Demo: [staging](https://pr2370.build.csb.dev/embed/currying-violet-cvgmq?fontsize=14&module=%2Fpublic%2Fanother%2Fjust-messing.js)

By default, the folder containing the open file is open.

Manually test the embed 😅 
(tested with nested directories as well)

## Checklist

- [NA] Documentation
- [x] Testing 
- [x] Ready to be merged 

## Known bug / missing feature

Nested folders don't remember their open states. Not sure if that's something we should fix here.

forgetful folders

### DIFF
--- a/packages/app/src/embed/components/File/index.js
+++ b/packages/app/src/embed/components/File/index.js
@@ -29,13 +29,16 @@ export default class File extends React.PureComponent<Props> {
   };
 
   render() {
-    const { title, depth, type, active, alternative } = this.props;
+    const { title, depth, type, active, alternative, onClick } = this.props;
     return (
       <div>
         <Entry
           alternative={alternative}
           active={active}
-          onClick={this.setCurrentModule}
+          onClick={() => {
+            onClick && onClick();
+            this.setCurrentModule();
+          }}
           type={type}
         >
           <LeftOffset depth={depth}>
--- a/packages/app/src/embed/components/File/index.js
+++ b/packages/app/src/embed/components/File/index.js
@@ -7,25 +7,18 @@ import EntryTitle from 'app/pages/Sandbox/Editor/Workspace/Files/DirectoryEntry/
 import { LeftOffset } from './elements';
 
 type Props = {
-  id: string,
-  setCurrentModule: (id: string) => void,
   title: string,
   depth: number,
   type: string,
   active?: boolean,
   alternative?: boolean,
+  onClick?: () => void,
 };
 
 export default class File extends React.PureComponent<Props> {
   static defaultProps = {
     active: false,
     alternative: false,
-  };
-
-  setCurrentModule = () => {
-    const { id, setCurrentModule } = this.props;
-
-    setCurrentModule(id);
   };
 
   render() {
@@ -35,11 +28,8 @@ export default class File extends React.PureComponent<Props> {
         <Entry
           alternative={alternative}
           active={active}
-          onClick={() => {
-            onClick && onClick();
-            this.setCurrentModule();
-          }}
           type={type}
+          onClick={onClick}
         >
           <LeftOffset depth={depth}>
             <EntryIcons type={type} />
--- a/packages/app/src/embed/components/Files/index.js
+++ b/packages/app/src/embed/components/Files/index.js
@@ -112,17 +112,16 @@ function Directory({
   depth,
 }) {
   /** directory should be open by default if currentModule is inside it */
+  const [open, setOpen] = React.useState(function() {
+    const currentModule = getCurrentModule(modules, currentModuleId);
+    const currentModuleTree = getCurrentModuleTree(directories, currentModule);
 
-  const currentModule = getCurrentModule(modules, currentModuleId);
-
-  const currentModuleTree = getCurrentModuleTree(directories, currentModule);
-
-  let openByDefault = false;
-  if (currentModuleTree.find(module => module.id === directory.id)) {
-    openByDefault = true;
-  }
-
-  const [open, setOpen] = React.useState(openByDefault);
+    let openByDefault = false;
+    if (currentModuleTree.find(module => module.id === directory.id)) {
+      openByDefault = true;
+    }
+    return openByDefault;
+  });
 
   return (
     <div>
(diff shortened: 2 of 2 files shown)


packages/app/src/embed/components/File/index.js
```diff
@@ -29,13 +29,16 @@ export default class File extends React.PureComponent<Props> {
...
+    const { title, depth, type, active, alternative, onClick } = this.props;
     return (
       <div>
         <Entry
           alternative={alternative}
           active={active}
-          onClick={this.setCurrentModule}
+          onClick={() => {
+            onClick && onClick();
+            this.setCurrentModule();
```
siddharthkp: merge local `onClick` and the one from props

packages/app/src/embed/components/Files/index.js
```diff
@@ -44,28 +47,21 @@ function Files({
...
-            setCurrentModule={setCurrentModule}
-          />
-          <Files
-            modules={modules}
-            directories={directories}
-            directoryId={d.shortid}
-            depth={depth + 1}
+      {sortBy(childrenDirectories, directory => directory.title).map(
+        directory => (
+          <Directory
```
siddharthkp: Created a new component to make the component easier to understand

packages/app/src/embed/components/Files/index.js
```diff
@@ -15,7 +18,7 @@ import { Container } from './elements';
 
 type Props = {
   modules: Array<Module>,
-  directories: Array<Directory>,
+  directories: Array<DirectoryType>,
```
MichaelDeBoey: You could write this as

```suggestion
  directories: DirectoryType[],
```

packages/app/src/embed/components/Files/index.js
```diff
@@ -83,4 +79,80 @@ function Files({
...
+    </div>
+  );
+}
+
 export default Files;
+
+/** Utils to help identify module tree */
+
+const getParentDirectory = (directories, child) => {
+  return directories.find(directory => {
```
MichaelDeBoey: This could be an implicit return

packages/app/src/embed/components/Files/index.js
```diff
@@ -83,4 +79,80 @@ function Files({
...
+    return directory.shortid === child.directoryShortid;
```
MichaelDeBoey: Same

packages/app/src/embed/components/Files/index.js
CompuIves: That's very nice!

packages/app/src/embed/components/Files/index.js
```diff
@@ -83,4 +79,80 @@ function Files({
...
+  setCurrentModule,
+  template,
+  entry,
+  modules,
+  directories,
+  depth,
+}) {
+  /** directory should be open by default if currentModule is inside it */
+
+  const currentModuleTree = getCurrentModuleTree(
```
CompuIves: This is a very nice addition!

packages/app/src/embed/components/File/index.js
CompuIves: It might be better to extract the `setCurrentModule` to the `onClick` handler of `File`, I noticed that when you click a directory we automatically open `src/index.js`. I assume that this happens because `this.setCurrentModule` is also called for directories, which resolves to an unknown module and then defaults to `src/index.js`.

siddharthkp:
> I noticed that when you click a directory we automatically open src/index.js.

🐞💀

packages/app/src/embed/components/Files/index.js
siddharthkp: Ah, I'm really conflicted about implicit returns for really long statements

The latter is shorter, but I'm not sure if it's easier to _read_/understand 🤷‍♀ 

Screenshot 2019-09-04 at 11 13 53 AM

@SaraVieira @Saeris how do you folks feel about this?

Saeris: Personally I always go with implicit returns. I'm fine either way.

packages/app/src/embed/components/Files/index.js
siddharthkp: ✅

packages/app/src/embed/components/Files/index.js
siddharthkp: > Personally I always go with implicit returns. I'm fine either way.

I try to break the rule when needed 😅, but I'm fine either way too 👍. Changed it to implicit returns

packages/app/src/embed/components/File/index.js
siddharthkp: Yep, good catch!

siddharthkp:
Looks ready to ship :shipit: 

There are a few other things to cleanup in this part of the codebase - the primary one is to move presentation code for Files to `common` so that we can keep them consistent all across. Will create a different PR for that 👍

packages/app/src/embed/components/Files/index.js
```diff
@@ -74,13 +70,86 @@ function Files({
...
+  const currentModule = getCurrentModule(modules, currentModuleId);
+
+  const currentModuleTree = getCurrentModuleTree(directories, currentModule);
+
+  let openByDefault = false;
+  if (currentModuleTree.find(module => module.id === directory.id)) {
+    openByDefault = true;
+  }
+
+  const [open, setOpen] = React.useState(openByDefault);
```
MichaelDeBoey: You can add the `openByDefault` calculation in the `useState`'s callback.

https://reactjs.org/docs/hooks-reference.html#lazy-initial-state

siddharthkp: Not sure if this counts as expensive, but can't hurt 😄

MichaelDeBoey: Yeah I know it's not really expensive, but it makes it more readable and clear imo.

PR #2376, opened 2019-09-04
✨ Add URL to Crash Report, Add CodeSadbox route in Development

Saeris:
Updated the crash report template to include the current URL so as to make it easier to hunt down where crashes are occurring. Also updated the template copy to make it more explicit that the user should use the template to write up a crash report.

When running the client locally, the `/codesadbox` route will now be available to make it easier to test changes to the error boundary components. This route is removed in production builds.


CompuIves:
Nice additions!

Issue #2249, opened 2019-08-04
Vanilla (Parcel): Default Open File is Unexpected / Unreliable


karlhorky:
# 🐛 bug report

## Description of the problem

Opening a Vanilla (Parcel) sandbox, it is not clear which file will open. I have been unable to find why this is in my quick investigation of the behavior, but its behavior seems to have the following characteristics:

1. A new Vanilla (Parcel) sandbox will open `src/index.js`, even though it is not listed under the `main` field in `package.json`
2. If this sandbox is forked by a user and this file removed (either with or without the reference in `index.html`), then the default file that will open will now be `index.html`
3. If the `src/styles.css` file is renamed to `src/index.css` then this file will now be the default file to open.

It seems like there should be a more clear, reliable list of default files. My suggestion would be:

1. Optional: The last files open, if any (see #1456 below)
2. Initial opened file setting (see #514 below)
3. The `main` field in `package.json`

If these three configuration options are not set, I would suggest not opening any file initially.

This may also apply to other templates, potentially in different ways.

## How has this issue affected you? What are you trying to accomplish?

I would like my sandboxes to have a reliable default initial file, based on clear rules.

### Your Environment

| Software         | Name/Version |
| ---------------- | ------------ |
| Сodesandbox      | PROD-1564761038-1a9fb01ba |
| Browser          | Chrome 76.0.3809.87 (Official Build) (64-bit) |
| Operating System | macOS Mojave 10.14.6 (18G84) |

### Related

Allow setting the initial opened file in `sandbox.config.js`: https://github.com/codesandbox/codesandbox-client/issues/514

Add sandbox configuration to change default opened file: https://github.com/codesandbox/codesandbox-client/issues/1254

Other related features: 

Array syntax for `main` field in `package.json`: https://github.com/codesandbox/codesandbox-client/issues/1739

Open previously-opened files on refresh: https://github.com/codesandbox/codesandbox-client/issues/1456

**Partially related:**

Vanilla (Parcel) Unreliable Behavior When index.html Under Other Path (for example, in src)
https://github.com/codesandbox/codesandbox-client/issues/2240

lbogdan:
Hey @karlhorky ,

Thanks for the really detailed description! (for the other opened issues, also 🙂) I'll brainstorm this with the team and hopefully we'll come up with a clear, intuitive enough solution.

andyrichardson:
Weirdly enough I'm having this same issue.

[...]

PR #2382, opened 2019-09-04
Use app logic for loading default file in embed

### LINKED ISSUE #2091: README.md does not display in embed when selected as "default module to show".
# 🐛 bug report

## Description of the problem

When sharing a CodeSandbox, I clicked `README.md` to display as the default module users see when the embed opens. However, a random file is displayed instead of the `README.md`.

### Steps to reproduce
- Open [CodeSandbox](https://codesandbox.io/embed/code-challenge-uhz9n)
- Click _Share_
- In the _Configure_ section of the modal, click on _Sandbox options_
- In the _Default module to show_ section, click on `README.md`

In my case, `tests/state/reducers/gameReducer.js` shows, not `README.md` from the root directory.

This CodeSandbox is a solution to an interview problem. When the reviewing developer opens my solution, I'd like them to show them the `README.md`. The `README.md` shows my ability to write documentation and includes notes about my implementation.

Outside of the reviewing developer, visitors to the embed may be unfamiliar with [Conway's Game of Life](https://en.wikipedia.org/wiki/Conway%27s_Game_of_Life). The `README.md` includes information about the purpose, rules, and gameplay.

### Your Environment

| Сodesandbox      | PROD-1561388488-a3b21c078|
| Browser          | Chrome Version 75.0.3770.100 (Official Build) (64-bit)
| Operating System |macOS Mojave 10.14.4

siddharthkp:

Bugfix, Fixes #2091

There is a concept of "mainModule" when sharing a sandbox. For node project without a clear `index.js` file, this will default to `README.md`. Example: [/s/code-challenge-uhz9n](https://codesandbox.io/s/code-challenge-uhz9n)

This is the file that the editor in sandbox will open by default.

However, in the case of embed, if the entry file doesn't exist, it falls back to the first available file (`modules[0]`). This often opens an unexpected (sometimes deeply nested) file. Example: [/embed/code-challenge-uhz9n](https://codesandbox.io/embed/code-challenge-uhz9n)

## What is the new behavior?

We do have the concept of the concept of [defaultOpenedFiles](https://github.com/codesandbox/codesandbox-client/blob/master/packages/common/src/templates/template.ts#L191-L195) which is a list of potential default options (including README)

Using that as a fallback before `modules[0]` is a better result. Demo: [/staging/embed/code-challenge-uhz9n](https://pr2382.build.csb.dev/embed/code-challenge-uhz9n)

**Doubt**: is it weird that `sandbox.entry` is set to a file that doesn't exist? Does this come from the files in the sandbox or template configuration? 🤔 

- [x] Tested embed without module in URL
- [x] Tested embed with module in URL
- [x] Testing embed with node sandbox
- [x] Testing embed with react sandbox
- [x] Checked if this affects `/s`, seems to be fine 

## Checklist

- NA Documentation
- [x] Testing 
- [x] Ready to be merged 


packages/app/src/embed/components/App/index.js
```diff
@@ -225,13 +225,16 @@ export default class App extends React.PureComponent<
...
+      sandbox.directories,
+      currentModulePath,
+      findMainModule(
         sandbox.modules,
         sandbox.directories,
-        currentModulePath,
-        findMainModule(sandbox.modules, sandbox.directories, sandbox.entry)
-      ) || sandbox.modules[0]
+        sandbox.entry,
+        sandbox.template
```
siddharthkp: `Notes for reviewer`: Pass `sandbox.template` to use in attempt to find `mainModule`

packages/app/src/embed/components/Content/index.tsx
```diff
@@ -387,7 +387,12 @@ export default class Content extends React.PureComponent<Props, State> {
     } = this.props;
 
     const mainModule = isInProjectView
-      ? findMainModule(sandbox.modules, sandbox.directories, sandbox.entry)
+      ? findMainModule(
+          sandbox.modules,
+          sandbox.directories,
+          sandbox.entry,
+          sandbox.template
```
siddharthkp: `Notes for reviewer`: Not sure what Content does with this information, but keeping it consistent with App

packages/common/src/sandbox/modules.ts
```diff
@@ -251,14 +252,30 @@ export const isMainModule = (
...
-  entry: string = 'index.js'
+  entry: string = 'index.js',
+  template?: TemplateType
 ) => {
   try {
+    // first attempt: try loading the entry file if it exists
     const module = resolveModule(entry, modules, directories);
-
     return module;
   } catch (e) {
```
siddharthkp: ugh nested try catch. Is there a better way to handle this? This looks ugly.

packages/common/src/sandbox/modules.ts
```diff
@@ -284,3 +301,19 @@ export const findCurrentModule = (
     mainModule
   );
 };
+
+export const resolveModuleWrapped = sandbox => (path: string) => {
```
siddharthkp: Moved these 2 utils from `templates/configuration/resolve-module-wrapped.ts`, needed to use one of them. This feels like a nicer home for them.

packages/common/src/sandbox/modules.ts
```diff
@@ -251,14 +252,30 @@ export const isMainModule = (
 export const findMainModule = (
   modules: Module[],
   directories: Directory[],
-  entry: string = 'index.js'
+  entry: string = 'index.js',
+  template?: TemplateType
 ) => {
   try {
+    // first attempt: try loading the entry file if it exists
     const module = resolveModule(entry, modules, directories);
```
MichaelDeBoey: You can just directly return this, no? 🤔

packages/common/src/sandbox/modules.ts
```diff
@@ -251,14 +252,30 @@ export const isMainModule = (
...
   } catch (e) {
-    return modules[0];
+    try {
+      // second attempt: try loading the first file that exists from
+      // the list of possible defaults in the template defination
+      const templateDefinition = getTemplateDefinition(template);
+
+      const defaultOpenedFiles = templateDefinition.getDefaultOpenedFiles({});
+
+      const defaultOpenModule = defaultOpenedFiles
```
MichaelDeBoey: Same

packages/common/src/sandbox/modules.ts
siddharthkp: Can do

packages/common/src/sandbox/modules.ts
siddharthkp: Can, but wanted to use the variable name as an excuse to label the value

siddharthkp:
@MichaelDeBoey @CompuIves @Saeris Is the right direction to go into? Running on a hunch here.

packages/common/src/sandbox/modules.ts
MichaelDeBoey: I think ESLint will complain too if I'm not mistaken

siddharthkp: Doesn't look like it 😇

CompuIves: Hmm yes I like this clear variable name, good documentation.

packages/common/src/sandbox/modules.ts
```diff
@@ -251,14 +252,30 @@ export const isMainModule = (
...
-
     return module;
   } catch (e) {
-    return modules[0];
+    try {
+      // second attempt: try loading the first file that exists from
+      // the list of possible defaults in the template defination
+      const templateDefinition = getTemplateDefinition(template);
+
+      const defaultOpenedFiles = templateDefinition.getDefaultOpenedFiles({});
```
CompuIves: Ah, normally we pass in the parsed configurations (like `package.json`) here, in case it's needed to determine the main file. We use the function `parseSandboxConfigurations` (https://github.com/CompuIves/codesandbox-client/blob/b228282987eb687bf4251c1fb564ec6467e32d6c/packages/app/src/app/store/actions.js#L173) to get these configurations, but that function needs `sandbox`, which is not available here.

It would be good to use the parsed configurations as well, to keep consistency with the editor, but we'd probably need to get another argument here with parsed configurations...

packages/common/src/sandbox/modules.ts
CompuIves: Hmm, if we use `resolveModuleWrapped` here we can do a nullcheck on the result, move on to the next flow if needed.

packages/common/src/templates/configuration/resolve-module-wrapped.ts
```diff
@@ -1,17 +1,10 @@
...
+} from '../../sandbox/modules';
 
-export const resolveModuleWrapped = sandbox => (path: string) => {
-  try {
-    return resolveModule(path, sandbox.modules, sandbox.directories);
-  } catch (e) {
-    return undefined;
-  }
-};
+// keeping this file around to reduce merge conflicts
```
CompuIves: Smart!

lbogdan:
Should we maybe address #2249 (and the referenced related ones in that issue's description) with this, too?

siddharthkp:
> Should we maybe address #2249 (and the referenced related ones in that issue's description) with this, too?

Yep, seems related. But the change would probably be in templates, not embed. Will attempt this in a different pull request

packages/common/src/sandbox/modules.ts
siddharthkp: Nice, I really like that

packages/common/src/sandbox/modules.ts
siddharthkp: I couldn't import that exact file because of a cyclic dependency. But was able to repurpose it's implementation.

packages/common/src/sandbox/modules.ts
```diff
@@ -249,34 +250,36 @@ export const isMainModule = (
...
-      // second attempt: try loading the first file that exists from
-      // the list of possible defaults in the template defination
-      const templateDefinition = getTemplateDefinition(template);
+export const findMainModule = (sandbox?: Sandbox) => {
+  const resolve = resolveModuleWrapped(sandbox);
 
-      const defaultOpenedFiles = templateDefinition.getDefaultOpenedFiles({});
+  // first attempt: try loading the entry file if it exists
+  const entryModule = resolve(sandbox.entry);
+  if (entryModule) return entryModule;
```
MichaelDeBoey: Put curly braces around body to make this less error-prone.

packages/common/src/sandbox/modules.ts
```diff
@@ -249,34 +250,36 @@ export const isMainModule = (
...
+
+  const defaultOpenedFiles = templateDefinition.getDefaultOpenedFiles(
+    parsedConfigs
+  );
+
+  const defaultOpenModule = defaultOpenedFiles
+    .map(path => resolve(path))
+    .find(module => module);
+
+  if (defaultOpenModule) return defaultOpenModule;
```
MichaelDeBoey: Same

packages/common/src/sandbox/modules.ts
siddharthkp: Like this:?

```js
if (entryModule) {
  return entryModule;
}
```

MichaelDeBoey: Yes that's what I ment indeed 🙂

siddharthkp: Done

siddharthkp:
Merged master (https://github.com/codesandbox/codesandbox-client/pull/2392) in, this should make the build work now.

PR #2392, opened 2019-09-06
Browser compatible jsonlint

siddharthkp:

It is a reduction in code complexity (hopefully)

We use the `jsonlint` package which requires `file` and `system` based on environment. 

To use `jsonlint` on the browser, we have to declare [`file`, `system`] as "external" for each build process that consumes `common`.

## What is the new behavior?

This pull request removes the file system dependencies of `jsonlint` by adding a forked version into our codebase.

@CompuIves Right now, I've created a new file `jsonlint.browser.js` next to the place it's used. Is there a convention for where we drop forked vendor-ish files?

## Checklist

- [NA] Documentation
- [x] Testing 
- [x] Ready to be merged 

### DIFF
--- a/packages/app/config/webpack.common.js
+++ b/packages/app/config/webpack.common.js
@@ -375,7 +375,7 @@ module.exports = {
   },
 
   // To make jsonlint work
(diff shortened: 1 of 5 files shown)


CompuIves:
Perfect!

PR #2398, opened 2019-09-06
Replaced skewed logo image

dattard21:
Added a transparent square logo in PNG format instead of older, skewed version.

Small update to website logo


CompuIves:
Makes sense to me, thanks for the contribution!

PR #2406, opened 2019-09-09
[WIP] Redesigned file tabs

siddharthkp:
work in progress 🚧 

[preview](https://codesandbox-client-git-common-tabs.codesandbox1.now.sh/?path=/story/components-filetabs--basic)


packages/common/src/components/FileTabs/index.tsx
```diff
@@ -1,16 +1,29 @@
...
-import { Container, Tab, Close } from './elements';
+import { Tabs, Tab, Close } from './elements';
+
+/**
+  Because we have two themes, we need to
+  segregate application of styles.
+
+  We wrap the Container with a ThemeProvider.
+
+  TODO: Find a better way of doing this, we're
```
MichaelDeBoey: We can add an extra `ThemeProvider` in the root (where we have `styled-components`' one), no? 🤔

packages/common/src/components/nu-theme/index.js
```diff
@@ -0,0 +1,36 @@
+// new theme based on redesign
+// this follows the System UI Theme Specification
+// Ref: https://system-ui.com/theme
+import components from './components';
+
+const colors = {
```
MichaelDeBoey: Maybe we can have a separate file where we got all the colors, ... and then 2 theme files (1 for `style-components` and 1 for `react-ui`), so the colors stay in sync in both theming systems? 🤔

PR #2420, opened 2019-09-12
Cleanup onCreateNode in the gatsby-node file

MichaelDeBoey:
After looking at the [`onCreateNode` function in the `gatsby-node` file](https://github.com/kentcdodds/kentcdodds.com/blob/5fcb4b2ad9e9cc55fda049faa08ef2be913c6afd/gatsby-node.js#L288-L443) of @kentcdodds' [personal website](https://github.com/kentcdodds/kentcdodds.com), I thought it would be a good idea to refactor our homepages `gatsby-node` file and put all mutations and default-values inside of this file's `onCreateNode` function.
This looks to me like a good idea so the components that are using those fields don't have to do it themselves and we don't have to repeat those steps in all the components.

Every component has to foresee their own default values or mutations to fields from within the corresponding Markdown file.

## What is the new behavior?
The `onCreateNode` function in the `gatsby-node` file takes care of foreseeing all default values and mutations so the components just have to query them and show them.

Go through all pages (blogposts, docs, jobs & legal pages) to see if everything is still working as it should be.

## Checklist
- [ ] Documentation N/A
- [x] Testing
- [x] Ready to be merged


SaraVieira:
Hey!

Sorry can you fix the conflicts? :(

MichaelDeBoey:
@SaraVieira I've fixed the merge conflicts 🙂

MichaelDeBoey:
@SaraVieira I've fixed the merge conflicts 🙂

packages/homepage/src/templates/docs.js
```diff
@@ -211,17 +211,20 @@ const Description = styled.p`
 // eslint-disable-next-line
 export default class Docs extends React.Component {
```
SaraVieira: Can you make this a functional component? Maybe after this one is merged

packages/homepage/src/templates/docs.js
```diff
@@ -263,33 +264,32 @@ export default class Docs extends React.Component {
 
 export const pageQuery = graphql`
   query Docs($slug: String!) {
-    allMarkdownRemark(
-      filter: { fields: { slug: { regex: "/docs/" } } }
-      sort: { fields: [fileAbsolutePath], order: ASC }
+    allDocs: allMarkdownRemark(
+      filter: { fileAbsolutePath: { regex: "/docs/" } }
+      sort: { fields: [fileAbsolutePath], order: [ASC] }
```
SaraVieira: What is the difference between `[ASC]` and `ASC`?

SaraVieira:
All looks good! I am gonna merge and if you would like it would be nice to move all the `css` to the elements file

Thank you!

packages/homepage/src/templates/docs.js
MichaelDeBoey: No real difference in this case, just a bit cleaner to specify it for each sorted field I think 🙂

packages/homepage/src/templates/docs.js
MichaelDeBoey: Will do in a follow-up PR 🙂

MichaelDeBoey:
@SaraVieira Will do that in a follow-up PR 🙂

PR #2427, opened 2019-09-14
add toggle statusbar

### LINKED ISSUE #2150: Hiding status bar
# 🌈 Feature

Would be nice to be able to hide status bar.

Looks like toggling it in "View › Appearance › Hide/Show Status Bar" as well as changing VSCode setting "Workbench › Status Bar: Visible" both do nothing to the status bar visibility. 

So I wonder if it's implemented at the moment.

Being able to hide it would free some additional space of the screen:

Screen Shot 2019-07-09 at 10 05 09

SaraVieira:
This adds ability to toggle the status bar

closes #2150

### DIFF
--- a/packages/app/src/app/pages/Sandbox/Editor/index.js
+++ b/packages/app/src/app/pages/Sandbox/Editor/index.js
@@ -134,10 +134,11 @@ class ContentSplit extends React.Component {
                 <Content match={match} />
               </SplitPane>
 
-              {vscode && statusBar && (
+              {vscode && (
                 <StatusBar
                   style={{
                     position: 'fixed',
+                    display: statusBar ? 'block' : 'none',
                     bottom: 0,
                     left: 0,
                     right: 0,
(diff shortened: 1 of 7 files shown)


packages/app/src/app/pages/Sandbox/Editor/index.js
```diff
@@ -132,9 +135,10 @@ class ContentSplit extends React.Component {
               </SplitPane>
 
               {vscode && (
-                <AOverride
+                <StatusBar
                   style={{
                     position: 'fixed',
+                    display: statusBar ? 'block' : 'none',
```
MichaelDeBoey: Maybe don't render the component if `statusBar` doesn't have a value? 🤔

```ts
{vscode && statusBar && (
  // ...
)}
```

SaraVieira: That's on porpuse because the sidebar won't remount again if I do that

MichaelDeBoey: Maybe add a comment to make this more clear then? 🤔
This way people who work on this piece in the future won't do it the way I suggested.

PR #2430, opened 2019-09-15
Add modals concept

christianalfoni:
This just adds the modal concept showed off here: https://www.youtube.com/watch?v=2DeNIaZC6E4

We can use this where it makes sense when we start refactoring components and VSCode effect


packages/app/src/app/overmind/factories.ts
```diff
@@ -80,3 +80,78 @@ export const withOwnedSandbox = <T>(
 
   return continueAction(context, payload);
 };
+
+export const createModals = <
+  T extends {
+    [name: string]: {
+      state?: IState;
+      result?: any;
```
MichaelDeBoey: Maybe use `unknown` type instead too add? 🤔

packages/app/src/app/overmind/factories.ts
```diff
@@ -80,3 +80,78 @@ export const withOwnedSandbox = <T>(
...
+    };
+  }
+>(
+  modals: T
+): {
+  state?: {
+    current: keyof T;
+  } & {
+    [K in keyof T]: T[K]['state'] & { isCurrent: IDerive<any, any, boolean> }
```
MichaelDeBoey: Same

packages/app/src/app/overmind/factories.ts
```diff
@@ -80,3 +80,78 @@ export const withOwnedSandbox = <T>(
...
+
+      return aggr;
+    },
+    {
+      state: {
+        current: null,
+      },
+      actions: {},
+    }
+  ) as any;
```
MichaelDeBoey: Same

packages/app/src/app/overmind/factories.ts
christianalfoni: Yeah, agree on unknown here is it affects the DX :)

packages/app/src/app/overmind/factories.ts
christianalfoni: This type will never be evaluated, it is just to "match a derived", so will always be defined from the out (in overmind/index.ts)

packages/app/src/app/overmind/factories.ts
christianalfoni: `any` is used here so that the function return type can decide what it is. This type has no effect on the consumption of the function :)

CompuIves:
Ready to merge, just one typing issue left!

PR #2437, opened 2019-09-17
notification indication, contributors fixed again, changed modules fixed

christianalfoni:


packages/app/src/app/overmind/effects/api/index.ts
```diff
@@ -62,12 +62,13 @@ export default {
...
     const sandbox = await api.get<Sandbox>(`/sandboxes/${id}`);
 
-    // We need to add savedCode property to track it
+    // We need to add client side properties for tracking
     return {
       ...sandbox,
       modules: sandbox.modules.map(module => ({
         ...module,
         savedCode: null,
+        isNotSynced: false,
```
CompuIves: Interesting, where do we use this?

packages/app/src/app/overmind/effects/live/clients.ts
```diff
@@ -39,7 +39,7 @@ class CodeSandboxOTClient extends Client {
   }
 
   sendOperation(revision, operation) {
-    this.onSendOperation(revision, operation);
+    this.onSendOperation(revision, operationToElixir(operation.toJSON()));
```
CompuIves: I wonder how `master` currently works without this 🤔. But nice to see it fixed now!

christianalfoni: Was it not opposite? I had merge conflict. I chose the wrong thing, it is handled up above, 2 sec

packages/app/src/app/overmind/effects/api/index.ts
christianalfoni: It is being used in FilePath, ImageViewer and FuzzySearch... maybe related to iPad only?

CompuIves: Oh wow, do we ever change this property? Might be a legacy thing from the redux days.

christianalfoni: We should track synced state on modules anyways, and make changedModuleshortids a derived, so this is a good reminder :)

packages/app/src/app/overmind/effects/live/clients.ts
christianalfoni: This is fixed now

PR #2438, opened 2019-09-17
Optimize SVG files

SaraVieira:
This optimzes all SVG files with SVGO


CompuIves:
Hmm, could we make this a step in Webpack maybe? This change makes it pretty hard to alter the svg files when you want to make small changes like `viewBox` and colors, kind of similar to committing minified JS. If it's too much work we can merge this in, just want to know if that possibility is there.

PR #2445, opened 2019-09-17
[Request for comments] Icons workflow

siddharthkp:
We will soon have a beautiful set of icons for the codesandbox app, might be a good time to think about how will we consume + maintain them.

Here's one way:

I've added a new package `@codesandbox/icons`

### Usage workflow

You should be able to import an icon by path if you know the name.

```js
import { CloseIcon, PlusIcon } from '@codesandbox/icons';

render(
  <div>
    <PlusIcon />
	<CloseIcon />
  </div>
)
```

Icon discovery can be painful. This is where having a story with all Icons would be very useful

Screenshot 2019-09-17 at 3 43 19 PM
_bonus points if there's a search with aliases: trash=delete=remove_

### Contributor workflow 

Icons are [designed in Figma](https://www.figma.com/file/PGNNkvzr96QGmyqg7zeYeD/Icons)

When you have a new icon, you can

Dump exported `.svg` files in `/icons/svg` folder.

And run `yarn build` which will 
- Run `svgr` with the `svgo` plugin on the icons
- create master file which exports named components
- merge PR
- update [the sandbox](https://6izxn.csb.app) to show new icons (codesandbox CI?)

Ideally, this shouldn't take more than a minute and Danny can introduce new icons when he creates a new one.

Open questions:
1. Do we want to version the icons library or publish to npm? We could also just treat it as an internal package for `codesandbox-client` and let lerna link them up.
2. Can we automate the work of pulling a svg from figma? If yes, that would mean we can treat figma as the source of truth and generate the rest (versioning would be super important in this case)

This pull request has a working draft that you can go through.


Saeris:
One thing that jumps out to me immediately is I would prefer that we use named exports and not default exports as best we can manage. Read here for some reasoning behind this: https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/

Might be worth the time to check out how Font Awesome handles their icons. While it would be nice to have a single import, it seems pretty likely to me that it would cause tree-shaking issues. For that reason I'd prefer to just export each icon from a top level index file, where we can also use ESM syntax to provide aliases for icons, ex:

```typescript
export { Trash, Trash as Delete, Trash as Remove } from "./icons/Trash"
```

I think internally we could use Lerna to link up the package so that in development of the client we always have the latest version of the icons library. However it would be good to publish it to NPM as well so that we can import it into prototyping projects within a sandbox too.

Overall looks like a great idea to me! I'm all for automating this workflow!

packages/icons/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "scripts": {
+    "storybook": "start-storybook -p 6007",
+    "build": "node scripts/build"
+  },
+  "license": "MIT",
+  "devDependencies": {
+    "@storybook/react": "5.2.0-beta.28",
+    "lodash.camelcase": "^4.3.0",
+    "lodash.upperfirst": "^4.3.1",
+    "prettier": "^1.18.2",
```
MichaelDeBoey: Prettier recommends using an exact version

https://prettier.io/docs/en/install.html

> We recommend pinning an exact version of prettier in your `package.json` as we introduce stylistic changes in patch releases.

packages/icons/scripts/build.js
```diff
@@ -0,0 +1,106 @@
...
+const glob = require('glob');
+const SVGO = require('svgo');
+const path = require('path');
+const prettier = require('prettier');
+const camelCase = require('lodash.camelcase');
+
+const svgo = new SVGO({});
+
+const svgPaths = glob.sync('./svg/*');
+const outputDir = './react/';
```
MichaelDeBoey: Maybe call it `__generated__`  or something, so it's more clear these files are generated? 🤔

packages/icons/scripts/build.js
```diff
@@ -0,0 +1,106 @@
...
+        fs.writeFileSync(path.join(outputDir, 'all.js'), contents, 'utf8');
+      }
+    })
+    .catch(error => {
+      console.log(filepath, error);
+    });
+});
+
+function getComponentSource(source) {
+  return `
```
MichaelDeBoey: Use implicit return

packages/icons/scripts/build.js
```diff
@@ -0,0 +1,106 @@
...
+    const Icon = props => (
+      ${getSvg(source)}
+    )
+
+    export default Icon
+  `;
+}
+
+function getNamedExport(name, source) {
+  return `
```
MichaelDeBoey: Same

packages/icons/.storybook/icon.stories.js
```diff
@@ -0,0 +1,44 @@
...
+  display: inline-block;
+  border: 2px solid transparent;
+  > label {
+    font-size: 14px;
+  }
+`;
+
+/** require all icon files */
+function requireAll(context) {
+  return context
```
MichaelDeBoey: Same

packages/icons/.storybook/icon.stories.js
```diff
@@ -0,0 +1,44 @@
...
+import styled from 'styled-components';
+import { storiesOf } from '@storybook/react';
+
+const stories = storiesOf('Icons', module);
+
+const context = require.context('./', false, /.js$/);
+const files = requireAll(context);
+
+stories.add('Gallery', () => {
+  return files.map(({ name, Component }) => {
```
MichaelDeBoey: Same

packages/icons/scripts/build.js
```diff
@@ -0,0 +1,106 @@
...
+const path = require('path');
+const prettier = require('prettier');
+const camelCase = require('lodash.camelcase');
+
+const svgo = new SVGO({});
+
+const svgPaths = glob.sync('./svg/*');
+const outputDir = './react/';
+
+const prettierOptions = { parser: 'babel', singleQuote: true };
```
MichaelDeBoey: Maybe add all options we use in [the rest of our codebase](https://github.com/codesandbox/codesandbox-client/blob/62e179b292e629e8a4cf28205b8b6c6136e0636c/.prettierrc)? 🤔

```suggestion
const prettierOptions = {
  endOfLine: 'lf',
  parser: 'babel',
  singleQuote: true,
  trailingComma: 'es5',
};
```

Or maybe get it from the root's `.prettierrc`? 🤔

packages/icons/package.json
siddharthkp: no feedback on the code please (yet), this is a draft 😇

MichaelDeBoey: Sorry for my enthusiasm 😁😇

siddharthkp:
> One thing that jumps out to me immediately is I would prefer that we use named exports and not default exports as best we can manage. Read here for some reasoning behind this: [humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module](https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/)

Thanks for the link! That helps

I'm curious if we even need the single icon files then. If we can get tree shaking on the master file - we could just have a single way of importing

```
import { Close, Delete } from '@codesandbox/icons'
```

CompuIves:
This is a great suggestion, and also a good candidate to hook up with CodeSandbox CI!

For publishing I don't think we need to publish for now, since it's only used internally. When it comes to how to resolve icons, I would prefer to have the (classic) way of:

```
import { TrashIcon } from '@codesandbox/icons/trash';
```

That way we certainly know that we get "treeshaking" as we only include the file that we need, I've been bitten before by other libraries where I thought we were treeshaking, but then Webpack opted out for some cases. I don't trust treeshaking enough to do all exports from a single file, additionally since we have TS auto import, importing an icon should be as simple as typing the icon name and letting TS do the rest.

The idea of letting Figma generate the files is veery intriguing, would be very cool to explore this! Also would be interesting if we could generate TS definitions as well.

siddharthkp:
> This is a great suggestion, and also a good candidate to hook up with CodeSandbox CI!

Yes please! Then I don't have to use storybook, I can just use a sandbox ✨

siddharthkp:
>  I would prefer to have the (classic) way of:
> 
> import { TrashIcon } from '@codesandbox/icons/trash';
> 
> That way we certainly know that we get "treeshaking" as we only include the file that we need

I want to spend time making sure we always get tree-shaking, fear driven development is dangerous!

CompuIves:
> I want to spend time making sure we always get tree-shaking, fear driven development is dangerous!

I agree with the sentiment, it would be good to take a closer look at this.

I'm not sure if I see the advantage of exporting everything from a single file, what is the benefit over exporting an icon per file?

siddharthkp:
> I'm not sure if I see the advantage of exporting everything from a single file, what is the benefit over exporting an icon per file?

Just convenience really of usage, really

```js
import { Refresh, Fullscreen, Close } from '@codesandbox/icons'

vs

import { Refresh } from '@codesandbox/icons/refresh'
import { Fullscreen } from '@codesandbox/icons/fullscreen'
import { Close } from '@codesandbox/icons/close'
```

MichaelDeBoey:
I would prefer the 1 line import too if possible _and_ treeshaking is available.

PR #2461, opened 2019-09-19
Fix ZEIT deployments for overmind

CompuIves:


packages/app/src/app/overmind/effects/zeit.ts
```diff
@@ -87,16 +87,20 @@ export default (() => {
...
     },
     async getUser(): Promise<ZeitUser> {
-      const response = await axios.get('https://api.zeit.co/www/user', {
-        headers: getDefaultHeaders(),
-      });
+      const response = await axios
+        .get('https://api.zeit.co/www/user', {
+          headers: getDefaultHeaders(),
+        })
+        .then(x => x.data.user);
```
christianalfoni: Remove the `.then` and rather....

packages/app/src/app/overmind/effects/zeit.ts
```diff
@@ -87,16 +87,20 @@ export default (() => {
...
 
-      return response.data;
+      return response;
```
christianalfoni:
```ts
return response.data.user
```

packages/app/src/app/overmind/effects/zeit.ts
```diff
@@ -87,16 +87,20 @@ export default (() => {
...
     },
     async deleteDeployment(id: string) {
-      return axios.delete(`https://api.zeit.co/v9/now/deployments/${id}`, {
-        headers: getDefaultHeaders(),
-      });
+      return axios
```
christianalfoni: Same here:

```ts
const response = await axios.delete<{ deployments: Deployment[] }>(...)

return response.data.deployments
```

packages/app/src/app/overmind/effects/zeit.ts
```diff
@@ -91,12 +91,16 @@ export default (() => {
...
-      return axios.delete(`https://api.zeit.co/v9/now/deployments/${id}`, {
-        headers: getDefaultHeaders(),
-      });
+      const response = await axios.delete(
+        `https://api.zeit.co/v9/now/deployments/${id}`,
+        {
+          headers: getDefaultHeaders(),
+        }
+      );
+      return response.data.deployments;
```
christianalfoni: I believe this now returns `any`. Should either:

```ts
const response = await axios.delete<{ deployments: Deployment[] }>(...)
```

Sorry, should have stated it in previous review ;-) You could also:

```ts
async deleteDeployment(id: string): Promise<Deployment[]> {
```

PR #2498, opened 2019-09-24
🔨 Switch CodeSadbox to useOvermind, Remove title, add label to Report

Saeris:
Small refactor to make the switch to Overmind.

Removed the title from the generated crash report, which will now require users to enter their own in order to be able to submit the report. Hopefully this makes crash reports more descriptive. Crash Reports now include the proper label automatically when they are generated.

@christianalfoni there appears to be an issue with Overmind in this component, as isLoggedIn returns false even when the user is in fact logged in (test this by running locally and navigating to http://localhost:3000/codesadbox from another page). Not quite sure how to solve this one, so I'd appreciate it if you can investigate! Probably has something to do with the error boundary.

### DIFF
--- a/packages/app/src/app/pages/common/ErrorBoundary/CodeSadbox/CodeSadbox.tsx
+++ b/packages/app/src/app/pages/common/ErrorBoundary/CodeSadbox/CodeSadbox.tsx
@@ -1,4 +1,3 @@
-// @ts-ignore
 import { Button } from '@codesandbox/common/lib/components/Button';
 import { dashboardUrl } from '@codesandbox/common/lib/utils/url-generator';
 import { useOvermind } from 'app/overmind';
@@ -6,6 +5,7 @@ import { Navigation } from 'app/pages/common/Navigation';
 import React, { useEffect } from 'react';
 import GoHome from 'react-icons/lib/go/home';
 import GoIssueOpened from 'react-icons/lib/go/issue-opened';
+// @ts-ignore
 import Dashboard from '-!svg-react-loader!@codesandbox/common/lib/icons/dashboard.svg';
 
 import { IFallbackComponentProps } from '../types';
--- a/packages/app/src/app/pages/common/ErrorBoundary/CodeSadbox/CodeSadbox.tsx
+++ b/packages/app/src/app/pages/common/ErrorBoundary/CodeSadbox/CodeSadbox.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import GoHome from 'react-icons/lib/go/home';
 import GoIssueOpened from 'react-icons/lib/go/issue-opened';
-import { inject, hooksObserver } from 'app/componentConnectors';
+import { useOvermind } from 'app/overmind';
 import { Button } from '@codesandbox/common/lib/components/Button';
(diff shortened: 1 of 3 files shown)


christianalfoni:
> Looks good! One question though: wouldn't it be better to put this onMount action on the error boundary component itself? Since this is just the default fallback component and we may use error boundaries elsewhere. Maybe not necessary, but worth considering.

Conceptually it makes sense :) Though technically it can not be defined there. Has to be defined on the Overmind configuration as that is where all the state and effects lives. The `withLoadApp` factory is like an "HOC" for the action, making sure that the app is initialized with a user etc. But yeah, the ideal goal here is to have something like:

```
effects.router.initialize({
  '/s/:sandboxId': actions.routeSandbox,
  '/search': actions.routeSearch,
  '/dashboard': actions.routeDashboard,
  ...
})
``` 

Now Overmind is hit on all the routes preparing the user and whatever data that is needed for the specific route. The components just render based on the current state of the app, as they do related to all other interactions with the app.

That means the error boundary becomes very dumb, same for all components that now on their "mount" does "actions.thisPageMounted" and "actions.thatPageMounted". We can ensure a user before even rendering anything. Stop having these weird flips on the header where you see you are logged out and then suddenly logged in etc. And lots of other benefits as the routing now just becomes an event in the application like any other event :)

christianalfoni:
Sorry, I misunderstood the question on my deleted response :) 

Yeah, I initially looked at that, but it is a class and needs the `connect` HOC. Would love to avoid using that, also I got some really weird typing errors I did not want to ignore or dive into. React HOC typing history is extremely bad :) 

But after putting it into "codesadbox" I think it makes more sense from the perspective of it being an actual route. The loading of the initial data and user is related to hitting routes, so if we later refactor to a "state first routing" approach this will still work. The `ErrorBoundary` is kept dumb and also the `codesadBox` component will just remove its effect and kept dumb as well :)

packages/app/src/app/pages/common/ErrorBoundary/CodeSadbox/CodeSadbox.tsx
```diff
@@ -1,71 +1,82 @@
...
-            )}
-            {/*
-          // @ts-ignore */}
-            <Button
-              small
-              target="_blank"
-              rel="noopener"
-              href={buildCrashReport({ error, trace })}
-            >
+export const CodeSadbox: React.FC<IFallbackComponentProps> = ({
```
MichaelDeBoey: I think using `FunctionComponent` would make the type a bit more clear.

```suggestion
export const CodeSadbox: React.FunctionComponent<IFallbackComponentProps> = ({
```

christianalfoni: Yeah, I am a bit split on this. I started using `FunctionComponent`, but went for `FC` due to lazyness. Totally agree it is more explicit, though this type is not really what tells us this is a component. It is the folder name, file name and not at least the JSX in the file. We would never like... "what is this thing? Oh, let me check the type... hm React.FC... hmmmm... oh, it is a component", you know long before you look at the type that it is a component 😄

Issue #2497, opened 2019-09-23
💥 Crash Report: <Please Add a Short Description of Crash Circumstances>


BradenLawrence:
<h1>💥 Crash Report</h1>

<h2>What were you trying to accomplish when the crash occurred?</h2>

> I was attempting to log in for the first time. From the homepage, I clicked on "sign in". I expected this to bring me to a login page where I could sign up or link my github account. Instead, it brought me to a code editor window with several error notifications in the bottom right. I didn't grab a screenshot in time, but they were all related to not being logged in. 

> I then attempted to sign in using my github account. After clicking the button to link my account, it brought me to this crash page.

> After this I tried to recreate the error by clicking the "sign in" button from the homepage again, but each time it simply brought me to a standard login page. Hope this helps!

<h3>Crash Details</h3>

<details>
<summary>Environment</summary>

| Browser |  Version  | Operating System |
| ------- | --------- | ---------------- |
| firefox | 69.0.0 | Ubuntu           |

**Route:**
https://codesandbox.io/dashboard/recent

</details>

<details>
<summary>Error Message</summary>

```bash
TypeError: a.me is undefined
value/</</<@https://codesandbox.io/static/js/app.c23775ad.js:438:183077
n@https://codesandbox.io/static/js/vendors~app.ab650e77.chunk.js:1:753765
../../node_modules/react-apollo/react-apollo.esm.js/C</t.prototype.render@https://codesandbox.io/static/js/vendors~app.ab650e77.chunk.js:1:753863
vs@https://codesandbox.io/static/js/common.8fd6ed86.chunk.js:1:2062663
ms@https://codesandbox.io/static/js/common.8fd6ed86.chunk.js:1:2062456
Ka@https://codesandbox.io/static/js/common.8fd6ed86.chunk.js:1:2096852
Fa@https://codesandbox.io/static/js/common.8fd6ed86.chunk.js:1:2081218
Na@https://codesandbox.io/static/js/common.8fd6ed86.chunk.js:1:2078238
Ra@https://codesandbox.io/static/js/common.8fd6ed86.chunk.js:1:2076798
po/<@https://codesandbox.io/static/js/common.8fd6ed86.chunk.js:1:2038699
../../node_modules/react-dom/node_modules/scheduler/cjs/scheduler.production.min.js/t.unstable_runWithPriority@https://codesandbox.io/static/js/common.8fd6ed86.chunk.js:1:2113545
[...]

PR #2499, opened 2019-09-24
fix: data.me is undefined error

jyash97:

Bug Fix

Some user reported these issues recently in #2493, #2495 , #2496 , #2497 . All the issues are related to the same component if we follow the stack trace. Though, I was not able to replicate the issue though.

## What is the new behavior?
Added a check which might help in these case.

tested locally it was working the same for me.

## Checklist

- [ ] Documentation
- [x] Testing 
- [ ] Ready to be merged  


jyash97:
Right @christianalfoni  I am about to create a PR for that fix it used to work previously but recently when we log out we are on the same page. I checked the history of a few pages to know what was changed but couldn't find it.

PR #2509, opened 2019-09-25
should hard reload on signout to avoid bad states

christianalfoni:


christianalfoni:
Hm, good question! I believe this has been there all along. We did do a general refactor, but the "signout" logic is very simple so there was no change there. So I think it is just introduction of new features that has increased the likelihood of getting into a bad state when signing out :)

PR #2510, opened 2019-09-25
moved instantiation of vscode to effect, written components, reviewed

christianalfoni:
This is a refactor which prepares for putting in the new VSCode. we can actually make this part of master to save some work later. Basically this is what I did:

- Moved instantiation of BrowserFS and VSCodeManager to the `vscode` effect
- Made sure no components are referencing the `manager` directly, except for **VSCode** and **MonacoVSCodeComponent** as these will be replaced later anyways
- I reviewed what is happening in those components and suggested an API of the `vscode` effect, with notes on what needs to be moved into that effect

Hopefully this helps draw the picture further on how this new VSCode concept should work :)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/2510/commits/f4b584d07c43e199f25b104e6c129a2b9dc4169f is at https://pr2510.build.csb.dev/s/new.

christianalfoni:
So the following should be tested:

- [ ] Open Sandbox
- [ ] Fork Sandbox
- [ ] Open Sandbox from Dashboard
- [ ] Open Sandbox from profile
- [ ] Create file
- [ ] Change content of file
- [ ] Save files
- [ ] Close tabs
- [ ] Delete file
- [ ] Create folder
- [ ] Delete folder
- [ ] Move files and folders
- [ ] Verify preview updates

Like... everything should be tested

christianalfoni:
ops

lbogdan:
[Embeds](https://pr2510.build.csb.dev/embed/new) fail with

image

christianalfoni:
I have tested the following:

- [ ] Open sandbox
- [ ] Fork sandbox
- [ ] Change code and verify preview updates
- [ ] Save code and verify preview updates
- [ ] Change code and refresh (keeps code)
- [ ] Create file, change code and save (refresh)
- [ ] Create folder (refresh)
- [ ] Move file, change code and save (refresh)
- [ ] Move directory, change code in a file and save (refresh)
- [ ] Delete file
- [ ] Delete directory with files in it
- [ ] Verify prettify works
- [ ] Verify linting works
- [ ] Verify module imports/exports are updated
- [ ] Verify typescript runs
- [ ] Verify added typings are updated
- [ ] Test LIVE

packages/app/src/app/overmind/effects/preview.ts
```diff
@@ -0,0 +1,52 @@
+import { blocker } from 'app/utils/blocker';
+
+import { Reaction } from '..';
+
+let _preview = blocker<any>();
```
CompuIves: Very clever!! Can we type this to the component?

packages/app/src/app/overmind/effects/preview.ts
```diff
@@ -0,0 +1,52 @@
+import { blocker } from 'app/utils/blocker';
```
CompuIves: In the future it would be nice if we could incorporate `@codesandbox/executors` in this. This was my goal with that project, that we'd have a single interface to send updates to files to, and the `executors` package would either let the browser know for the client sandboxes or the server for contianer sandboxes. This way we abstract that difference away.

packages/app/src/app/overmind/effects/vscode/sandboxFsSync/index.ts
```diff
@@ -0,0 +1,357 @@
...
+          ...m,
+          type: 'file',
+        };
+      }
+    });
+
+    sandbox.directories.forEach(d => {
+      const path = getDirectoryPath(sandbox.modules, sandbox.directories, d.id);
+
+      d.path = path;
```
CompuIves: So we mutate the directory here on the sandbox? 🤔

packages/app/src/app/overmind/effects/vscode/sandboxFsSync/index.ts
```diff
@@ -0,0 +1,357 @@
...
+  public mkdir(fs: SandboxFs, directory: Directory) {
+    const copy = json(directory);
+
+    mkdir(fs, copy);
+    this.send('mkdir', copy);
+  }
+
+  private syncSandbox() {
+    // eslint-disable-next-line
+    console.log('## SYNCING SANDBOX AND TYPINGS WITH WORKERS');
```
CompuIves: Will need to remove this one later when we merge 😄

packages/app/src/app/overmind/effects/vscode/sandboxFsSync/index.ts
```diff
@@ -0,0 +1,357 @@
...
+    this.send('sandbox-fs', json(this.options.getSandboxFs()));
```
CompuIves: Maybe we should rename this to something different. It's not really the sandbox fs (I expect to get an interface with `readFile` etc), but rather a representation. Maybe we should rename this to something like `sandboxFiles`?

packages/app/src/app/overmind/effects/vscode/sandboxFsSync/utils.ts
```diff
@@ -0,0 +1,34 @@
+import { Directory, Module, SandboxFs } from '@codesandbox/common/lib/types';
+import { json } from 'overmind';
+
+export const writeFile = (fs: SandboxFs, module: Module) => {
+  fs[module.path] = module;
+};
+
+export const rename = (fs: SandboxFs, fromPath: string, toPath: string) => {
+  Object.keys(fs).forEach(path => {
+    if (path.startsWith(fromPath)) {
```
CompuIves: This check wouldn't work in cases like this:

Let's say I rename `/test/a` to `/test/b`. In this case if there's a file called `/test/another.js` it would be renamed as well, to `/test/bnother.js`. We can fix it by changing the check to see if `path.startsWith(query + '/)` or `path.startsWith(query + <end of line>)`.

packages/app/src/app/overmind/effects/vscode/sandboxFsSync/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+
+      delete fs[path];
+      fs[newPath] = json(module);
+    }
+  });
+};
+
+export const rmdir = (fs: SandboxFs, directory: Directory) => {
+  Object.keys(fs).forEach(path => {
+    if (path.startsWith(directory.path)) {
```
CompuIves: Same as the comment above.

packages/app/src/app/overmind/effects/vscode/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+    return undefined;
+  }
+
+  return model.uri.path.replace(/^\/sandbox/, '');
+}
+
+export function getCode(editor) {
+  const activeEditor = editor.getActiveCodeEditor();
+
+  if (!activeEditor) return '';
```
CompuIves: Maybe makes sense to return `null` here? Unless we used to return `''` as well, then it's probably better to return the same for compat reasons.

packages/app/src/app/overmind/internalActions.ts
```diff
@@ -286,13 +279,6 @@ export const setCurrentSandbox: AsyncAction<Sandbox> = async (
   });
 
   effects.executor.setupExecutor();
-
-  /*
-    There seems to be a race condition here? Verify if this still happens with Overmind
-  */
-  if (oldSandboxId !== state.editor.currentId) {
-    delete state.editor.sandboxes[oldSandboxId];
```
CompuIves: Did we verify this? I think the case is when you press the back button after navigating from sandbox to sandbox we don't refetch the sandbox info, or we don't join the team live session.

packages/app/src/app/overmind/namespaces/editor/actions.ts
```diff
@@ -72,23 +78,18 @@ export const sandboxChanged: AsyncAction<{ id: string }> = withLoadApp<{
...
 
-  if (state.live.isLive) {
-    actions.live.internal.disconnect();
+  // This happens when we fork. This can be avoided with state first routing
+  if (state.editor.isForkingSandbox) {
+    return;
   }
 
-  if (state.editor.sandboxes[newId] && !state.editor.sandboxes[newId].team) {
-    const sandbox = await effects.api.getSandbox(newId);
```
CompuIves: When opening a team sandbox we automatically open a live session, we need to make sure this still happens. I'll test it later.

CompuIves: Wow this works really well, nice work!

packages/app/src/app/overmind/namespaces/editor/actions.ts
```diff
@@ -98,11 +99,14 @@ export const sandboxChanged: AsyncAction<{ id: string }> = withLoadApp<{
   try {
     const sandbox = await effects.api.getSandbox(newId);
 
-    if (state.editor.sandboxes[newId]) {
-      actions.internal.updateCurrentSandbox(sandbox);
-    } else {
-      actions.internal.setCurrentSandbox(sandbox);
-    }
+    await effects.vscode.closeAllTabs();
```
CompuIves: I love this.

packages/app/src/app/overmind/namespaces/editor/actions.ts
```diff
@@ -178,16 +207,20 @@ export const codeChanged: Action<{
     return;
   }
 
-  if (state.live.isLive && !noLive) {
-    state.live.receivingCode = true;
-    effects.live.sendCodeUpdate(moduleShortid, module.code || '', code);
-    state.live.receivingCode = false;
+  if (state.live.isLive) {
+    effects.live.sendCodeUpdate(moduleShortid, module.code, event);
```
CompuIves: Interesting that we did a `receivingCode` here before. I don't think it's needed while reading this code.

packages/app/src/app/overmind/namespaces/editor/actions.ts
```diff
@@ -178,16 +207,20 @@ export const codeChanged: Action<{
...
   }
 
   actions.editor.internal.setModuleCode({
     module,
     code,
   });
+
+  const { isServer } = getTemplate(state.editor.currentSandbox.template);
+
+  if (!isServer && state.preferences.settings.livePreviewEnabled) {
```
CompuIves: Nice, glad the check is still here. Good eye for detail!

packages/app/src/app/overmind/namespaces/editor/actions.ts
```diff
@@ -293,11 +330,15 @@ export const moduleSelected: Action<{
       );
     }
 
+    if (module.shortid === state.editor.currentModuleShortid) {
```
CompuIves: module can probably be null here?

packages/app/src/app/overmind/namespaces/editor/actions.ts
```diff
@@ -316,9 +357,14 @@ export const moduleSelected: Action<{
...
 
       effects.live.sendUserCurrentModule(module.shortid);
+
+      if (!state.editor.isInProjectView) {
+        actions.editor.internal.updatePreviewCode();
+      }
     }
   } catch (error) {
-    state.editor.currentModuleShortid = null;
+    // Do nothing, it is most likely VSCode selecting a file
```
CompuIves: Haha, I love this text. I do think this is the cause for not unselecting a file when you open a file that's not a module (try CMD+clicking) op `react-dom` from the editor.

packages/app/src/app/overmind/namespaces/editor/actions.ts
```diff
@@ -636,7 +705,6 @@ export const previewActionReceived: Action<{
       actions.editor.addNpmDependency({
         name,
       });
-      actions.forceRender();
```
CompuIves: No need?

packages/app/src/app/overmind/namespaces/editor/internalActions.ts
```diff
@@ -338,7 +322,27 @@ export const forkSandbox: AsyncAction<{
       });
     }
 
-    await actions.internal.setCurrentSandbox(forkedSandbox);
+    // When we have a server we want to set a brand new sandbox,
+    // as the preview also needs to be managed correctly
+    if (getTemplateDefinition(forkedSandbox.template).isServer) {
+      await effects.vscode.closeAllTabs();
+      actions.internal.setCurrentSandbox(forkedSandbox);
+      state.editor.modulesByPath = effects.vscode.fs.create(forkedSandbox);
```
CompuIves: We can put thisline out of the if statement (happens in both)

packages/app/src/app/overmind/namespaces/editor/internalActions.ts
```diff
@@ -338,7 +322,27 @@ export const forkSandbox: AsyncAction<{
       });
     }
 
-    await actions.internal.setCurrentSandbox(forkedSandbox);
+    // When we have a server we want to set a brand new sandbox,
+    // as the preview also needs to be managed correctly
+    if (getTemplateDefinition(forkedSandbox.template).isServer) {
+      await effects.vscode.closeAllTabs();
```
CompuIves: Is this really needed for servers?

packages/app/src/app/overmind/namespaces/editor/internalActions.ts
```diff
@@ -338,7 +322,27 @@ export const forkSandbox: AsyncAction<{
...
+      actions.internal.setCurrentSandbox(forkedSandbox);
```
CompuIves: It's interesting, from a first glance this seems to imply that some more is happening in this function. Might be a good thing in the future to rename this or something, not now though 😄

yeion7:
@christianalfoni any chance of make a video (doc, etc) about this migration? I think this unlock a lot of cool thing for codesandbox and I would like to know better how works

christianalfoni:
> @christianalfoni any chance of make a video (doc, etc) about this migration? I think this unlock a lot of cool thing for codesandbox and I would like to know better how works

Would love to do that! We are thinking a lot about how to target the "contribution" audience and really want to find a good way to create a developer community around Codesandbox 😄 

I will try to create some sort of summary of this PR and some plans ahead and share it 👍

packages/app/src/app/overmind/effects/vscode/utils.ts
christianalfoni: It is actually no being used for anything, removed it 😄

packages/app/src/app/overmind/internalActions.ts
christianalfoni: Whenever we change Sandbox now it runs the exact same logic. It downloads it, sets it, restarts live etc. There is no difference it being there or not. So yeah, I do not see why we need this logic 😄

packages/app/src/app/overmind/namespaces/editor/actions.ts
christianalfoni: Yeah, this is old stuff, I will remove it 😄

packages/app/src/app/overmind/namespaces/editor/actions.ts
christianalfoni: Hm, if it is `null` here I would state it is a bug. Cause with `path` (from VSCode) it will throw an error above and it is caught. With an `id` it is actually you selecting the file in the explorer, so it has to be there

packages/app/src/app/overmind/namespaces/editor/actions.ts
christianalfoni: Hehe, yeah, this is actually changed to:

```
// You jumped to a file not in the Sandbox, for example typings
```
now 😂

packages/app/src/app/overmind/namespaces/editor/actions.ts
christianalfoni: Nopes `forceRender` was used to trigger the preview, but now trigger preview imperatively through the `effects.preview` API 😄

packages/app/src/app/overmind/namespaces/editor/internalActions.ts
christianalfoni: BElieve this is outdated?

packages/app/src/app/overmind/namespaces/editor/internalActions.ts
christianalfoni: Oh yeah, fixed this now. Does not seem we need it! 👏

packages/app/src/app/overmind/namespaces/editor/internalActions.ts
christianalfoni: Yeah, lets save this for the refactor after components are ready. I will refactor the namespaces and the actions 😄

PR #2511, opened 2019-09-25
Cleanup homepage templates

MichaelDeBoey:
This is the follow-up of #2420, where I've make the `docs` template a function component and put all styles in their own `.elements` file.
 
- `docs` is a class component
- most styles are in the templates themselves

## What is the new behavior?
- `docs` is a function component
- all styles are in their own `.elements` file.

Check if the pages still look the same

## Checklist
- [ ] Documentation N/A
- [x] Testing
- [x] Ready to be merged


MichaelDeBoey:
Always a pleasure 🙂

Issue #2501, opened 2019-09-24
Design issue: Edit menu drags down itself where it is unaccessible


meetzaveri:
# 🐛 bug report

## Description of the problem
In last row of my sandbox screen, when I click on any sandbox's edit button and hover on it, then it just dragged down to the screen to be cut off half from the whole sandbox screen 

## How has this issue affected you? What are you trying to accomplish?
I was trying to perform action of move to trash, but when I hover it just moved/dragged down itself forcefully

### Demo
![chrome-capture](https://user-images.githubusercontent.com/21293048/65501455-a989d800-dede-11e9-81ef-c654b3800c4e.gif)

### Your Environment
OS : Ubuntu | 18.04
Browser : Chrome

PR #2512, opened 2019-09-25
show popover

yeion7:

Bugfix

When the transition ends the height and width is `auto`, because of this we don't have the menu measurements

#2501

## What is the new behavior?

This is a quick fix, I think this requires a better refactor of `ContextMenu`

before:

fTs0mvC5Os

after:

l4Et5IP3Hz

## Checklist

- [ ] Documentation
- [ ] Testing 
- [x] Ready to be merged 


SaraVieira:
This is awesome! Thank you!

PR #2523, opened 2019-09-26
🔨 Switch GithubIntegration to use useOvermind

MichaelDeBoey:


MichaelDeBoey:
@christianalfoni `signInGithubClicked` expect to get an argument, but got `{ useExtraScopes: true }` as argument.

Should I delete the argument or should I update the type? 🤔

packages/app/src/app/pages/common/GithubIntegration/index.tsx
```diff
@@ -1,36 +1,36 @@
...
+  } = useOvermind();
+
+  return (
+    <Integration
+      bgColor="#4078c0"
+      description={small ? 'Commits & PRs' : 'Committing & Pull Requests'}
+      Icon={GitHubLogo}
+      loading={isLoadingGithub}
+      name="GitHub"
+      onSignIn={() => signInGithubClicked({ useExtraScopes: true })}
```
MichaelDeBoey: @christianalfoni `signInGithubClicked` expect to get an argument, but got `{ useExtraScopes: true }` as argument.

Should I delete the argument or should I update the type? 🤔

MichaelDeBoey: Looks like `useExtraScopes` isn't used (it's always passed with a `true` value), so will update it here.
https://github.com/codesandbox/codesandbox-client/blob/cb965d440eb569d569e97ded524d54d142f241ad/packages/app/src/app/overmind/actions.ts#L163-L167

Came across this one too btw, don't know it that one's still used? 🤔
https://github.com/codesandbox/codesandbox-client/blob/cb965d440eb569d569e97ded524d54d142f241ad/packages/app/src/app/store/actions.js#L290-L308

PR #2525, opened 2019-09-26
🔨 Switch TeamInvite to use useOvermind

MichaelDeBoey:


packages/app/src/app/pages/common/Navigation/Notifications/notifications/TeamInvite.tsx
```diff
@@ -1,86 +1,90 @@
...
+        <Buttons>
+          <Mutation
+            variables={{ teamId }}
+            mutation={REJECT_TEAM_INVITATION}
+            refetchQueries={['RecentNotifications']}
+            onCompleted={() => {
+              track('Team - Invitation Rejected');
+              notificationAdded({
+                message: `Rejected invitation to ${teamName}`,
+                type: 'success',
```
MichaelDeBoey: @christianalfoni It seems like `message` is renamed to `title`, `type` is renamed to `notificationType` _**and**_ a value (`number`) for `timeAlive` needs to be provided.
https://github.com/codesandbox/codesandbox-client/blob/cb965d440eb569d569e97ded524d54d142f241ad/packages/app/src/app/overmind/actions.ts#L28-L38

Any advice what I should do with the `timeAlive` value? 🤔
Should I set a default value or should I provide a value for them? 🤔

christianalfoni: Actually, since we now have more control I would suggest adding a new action called: `rejectTeamInvitation`. This action does the tracking and uses the `effects.notificationsToast.success(...)` instead. This cleans up the component a bit and we see more of what is going on when this scenario occurs :)

But yeah, would be awesome if you do a search for the usage of this `notificationAdded`, it is really deprecated. But yeah, for this specific scenario... new action :)

packages/app/src/app/overmind/actions.ts
```diff
@@ -224,3 +224,21 @@ export const refetchSandboxInfo: AsyncAction = async ({
     await actions.editor.internal.initializeLiveSandbox(sandbox);
   }
 };
+
+export const acceptTeamInvitation: Action<{ teamName: string }> = (
+  { effects },
+  { teamName }
+) => {
+  effects.analytics.track('Team - Invitation Accepted', {});
```
MichaelDeBoey: @christianalfoni I could use `actions.track` instead? 🤔

christianalfoni: This `track` thing is all over the place currently 😂 It is exposed as an action, it is imported directly into components and it is an effect.

We ideally want this to only be used as an effect. That way we later get the freedom to do things like `withTrack('trackThis', myActionDefinition)`. So kinda like how you would use HOC in components to give functionality, the same is possible with actions in Overmind. 

So yeah, go for effect and then we try to clean up the other stuff :)

MichaelDeBoey: Then this one's ready to be merged 🙂

PR #2529, opened 2019-09-26
🔨 Switch SignInButton to use useOvermind

MichaelDeBoey:


packages/app/src/app/pages/common/SignInButton/index.tsx
```diff
@@ -1,20 +1,21 @@
...
+
+type Props = Omit<ComponentProps<typeof Button>, 'onClick' | 'small'>;
+export const SignInButton: FunctionComponent<Props> = props => {
+  const {
+    actions: { signInClicked },
+  } = useOvermind();
 
   return (
-    <Button small onClick={signInClicked} {...props}>
+    <Button {...props} small onClick={signInClicked}>
```
MichaelDeBoey: @christianalfoni `signInClicked` needs an object with an `useExtraScopes` value (`boolean`).
https://github.com/codesandbox/codesandbox-client/blob/cb965d440eb569d569e97ded524d54d142f241ad/packages/app/src/app/overmind/actions.ts#L81-L84

I can either provide a `false` value _or_ I can make  the `options` have a default-value of empty object (`{}`) in the action.
What's the preferred one? 🤔

christianalfoni: Oh, hm, since there is a payload typed it will be required even though you set a default payload object in the action. Just the way the typing works. So I would say that you explicitly send in `false` :)

MichaelDeBoey: Also if I make `useExtraScopes`? 🤔
Then I'm required to at least pass an empty object to as payload for `signInClicked`? 🤔

PR #2530, opened 2019-09-26
🔨 Switch NewSandboxModal to use useOvermind

MichaelDeBoey:


SaraVieira:
Hey!

This looks good but we are now working on the new modal so it doesn't really make sense to merge it as it will all be replaced and then I would cry over the conflicts :(


Sorry

MichaelDeBoey:
@SaraVieira No problem, I'm with you 😉

PR #2532, opened 2019-09-26
🔨 Switch UserMenu to use useOvermind

MichaelDeBoey:


packages/app/src/app/pages/common/UserMenu/index.tsx
```diff
@@ -1,44 +1,51 @@
...
-          <ProfileImage
-            alt={user.username}
-            width={30}
-            height={30}
-            src={user.avatarUrl}
+      {userMenuOpen && (
+        <HoverMenu onClose={() => userMenuClosed()}>
+          <Menu
+            openPreferences={() =>
+              modalOpened({ message: null, modal: 'preferences' })
```
MichaelDeBoey: @christianalfoni Just like https://github.com/codesandbox/codesandbox-client/pull/2544#discussion_r328837587

Maybe make `message` optional? 🤔

packages/app/src/app/pages/common/UserMenu/index.tsx
```diff
@@ -1,44 +1,51 @@
...
+            }
+            openStorageManagement={() => gotUploadedFiles()}
+            signOut={() => signOutClicked()}
+            username={user.username}
+            curator={user.curatorAt}
+            openFeedback={() =>
+              modalOpened({ message: null, modal: 'feedback' })
```
MichaelDeBoey: @christianalfoni Just like https://github.com/codesandbox/codesandbox-client/pull/2544#discussion_r328837587

Maybe make `message` optional? 🤔

packages/app/src/app/pages/common/UserMenu/index.tsx
```diff
@@ -1,44 +1,51 @@
...
-            width={30}
-            height={30}
-            src={user.avatarUrl}
+      {userMenuOpen && (
+        <HoverMenu onClose={() => userMenuClosed()}>
+          <Menu
+            openPreferences={() =>
+              modalOpened({ message: null, modal: 'preferences' })
+            }
+            openStorageManagement={() => gotUploadedFiles()}
```
MichaelDeBoey: @christianalfoni `gotUploadedFiles` needs an object with a `message` key.
Do I also provide a `null` or do you have a more descriptive one in mind? 🤔

packages/app/src/app/pages/common/UserMenu/index.tsx
christianalfoni: Yeah, we should make that optional and also the "modal" should be typed explicitly with types of modals you can open, now I believe it is a string? Would have to search codebase to find all possible modals though. 

Can maybe do that in a new PR? :)

packages/app/src/app/pages/common/UserMenu/index.tsx
christianalfoni: The `gotUploadedFiles` is a bit tricky in terms of typing. Cause with `Action` and `Action<string>` we are able to say... "no argument" or "argument of this type". Also we want to refactor all these `{ singleValue }`  payloads to plain `singleValue`, so I think you can just do that here now:

```ts
export const gotUploadedFiles: AsyncAction<{
  message: string;
}> 
```

To:

```ts
export const gotUploadedFiles: AsyncAction<string | void>
```

`void` in Action/AsyncAction means "no argument". I am not sure if it works correctly, cause we evaluate that typing to create the correct signature for the callable action. If it does not work I would just do:

```ts
export const gotUploadedFiles: AsyncAction<string>
```

And pass in `null` as argument. Just make sure any other use of the action passes their values correctly :) 

Does it make sense? :)

packages/app/src/app/pages/common/UserMenu/index.tsx
christianalfoni: We made it optional here: https://github.com/codesandbox/codesandbox-client/pull/2553 , but yeah, the explicit typing of modal types can still be a new PR :)

packages/app/src/app/pages/common/UserMenu/index.tsx
SaraVieira: done :)

packages/app/src/app/pages/common/UserMenu/index.tsx
MichaelDeBoey: I'll make a new PR for the `modal` names 😉

packages/app/src/app/pages/common/UserMenu/index.tsx
MichaelDeBoey: Makes sense to me. 🙂

If no argument at all, I can just use `Action` or `AsyncAction` right? 🤔

MichaelDeBoey: Since we're setting `state.currentModalMessage`, the message of `gotUploadedFiles` is required.

https://github.com/codesandbox/codesandbox-client/blob/c22379febc3bbace5539b81baa070bdba50cbdec/packages/app/src/app/overmind/namespaces/files/actions.ts#L220-L237

packages/app/src/app/pages/common/UserMenu/index.tsx
MichaelDeBoey: PR created: https://github.com/codesandbox/codesandbox-client/pull/2554

PR #2548, opened 2019-09-26
Moreperf

christianalfoni:
Okay, excited about this update!

1. The `modulesByPath` is now only tracking properties that actually affects its path. This is a huge boost in itself as this derived became dirty on any code change, always been like this actually
2. I completely removed the "1 second" sync of files. Instead we have a reaction looking at `modulesByPath`. Whenever it changes, a sync is triggered
3. I found a bug deep inside Overmind related to derived, so that was also fixed. Is is a pretty extreme edge case, but the `modulesByPath` was affected by it
4. Cleaned some typing stuff

I also found some weird logic on templates, as you can see in the change. It was treating everything from `modulesByPath` as actual modules, but there are directories there as well. So yeah, can just test it, but it is more specific now.


christianalfoni:
No idea why `build:deps` fails here as there is a @ts-ignore on that wrong typing. Could not figure out where that method was defined

packages/app/src/app/overmind/namespaces/editor/state.ts
```diff
@@ -158,7 +168,14 @@ export const state: State = {
 
       // If this is a single directory with no children
       if (!Object.keys(modulesObject).some(p => dirname(p) === path)) {
-        modulesObject[path] = { ...d, type: 'directory' };
+        // We need to track title and directoryShortid due to
+        // that affectings its path
+        modulesObject[path] = {
```
CompuIves: I believe that BrowserFS also uses `updatedAt` and `createdAt` to set the `mtime` and created time. `mtime` is used by the file watcher of VSCode.

christianalfoni:
We did another iteration on new pr, so closing this

PR #2554, opened 2019-09-27
Restrict modal opened action

MichaelDeBoey:
As requested by @christianalfoni in https://github.com/codesandbox/codesandbox-client/pull/2532#discussion_r328996362

Not sure if I should extract the new `ModalName` type into `@codesandbox/common` so I just kept it in `actions.ts` for now.


SaraVieira:
It looks good here and then if it gets too big we can move it but i don't think it will :)

PR #2562, opened 2019-09-27
🔨 Switch SandboxConfig to use useOvermind

MichaelDeBoey:


packages/app/src/app/pages/Sandbox/Editor/Workspace/Project/SandboxConfig/SandboxConfig.tsx
```diff
@@ -1,79 +1,81 @@
...
-          modalOpened({ modal: 'deleteSandbox' });
-        }
-      };
+    addedTemplate({
+      template: {
+        color:
+          (customTemplate && customTemplate.color) ||
+          templates.default(template).color(),
+        title,
+        description,
```
MichaelDeBoey: @christianalfoni `addedTemplate` needs a `CustomTemplate`, but that doesn't have  a `description`.

https://github.com/codesandbox/codesandbox-client/blob/3b2ca3f5e03388779287f279020db821edac4a3b/packages/common/src/types/index.ts#L119-L125

Should I add it or delete it here? 🤔

MichaelDeBoey: `id` & `url` are also missing, so maybe change the type of `addedTemplate`'s `template`? 🤔

https://github.com/codesandbox/codesandbox-client/blob/400bf0a1d0e41f481a519cd1a1f5568a46b24c3c/packages/app/src/app/overmind/namespaces/workspace/actions.ts#L243-L262

christianalfoni: Hm, the templating typing has been confusing me a bit. I would log out the template and just look at the data. I think we are also confusing `CustomTemplate` with `Template` a bit in the code. But yeah, this I believe to be `CustomTemplate`. 

@SaraVieira You have comments on this? 😄

SaraVieira: Yeah, this is the CustomTemplate, the description is not mandatory

There will probably some issues with the ts on this because it was right when I started learning ts :/

MichaelDeBoey: @SaraVieira If I could figure out what the `` POST endpoint is accepting, that could maybe help me out.
https://github.com/codesandbox/codesandbox-client/blob/2f6577e466aac8d7d3781817810b2c6c2b9fa9d0/packages/app/src/app/overmind/effects/api/index.ts#L432-L441

Another thing I could do is adding an optional `description` to `CustomTemplate` and make `id` & `url` optional too.

Like I said:
- `CustomTemplate` doesn't have a `description`, which is provided
- `CustomTemplate` has a mandatory `id` & `url`, which aren't provided

MichaelDeBoey: I think we should create a new `CustomTemplateToCreate` type, since I think a `CustomTemplate` should have an `id`, but that would be set by the API itself, just like the `url` I think? 🤔

PR #2566, opened 2019-09-27
🔨 Switch Author to use useOvermind

MichaelDeBoey:


packages/app/src/app/pages/Sandbox/Editor/Workspace/Project/Author.tsx
```diff
@@ -1,9 +1,16 @@
...
+import React, { FunctionComponent } from 'react';
+
 import { Item, UserLink } from './elements';
 
-export const Author = ({ author: { username, avatarUrl, subscription } }) => (
+type Props = {
+  author: User;
+};
+export const Author: FunctionComponent<Props> = ({
+  author: { username, avatarUrl, subscription },
```
MichaelDeBoey: @christianalfoni `subscription` doesn't exist on type `User`
https://github.com/codesandbox/codesandbox-client/blob/cb965d440eb569d569e97ded524d54d142f241ad/packages/common/src/types/index.ts#L154-L172

I guess adding it is the way to go? 🤔

SaraVieira: yeah! Thats the one that exists I think

PR #2567, opened 2019-09-27
🔨 Switch Keywords to use useOvermind

MichaelDeBoey:


packages/app/src/app/pages/Sandbox/Editor/Workspace/Project/Keywords.tsx
```diff
@@ -1,77 +1,74 @@
...
-          removedTags.forEach(tag => {
-            tagRemoved({ tag });
-          });
-        } else {
-          tagAdded();
-        }
-      };
+  const changeTags = (newTags: string[], removedTags: string[]) => {
+    if (tags.length > 5) {
+      notificationAdded('You can have a maximum of 5 tags', 'error');
```
MichaelDeBoey: @christianalfoni `notificationAdded` has the following signature:
https://github.com/codesandbox/codesandbox-client/blob/cb965d440eb569d569e97ded524d54d142f241ad/packages/app/src/app/overmind/actions.ts#L28-L38

Should I still use `notificationAdded` or is another method preferred? 🤔

SaraVieira: That is the way, I think time alive was supposed to be an optional number because its all over the code without it

@christianalfoni

MichaelDeBoey: Then `timeAlive` should have a default of `1` too I guess? 🤔

christianalfoni: Ah, yeah, but in this case we have logic inside the component controllin how many tags can be added. This should be moved to `tagAdded` action. That should check number of tags and then directly use the `notifcationsToast.error` method.

This just ensures that if we ever change this component or allow adding tags from somewhere else we have this logic centralized :)

christianalfoni: Actually... just create a new action `tagsChanged` and use that directly. Move the logic into that action. Remove `tagRemoved` and `tagAdded` 😄

Issue #2465, opened 2019-09-19
Wrong time information rendered in <SandboxGrid />


janmyler:
# 🐛 bug report

The time information — _Created|Edited|Deleted … ago_ — is rendered incorrectly. 

## Description of the problem

The date string passed to `distanceInWordsToNow` seems to be in ISO 8601 format in UTC (e.g. `2019-09-19 11:29:27.280934`), however, the time zone information is missing, therefore it is parsed as a local date-time.

This results in showing incorrect information depending on user's local time zone.

A fix would be simply adding _Z_ as the zone offset for UTC. Not sure if such fix at the component level is clean, though. If the dates are in UTC, they should have the time zone specified probably. Even though, that might have consequences in other parts of the system which I am not aware of.

### Screenshots

Showing “Edited 2 minutes ago…” if I change my time zone to GMT (UTC) 👍
<img width="1140" alt="Screen Shot 2019-09-19 at 13 13 02" src="https://user-images.githubusercontent.com/718501/65247924-438f0080-db2c-11e9-9b83-abf439318ea5.png">

Showing “Edited about 9 hours ago…” if I use my local time zone 👎
<img width="1168" alt="Screen Shot 2019-09-19 at 22 13 46" src="https://user-images.githubusercontent.com/718501/65247931-47228780-db2c-11e9-99b3-d86b4d639824.png">

janmyler:
Closing for the issue has been fixed.

PR #2590, opened 2019-09-30
show right time zone

yeion7:

Fix

The sandbox dates (`insertedAt`, `updatedAt`, `deletedAt`) is received with format `2019-09-19 11:29:27.280934` this format now allow parsed with the right timezone

#2465 

## What is the new behavior?

There are two changes

1. Add `date-fns` to `app` scope
2. use [date-fns-tz](https://date-fns.org/v2.2.1/docs/Time-Zones) to parse the dates as utc dates and then `formatDistanceToNow` can parsed with the user time zone

*Before:*
Screen Shot 2019-09-19 at 20 33 56

*After*
Screen Shot 2019-09-19 at 20 33 26

## Checklist

- [ ] Documentation
- [ ] Testing 
- [x] Ready to be merged 

### DIFF
--- a/packages/app/tsconfig.check.json
+++ b/packages/app/tsconfig.check.json
@@ -2,6 +2,7 @@
   "extends": "./tsconfig.json",
   "exclude": ["codesandbox-browserfs/*"],
   "compilerOptions": {
-    "allowJs": false
+    "allowJs": false,
+    "skipLibCheck": true
   }
 }
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -104,6 +104,8 @@
     "console": "^0.7.2",
     "console-feed": "CompuIves/console-feed#build2",
     "css-modules-loader-core": "^1.1.0",
+    "date-fns": "^2.4.1",
+    "date-fns-tz": "^1.0.7",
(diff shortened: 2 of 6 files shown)


packages/app/tsconfig.check.json
```diff
@@ -2,6 +2,7 @@
   "extends": "./tsconfig.json",
   "exclude": ["codesandbox-browserfs/*"],
   "compilerOptions": {
-    "allowJs": false
+    "allowJs": false,
+    "skipLibCheck": true
```
yeion7: @SaraVieira i've fixed the `date-fns` typings error enable this setting, but i don't know if is ok for the codesandbox setup

christianalfoni: What was the typing errors? Should not be any issue using `date-fns` with Typescript?

yeion7: yes, it's one error with the `date-fns` typing definition

SaraVieira:
I am okay with the skip, we shouldn't fail because of libs

About the thing thank you so much this is because I removed moment and didn't notice this since :/

Thank you so much 🌈

PR #2591, opened 2019-10-01
Perf performance on file sync

christianalfoni:


packages/app/src/app/overmind/internalActions.ts
```diff
@@ -257,8 +258,6 @@ export const setCurrentSandbox: AsyncAction<Sandbox> = async (
 
   effects.executor.setupExecutor();
 
-  effects.fsSync.syncCurrentSandbox();
```
CompuIves: Should this be removed? I think whenever the sandbox changes we need to send the new files to the worker. Oh wait, we look at the structure change now, so it happens automatically... Okay, nice! Keeping this comment here for documentation purposes.

packages/app/src/app/overmind/namespaces/editor/state.ts
```diff
@@ -157,13 +165,34 @@ export const state: State = {
...
+      if (!Object.keys(paths).some(p => dirname(p) === path)) {
+        paths[path] = { shortid: d.shortid, type: 'directory' };
       }
     });
 
-    return modulesObject;
+    return paths;
   },
+  modulesByPath: ({ currentSandbox, modulePaths }) =>
+    Object.keys(modulePaths).reduce((aggr, path) => {
```
CompuIves: We can make this a bit faster by having something like this:

```
const finalObj = {};

Object.keys(modulePaths).forEach(m => {
  finalObj[...] = m;
});

return finalObj;
```

This way we don't create a new object and merge for all dirs and modules (so max ~300 times) and should be a bit faster.

packages/app/src/app/overmind/internalActions.ts
christianalfoni: Jep, exactly :)

packages/app/src/app/overmind/namespaces/editor/state.ts
christianalfoni: ah, but the populated `type` property of "file" or "directory" I believe to be used inside BrowserFS?

CompuIves: Yes, oh, this is just pseudo code. What comes after the `=` should stay the same.

christianalfoni: Hm... then I do not see how they differ? :) You create a new object and populate it in both scenarios? The reducer is not returning a new object every time, it is returning the same object, just populated with a new property :)

christianalfoni: But reducers... they do not express populating arrays/objects the best. Your way is more explicit for sure. I can change it to make it more readable :)

CompuIves: Ahhh, I've been misreading this. Sorry! I thought it was saying

```
return {
	...acc,
	[m.path]: {...}
}
```

Whoops sorry!

christianalfoni:
Okay, so the last fixes I did was:

- Fixed a subtle bug that became apparent with this change. When forking now we also populate the modules with client side state. This might have caused specific issues just after forking
- Each module has now `errors` and `corrections` which is pushed/cleared with the existing manipulation of `editor.errors` and `editor.corrections`. Have to keep the editor state because of usage inside `codesandbox.main.js`
- Cleaned the typing so that we only have `ModuleError` and `ModuleCorrection`
- `ModuleEntry` now looks specifically on the module for errors
- `ModuleEntry` now compares `savedCode` and `code` to figure out changes, this also avoid the initial render of all modules when `changedModuleShortids` change
- Changed to `PureComponent` on both `DirectoryChildren` and `ModuleEntry`

packages/app/src/app/overmind/namespaces/editor/actions.ts
```diff
@@ -490,34 +503,27 @@ export const previewActionReceived: Action<{
...
-        column: action.column,
-        line: action.line,
-        columnEnd: action.columnEnd,
-        lineEnd: action.lineEnd,
-        message: action.message,
-        title: action.title,
-        path: action.path,
-        source: action.source,
-      };
+      const error: ModuleError = action;
```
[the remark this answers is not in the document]
christianalfoni: Yup! Cool, let me fix that :)

PR #2593, opened 2019-10-01
This one weird trick will fix console height issues

### LINKED ISSUE #2522: Trying to resize the console minimizes it
# 🐛 bug report

## Description of the problem
Resizing the console doesn't work in the editor. When you select the console edge and try to drag it to resize the panel, it immediately closes/minimizes on click.

### Your Environment

| Сodesandbox      | PROD-1569330198-771afd31e
| Browser          | Chrome
| Operating System | OS X

siddharthkp:

Fix #2522 

Oh boy, strap in

This bug was introduced in #2451 which makes ESLint happy

We have an eslint rule - [no-access-state-in-setstate.md](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-access-state-in-setstate.md) which asks for a straightforward change like this - 

```diff
- this.setState({
+ this.setState(state => ({
    startY: event.clientY,
	... a few more
  })
+ )
```

The trouble is that passing a callback runs it asynchronously.

This is the warning we see in dev mode:

```
Warning: This synthetic event is reused for performance reasons. If you're seeing this, you're accessing the property `clientY` on a released/nullified synthetic event. This is set to null. If you must keep the original synthetic event around, use event.persist(). See https://fb.me/react-event-pooling for more information
```

Following the docs:

> The SyntheticEvent is pooled. This means that the SyntheticEvent object will be reused and all properties will be nullified after the event callback has been invoked. This is for performance reasons. As such, you cannot access the event in an asynchronous way.

> If you want to access the event properties in an asynchronous way, you should call event.persist() on the event, which will remove the synthetic event from the pool and allow references to the event to be retained by user code.

## What is the new behavior?

Added `event.persist()` which makes it available for height calculations. (and also makes it future-proof for concurrent mode)

There might be performance regressions because of this which ~~I haven't tested yet~~ aren't noticeable to the eye

## Open questions

Should we go this route or should we undo this and use the `setState({object})` and disable the eslint rule for this block.

- [x] Resized the console after making the fix
- [x] Test for performance regressions in sync mode

## Checklist

- [NA] Documentation
- [x] Testing 
- [ ] Ready to be merged 

### DIFF
--- a/packages/app/src/app/components/Preview/DevTools/index.tsx
+++ b/packages/app/src/app/components/Preview/DevTools/index.tsx
@@ -274,6 +274,7 @@ export class DevTools extends React.PureComponent<Props, State> {
     event: React.MouseEvent & { clientX: number; clientY: number }
   ) => {
     if (!this.state.mouseDown && typeof this.state.height === 'number') {
+      event.persist();
       unFocus(document, window);
       // @ts-ignore
       this.setState(state => ({


yeion7:
maybe we could get the `clientY` value before call `setState`, in that way we don't persist the all event object, just the `clientY` value

```
 handleMouseDown = (
    event: React.MouseEvent & { clientX: number; clientY: number }
  ) => {
    if (!this.state.mouseDown && typeof this.state.height === 'number') {
      const { clientY } = event
      unFocus(document, window);
      // @ts-ignore
      this.setState(state => ({
        startY: clientY,
        startHeight: state.height,
        mouseDown: true,
      }));
      if (this.props.setDragging) {
        this.props.setDragging(true);
      }
    }
  };
```

siddharthkp:
@yeion7 Aaah, totally could be. Would you like to fork and give it a try?

yeion7:
Sure!

siddharthkp:
Carried forward in https://github.com/codesandbox/codesandbox-client/pull/2595

Issue #2598, opened 2019-10-01
Live session crash report


wmain:
<h1>💥 Crash Report</h1>

  <h2>Join a live session</h2>

  <h3>Link to sandbox: https://codesandbox.io/live/nZY8K4() (optional)</h3>

  <h3>Crash Details</h3>

  <details>
  <summary>Environment</summary>

  | Browser |  Version  | Operating System |
  | ------- | --------- | ---------------- |
  | chrome | 77.0.3865 | OS X 10.14.5           |

  **Route:**
  https://codesandbox.io/live/nZY8K4

  </details>

  <details>
  <summary>Error Message</summary>

  ```bash
  TypeError: Cannot read property 'editorService' of undefined
  TypeError: Cannot read property 'editorService' of undefined
    at t.destroyMonaco (https://codesandbox.io/static/js/default~app~embed.df990aeb.chunk.js:276:41612)
    at t.value (https://codesandbox.io/static/js/default~app~embed.df990aeb.chunk.js:276:42384)
    at https://codesandbox.io/static/js/common.19be7978.chunk.js:1:2070092
    at Is (https://codesandbox.io/static/js/common.19be7978.chunk.js:1:2070133)
    at Ps (https://codesandbox.io/static/js/common.19be7978.chunk.js:1:2070214)
    at zs (https://codesandbox.io/static/js/common.19be7978.chunk.js:1:2071898)
    at https://codesandbox.io/static/js/common.19be7978.chunk.js:1:2090729
    at t.unstable_runWithPriority (https://codesandbox.io/static/js/common.19be7978.chunk.js:1:2113916)
    at co (https://codesandbox.io/static/js/common.19be7978.chunk.js:1:2038799)
    at Ua (https://codesandbox.io/static/js/common.19be7978.chunk.js:1:2088532)
    at Ra (https://codesandbox.io/static/js/common.19be7978.chunk.js:1:2077169)
    at https://codesandbox.io/static/js/common.19be7978.chunk.js:1:2039070
    at t.unstable_runWithPriority (https://codesandbox.io/static/js/common.19be7978.chunk.js:1:2113916)
    at co (https://codesandbox.io/static/js/common.19be7978.chunk.js:1:2038799)
    at po (https://codesandbox.io/static/js/common.19be7978.chunk.js:1:2039016)
    at ho (https://codesandbox.io/static/js/common.19be7978.chunk.js:1:2038951)
[...]

PR #2599, opened 2019-10-01
prevent live session crash

yeion7:

bugfix

#2598 

I'm not sure about the cause of the error, but when the editor is unmounting, it doesn't find the editor instance

## What is the new behavior?

Check is editor exists

## Checklist

- [ ] Documentation
- [ ] Testing 
- [x] Ready to be merged 


CompuIves:
Hey! After checking this PR I found that the error happens because of another error that happens, it was quite a critical error so I went ahead and put a fix in right away here: #2608. I hope that that will fix this issue as well, I'm deploying it in 10min so we can see for sure.

PR #2600, opened 2019-10-01
fix eslint issues in url-listeners

armujahid:

Nothing. Just ESLint fixes in url-listeners.js


CompuIves:
Good fix, thanks a lot!

PR #2638, opened 2019-10-04
🔨 Refactored 🧠 Overmind : /editor/content/preview/index to replace Cerebral with Overmind

chinmay17:
## What kind of change does this PR introduce?

Refactors code as a part of hacktoberfest mentioned in #2621.
@Saeris @christianalfoni

## What is the current behavior?

/editor/content/preview/index.tsx was using app/componentConnectors

## What is the new behavior?

uses `useOvermind` from `app/overmind`

## What steps did you take to test this?

Tested the Preview component by making changes in the code, adding/removing dependencies

## Checklist

- [ ] Documentation
- [x] Testing 
- [x] Ready to be merged 
- [ ] Added myself to contributors table

PR #2641, opened 2019-10-04
update Overmind to get correct reaction type

christianalfoni:
This update fixes typing of Reaction and also Overmind will now pass the returned result of the first callback as the argument to the next callback.

Should verify by running project and play around a bit, but we we have our own wrapper currently which should not be affected by this. The good thing is that the change of Reaction now fits with the way we use the wrapper as well.


SaraVieira:
@christianalfoni typecheck is failing

christianalfoni:
Closed as fixed in: https://github.com/codesandbox/codesandbox-client/pull/2638#pullrequestreview-297432160

PR #2675, opened 2019-10-07
Make the status bar less distracting

siddharthkp:

Enhancement. I finally found out how to change the color of the distracting status bar that vscode puts on the bottom.

I suggest we change it to a dark black (chose `#242424` from the redesign) in the default codesandbox theme

It's distracting blue

Screenshot 2019-10-07 at 11 54 04 AM

## What is the new behavior?

It's subtle black

Screenshot 2019-10-07 at 12 59 30 PM

~~See deployment for Real Feel™~~
@CompuIves [the deployment](https://pr2675.build.csb.dev/s/new) doesn't show theme from local, I'm guessing it uses the published theme?

⚠️  Important: This is a breaking change for the default theme. That's the intention, merge if you agree. (tagged a bunch of folks for opinions)

Tested the sandbox, looks great

## Checklist

- [NA] Documentation
- [x] Testing 
- [ ] Ready to be merged 


CompuIves:
I like this, @DannyRuchtie what do you think?

CompuIves:
Oh @siddharthkp, you also need to change it here: https://github.com/CompuIves/codesandbox-client/blob/2acdfce8957c80431e86f94a6d64e754ab13f890/standalone-packages/vscode-extensions/out/extensions/ngryman.codesandbox-theme-0.0.1/themes/CodeSandbox-color-theme.json (this file is actually used in the theme picker)

siddharthkp:
@CompuIves thanks! Should we keep the change in common/themes as well or is that basically dead code (unless that key is explicitly used?)

CompuIves:
Basically dead code, only used for the iPad syntax highlighting.

siddharthkp:
@CompuIves Changed a few versions to `v10`, there are some more that I don't think are related?

Like these: [overmind/effects/zeit.ts](https://github.com/codesandbox/codesandbox-client/blob/9eefb01c205db074df242cad5c8df48baad63830/packages/app/src/app/overmind/effects/zeit.ts#L98) and [deployment/actions.js](https://github.com/codesandbox/codesandbox-client/blob/01da61e27b8acc98ad45be281264e38ce55cf7fa/packages/app/src/app/store/modules/deployment/actions.js#L258)

Aaaand should I bother changing the package version of `vscode-extensions` or is that a outdated?

CompuIves:
😂 nice commit. One last thing, we now need to invalidate the cache of VSCode, we can do this by changing all `vscode23` to `vscode24` with find and replace.

siddharthkp:
@CompuIves that didn't do it, there's more to it?

CompuIves:
Heh, this should be it! Hmm, okay maybe it could be that VSCode also does a bit of caching.

siddharthkp:
@CompuIves I don't think this made any different 🤔

PR #2680, opened 2019-10-07
Fix material-ui integration test

siddharthkp:
The integration test for material-ui is broken on master because we use `latest` as version for dependencies and there seems to be a breaking change.

Sandbox: https://codesandbox.io/s/n5wy74w8vl

Until we fix this, all branches from master also show a broken state which isn't useful. 

I suggest we disable this specific integration test and bring it back after fixing it. Having a green master is the base of our test suite!

Screenshot 2019-10-07 at 4 45 44 PM


siddharthkp:
@SaraVieira 🥳

PR #2683, opened 2019-10-07
fix: fork errors when the limit is reached

jyash97:
Bug Fix

Note: This happens when the sandbox limit is reached i.e. 50.

When we click `CMD + S` on new sandbox template `saveModule` is called which sends an unnecessary API call for the sandbox. Check the below gif for more details. You can also see the 2 extra errors coming on the right side.
ForkError

## What is the new behavior?
Does not call the API and shows the other errors.

## Checklist

- [ ] Documentation - N/A
- [x] Testing 
- [ ] Ready to be merged 

Do lmk if this wasnt require I can close the PR. Also how to connect to overmind devtools ?

### DIFF
--- a/packages/app/src/app/overmind/namespaces/editor/internalActions.ts
+++ b/packages/app/src/app/overmind/namespaces/editor/internalActions.ts
@@ -83,6 +83,10 @@ export const saveCode: AsyncAction<{
     return;
   }
 
+  if (state.editor.currentId === 'new' && !state.editor.isForked) {
+    return;
+  }
+
   if (state.preferences.settings.experimentVSCode) {
     await actions.editor.codeChanged({
       code,
@@ -314,9 +318,11 @@ export const forkSandbox: AsyncAction<{
     }
 
     await actions.internal.setCurrentSandbox(forkedSandbox);
+    state.editor.isForked = true;
     effects.notificationToast.success('Forked sandbox!');
     effects.router.updateSandboxUrl(forkedSandbox);
   } catch (error) {
+    state.editor.isForked = false;
     console.error(error);
     effects.notificationToast.error('Sorry, unable to fork this sandbox');
   }
--- a/packages/app/src/app/overmind/namespaces/editor/internalActions.ts
+++ b/packages/app/src/app/overmind/namespaces/editor/internalActions.ts
@@ -83,10 +83,6 @@ export const saveCode: AsyncAction<{
     return;
   }
 
-  if (state.editor.currentId === 'new' && !state.editor.isForked) {
-    return;
-  }
-
   if (state.preferences.settings.experimentVSCode) {
     await actions.editor.codeChanged({
       code,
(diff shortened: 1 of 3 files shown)


packages/app/src/app/overmind/namespaces/editor/internalActions.ts
```diff
@@ -83,6 +83,10 @@ export const saveCode: AsyncAction<{
     return;
   }
 
+  if (state.editor.currentId === 'new' && !state.editor.isForked) {
```
SaraVieira: I think this will only work in the react one :/

SaraVieira: Can you explain why you check for the currentID?

jyash97: Thanks for the quick review.

> I think this will only work in the react one :/

Good catch, I forgot to test for other templates

> Can you explain why you check for the currentID?

Consider doing `CMD + S` on an already saved app, `isForked` will be false and would just do `return` and doesn't call the API.


Is there a way to know when the app would be forked? We can add that logic over here and that would work for other templates as well.

CompuIves: Would we be able to exit the flow if an error occurs in the fork itself? Something like an early return? That would seem like the most probably solution to me.

jyash97: `isForked = false` tells that there was an error while forking and we can use it to return early. And this would work currently if the fork is unsuccessful the only issue that exists we need to fix the condition for `currentId === 'new'`.

jyash97: I got an idea do let me know if this sounds good: 

We can create a state variable called `hasForkError` and set it to `true`  when the Fork API throws or returns an error. And exit the flow when this is `true` something like this: 

```
if (hasForkError) {
   return;
}
```

cc @SaraVieira  @CompuIves

CompuIves: Hmm, would it be possible to do this:

```
try {
    await actions.editor.internal.forkSandbox({
      sandboxId: state.editor.currentId,
    });
} catch (e) {
  return;
}
```

here:

https://github.com/CompuIves/codesandbox-client/blob/7e67db3c72b7c2099d33296c437c4a60be47d8ac/packages/app/src/app/overmind/factories.ts#L74-L76

?

CompuIves: Ah, I see we'd need to do that at multiple places then, here as well:

https://github.com/CompuIves/codesandbox-client/blob/7e67db3c72b7c2099d33296c437c4a60be47d8ac/packages/app/src/app/overmind/namespaces/editor/internalActions.ts#L301

jyash97: Great @CompuIves  this works properly. Thanks for helping me with this. I missed the `factories` file, now I know how things work when we save. You can check the implementation if it looks good.

jyash97:
Ping @CompuIves  Any updates on this?

jyash97:
Closing this as the limit restriction are removed :)

Issue #2107, opened 2019-06-28
Always prompt user on ctrl/cmd-w (even with all files saved) to prevent accidental closing


karlhorky:
# 🌈 Feature

Always prompt user on ctrl/cmd-w (even with all files saved) to prevent accidental closing. Maybe with a setting, in case some people find it annoying?

I've closed the PWA so many times accidentally now, and it takes me a long time to get back in - need to go over the dashboard again each time :(

One way of doing this would be to always set `onbeforeunload`, even if a file hasn't been saved.

Related: https://github.com/codesandbox/codesandbox-client/issues/1044

parkerault:
I second this. I reflexively hit ctrl-w to close an editor, and the user prompt catches it when I have an unsaved file open, but if everything is saved it closes the browser tab. This happens to me constantly and it drives me crazy.

karlhorky:
not stale

SaraVieira:
hey Karl, 

Same deal as the other two and moved to our backlog. This will be a setting when implemented since it's a super intrusive action to stop people from closing the tab without their consent but a setting that's off by default does not hurt

karlhorky:
Ok, thanks for the heads up :)

Issue #2089, opened 2019-06-24
Hotkey defaults that work in the browser (for common actions)


karlhorky:
# 🌈 Feature

Would the CodeSandbox team be interested in implementing hotkey defaults that work in browsers for common actions such as switching between tabs (VS Code Settings "View: Open Next Editor" and "View: Open Previous Editor").

For example, I would like to be able to switch to the next and previous tabs, but they are currently set to hotkeys like <kbd>cmd</kbd>-<kbd>option</kbd>-<kbd>left/right</kbd>, which collide with Chrome's bindings:

<img width="623" alt="Screen Shot 2019-06-24 at 17 30 08" src="https://user-images.githubusercontent.com/1935696/60032882-ca584c80-96a7-11e9-9268-5fd9182ddc76.png">

Since Chrome won't allow rebinding these keys, it would be cool if CodeSandbox had a binding for this out-of-the-box.

I know that I can just bind them myself, but I am suggesting to do this for all users to avoid having to do this.

## Un-rebindable Keybindings

A list of keybindings that do not work in various platform + browser combinations (I've been testing them with [JavaScript Event Key Tester](https://unixpapa.com/js/testkey.html) with all suppressions enabled):

#### macOS

- command-option-left (Previous Tab)
- command-shift-[ (Previous Tab)
- command-option-right (Next Tab)
- command-shift-] (Next Tab)
- command-w (Close Tab)
- command-t (Open New Tab)
- command-n (Open New Window)

---

**References:**

Prior Art: https://github.com/codesandbox/codesandbox-client/issues/1044#issuecomment-413623912

> I wish we could have native behaviour here, but Chrome doesn't allow us to override CMD + W. This is frustrating for me, for now I 'fixed' it by implementing CTRL + W for Mac as alternative...

Issue "Ctrl + Tab to switch between two files": #761 (PR with implementation: #817)

CompuIves:
Hmm yes this seems like a good idea. No use of keybindings that don't work 😄. If you're interested in tackling this, could you compile a list of keybindings that need to be replaced with the corresponding replacements? We can then put that as default keybindings for VSCode.

karlhorky:
Cool, I've started to compile a list above in the description, maybe we can put the "help wanted" label on this. It's going to be a bit of a job, especially for a cross-platform list.

karlhorky:
[...]

Issue #1997, opened 2019-05-29
CTRL/CMD + 1/2/3/4/5... to switch to another Chrome tab not working


alexluong:
# ❔ Question

I know it must be difficult to integrate editor's shortcuts and browser's ones. For example, `CTRL/CMD + W` is a known issue here and cannot be fixed.

My use case is actually the opposite of the situation above. I'd prefer Chrome's shortcuts to take precedence when using `CTRL/CMD + NUMBER` to switch tabs while editing. Is there anyway I can achieve this? Thank you!

### Your Environment

| Software         | Name/Version |
| ---------------- | ------------ |
| Browser          | Chrome
| Operating System | Mac OS

Saeris:
Hey there!

You can remap keybindings inside the editor under File > Preferences > Keyboard Shortcuts (or by pressing CTRL/CMD + K, CTRL/CMD + S), where you should be able to unbind the hotkeys you're looking to use in Chrome. Finding all of them may take a bit of effort however.

![image](https://user-images.githubusercontent.com/3144549/58623310-8e85bf00-8282-11e9-87ef-faf8e00ed171.png)

If this workaround doesn't solve your issue then we may need to investigate a fix.

Hope this helps!

alexluong:
@Saeris Hey thanks for the replies! It works for me now. I'll close the issue. Appreciate the help.

Issue #1044, opened 2018-08-16
`CTRL/CMD + W` to close active editor tab


farskid:
# 🌈 Feature

for now if you use a combination of `CTRL/CMD + W`, it closes the whole browser tab.

I'd rather have it close active editor tab same as my editor.

@CompuIves Would be happy to work on it and send a PR.

CompuIves:
I wish we could have native behaviour here, but Chrome doesn't allow us to override `CMD + W`. This is frustrating for me, for now I 'fixed' it by implementing `CTRL + W` for Mac as alternative...

farskid:
Thanks for the info :) I'll close the issue.

PR #2716, opened 2019-10-09
Add support for native keyboard shortcuts in fullscreen

patrikholcak:

So I’ve always dreamt of having just browser installed and code everything inside of it. But something was always stopping me from doing so. Lack of a good IDE, slow response, etc. I was really happy to see CodeSandbox implement VSCode and even more happy to see how snappy it is! But still, I was sometimes frustrated when I forgot I was using a browser-based IDE and that the system shortcuts don’t work. My primary cause of headaches was always hitting <kbd>⌘+W</kbd> 🙄. Never got used to it. And it looks like I’m [not](https://github.com/codesandbox/codesandbox-client/issues/1044) [the](https://github.com/codesandbox/codesandbox-client/issues/1997) [only](https://github.com/codesandbox/codesandbox-client/issues/2089) [one](https://github.com/codesandbox/codesandbox-client/issues/2107).

So when I stumbled upon [this](https://developer.mozilla.org/en-US/docs/Web/API/Keyboard/lock) _experimental_ Keyboard lock API, I immediately started thinking about how I can implement this into CodeSandbox and if it would work. As it turns out, it does! **

** In fullscreen, where [supported](https://caniuse.com/#feat=mdn-api_keyboard_lock).

This experiment introduces new fullscreen mode (View → Toggle Fullscreen or <kbd>⌥+⌃+F</kbd>), which calls the Keyboard lock api to disable browser shortcuts when active. When active, keyboard shortcuts like New Editor/Close current Editor/Reopen last closed) work the same as they would in desktop VSCode.

> **NOTE:** Everything is subject to change, this is just an experiment. I always code in fullscreen so this was a natural solution for my use-case but it might not be valid for you or at all. In that case, feel free to close this PR or add commits to it. 🙂 

Screenshot 2019-10-09 at 08 11 44

_Activating the Fullscreen mode adds notification with information how to get out of it_

Hitting one of the browser default shortcuts performs browser action.

## What is the new behavior?

Hitting one of the browser default shortcuts doesn’t perform browser action.

## What steps did you take to test this? This is required before we can merge.

None yet, this is mostly an experiment 😇

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      
### DIFF
--- a/packages/app/src/app/vscode/index.ts
+++ b/packages/app/src/app/vscode/index.ts
@@ -1,9 +1,9 @@
-import {
-  convertTypeToStatus,
-  notificationState,
-} from '@codesandbox/common/lib/utils/notifications';
+import { notificationState } from '@codesandbox/common/lib/utils/notifications';
 import { blocker } from 'app/utils/blocker';
-import { NotificationMessage } from '@codesandbox/notifications/lib/state';
+import {
+  NotificationMessage,
+  NotificationStatus,
+} from '@codesandbox/notifications/lib/state';
 
 import { KeyCode, KeyMod } from './keyCodes';
 import bootstrap from './dev-bootstrap';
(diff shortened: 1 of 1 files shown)


CompuIves:
This is geeeeniuuus!!!!! Going to play with this now!!

patrikholcak:
@CompuIves did you just merge it? 😅This was supposed to be an experiment IVES!
Next step is to research a way to remap keybind to close the current editor to CMD+W in fullscreen

PR #2722, opened 2019-10-09
Update users.json

joepio:
Ontola is a proud user of CodeSandbox! Jullie zijn awesome.


packages/homepage/content/users.json
```diff
@@ -102,5 +102,10 @@
...
+  },
 ]
```
SaraVieira: It's a json file you can't add an ending comma :(

siddharthkp: Fixed that for you!

PR #2724, opened 2019-10-10
Live: Don't send userEnter notification if notifications are turned off

### LINKED ISSUE #2719: Live mode - Switch "Hide notifications" doesn't have any effect
# 🐛 bug report

## Description of the problem

In live mode, we tried to hide notifications in classroom mode. Professors and students could still see all the notifications about people joining and leaving the session (which is very annoying).

### Your Environment

| Сodesandbox      | PROD-1570545059-ce110646c
| Browser          | Firefox 70.0b13
| Operating System | Windows 10

siddharthkp:

Fix #2719

We send a notification about user entering the session even if it's turned off in notifications. Seems like the check was missing from `onUserEnter` (it's present in `onUserLeft`)

It might be worth looking at other notifications as well. 

## What steps did you take to test this? This is required before we can merge.

Could not test this on local (can't start live session on local), will test on staging

## Checklist

- [NA] Documentation
- [ ] Testing 
- [ ] Ready to be merged

### DIFF
--- a/packages/app/src/app/overmind/namespaces/live/liveMessageOperators.ts
+++ b/packages/app/src/app/overmind/namespaces/live/liveMessageOperators.ts
@@ -64,10 +64,12 @@ export const onUserEntered: Operator<
 
   const user = data.users.find(u => u.id === data.joined_user_id);
 
-  effects.notificationToast.add({
-    message: `${user.username} joined the live session.`,
-    status: NotificationStatus.NOTICE,
-  });
+  if (!state.live.notificationsHidden) {
+    effects.notificationToast.add({
+      message: `${user.username} joined the live session.`,
+      status: NotificationStatus.NOTICE,
+    });
+  }
 });
 
 export const onUserLeft: Operator<


packages/app/src/app/overmind/namespaces/live/liveMessageOperators.ts
```diff
@@ -64,10 +64,12 @@ export const onUserEntered: Operator<
 
   const user = data.users.find(u => u.id === data.joined_user_id);
 
-  effects.notificationToast.add({
-    message: `${user.username} joined the live session.`,
-    status: NotificationStatus.NOTICE,
-  });
+  if (!state.live.notificationsHidden) {
+    effects.notificationToast.add({
```
christianalfoni: Actually, we can make `effects.notificationToast` aware of this state.

By going to the `overmind/effects/notificationToast` we can add an `initialize` method. This is just a convention.

Something like:

```ts
let _shouldNotify: () => boolean
export default {
  initialize(shouldNotify) {
    _shouldNotify = shouldNotify
  },
  ...existingMethods
}
```

Now, when Overmind is initialized we wire up our effects. In the case of our toasts effect we can now (overmind/onInitialize.ts):

```ts
export const onInitialize: OnInitialize = async (
  { state, effects, actions },
  overmindInstance
) => {
  effects.notificationsToast.initialize({
    shouldNotify: () => !state.live.isLive || (state.live.isLive && !state.live.notificationsHidden)
  })
  ...
}
```

We do not have to do this now, maybe overkill. But this will put that responsibility at a single point in the codebase, which we could expand later as well :)

siddharthkp: I like it! 

Creating another pull request after this one because I have no idea how to test the changes 😅

CompuIves: Wouldn't it make sense to only hide the live notifications when you press the "silence notifications" option?

PR #2736, opened 2019-10-12
update overmind and fix reaction

christianalfoni:
This inserts latest version of Overmind and also fixes a wrongful reaction. The typing there was throwing so have no idea how this got into `master`?


packages/app/src/app/overmind/internalActions.ts
```diff
@@ -64,7 +64,7 @@ export const setPatronPrice: Action = ({ state }) => {
 
 export const setSignedInCookie: Action = ({ state }) => {
   document.cookie = 'signedIn=true; Path=/;';
-  identify('signed_in', true);
+  identify('signed_in', 'true');
```
CompuIves: This one should be fine to have boolean now. We had it as boolean before, so for consistency I think we should have boolean again.
```suggestion
  identify('signed_in', true);
```

packages/app/src/app/overmind/effects/jwt.ts
```diff
@@ -16,7 +16,7 @@ export default {
     document.cookie = `signedIn=; Path=/; expires=Thu, 01 Jan 1970 00:00:01 GMT;`;
     document.cookie = `jwt=; Path=/; expires=Thu, 01 Jan 1970 00:00:01 GMT;`;
 
-    identify('signed_in', false);
+    identify('signed_in', 'false');
```
CompuIves:
```suggestion
    identify('signed_in', false);
```

PR #2741, opened 2019-10-12
Appreciate property of savedCode more

CompuIves:
`savedCode` can be `null` whenever `savedCode === code`. However, we often get values this way: `savedCode || code || ''`. This doesn't work when `savedCode` is `''`, and `code` is a filled string. I'm not sure if this has caused many bugs before, but this definitely can cause inconsistencies. This change adds explicit checks.

I also created specific types for the API, also updated some existing types to use `null` instead of `undefined` (elixir doesn't know what undefined is). We need to look out for cases where properties are added to sandbox which are not included in the API, this can cause performance problems (all components using the sandbox update). This is an attempt at making that more clear, but maybe we should just put the whole API response inside the API typings and then make the sandbox type extend that type? That would explicitly show that fields are added in the end result.

### DIFF
--- a/packages/app/src/app/overmind/effects/api/types.ts
+++ b/packages/app/src/app/overmind/effects/api/types.ts
@@ -1,0 +1,28 @@
+import { Sandbox } from '@codesandbox/common/lib/types';
+
+export interface IModuleAPIResponse {
+  id: string;
+  shortid: string;
+  code: string | null;
+  directoryShortid: string | null;
+  isBinary: boolean;
+  sourceId: string;
+  title: string;
+  insertedAt: string;
+  updatedAt: string;
+}
+
+export interface IDirectoryAPIResponse {
+  id: string;
+  shortid: string;
+  directoryShortid: string | null;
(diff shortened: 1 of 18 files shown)


packages/app/src/app/overmind/effects/api/types.ts
```diff
@@ -0,0 +1,28 @@
...
+  id: string;
+  shortid: string;
+  directoryShortid: string | null;
+  sourceId: string;
+  title: string;
+  insertedAt: string;
+  updatedAt: string;
+}
+
+export type SandboxAPIResponse = Omit<Sandbox, 'environmentVariables'> & {
```
CompuIves: I wonder what you thiink of this @christianalfoni

christianalfoni:
I think we should check out the latest `graphql` implementation of Overmind, it also includes automatic typing of responses and variables. Since the communication of graphql is moved to an effect we could in theory replace our current `api` with graphql, with automatic typing etc.

We have to consider it at least :)

PR #2750, opened 2019-10-12
Add dev auth for localhost

CompuIves:
Before it wouldn't be possible to sign in a dev environment on `localhost` without copying over the `jwt`. This allows you to actually do it.

Instead of the normal flow we open a new path (only for dev) called `/auth/dev`. This popup prompts you to open `codesandbox.io/cli/login`, which will generate a token we can use to sign in. If you give this token we will do the API call to do get the `jwt` to return that.

This way we can test the sign in flow in localhost!

I tested this with my own account.

AuthFlow


jyash97:
This is really Awesome 💯 🔥

PR #2785, opened 2019-10-14
Overflow styling for templates | HacktoberFest

twanmulder:
Fixed overflow styling for Template /Dependencies dropdown.

This PR changes the overflow property, from the ReactShow component inside a WorkspaceItem, to 'hidden'.

ezgif com-resize

When a user is on the editor page and on the Template info side-menu, the template dropdown doesn't properly set the overflow property to hidden.

## What is the new behavior?
ezgif com-resize (1)

The property used to be assigned using props.

Toggling a dropdown could toggle the overflow property between 'initial' and 'hidden', but this property can't be transitioned. 

That's why the default is changed to 'hidden', so that the toggling of the dropdown will always "look pretty".

## Checklist

- [ ] Documentation **N/A**
- [x] Testing 
- [x] Ready to be merged
      
### DIFF
--- a/packages/app/src/app/pages/Sandbox/Editor/Workspace/WorkspaceItem/index.tsx
+++ b/packages/app/src/app/pages/Sandbox/Editor/Workspace/WorkspaceItem/index.tsx
@@ -43,15 +43,7 @@ export class WorkspaceItem extends React.Component<Props, State> {
   toggleOpen = () => this.setState(state => ({ open: !state.open }));
 
   render() {
-    const {
-      children,
-      title,
-      keepState,
-      disabled,
-      actions,
-      style,
-      showOverflow,
-    } = this.props;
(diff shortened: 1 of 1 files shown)


Saeris:
So there's a prob;em we've encountered with this component that needs to be tested and solved for: that `overflow: hidden` property clips all nested components, so in the example you demonstrate here in your gifs, if you click on the Template Color Picker, that modal gets clipped, which isn't ideal.

CompuIves:
Hmm yes, a fix for this would be to move the color picker to a React portal that's rendered on the body instead of in the div. I think we should apply this fix of this PR and do that, as I think it's also conventional for popups to mount on body instead of relative to the parent (I think Reakit does this too)

SaraVieira:
Hey!

Thank you so much for your help but we tried this and as drake said there are some problems :(

We don't quite know how to fix it yet so I am gonna close for now :(

PR #2787, opened 2019-10-15
🔨 Refactored 🧠 Overmind Hacktober | /app/pages/Sandbox/Editor/Workspace/Files/DirectoryEntry/index.js

malwilley:
## What kind of change does this PR introduce?

Refactor `/app/pages/Sandbox/Editor/Workspace/Files/DirectoryEntry/index.js` to use TS/hooks.

@Saeris @christianalfoni

## What is the current behavior?

https://github.com/codesandbox/codesandbox-client/issues/2621

## What steps did you take to test this? This is required before we can merge.

1. Checked sandbox file directory locally.
2. Ran `yarn lint && yarn typecheck`

## Checklist

- [ ] Documentation N/A
- [x] Testing 
- [x] Ready to be merged
- [ ] Added myself to contributors table

PR #2790, opened 2019-10-15
add script to test prs

christianalfoni:
Allows you to easily test a PR:

Screenshot 2019-10-15 at 10 03 16

Note the number of the PR, in this example: `2787`

In the root of client run: `yarn test:pr 2787`

It will automatically checkout the PR and update and build the deps. This is just to ensure the environment is correct related to testing.

If you want to make changes you can do so by:

- Just push the automatically created branch to Github as normal
- Create a new PR, but choose the existing PR (2787) as the source
- This creates a PR on the repo of the contributor and the contributor has to merge in this change, so write a comment on the original PR pointing to what needs to be merged in

I think this creates a really good collaborative feeling for the contributor. We are helping them helping us, instead of overriding what they are doing.

### DIFF
--- a/testpr-script.js
+++ b/testpr-script.js
@@ -1,0 +1,63 @@
+const { exec, spawn } = require('child_process');
+const { argv } = require('yargs');
+const username = require('username');
+
+const id = argv._[0];
+
+function spawnPromise(command, args) {
+  return new Promise((resolve, reject) => {
+    const p = spawn(command, args);
+
+    p.stdout.on('data', data => {
+      console.log(data.toString());
+    });
+
+    p.stderr.on('data', data => {
+      console.error(data.toString());
+    });
+
+    p.on('exit', code => {
+      if (code === 0) {
+        resolve();
+      } else {
+        reject();
+      }
+    });
+  });
+}
+
+async function test(prId) {
+  const branchName = `pr-${username.sync()}-${prId}`;
+
+  exec(
+    `git fetch origin pull/${prId}/head:${branchName}`,
+    (error, stdout, stderror) => {
+      if (error) {
+        throw error;
+      }
+
+      if (stderror) {
+        throw new Error('Something went wrong:' + stderror);
+      }
+
+      console.info(stdout);
+
+      spawnPromise('git', ['checkout', branchName])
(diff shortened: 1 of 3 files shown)


testpr-script.js
```diff
@@ -0,0 +1,63 @@
...
+      }
+    });
+  });
+}
+
+async function test(prId) {
+  const branchName = `pr-${username.sync()}-${prId}`;
+
+  exec(
+    `git fetch origin pull/${prId}/head:${branchName}`,
```
MichaelDeBoey: @christianalfoni This wont work on forks I guess? 🤔

christianalfoni: It is for forks :) Take a look at: https://github.com/codesandbox/codesandbox-client/pull/2787

MichaelDeBoey: I mean if I run locally `yarn test:pr 2787` this wont work, since my `origin` is my own fork of the repo and my `upstream` is the original repo

christianalfoni: Oh, I see :) Yeah, the intention was that the devs who merge into `master` has a chance to verify the PR before they do so :)

MichaelDeBoey: If it's only for CodeSandbox-people, my comment is obsolete 😉

testpr-script.js
```diff
@@ -43,9 +43,7 @@ async function test(prId) {
       console.info(stdout);
 
       spawnPromise('git', ['checkout', branchName])
-        .then(() => spawnPromise('yarn', ['install']).then(() => {
-            return spawnPromise('yarn', ['build:deps']);
-          }))
+        .then(() => spawnPromise('yarn', ['build:deps']))
```
MichaelDeBoey: But what if they _are_ changed? 🤔

christianalfoni: You will get an error either building when you start the project or when building the deps :) Just as you normally do

testpr-script.js
christianalfoni: I did not want to use the term `Codesandbox-people`, hehe... "merge people" 😂

MichaelDeBoey: Same people, different name 😉

PR #2802, opened 2019-10-15
Hide animations on prefers-reduced-motion

yeion7:

fix

when user select `reduce motion` in the SO, we don't disable the animations

## What is the new behavior?

Disable animations if user preferences is set to reduce motion,

https://developers.google.com/web/updates/2019/03/prefers-reduced-motion

1. Open homepage
2. Disable animations
3. All animations are hidden

## Checklist

- [ ] Documentation
- [ ] Testing 
- [x] Ready to be merged
      
### DIFF
--- a/packages/homepage/src/css/global.css
+++ b/packages/homepage/src/css/global.css
@@ -50,3 +50,13 @@ twitter-widget {
   margin-left: auto;
   margin-right: auto;
 }
+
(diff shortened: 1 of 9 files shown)


SaraVieira:
THIS IS AMAZING!

Thank you!!

Merging it :D

PR #2827, opened 2019-10-17
Validate if a folder/file already exist

yeion7:

Fix

You can create a file with a name that already exist but you'll get an error

## What is the new behavior?

Validate if already exist a folder/file with the same name
Validate if name includes `/` or `\`

1. Try to rename a folder or file, name it with a name that already exist
2. Try to create a new folder or file, name it with a name that already exist
3. In the above cases you'll get an error

## Checklist

- [ ] Documentation
- [ ] Testing 
- [x] Ready to be merged
      

CompuIves:
Wow, this is really nice. Also thoughtful on the memoize function. Thanks a lot for this contribution, this is really good!

PR #2843, opened 2019-10-18
🔨 Switch Project to use useOvermind

MichaelDeBoey:
Follow-up of #2782

### DIFF
--- a/packages/app/src/app/overmind/namespaces/workspace/actions.ts
+++ b/packages/app/src/app/overmind/namespaces/workspace/actions.ts
@@ -177,17 +177,15 @@ export const sandboxDeleted: AsyncAction = async ({
   effects.router.redirectToSandboxWizard();
 };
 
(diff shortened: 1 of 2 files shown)


Saeris:
Sorry, looking through this I'm not seeing any value add here.

MichaelDeBoey:
@Saeris These PRs are all my refactoring to `useOvermind` branches I already created.
I just didn't push them, 'cause the #Hacktoberfest issue came and I wanted to give everybody a chance to win their #Hacktoberfest t-shirt.
Now I already pushed some branches (and rebased them onto `master`), because I think their are still some small improvements here.

In this PR is changed the signature of `sandboxPrivacyChanged` to accept `0 | 1 | 2` instead of `{privacy: 0 | 1 | 2}`.
This because we want to change all Overmind `actions`/`effects` that only have 1 argument to accept only that type instead of an object (@christianalfoni said it somewhere, but can't find the comment anymore 😕)

PR #2849, opened 2019-10-18
🔨 Switch DashboardChangelog to use useOvermind

MichaelDeBoey:
Although `DashboardChangelog` isn't used, so I think this component can be deleted? 🤔


MichaelDeBoey:

Here I converted `Helmet` to a `tsx` file.
But I also think `DashboardChangelog` can just be removed, since I don't find any reference where it's used.

PR #2635, opened 2019-10-03
🔨 Switch LikeHeart to use useOvermind

vanya829:
## What kind of change does this PR introduce?
Refactor LikeHeart to functional components with useOvermind hook

Refactor
## What is the current behavior?

#2621
## Checklist

- [ ] Documentation
- [ ] Testing 
- [x] Ready to be merged 
- [x] Added myself to contributors table

PR #2855, opened 2019-10-18
🔨 Switch LikeHeart to use useOvermind

MichaelDeBoey:
Follow-up of #2635


MichaelDeBoey:

Here I also extracted `MaybeTooltip` into its own file and made `LikeHeart`'s props more correct.

PR #2650, opened 2019-10-04
🔨 Refactored 🧠 Overmind Hacktober | /app/pages/Dashboard/Content/Sandboxes/Filters/FilterOptions/index.tsx : refactor to replace Cerebral with Overmind

sakthivel-tw:
## What kind of change does this PR introduce?

Refactors code as a part of hacktoberfest mentioned in #2621.
@Saeris @christianalfoni @SaraVieira 

## What is the current behavior?

`/Sandboxes/Filters/FilterOptions/index.tsx` was using `inject` and `observer` from `app/componentConnectors`

## What is the new behavior?

uses `useOvermind` from `app/overmind`

## What steps did you take to test this?

ran `yarn lint` and `yarn typecheck`

## Checklist

- [ ] Documentation N/A
- [X] Testing 
- [X] Ready to be merged 
- [ ] Added myself to contributors table

PR #2857, opened 2019-10-18
🔨 Switch FilterOptions to use useOvermind

MichaelDeBoey:
Follow-up of #2650


MichaelDeBoey:

Here I also made `Filters`' and `Content`'s props more correct.

PR #2886, opened 2019-10-20
add codesandboxer-fs docs

Noviny:

Small docs update to include `codesandboxer-fs` as a way to upload files (I think this is realistically more useful than the `react-codesandboxer` docs beneath it)
## Checklist

- [x] Documentation
- [x] Testing 
- [x] Ready to be merged
      
Quick note on this: the fragmentation of tools is weird here, dating back to legacy stuff on `codesandboxer` being built out mostly independently of codesandbox itself more as an extension. It seems like it would make sense to make the `codesandbox` CLI itself work for both directories and files, to avoid recommending two different installs, but this was a quick fix to at least document this capability.

### DIFF
--- a/packages/homepage/content/docs/2-importing.md
+++ b/packages/homepage/content/docs/2-importing.md
@@ -177,7 +177,7 @@ example sandbox of that is here:
 You can export a local component to CodeSandbox by using our other
 [CLI](https://github.com/codesandbox/codesandboxer/tree/master/packages/codesandboxer-fs).
 
-You can install our CLI by running `npm install -g codesandbox`. Then you can
+You can install our CLI by running `npm install -g codesandboxer-fs`. Then you can
 export a project by running `codesandboxer {filePath}`.
 
 ```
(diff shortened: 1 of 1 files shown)


packages/homepage/content/docs/2-importing.md
```diff
@@ -172,6 +172,22 @@ example sandbox of that is here:
 
 <iframe src="https://codesandbox.io/embed/9loovqj5oy?editorsize=70&fontsize=14&hidenavigation=1&runonclick=1" style="width:100%; height:500px; border:0; border-radius: 4px; overflow:hidden;" sandbox="allow-modals allow-forms allow-popups allow-scripts allow-same-origin"></iframe>
 
+## Import single Component as a Sandbox
+
+You can export a local component to CodeSandbox by using our other
+[CLI](https://github.com/codesandbox/codesandboxer/tree/master/packages/codesandboxer-fs).
+
+You can install our CLI by running `npm install -g codesandbox`. Then you can
```
CompuIves: Should be this, right?

```suggestion
You can install our CLI by running `npm install -g codesandboxer-fs`. Then you can
```

Noviny: Yep!

SaraVieira:
Thank you so much!

Issue #2911, opened 2019-10-23
Typescript: modules not founds


dagatsoin:
# 🐛 bug report

## Description of the problem
I reopened a project wrote two weeks ago. All my typescript modules became not founds.

## How has this issue affected you? What are you trying to accomplish?
![image](https://user-images.githubusercontent.com/3670794/67411775-6fcfee00-f5be-11e9-9bbd-191472d39845.png)
As soon I make a modification, all my typescript modules becomes not found.
If I refresh the page, I got back my modules back. But it return not found as soon I make a change.

### Link to sandbox: [link]() (optional)
https://codesandbox.io/s/ravioli-mrenz

### Your Environment

| Software         | Name/Version |
| ---------------- | ------------ |
| Сodesandbox      | PROD-1571674162-0e9c0cfe9
| Browser          | Chrome Version 77.0.3865.120 
| Operating System | OSX High Siez

lbogdan:
Closing this, as it should be fixed in #2913 , and we just deployed it. Please reopen if it's still an issue for you.

Issue #2836, opened 2019-10-17
tsconfig options ignored after saving


jeremy-deutsch:
# 🐛 bug report

## Description of the problem

When tsconfig has strictness flags in `compilerOptions` (such as `strictNullChecks`, `strictBindCallApply`, or just `strict`), after you save the file, the editor stops showing type errors that reflect those flags. If unsaved changes are made to `tsconfig.json` in a separate tab, the tab with code will show the correct errors until the files are saved again.

## How has this issue affected you? What are you trying to accomplish?

TypeScript's `strictNullChecks` compiler flag is one of the main reasons I use TypeScript. Not being able to use it makes it a lot harder to safely traverse complicated data structures.

### Your Environment

| Software         | Name/Version |
| ---------------- | ------------ |
| Сodesandbox      | PROD-1571237038-a19d42751
| Browser          | Firefox 69.0.2 (64-bit)
| Operating System | ubuntu 16.0.4 LTS

lbogdan:
Hey @jeremy-deutsch ,

Can you please detail a bit what's the current vs. expected behavior? I here see the expected `null` check errors inside your sandbox:

![image](https://user-images.githubusercontent.com/2083930/67100040-ab864480-f1c7-11e9-993e-22ffa46dd342.png)

jeremy-deutsch:
That’s odd. I’ll look into it more and post screenshots once I’m at a computer, but for now, could you try opening `index.ts` and hitting ctrl-s/cmd-s, potentially after making some trivial edits? That tends to make the type errors disappear for me.

jeremy-deutsch:
Yeah, confirmed that (I'll update the description).

Before hitting ctrl-s:
![Screenshot from 2019-10-18 09-18-05](https://user-images.githubusercontent.com/13303900/67110746-4bbc7900-f188-11e9-940e-180f596386ed.png)

After hitting ctrl-s:
![Screenshot from 2019-10-18 09-18-31](https://user-images.githubusercontent.com/13303900/67110748-4c550f80-f188-11e9-817c-de65e3b9b02a.png)

lbogdan:
Closing this, as it should be fixed in #2913 , and we just deployed it. Please reopen if it's still an issue for you.

Issue #2822, opened 2019-10-16
Cannot use JSX unless the '--jsx' flag is provided as soon as new dependency is added


zivanovicb:
# 🐛 bug report

## Description of the problem
I create a React + TS sandbox, and everything looks fine - no errors. As soon as I add a new dependency [this](https://imgur.com/a/ITCHHxT) appears. If I refresh, there are no errors. Then I add a new package, and the error is there again.

## How has this issue affected you? What are you trying to accomplish?
Red is annoying.

### Your Environment

| Software         | Name/Version |
| ---------------- | ------------ |
| Сodesandbox      | PROD-1571237038-a19d42751
| Browser          | Chrome -  Version 77.0.3865.90 (Official Build) (64-bit)
| Operating System | macOS Mojave 10.14.5

shobhitg:
+1 I am seeing this issue too. I am using Chrome Version 77.0.3865.90.

SAPikachu:
+1 My project had been working fine until today, and then this error happens in all TSX files. I also see "Module XXX can only be default-imported using the 'esModuleInterop' flag" error in some of the files. I am using Firefox 69. Weirdedly it is working if I try to open the sandbox without logging in.

burtek:
+1 same here

8clever:
+1

wolverineks:
have you checked the version of typescript in package.json?
i had this error, then i changed it to `"typescript": "3.6.2"` and this error went away

joshuaaron:
+ 1 for this and the esModuleInterop flag error, even though I've set this to true in my tsconfig. Pretty annoying bug.

acerola1:
+1

zizizheng:
+1

PR #2913, opened 2019-10-23
Fix typescript error

yeion7:

fix

#2822

Possibly related

#2911 #2836

1. Create a React TS sandbox
2. Fork sandbox
3. Save file or add new dependency

## Checklist

- [ ] Documentation
- [ ] Testing 
- [x] Ready to be merged
      
### DIFF
--- a/packages/app/src/app/overmind/onInitialize.ts
+++ b/packages/app/src/app/overmind/onInitialize.ts
@@ -9,7 +9,7 @@ export const onInitialize: OnInitialize = (
   effects.fsSync.initialize({
     onModulesByPathChange(cb: (modulesByPath: any) => void) {
       overmindInstance.reaction(
-        ({ editor }) => editor.modulePaths,
+        ({ editor }) => editor.modulesByPath,
         modulesByPath => cb(modulesByPath)
       );
     },


yeion7:
@christianalfoni 🙏🏻

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/2913/commits/e21dc1d9be08a050138ad97a2e2040876e6e92cd is at https://pr2913.build.csb.dev/s/new.

TrevorSayre:
@lbogdan considering the context of the PR, perhaps this link instead: https://pr2913.build.csb.dev/s/react-ts

CompuIves:
Wow this is interesting, we decided to only do `modulePaths` as an optimization (so we don't have to send new files for every code change, as the workers only care about saved code), but it seems like we did something wrong there. Really good find @yeion7! Now I'm curious what caused this bug in the first place, I wonder what we're actually sending to the workers after adding a dependency...

CompuIves:
Hmm, after the fork modulesByPath is very different... 
image

CompuIves:
This could also be the cause of the sudden empty editor state (@christianalfoni). Let's merge this in now for a quick fix, and take a closer look whether we can optimize this. We prefer not to send all files on a single keypress, so this is not really sustainable.

CompuIves:
This is an incredible find @yeion7, this is super impressive! Thanks for digging in!

yeion7:
Thanks!

@CompuIves there are some documentation about how this (workers communication) works at CodeSandbox?

TrevorSayre:
Just to confirm, is this not yet live? What is the best way to track when a PR is released into the wild?

lbogdan:
> Just to confirm, is this not yet live?

@TrevorSayre We just deployed it.

> What is the best way to track when a PR is released into the wild?

We're currently working on making production always reflect `master`.

PR #2928, opened 2019-10-27
Make the reconnect action of live more robust

CompuIves:
Now you stay as an owner when you reconnect, and we detect the disconncet state more easily.

### DIFF
--- a/packages/app/src/app/overmind/namespaces/live/liveMessageOperators.ts
+++ b/packages/app/src/app/overmind/namespaces/live/liveMessageOperators.ts
@@ -3,6 +3,7 @@ import {
   Module,
   Directory,
   Selection,
+  LiveDisconnectReason,
 } from '@codesandbox/common/lib/types';
 import { mutate, json } from 'overmind';
 import { Operator } from 'app/overmind';
@@ -400,9 +401,7 @@ export const onConnectionLoss: Operator<LiveMessage> = mutate(
 );
 
 export const onDisconnect: Operator<
-  LiveMessage<{
-    reason: 'close' | 'inactivity';
-  }>
+  LiveMessage<{ reason: LiveDisconnectReason }>
 > = mutate(({ state, actions }, { data }) => {
   actions.live.internal.disconnect();
   state.editor.currentSandbox.owned = state.live.isOwner;
--- a/packages/app/src/app/overmind/namespaces/live/liveMessageOperators.ts
+++ b/packages/app/src/app/overmind/namespaces/live/liveMessageOperators.ts
@@ -1,5 +1,4 @@
 import {
-  LiveMessageEvent,
   LiveMessage,
   Module,
   Directory,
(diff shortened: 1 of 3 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/2928/commits/be8341029f6a153544e7910c19dcbcf265f4a084 is at https://pr2928.build.csb.dev/s/new.

packages/app/src/app/overmind/namespaces/live/liveMessageOperators.ts
```diff
@@ -10,21 +9,23 @@ import { Operator } from 'app/overmind';
 import { camelizeKeys } from 'humps';
 import { NotificationStatus } from '@codesandbox/notifications/lib/state';
 
-export const onJoin: Operator<LiveMessage<LiveMessageEvent.JOIN>> = mutate(
-  ({ effects, state }) => {
-    effects.notificationToast.success(
-      state.live.isTeam ? 'Connected to Live Team!' : 'Connected to Live!'
-    );
+export const onJoin: Operator<
+  LiveMessage<{ status: 'connected'; live_user_id: string }>
```
christianalfoni: Why remove the Event type? :)

packages/app/src/app/overmind/namespaces/live/liveMessageOperators.ts
```diff
@@ -10,21 +9,23 @@ import { Operator } from 'app/overmind';
...
+> = mutate(({ effects, state }, { data }) => {
+  state.live.liveUserId = data.live_user_id;
```
christianalfoni: IS this not when anyone joins? Or is it when only you join?

packages/app/src/app/overmind/namespaces/live/liveMessageOperators.ts
```diff
@@ -398,11 +401,10 @@ export const onConnectionLoss: Operator<LiveMessage> = mutate(
 
 export const onDisconnect: Operator<
   LiveMessage<{
-    reason: string;
+    reason: 'close' | 'inactivity';
```
christianalfoni: Would add a type here as well, like the rest of them :)

A bit confused that you can give it a dictionary type here? I believe this is the "event.type"?

packages/app/src/app/overmind/namespaces/live/liveMessageOperators.ts
CompuIves: Only when you join. good catch!

packages/app/src/app/overmind/namespaces/live/liveMessageOperators.ts
CompuIves: I wanted to add payload to the message (`status` and `live_user_id`), is there another way to do this?

christianalfoni: Ah, `LiveMessageEvent.JOIN` was wrong here... sorry!

Issue #2391, opened 2019-09-06
Create a file and directories with a qualified path


atif089:
# 🌈 Feature

While creating a file if I input something that resembles a path then I get an error `has invalid format`. So I then have to create individual directories and then eventually create the file.

For example - let's say the filename I enter to create a new file is `components/TimePicker/index.js`

What I would like to happen is to create a folder `components` and then a folder `TimePicker` inside components and then the file `index.js` inside that folder.

Currently, I have to do all of this manually which is a bit churning.

This VS Code extension lets me do that so I'm not sure how easy/difficult it would be to port that code because it uses node fs module which I'm not sure if can be used in the browser.

https://github.com/sleistner/vscode-fileutils/blob/master/src/controller/NewFileController.ts

PR #2932, opened 2019-10-28
Create nested files/folders

yeion7:

feature

You can't create nested files or folders

## What is the new behavior?

if you input a valid path you can create nested files and folders #2391

*flow*

create nested files

1. create optimistic folders
2. create optimistic file
3. upload file 
4. open file (in this way, we don't need to wait after upload the full folder structure)
5. upload folders
6. move file to folder

1. Create a nested folter structure
2. Create a nested file structure
3. Rename a file and set a new folder structure

## Checklist

- [ ] Documentation
- [ ] Testing 
- [x] Ready to be merged
      

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/2932/commits/aece73891678010ed6a77ca4a85732add4cf1f46 is at https://pr2932.build.csb.dev/s/new.

CompuIves:
WOW. THIS IS AWESOME!

christianalfoni:
This certainly is impressive 😄 I suggest we wait bringing this in until the end of the week as the `vscodeeffect` branch touches a lot of these files. Just avoid having extra "tension" getting that in.

arthurdenner:
@yeion7, could you solve the conflicts on this to see if we can get it merged, please?

Awesome work, btw!

christianalfoni:
From discussion:

**christianalfoni**
*"Hmf... I have a really bad feeling about this nested directory stuff. We are throwing complexity on top of already way too complex stuff. The code becomes incredibly intricate, because:

- The directories evaluated from the filename might or might not contain new directories. It can also contain a mix of both. First part exists, but second needs to be created
- We have to introduce denormalize, which does not match our current optimistic logic
- A rename of a file can now also be a "move the file", but this move might be to a non existing directory, or partly non existing directory. The file might also be renamed and moved
- Reverting any failures is already quite complex and becomes just a mess with this added feature
- It will be so much easier to introduce this later, when all files and folders are based on paths. And maybe we will even use the VSCode explorer which has all this stuff built in

I just feel very strongly that this is going to bite us in the ass and become a distraction. The complexity cost is just too high with the current state of the code."*

In short summary this is an awesome PR 👍 😄 , but we are in the middle of changing out the concept of files and directories to reduce complexity and support new functionality. The complexity cost is just too high with the current state of the code. 

Feel really bad about closing this, but I hope our considerations came through. It is not you, it is us! 😂  This feature will be introduced shortly, but in a different form!

yeion7:
@christianalfoni Thanks for the explanation, I was thinking in the same way about this PR 👍

PR #2959, opened 2019-11-01
embed: show alias instead of id as fallback title

jyash97:
Improvment

In embed, we show `id` as a fallback if the title is not defined

## What is the new behavior?
Show `alias` as a fallback. Not sure if this makes sense but for consistency between app and embed it will make sense. As on app, the title is set to alias by default.

1. Local testing

## Checklist

- [ ] Documentation
- [x] Testing 
- [x] Ready to be merged
      

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/2959/commits/f800ca988ade780d34b09639011106f1627cf549 is at https://pr2959.build.csb.dev/s/new.

PR #2961, opened 2019-11-02
fix: show like button on showcase title when the user is logged in

jyash97:

Improvement. Not sure if this was intentional as the like button doesn't work but I guess its better to not show like in Editor at the top.

Like button visible in showcase even when the user is logged out.

## What is the new behavior?
Show like button only when logged in, like in editor.

- Tested this page locally

## Checklist

- [ ] Documentation- N/A
- [x] Testing 
- [x] Ready to be merged
      
### DIFF
--- a/packages/app/src/app/pages/Profile/Showcase/SandboxInfo/index.js
+++ b/packages/app/src/app/pages/Profile/Showcase/SandboxInfo/index.js
@@ -13,12 +13,13 @@ import {
   PlayButtonContainer,
 } from './elements';
 
-function SandboxInfo({ sandbox }) {
+function SandboxInfo({ sandbox, isLoggedIn }) {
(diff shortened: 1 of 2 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/2961/commits/16233aa66de8757d455071b55721db1a0df31c9d is at https://pr2961.build.csb.dev/s/new.

CompuIves:
This is really great (and that's some ancient code), thanks for this fix!

jyash97:
Hey right, the UI also seems different for showcase modal with white background. Maybe enhancing it can be the next steps. 

On other do lmk if there is something I can help you with. Always a pleasure working on CodeSandbox :)

PR #2967, opened 2019-11-03
Fix SandboxInfo

MichaelDeBoey:
Follow-up of #2961

### DIFF
--- a/packages/app/src/app/pages/Profile/Showcase/SandboxInfo/elements.ts
+++ b/packages/app/src/app/pages/Profile/Showcase/SandboxInfo/elements.ts
@@ -1,6 +1,7 @@
-import styled from 'styled-components';
 import delayEffect from '@codesandbox/common/lib/utils/animation/delay-effect';
 import { Link } from 'react-router-dom';
+import styled from 'styled-components';
+
 import { LikeHeart } from 'app/pages/common/LikeHeart';
 
 export const Container = styled.div`
@@ -18,7 +19,6 @@ export const Container = styled.div`
 `;
 
 export const Title = styled.h1`
-  font-weight: 400;
   margin: 0;
   padding: 0;
   box-sizing: border-box;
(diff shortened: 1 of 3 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/2967/commits/a6a71bd094ee4be53b4d95891613dee0a9d063aa is at https://pr2967.build.csb.dev/s/new.

packages/app/src/app/pages/Profile/Showcase/SandboxInfo/index.tsx
```diff
@@ -1,19 +1,30 @@
...
 } from './elements';
+import SvgButton from './play-button.svg';
+
+type Props = {
+  sandbox: Sandbox;
+};
+export const SandboxInfo: FunctionComponent<Props> = ({ sandbox }) => {
+  const {
+    state: { isLoggedIn },
+  } = useOvermind();
```
CompuIves: Hmm, I do like the fact that this component is "standalone" without the store, so I'd prefer to keep `useOvermind` out if it's not really needed.

packages/app/src/app/pages/Profile/Showcase/SandboxInfo/elements.ts
```diff
@@ -18,7 +19,6 @@ export const Container = styled.div`
 `;
 
 export const Title = styled.h1`
-  font-weight: 400;
```
CompuIves: This one wasn't needed?

MichaelDeBoey: A bit further down we have

```css
font-weight: 300;
```

so this one's obsolete

packages/app/src/app/pages/Profile/Showcase/SandboxInfo/index.tsx
MichaelDeBoey: This component is only used once _and_ we've done a lot of refactors where we used `useOvermind` as near as possible to the needing code.

MichaelDeBoey:
Ping @CompuIves

PR #2980, opened 2019-11-05
Add back user feedback survey behaviour

CompuIves:
With the move to Overmind we forgot to port over our userFeedbackSurvey behaviour from Cerebral. This puts it back in.

Tested the flow by removing the `if` check and confirming that the notification asking about the survey would pop up. Then verified that the modal opened with the typeform.

### DIFF
--- a/packages/app/src/app/overmind/internalActions.ts
+++ b/packages/app/src/app/overmind/internalActions.ts
@@ -9,6 +9,8 @@ import {
 } from '@codesandbox/common/lib/types';
 import { identify, setUserId } from '@codesandbox/common/lib/utils/analytics';
 
+import { notificationState } from '@codesandbox/common/lib/utils/notifications';
+import { NotificationStatus } from '@codesandbox/notifications';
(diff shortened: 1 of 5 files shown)


packages/app/src/app/overmind/internalActions.ts
```diff
@@ -68,6 +70,33 @@ export const setSignedInCookie: Action = ({ state }) => {
   setUserId(state.user.id);
 };
 
+export const showUserSurveyIfNeeded: Action = ({ state, effects, actions }) => {
+  if (state.user.sendSurvey) {
+    // Let the server know that we've seen the survey
+    effects.api.markSurveySeen();
+
+    notificationState.addNotification({
```
christianalfoni: Use `effects.notificationToast.add` API instead :)

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/2980/commits/390072ef9634614f2fd897a239911d603d1be8a6 failed.

PR #2981, opened 2019-11-05
Fix showing screenshot in the dashboard

CompuIves:
Screenshots don't get updated when a sandbox updates. For example, when you have two sandboxes next to eachother in the dashboard, and the first sandbox gets deleted, then the second sandbox gets the screenshot of the first sandbox. This is because our memoized state is not updated all the time. This makes that possible.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/2981/commits/d1ad9cf2ea543bdd10ffcc521d265a60defb32d6 is at https://pr2981.build.csb.dev/s/new.

CompuIves:
Yay!

PR #2997, opened 2019-11-07
refactored analytics, introduced vero and managing anonymous ids

christianalfoni:
What I did:

- I kept the public facing methods like `setUserId` to avoid refactoring a bunch of stuff. I rather added `setAnonymousUserId`
- I moved all services into their own files
- Vero has an internal state which checks if you have already signed in anonymously, which means it does not have an explicit `reidentify`, rather works like the rest setting an anonymous id or a user id (where it uses reidentify if already used anonymous)

**Note!** I have only tested that the scripts load and the correct public method is called. I have not tested in detail as I do not have access to all services etc. This should be tested :)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/2997/commits/592c45ebba8117298e2071e92f6956aac58a8308 is at https://pr2997.build.csb.dev/s/new.

packages/app/src/app/overmind/internalActions.ts
```diff
@@ -65,8 +66,6 @@ export const setPatronPrice: Action = ({ state }) => {
 
 export const setSignedInCookie: Action = ({ state }) => {
```
CompuIves: I think this is called twice, so we might need to add 

```
    effects.analytics.identify('signed_in', true);
    effects.analytics.setUserId(state.user.id);
```

There too.

packages/common/src/utils/analytics/utils.ts
```diff
@@ -0,0 +1,90 @@
...
+  'revealFilesInOS',
+  'keybindings.editor.searchKeyBindings',
+  'notifications.hideList',
+  'workbench.action.terminal.focus',
+  'workbench.action.console.focus',
+  'workbench.action.openRecent',
+  'code-runner.run',
+];
+
+export const DO_TRACK =
```
CompuIves: I think we need the inverse on this. Right now people who have DO_NOT_TRACK will only get DO_TRACK 😅. I still prefer to call it `DO_NOT_TRACK` since that's the feature name https://donottrack-doc.com/en/. I think I'll rename it to `DO_NOT_TRACK_ENABLED`.

PR #2941, opened 2019-10-29
Embed: Make preview/editor resizable

siddharthkp:
Add a slider to give the user the option to resize the editor/preview panels

[Figma file](https://www.figma.com/file/4vmHo4fnPHlEk4qRaKQs2q/Embed?node-id=133%3A22)

Other changes worth mentioning: 
- Removed the header
- Vertical mode is gone, mobile is also horizontally split
- Like button is floating on preview, copy link button is removed
- Added actions to refresh/open in new tab when navigation is hidden
- Disable "corner swipe to go back" for mobile

### Demo

https://pr2941.build.csb.dev/embed/5pj49

In a fake medium article: https://bygyp.csb.app/

## What steps did you take to test this?

- [x] Testing embed applies new styles
- [x] Testing sandbox remains the same

## Checklist

- [NA] Documentation
- [x] Testing 
- [x] Ready to be merged 
- [NA] Added myself to contributors table

PR #2998, opened 2019-11-07
Embed: Change default of hideNavigation to true (with backward compatibility)

siddharthkp:
One line change = fun line change

Goes well with https://github.com/codesandbox/codesandbox-client/pull/2941

By default, we show navigation bar in the embed. It's useful for sandboxes with multiple routes

That default hasn't aged well, majority of sandboxes don't have routes.

Ideally, we'd like to keep the navigation hidden unless an explicit flag asks for it. Example: `showNavigation=true`.

But that would change the behavior of existing embeds aka 🚨 backward breaking change 🚨

We can however, set `hideNavigation=1` for every new embed in the share sheet. This adds some fluff to the url but maintains backward compatibility.

### DIFF
--- a/packages/app/src/app/pages/common/Modals/ShareModal/index.js
+++ b/packages/app/src/app/pages/common/Modals/ShareModal/index.js
@@ -35,7 +35,7 @@ class ShareView extends React.Component {
     testsView: false,
     defaultModule: null,
     autoResize: false,
-    hideNavigation: false,
+    hideNavigation: true,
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/2998/commits/d67eb4c3cf5bcd925d4670324a302cc66d71959d is at https://pr2998.build.csb.dev/s/new.

Issue #323, opened 2017-11-23
.gitignore not honoured for Git commits


CompuIves:
Git ignored files still get committed.

cloggy45:
Any information as to what is causing this?

cloggy45:
Further to our little discord chat, I will look into this!

ckhicks:
This appears to still be happening - I had to remove files manually from my repo due to my gitignore not being respected. Any ideas?

cloggy45:
I looked into this problem, but I've recently been swamped with coursework.

ckhicks:
No worries, I'm a proud/paid supporter of this amazing framework so this is a small issue in light of how great the toolset is overall! Thanks for looking into it.

simplywondrous:
I'm having this same issue, where the .gitignore isn't working.

sun2rise:
Sure this is not a big issue, but if you want to use codesandbox for real projects with real data (like me), the fact that it will push to github all your keys is probably something quite limiting at the moment

janweinen:
Tried it today but the problem persists.

webskydavid:
Github support on codesandbox is still in beta phase. I think it will be fully supported in the future. It will be nice to have this option, and also a way to select files that are committed. :)

MichaelDeBoey:
Closed by #3003

srguiwiz2:
This still seems not to be working right, as described https://github.com/codesandbox/codesandbox-client/pull/3003#issuecomment-782754885 .  Not sure whether that deserves a new issue or reopening.

rayfoss:
@srguiwiz2 You set up an automatic clone of the CSB repo on another computer, just to fix Git issues? Care to elaborate on the architecture, is it headless, are you using the CSB API, do you need to re-authenticate occasionally, are you running into issues?

I've found that the .gitignore file is very special in CSB, the sync daemon ignores changes to it in the container working directory, but changes are sent down from CSB. I'm personally dealing with issues regarding #4504 , the images aren't corrupting, but it appears that CSB injects metadata to all images, modifying them, meaning I can never have the git repository and the CSB have  the same version of images. Which complicates workarounds like [rsync'ing the git repo](https://www.npmjs.com/package/g4c) with scripts

PR #3003, opened 2019-11-08
Fix honoured gitignore

yeion7:

fix

#323 

1. Create a sandbox or fork https://pr3003.build.csb.dev/s/falling-sun-uth2m
2. Add `.gitignore`
3. Create a repositorio from sandbox

## Checklist

- [ ] Documentation
- [ ] Testing 
- [x] Ready to be merged
      
### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -129,6 +129,7 @@
     "http-browserify": "^1.7.0",
     "https-browserify": "^1.0.0",
     "humps": "CompuIves/humps",
+    "ignore": "^5.1.4",
     "immer": "^3.2.0",
     "immutability-helper": "^2.6.6",
     "instantsearch.css": "^7.1.0",
--- a/packages/app/src/app/overmind/effects/git/export-to-github.ts
+++ b/packages/app/src/app/overmind/effects/git/export-to-github.ts
@@ -9,7 +9,8 @@ export default async function deploy(sandbox: Sandbox) {
     sandbox,
     sandbox.modules,
     sandbox.directories,
-    false
+    false,
+    true
   );
 
   if (!zipFile) {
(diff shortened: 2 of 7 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3003/commits/6c944bfceb428b9d0612c2c41f16f4a8624c2d55 is at https://pr3003.build.csb.dev/s/new.

CompuIves:
Man you're really good, thanks a lot for building this!

wannikid:
This fix only works for me when I export files for the first time to Github. Any following commit to the same repository however still uploads the excluded files. As an example, my .env file does not get exported to Github at first, but when I add a .env file to my sandbox later on, it is commited to Github despite being mentioned the .gitignore file. Is that the intended behaviour?

tjxnor:
Ditto @wannikid. I had a project already connected to github. Added the .gitignore and .env files later on. On making a new commit, the entries in .gitignore are not honoured.

srguiwiz2:
As git works, once committed, merely making a `.gitignore` entry won't automatically remove it.  To achieve that, one has to do a `git rm --cached path/filename.ext` and a commit, and if necessary a push.

See the answer at

https://stackoverflow.com/questions/6030530/git-ignore-not-working-in-a-directory/6030686#6030686

and a mention at

https://git-scm.com/docs/gitignore

Therefore I will now make sure to write into `.gitignore` early, and commit it, before adding files that should not be committed.

srguiwiz2:
While I got it to work once, `.gitignore` in codesandbox is overall not functioning for me.

As hints for reproducing the problems:  Overall we are using private setting, in case that matters for reproducibility.  We have made a very simple static template on a private repository we are starting from.

There is proof It is a valid `.gitignore` we use, because on a command line on a computer that has a clone of that same repository, `.gitignore` works as expected.

PR #3014, opened 2019-11-11
React Refresh

CompuIves:
Working version of React Refresh. Description will come, question for now: do we want to enable this by default for every project with React >16.9.0 or only when `react-refresh` is installed? (I prefer the first...)

This also includes some refactors to how we decide between CRA v1 and CRA v3, because I found that the code was becoming messy.

Test case: https://pr3014.build.csb.dev/s/boring-mendeleev-ge9xy 

This does 3 different steps if React 16.9.0 is installed and `react-refresh` is installed:

1. It adds a new module that gets evaluated before all the code (https://github.com/codesandbox/codesandbox-client/blob/ef46f7d89fca0e565564be2098bdaa895f2ccce9/packages/app/src/sandbox/eval/presets/create-react-app/v3.ts#L23-L32)
2. It adds another new module with helpers that we use related to react refresh (https://github.com/codesandbox/codesandbox-client/blob/ef46f7d89fca0e565564be2098bdaa895f2ccce9/packages/app/src/sandbox/eval/transpilers/react/refresh-transpiler.ts#L6-L62)
2. It adds a new transpiler for user code, which wraps the user code in a try catch that initializes refresh and adds code that calls the enqueueUpdate if all exports are react components. https://github.com/codesandbox/codesandbox-client/blob/ef46f7d89fca0e565564be2098bdaa895f2ccce9/packages/app/src/sandbox/eval/transpilers/react/refresh-transpiler.ts#L96-L99


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3014/commits/bf4c5ee726abb98975e7f3421a3f9475d653ab84 failed.

CompuIves:
Last TODO: try to find a way to circumvent a full refresh when the root of the file (eg. `App.js`) doesn't expose any exports. When there are no exports, we don't call `module.hot.accept`, which forces a refresh.

Bnaya:
I'm getting very unexpected `Invalid value used as weak map key` that seems coming from inside react refresh
Screen Shot 2020-01-12 at 0 37 14

See sandbox:
https://codesandbox.io/s/invalid-value-used-as-weak-map-key-broken-react-fresh-y2mx3

CompuIves:
Good find @Bnaya, this might be a bug in React Refresh. I've asked the React team and will follow up once I know more.

Bnaya:
Thanks!
Can you add a kill switch for react fresh?
I had to downgrade React to 16.8 to keep working on my sandbox

CompuIves:
It was indeed something in React Refresh! The fix will be live within 10 min, sorry for the inconvenience!

> Can you add a kill switch for react fresh?

I can add a dependency check of some sort for this, although I'd prefer to first see if there are more issues. This is the only one I saw so far.

PR #2912, opened 2019-10-23
New homepage

SaraVieira:
YAY?

PR #3018, opened 2019-11-11
Embeds: add secret url flag to hide dev tools

siddharthkp:
Secret URL flag to hide dev tools in embed `?hidedevtools=1`

Useful for https://github.com/codesandbox/codesandbox-client/pull/2912

⚠️ Not added to the share sheet. We need to make it nicer so that it doesn't conflict with expandDevTools


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3018/commits/16b24dba340c38709c76d050b8c575814d52788e is at https://pr3018.build.csb.dev/s/new.

packages/app/src/embed/components/Actions/index.js
```diff
@@ -46,9 +46,14 @@ export function GlobalActions({
   );
 }
 
-export function NavigationActions({ refresh, openInNewWindow, isDragging }) {
+export function NavigationActions({
```
MichaelDeBoey: Maybe make this a `FunctionComponent` with implicit return? 🤔

packages/app/src/embed/components/Actions/index.js
```diff
@@ -13,13 +13,13 @@ import {
 export function GlobalActions({
```
MichaelDeBoey: Same

packages/app/src/embed/components/Actions/index.js
siddharthkp: What's the benefit?

MichaelDeBoey: It's a React component, right? 🤔

Since we're converting every component to TypeScript and typing it as `FunctionComponent`, I think we should follow the "boyscout rule" and make this a `FunctionComponent`.

siddharthkp: sorry, I'm lost

2 questions:

1. this isn't a ts file yet, so can't type it just yet
2. does `FunctionComponent` require an explicit return?

MichaelDeBoey: 1. Indeed, I think you should make this a TS file
2. No it doesn’t enforce it, but it will make it have 1 less indentation

siddharthkp: 1. I'm still learning typescript so not yet, but soon?
2. I'll skip the aesthetic value if that's alright

PR #3032, opened 2019-11-15
Ignore error where an error is thrown when a module is not found

CompuIves:
Fixes https://sentry.io/organizations/codesandbox/issues/1333824380/?project=155188&query=first-release%3APROD-1573738286-c05030adf.

The module itself is not crucial, if it doesn't exist we also don't need to set the error array to length 0.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3032/commits/594633f66092e922db31abb7079ec28b73840b6e is at https://pr3032.build.csb.dev/s/new.

PR #2654, opened 2019-10-04
Refactored 🧠 Overmind Hacktober | patron/subscription: /app/pages/Patron/PricingModal/PricingChoice/ChangeSubscription/index.ts

NileshPatel17:
## What kind of change does this PR introduce?
This is part of the requested refactor for hacktoberfest mentioned in #2621.

## What is the current behavior?
Previously it was utilizing inject and hooksObserver from app/componentConnectors.
#2621

## What is the new behavior?
Prefers useOvermind over inject

## What steps did you take to test this?
* yarn:lint: There were several linting errors, but none relevant to the scope of this PR
* yarn:test: Ran successfully

## Checklist

- [ ] Documentation
- [x] Testing 
- [x] Ready to be merged 
- [ ] Added myself to contributors table

PR #3055, opened 2019-11-21
🔨 Switch ChangeSubscription to use useOvermind

MichaelDeBoey:
Follow-up of #2654

Things I did extra:
- Change `updateSubscriptionClicked`'s signature to accept a `string` instead of `{ coupon: string }`, because it only has 1 argument
- Type `StripeInput`'s `onChange` prop
- Type `LinkButton`'s `onClick` prop


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3055/commits/8e836b74594d192eef15bea23d9c43361371db47 is at https://pr3055.build.csb.dev/s/new.

PR #3057, opened 2019-11-21
🔨 Switch NewSandbox to use useOvermind

MichaelDeBoey:
Follow-up of #2658 & #2748

Things I did extra:


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3057/commits/036fbe829f76092de36f0c07c8d9170227a6b29a is at https://pr3057.build.csb.dev/s/new.

MichaelDeBoey:
With pleasure! 🙂

PR #2656, opened 2019-10-04
🔨 Refactored 🧠 Overmind Hacktober | /app/pages/CliInstructions/index.tsx : refactor to replace Cerebral with Overmind

iwgx:
## What kind of change does this PR introduce?

Refactors code as a part of hacktoberfest mentioned in #2621.
@Saeris @christianalfoni 

## What is the current behavior?

`packages/app/src/app/pages/CliInstructions/index.tsx` was using `inject` and `hooksObserver` from `app/componentConnectors`

## What is the new behavior?

uses `useOvermind` from `app/overmind`

## What steps did you take to test this?

run `yarn test`, `yarn typecheck` and through `lint-staged` also tried to run the app with `yarn start`

## Checklist

- [ ] Documentation
- [x] Testing 
- [x] Ready to be merged 
- [ ] Added myself to contributors table

PR #3058, opened 2019-11-21
🔨 Switch CLIInstructions to use useOvermind

MichaelDeBoey:
Follow-up of #2656

Things I did extra:


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3058/commits/bc443779f64c289db6b33ec08bf3fd6d7d6a249b is at https://pr3058.build.csb.dev/s/new.

SaraVieira:
Danke!

MichaelDeBoey:
With pleasure! 🙂

PR #2864, opened 2019-10-19
🔨 Refactor, 🧠 Overmind Hactoberfest | Refactor /app/pages/Curator/index.js to use Overmind #2841

patelrr:
## What kind of change does this PR introduce?

Refactors code as a part of hacktoberfest mentioned in #2621.
@Saeris @christianalfoni

## What is the current behavior?

app/pages/Curator/index.js was using inject and hooksObserver from app/componentConnectors

## What is the new behavior?

Converted into tsx and uses useOvermind from app/overmind

## What steps did you take to test this? 

1. Run yarn test
2. Run yarn lint
3. Run yarn typecheck

## Checklist

- [ ] Documentation
- [x] Testing 
- [x] Ready to be merged
      
- [ ] Added myself to contributors table

PR #3061, opened 2019-11-22
🔨 Switch Curator to use useOvermind

MichaelDeBoey:
Follow-up of #2864

Things I did extra:
- Make `DelayedAnimation`'s `delay` prop really optional
- Change `popularSandboxesMounted`'s signature to accept a `string` instead of `{ date: string }`, because it only has 1 argument
- Change `pickSandboxModal`'s signature to get its argument types directly from `PickedSandboxDetails` instead of hardcoding them
- Export `Curator` by a named export  instead of a `default export`

### DIFF
--- a/packages/app/src/app/pages/Curator/index.tsx
+++ b/packages/app/src/app/pages/Curator/index.tsx
@@ -76,8 +76,7 @@ export const Curator: FunctionComponent = () => {
         </SubTitle>
 
         <Buttons>
-          <span>Most popular sandboxes in the:</span>
-
+          Most popular sandboxes in the:
           <Button
             onClick={() =>
               fetchPopularSandboxes(getTime(subWeeks(new Date(), 1)))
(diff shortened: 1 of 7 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3061/commits/a6e2290b677395fa70c931fa01bcdff0489e84a6 is at https://pr3061.build.csb.dev/s/new.

packages/app/src/app/pages/Curator/index.tsx
```diff
@@ -65,7 +76,8 @@ const Curator: React.FC = () => {
         </SubTitle>
 
         <Buttons>
-          Most popular sandboxes in the:
+          <span>Most popular sandboxes in the:</span>
```
christianalfoni:
```suggestion
          Most popular sandboxes in the:
```

Not sure why you added the `span`? This causes a visual bug, missing margin. So needs to be reverted. Other than that, looks great 😄

SaraVieira:
Looks good!

PR #2663, opened 2019-10-05
🔨 Refactored 🧠 Overmind Hacktober | /Sandbox/Editor/Workspace/items/Live/index.js : refactor to replace Cerebral with Overmind

stackyism:
## What kind of change does this PR introduce?

Refactors code as a part of hacktoberfest mentioned in #2621.
@Saeris @christianalfoni

## What is the current behavior?

`/Sandbox/Editor/Workspace/items/Live/index.js` was using `inject` and `observer` from `app/componentConnectors`

## What is the new behavior?

uses `useOvermind` hook from `app/overmind`

## What steps did you take to test this?

Opened codesandbox on local with my logged-in user by replacing `jwt` token.
And while logged in tested the successful rendering of LiveInfo Component inside Live/index.js.
Successfully ran `yarn typecheck`, `yarn lint` and `yarn test`.

## Checklist

- [ ] Documentation:N/A
- [x] Testing 
- [x] Ready to be merged 
- [ ] Added myself to contributors table: N/A

PR #3063, opened 2019-11-22
🔨 Switch Live to use useOvermind

MichaelDeBoey:
Follow-up of #2663

Things I did extra:
- Change `createLiveClicked`'s signature to accept a `string` instead of `{ sandboxId: string }`, because it only has 1 argument
- Change `onChatEnabledChange`'s signature to accept a `boolean` instead of `{ enabled: boolean }`, because it only has 1 argument
- Remove unnecessary use of a `Fragment` (`<></>`)

### DIFF
--- a/packages/app/src/app/overmind/namespaces/live/actions.ts
+++ b/packages/app/src/app/overmind/namespaces/live/actions.ts
@@ -39,9 +39,10 @@ export const roomJoined: AsyncAction<{
   state.live.isLoading = false;
 });
 
-export const createLiveClicked: AsyncAction<{
-  sandboxId: string;
-}> = async ({ state, effects, actions }, { sandboxId }) => {
+export const createLiveClicked: AsyncAction<string> = async (
+  { actions, effects },
+  sandboxId
+) => {
   effects.analytics.track('Create Live Session');
 
   const roomId = await effects.api.createLiveRoom(sandboxId);
@@ -186,9 +187,9 @@ export const onSendChat: Action<{ message: string }> = (
   effects.live.sendChat(message);
 };
 
-export const onChatEnabledChange: Action<{ enabled: boolean }> = (
+export const onChatEnabledChange: Action<boolean> = (
   { effects, state },
-  { enabled }
+  enabled
 ) => {
   effects.analytics.track('Enable Live Chat');
 
--- a/packages/app/src/app/overmind/namespaces/live/actions.ts
+++ b/packages/app/src/app/overmind/namespaces/live/actions.ts
@@ -40,7 +40,7 @@ export const roomJoined: AsyncAction<{
 });
 
 export const createLiveClicked: AsyncAction<string> = async (
-  { actions, effects },
+  { actions, effects, state },
   sandboxId
 ) => {
   effects.analytics.track('Create Live Session');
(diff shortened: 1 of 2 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3063/commits/68032d3c3887f56aac9b25369b35559c48249732 is at https://pr3063.build.csb.dev/s/new.

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Live/index.tsx
```diff
@@ -1,56 +1,55 @@
...
+        roomInfo,
       },
+      isLoggedIn,
     },
   } = useOvermind();
-  const hasUnsyncedModules = !isAllModulesSynced;
-
-  const showPlaceHolder = (!isLive && !owned) || !isLoggedIn;
 
-  if (showPlaceHolder) {
```
CompuIves: I'd keep the var here in. This has a documentational purpose (what do we show if this is true?).

packages/app/src/app/overmind/namespaces/live/actions.ts
```diff
@@ -23,9 +23,10 @@ export const roomJoined: AsyncAction<{
   state.live.isLoading = false;
 });
 
-export const createLiveClicked: AsyncAction<{
-  sandboxId: string;
-}> = async ({ state, effects, actions }, { sandboxId }) => {
+export const createLiveClicked: AsyncAction<string> = async (
```
CompuIves: Interesting, wouldn't this be less explicit (eg. I now need to know what's passed by going to the function call) and scalable (what if we want to add another option)? What's your opinion on this @christianalfoni?

MichaelDeBoey: I did this in recommendation of @christianalfoni in an earlier PR, but can't find it anywhere tho 😕

packages/app/src/app/pages/Sandbox/Editor/Workspace/items/Live/index.tsx
MichaelDeBoey: Seems reasonable 🙂

packages/app/src/app/overmind/namespaces/live/actions.ts
SaraVieira: Yeah, I remember this was christian idea in new code as it makes it cleaner

SaraVieira:
@CompuIves Can you please review this?

packages/app/src/app/overmind/namespaces/live/actions.ts
MichaelDeBoey: Yeah @christianalfoni said we should defo do it in new code.
But if I wanted, I could refactor them on the go

Can't find the comment where he said that anymore tho.
Too many PRs 😂😇

christianalfoni: Ah, yeah, there is no reason to pass an object anymore to actions. It was a requirement in Cerebral. 

You could argue that the typing does not explicitly say `sandboxId` anymore, but `payload`. I just favour reduced syntax as if you were to write this action from scratch you would type it with `string`, because of our lazyness as developers 😂

SaraVieira:
YAY! All looks good to me too! Thank you!

MichaelDeBoey:
With pleasure! 🙂

Issue #3059, opened 2019-11-22
Tooltips overlap reference elements in console UI


atomiks:
# 🐛 bug report

![overlapping tippy](https://i.imgur.com/jObyzgh.png)

You can fix this by setting `boundary="window"` prop on your tippy tooltips. 

Popper v2 will have a smarter default to prevent this: https://github.com/FezVrasta/popper.js/pull/793 🎉

atomiks:
Also if you update to latest Tippy you can have singleton tippies that look cool and improve the UX of delayed tooltips: 

gif: https://twitter.com/thekitze/status/1196864091786223617
docs: https://github.com/atomiks/tippy.js-react#-singleton
demo: https://atomiks.github.io/tippyjs/#singleton-1

I'm working on a PR for these two things, but might take a while as I'm learning the codebase and where to find things 😆

siddharthkp:
Singleton tippies look great!

> I'm working on a PR for these two things, but might take a while as I'm learning the codebase and where to find things 😆

 I'll add it to my list but to be honest, same here 😄

PR #3065, opened 2019-11-22
fix(Tooltip): fix overlapping and use tippy singletons where applicable

atomiks:

Improves UX of tooltips and fixes overlapping issues. `updateDuration` (position update transition) can be changed to a more suitable value if need be

See #3059 

## What is the new behavior?

See #3059 

Unfortunately it seems like importing CSS files in `Tooltip/index.tsx` is breaking Jest. Any help on this? 

```bash
yarn test
yarn lint
```

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      
### DIFF
--- a/packages/app/src/app/pages/Sandbox/Editor/Header/Buttons/RefreshButton/UpdateFound/UpdateFound.tsx
+++ b/packages/app/src/app/pages/Sandbox/Editor/Header/Buttons/RefreshButton/UpdateFound/UpdateFound.tsx
@@ -23,8 +23,8 @@ const useInterval = (callback = () => {}, delay: number) => {
 };
 
 export const UpdateFound = (props: any) => {
-  const [isVisible, setVisibility] = useState(true);
-  useInterval(() => setVisibility(false), isVisible ? 60000 : null);
+  const [visible, setVisibility] = useState(true);
+  useInterval(() => setVisibility(false), visible ? 60000 : null);
 
   return (
     <UpdateContainer {...props}>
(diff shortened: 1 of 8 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3065/commits/8031d3ed8a5f31412c2a6450521fd955a6ba207a is at https://pr3065.build.csb.dev/s/new.

CompuIves:
Nice!! Thanks for this contribution, I really appreciate it, especially since you're the author of the library and helping us with this. 

I pushed a fix for the tests. Did you also test the behaviour by running locally? As soon as that's confirmed we can merge this!

atomiks:
> Did you also test the behaviour by running locally? As soon as that's confirmed we can merge this!

I'm not sure what you mean here. I developed it locally but couldn't run the Jest tests as I mentioned :-(. You can manually test it in the deployed preview to see the behavior.

Also something to note: there is an accessibility problem. We added `tabindex="0"` to the wrapping `<span>` in the prev major, but now that doesn't get added anymore. It should be added back, or preferrably the wrapper `<span>` is removed so that `<Tippy />` uses the native element which is much better for the screen readers and keyboard nav

arthurdenner:
@atomiks, could you solve the conflicts here to see if we can get it merged, please?

SaraVieira:
All seems to work great and I fixed the conflicts 🗡 


Thank you so much!

PR #3050, opened 2019-11-20
Google Fonts picker - remove need for google api

SaraVieira:
Google api was needed because of the display of the font in his style and with that removed there is no need to use an api key

PR #2768, opened 2019-10-13
🔨 Refactor, 🧠 Overmind | pages/Sandbox/Editor/Workspace/Dependencies/AddVersion/index.js

Infi-Knight:
## What kind of change does this PR introduce?

Refactors code as a part of hacktoberfest mentioned in #2621.
@Saeris @christianalfoni

## What is the current behavior?

`app/pages/Sandbox/Editor/Workspace/Dependencies/AddVersion/index.js` was using Cerebral

## What is the new behavior?

`AddVersion` now uses `overmind` hook from the new state management tool, Overmind.

## What steps did you take to test this? This is required before we can merge.

1. Replaced inject and hooksObserver with useOvermind
2. Convert `AddVersion` class component to functional component.
3. Add `searchDependencies` to the list of values the `ModalName` type can take in `app/overmind/actions.ts`
4. Use named export in `AddVersion` and update `pages/Sandbox/Editor/Workspace/Dependencies/index.tsx` accordingly.
5. Ran `yarn test`
6. Ran `yarn lint`
## Checklist

- [ ] Documentation
- [x] Testing 
- [x] Ready to be merged
      
- [ ] Added myself to contributors table

PR #2671, opened 2019-10-06
🔨 Refactored 🧠 Overmind Hacktober | /app/pages/Sandbox/Editor/Workspace/Dependencies/AddVersion/index.js 

silltho:
## What kind of change does this PR introduce?

Refactor code for hacktoberfest #2621 
@Saeris @christianalfoni @SaraVieira

## What is the current behavior?

`Sandbox/Editor/Workspace/Dependencies/AddVersion/index.js`  was using `inject` and `observer` from `app/componentConnectors`

## What is the new behavior?

uses `useOvermind` from `app/overmind`

## What steps did you take to test this?

ran `yarn typecheck`and `yarn lint`

## Checklist

- [ ] Documentation N/A
- [X] Testing 
- [X] Ready to be merged 
- [X] Added myself to contributors table

PR #2670, opened 2019-10-06
🔨 Refactored and 🧠 Overmind ...Dependencies/elements and ...Dependencies/index

mkamranhamid:
…
## What kind of change does this PR introduce?

I refactored `/app/pages/Sandbox/Editor/Workspace/Dependencies/index.js` and `/app/pages/Sandbox/Editor/Workspace/Dependencies/index.js` and when I run the app I was getting some errors so I had to update `app/pages/Sandbox/Editor/Workspace/itemsNotOwnedSandboxInfo/NotOwnedSandboxInfo` which was importing `Dependencies` as default export, changed it to import it as named export.
@Saeris and @christianalfoni please review it.

## What is the current behavior?

Previously it was utilizing inject, hooksObserver and default export now we are using `useOvermind` with named export. 
[#2621](https://github.com/codesandbox/codesandbox-client/issues/2621)

## What is the new behavior?

Using `useOvermind` instead of `inject` and `hooksObserver`

## What steps did you take to test this?

`yarn lint`
`yarn test`
Everything runs like a charm. It had some warning but not related to this PR

## Checklist

- [ ] Documentation N/A
- [x] Testing 
- [x] Ready to be merged  N/A
- [ ] Added myself to contributors table  N/A

I am not able to add 🔨 Refactor, 🧠 Overmind, and #Hacktoberfest labels. Can you please add them.

PR #3066, opened 2019-11-22
🔨 Switch Dependencies to use useOvermind

MichaelDeBoey:
Follow-up of #2768, #2670, #2671, #2328 & #3050

Things I did extra:
- Change `npmDependencyRemoved`'s signature to accept a `string` instead of `{ name: string }`, because it only has 1 argument
- Change `externalResourceAdded`'s & `externalResourceRemoved`'s signature to accept a `string` instead of `{ resource: string }`, because it only has 1 argument
- Type `FontList`'s `updateSearch` function better
- Make `FontPicker`, `AddFont`, `ExternalFonts` & `ExternalResource` a `FunctionComponent`
- Just use a single file for `ExternalFonts` to not pollute the project files unnecessarily


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3066/commits/ac00327e15b471ac93174be83641cf5ce661fc95 is at https://pr3066.build.csb.dev/s/new.

MichaelDeBoey:
Thanks! 🙂

PR #3067, opened 2019-11-22
🔨 Switch CreateNewSandbox to use useOvermind

MichaelDeBoey:
Follow-up of #2720

Things I did extra:
- Make `CreateNewSandbox` a `FunctionComponent` (instead of `SFC`, since it's [deprecated](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f13238a5d2086cd87f3209704f8897f7f701d957/types/react/index.d.ts#L497-L503)) and improve it's `Props` type
- Cleanup `CreateNewSandbox` a bit (`Spring`'s `from` & `to` prop + the way `mostUsedSandboxComponent` is created)
- Export `CreateNewSandbox` by a named export  instead of a `default export`

### DIFF
--- a/packages/app/src/app/overmind/namespaces/dashboard/actions.ts
+++ b/packages/app/src/app/overmind/namespaces/dashboard/actions.ts
@@ -65,7 +65,7 @@ export const searchChanged: Action<{ search: string }> = (
 };
 
 export const createSandboxClicked: AsyncAction<{
+  body: { collectionId: string };
   sandboxId: string;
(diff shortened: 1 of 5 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3067/commits/c000f60f77d96b15eb52af7d25bf3862334559c4 is at https://pr3067.build.csb.dev/s/new.

PR #2737, opened 2019-10-12
🔨 Refactored 🧠 Overmind Hacktober | /app/pages/Dashboard/Sidebar/TemplateItem/TemplateItem.tsx : refactor to remove Cerebral

sakthivel-tw:
## What kind of change does this PR introduce?

Refactors code as a part of hacktoberfest mentioned in #2621.
@Saeris @christianalfoni @SaraVieira

## What is the current behavior?

`/app/pages/Dashboard/Sidebar/TemplateItem/TemplateItem.tsx` was using inject and observer from app/componentConnectors

## What is the new behavior?

No reference to `signals` or `store` from inject and removed the dependencies.

## What steps did you take to test this? This is required before we can merge.

ran `yarn lint` and `yarn typecheck`

## Checklist

- [ ] Documentation
- [x] Testing 
- [x] Ready to be merged
      
- [x] Added myself to contributors table - A different user was added to the contributors list instead of me. https://github.com/codesandbox/codesandbox-client/pull/2650

PR #3070, opened 2019-11-22
🔨 Switch TemplateItem to use useOvermind

MichaelDeBoey:
Follow-up of #2737

Things I did extra:


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3070/commits/f3ea7064e4ef2061c728ea9b49ef082ecd956f85 is at https://pr3070.build.csb.dev/s/new.

packages/app/src/app/pages/Dashboard/Sidebar/TemplateItem.tsx
```diff
@@ -0,0 +1,80 @@
...
+  currentPath,
+  isOver,
+  teamId,
+}) => {
+  const path = teamId
+    ? `/dashboard/teams/${teamId}/templates`
+    : `/dashboard/templates`;
+
+  return connectDropTarget(
+    <Item
```
christianalfoni: This needs to be wrapped in a `div`, cause the drag/drop library does not support it being a component 😄

packages/app/src/app/pages/Dashboard/Sidebar/TemplateItem.tsx
```diff
@@ -0,0 +1,80 @@
...
+    return {
+      [MAKE_TEMPLATE_DROP_KEY]: true,
+      teamId,
+    };
+  },
+};
+
+const collectTarget: DropTargetCollector<CollectedProps> = (
+  { dropTarget },
+  { canDrop, isOver }
```
christianalfoni: You still have to use `connectMonitor` and `monitor` here. The reason is that internally they use `this`. When you destructure, the functions looses their `this` context 😄

PR #2759, opened 2019-10-13
Refactor | Overmind: Refactor /app/pages/common/Modals/PreferencesModal/CodeFormatting/Prettier/index.js

NileshPatel17:
## What kind of change does this PR introduce?
Refactor code for hacktoberfest #2621

## What is the current behavior?
```/app/pages/common/Modals/PreferencesModal/CodeFormatting/Prettier/index.js```  was using inject and observer from app/componentConnectors

## What is the new behavior?
uses useOvermind from app/overmind and js file converted to tsx file

## What steps did you take to test this? This is required before we can merge.
1. yarn:test: Ran successfully
2. yarn:lint: There were several linting errors, but none relevant to the scope of this PR
## Checklist

- [ ] Documentation
- [x] Testing 
- [x] Ready to be merged
      
- [ ] Added myself to contributors table

PR #3073, opened 2019-11-22
🔨 Switch CodeFormatting to use useOvermind

MichaelDeBoey:
Follow-up of #2759

Things I did extra:
- Export `Prettier` by a named export  instead of a `default export`
- Type `Prettier`'s `bindValue` function


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3073/commits/89537fc11bca4064d379a071dd4891fce87b3e79 is at https://pr3073.build.csb.dev/s/new.

PR #3085, opened 2019-11-26
✨ Add Prototyping Sandbox Components to Common

Saeris:
Extracted Common components from #2225 and new components from the Prototyping sandbox into a separate commit to make it easier to migrate Profiles development.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3085/commits/b48579eab2b3cb30697b9e94c654cf888858e7a0 failed.

packages/common/package.json
```diff
@@ -62,7 +62,7 @@
     "react": "^16.8.6",
     "react-icons": "^2.2.7",
     "react-input-autosize": "^2.2.1",
-    "react-router-dom": "^4.3.1",
+    "react-router-dom": "^5.0.1",
```
siddharthkp: oh boy, there's a router in common 🤔 curious if it clashes with the router in the app

Saeris: We should keep these in sync, but the reason why is we're building primitives around React Router's Link component. Don't think we actually have any route tables in Common.

siddharthkp: aah, okay. It might be worth looking into how to move that control to the app instead, it's not ideal for common components to have routing preferences

(not urgent/critical, but a good to have)

CompuIves:
Would it be possible to leave the buttons and template related code out of this PR for now? I've also changed this code in the `follow-templates` branch and I think this will be a source of possible merge conflicts. Other than that it looks nice!

Saeris:
Build is passing now but there's at least one bug that's causing the Dashboard to crash that I'm going to look into. We're not using `common/Button` for every button in the app it would seem, at least the typecheck errors I had to fix were all related to using inline styles which is a practice we should move away from anyways. The only place I'm observing visual bugs with the Button component is in CodeSadbox, which I'll debug too.

packages/app/src/app/pages/Dashboard/Content/CreateNewSandbox/NewSandboxModal/Imports/StackbitButton.tsx
```diff
@@ -32,13 +32,18 @@ function openStackbit(username: string) {
...
-  style?: React.CSSProperties;
 }
 
-export const StackbitButton = ({ username, style }: Props) => (
-  <Button style={style} small onClick={() => openStackbit(username)}>
+export const StackbitButton: React.FC<IStackbitButtonProps> = ({
+  username,
+}) => (
+  <Button
+    css="margin-top: 1rem; float: right;"
```
SaraVieira: I don't think this works, what is this css?

Saeris: the CSS prop is part of styled components, the babel plugin reads that prop and generates classnames, it's an alternative to using `styled`. Otherwise you have to include `style` in your component interface, if you don't typescript yells at you. This is the preferred way to do inline styles, though we should avoid it as much as possible.

Saeris: for reference:

https://www.styled-components.com/docs/api#css-prop

CompuIves:
There are still some subtle differences in button styling:

New:
PixelSnap 2019-11-28 at 15 24 48

Old:
PixelSnap 2019-11-28 at 15 25 43

Is it possible to not have the style changes to the button component? I think we should reserve that for the design system update.

Saeris:
> Is it possible to not have the style changes to the button component? I think we should reserve that for the design system update.

Yeah I think we can safely revert those style changes. Looks to me like the only difference is the padding, correct?

packages/app/src/app/pages/Dashboard/Sidebar/index.js
```diff
@@ -135,12 +135,7 @@ class Sidebar extends React.Component {
         </Route>
 
         <div style={{ margin: '2rem', fontSize: '.875rem' }}>
-          <Button
-            style={{ display: 'block' }}
```
CompuIves: Shouldn't we need to keep `display: 'block'` here?

Saeris: No because it displays the same with that extra style being set or not. The `block` prop ensures that in this particular instance, the button fills 100% of the width of it's container. I think this would only matter if the underlying element was an inline element, such as an `a` tag, which `block` also solves, as it sets the display of the element to `flex` versus `inline-flex`.

Saeris:
Please check again if you can find any more errors. Should have fixed all the bugs you reported with the latest commit.

jyash97:
Button's background is the same in all themes.  If this is intentional, ignore the comment.

Check the difference in `Share`, `Fork` and `Sign In` buttons:

Screenshot from main site: 
image

Screenshot from deployed PR URL:
image

MichaelDeBoey:
I don't think this PR is still relevant, since we now have `@codesandbox/components` and a lot is already changed to that? 🤔

CC @SaraVieira

SaraVieira:
Sorry drake, closing this as we already a components package 

Thanks!!!

PR #3096, opened 2019-11-28
Pro checkout

siddharthkp:

- This PR adds a `/pro` page to the app for signing up the Pro plan
- Adds an `itemId` to the modalOpened action so that we can directly open a tab in the preference modal


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3096/commits/99b8b9c7519d1d212212e3ff672a85f3820e3748 is at https://pr3096.build.csb.dev/s/new.

packages/app/src/app/pages/Pro/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+    },
+  } = useOvermind();
+
+  useEffect(() => {
+    patronMounted();
+  }, [patronMounted]);
+
+  // if you're already a patron, you shouldn't
+  // try to get pro.
+  if (isPatron) location.href = '/patron';
```
MichaelDeBoey:
```suggestion
  if (isPatron) {
    location.href = '/patron';
  }
```

siddharthkp: outdated :)

siddharthkp: don't look at Draft PRs :P

packages/app/src/app/pages/Pro/index.tsx
```diff
@@ -103,7 +125,7 @@ const Pro: React.FC = () => {
                           </Button>
 
                           <HelpText>
-                            You will be billed on the <b>30th</b> of each month.
+                            You will be billed on the <b>29th</b> of each month.
```
MichaelDeBoey: Shouldn't this be `28th`, since February can also have only 28 days? 🤔

packages/app/src/app/overmind/actions.ts
```diff
@@ -79,13 +79,15 @@ type ModalName =
...
-) => {
+export const modalOpened: Action<{
+  modal: ModalName;
+  message?: string;
+  itemId?: string;
+}> = ({ state, effects }, { modal, message, itemId }) => {
   effects.analytics.track('Open Modal', { modal });
   state.currentModalMessage = message;
   state.currentModal = modal;
+  state.preferences.itemId = itemId;
```
MichaelDeBoey: This means `currentModal` needs to be set to `'preferences'`, so not sure if this is the right approach 🤔

packages/app/src/app/pages/Pro/index.tsx
siddharthkp: copied pasted from `/patron` 🤷‍♀

packages/app/src/app/overmind/actions.ts
siddharthkp: Yeah, it's not my favorite. 

Added a check for `currentModal` similar to `modalClosed`. Ideally I'd like to send `data` which is deepmerged in to the object

```js
modalOpened({
  modal: 'preferences',
  data: { itemId: 'paymentInfo' }
});
```

But this would only work for modals, not the rest of the overmind actions. Would talk to @christianalfoni for best practices in a different pull request

MichaelDeBoey: I think adding a `data` key that will be merged is indeed a much cleaner way to accomplish this 🙂

siddharthkp:
It's done, ship it! 🚢

packages/app/src/app/pages/Pro/index.tsx
siddharthkp: turns out they were completely wrong 😅. we charge folks the same day they got their subscription :P

PR #3098, opened 2019-11-29
fix: cross-origin error

### LINKED ISSUE #667: how to get rid of cross-origin error?

AndyOGo:
Set [CSP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) for [`'script-src'`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src) to allow to run JavaScript from `eval` or `Function` Constructor and access stacktrace beyond cross-origin boundary.

Another great article about CSP and `eval`:
https://developer.chrome.com/extensions/contentSecurityPolicy#JSEval

fixes #667 

Bugfix

JavaScript executed by `eval` or `Function` constructor is not allowed to access stuff like stacktraces beyond own origin according to default CSP rules. For more details see issue description at #667 

## What is the new behavior?

CSP headers are reset, to allow JS within eval and Function constructor to access stuff like stacktraces cross-origin:

```html
<meta http-equiv="Content-Security-Policy" content="script-src 'self' 'unsafe-eval'">
```

## Checklist

- [ ] Documentation
- [x] Testing 
- [ ] Ready to be merged
      
## Sample code to reproduce

Here is a reproduction script:
https://codesandbox.io/embed/ly0q02y9q9

```js
import React, { createElement, PureComponent } from "react";
import ReactDOM from "react-dom";
import val from "@skatejs/val";

import "./styles.css";

const h = val(createElement);

class CECrossOriginError extends HTMLElement {
  // properties getter
  static get observedAttributes() {
    return ["error"];
  }

  set error(value) {
    const error = Boolean(value);
    if (error) this.setAttribute("error", "");
    else this.removeAttribute("error");
  }

  get error() {
    return this.hasAttribute("error");
  }

  connectedCallback() {
    this.render();
  }

  attributeChangedCallback() {
    this.render();
  }

  render() {
    const { error } = this;

    this.className = error ? "has-error" : "";
  }
}

// Register the new element with the browser.
customElements.define("ce-cross-origin-error", CECrossOriginError);

class CrossOriginError extends PureComponent {
  render() {
    /** @jsx h */
    // can be fixed by extracting children, like:
    // const { children, ...props } = this
    // return <CECrossOriginError {...props}>{children}</CECrossOriginError>;
    const { props } = this;

    return <CECrossOriginError {...props} />;
  }
}

function App() {
  return (
    <div className="App">
      <h1>Hello CodeSandbox</h1>

      <CrossOriginError>foo bar</CrossOriginError>
    </div>
  );
}

const rootElement = document.getElementById("root");
ReactDOM.render(<App />, rootElement);
```

### DIFF
--- a/packages/app/src/app/index.html
+++ b/packages/app/src/app/index.html
@@ -4,6 +4,8 @@
 <head>
 	<meta charset="utf-8">
 	<meta name="viewport" content="width=device-width, initial-scale=1">
+  <!-- fix cross-origin error, by allowing to run JavaScript from eval or Function constructor source -->
+  <meta http-equiv="Content-Security-Policy" content="script-src 'self' 'unsafe-eval'">
 	<meta property="og:type" content="website">
 	<meta property="og:author" content='https://ivesvh.com'>
 	<meta property="og:title" content="CodeSandbox">
--- a/packages/app/src/app/index.html
+++ b/packages/app/src/app/index.html
@@ -5,7 +5,7 @@
 	<meta charset="utf-8">
 	<meta name="viewport" content="width=device-width, initial-scale=1">
   <!-- fix cross-origin error, by allowing to run JavaScript from eval or Function constructor source -->
-  <meta http-equiv="Content-Security-Policy" content="script-src 'self' 'unsafe-eval'">
+  <meta http-equiv="Content-Security-Policy" content="script-src 'self' 'unsafe-eval' 'unsafe-inline' https://*.googleapis.com">
 	<meta property="og:type" content="website">
 	<meta property="og:author" content='https://ivesvh.com'>
 	<meta property="og:title" content="CodeSandbox">
--- a/packages/app/src/app/index.html
+++ b/packages/app/src/app/index.html
@@ -5,7 +5,7 @@
 	<meta charset="utf-8">
 	<meta name="viewport" content="width=device-width, initial-scale=1">
   <!-- fix cross-origin error, by allowing to run JavaScript from eval or Function constructor source -->
-  <meta http-equiv="Content-Security-Policy" content="default-src 'self' * data: mediastream: blob: filesystem:; script-src 'self' 'unsafe-eval' 'unsafe-inline' * data: mediastream: blob: filesystem:; style-src 'self' 'unsafe-eval' 'unsafe-inline' * data: mediastream: blob: filesystem:">
+  <meta http-equiv="Content-Security-Policy" content="default-src 'self' 'unsafe-eval' 'unsafe-inline' * data: mediastream: blob: filesystem:">
 	<meta property="og:type" content="website">
 	<meta property="og:author" content='https://ivesvh.com'>
 	<meta property="og:title" content="CodeSandbox">
--- a/packages/app/src/app/index.html
+++ b/packages/app/src/app/index.html
@@ -5,7 +5,7 @@
 	<meta charset="utf-8">
 	<meta name="viewport" content="width=device-width, initial-scale=1">
   <!-- fix cross-origin error, by allowing to run JavaScript from eval or Function constructor source -->
-  <meta http-equiv="Content-Security-Policy" content="script-src 'self' 'unsafe-eval' 'unsafe-inline' https://*.googleapis.com">
+  <meta http-equiv="Content-Security-Policy" content="default-src: 'self' data: mediastream: blob: filesystem: *; script-src 'self' 'unsafe-eval' 'unsafe-inline' *">
 	<meta property="og:type" content="website">
 	<meta property="og:author" content='https://ivesvh.com'>
 	<meta property="og:title" content="CodeSandbox">
--- a/packages/app/src/app/index.html
+++ b/packages/app/src/app/index.html
@@ -5,7 +5,7 @@
 	<meta charset="utf-8">
 	<meta name="viewport" content="width=device-width, initial-scale=1">
   <!-- fix cross-origin error, by allowing to run JavaScript from eval or Function constructor source -->
-  <meta http-equiv="Content-Security-Policy" content="default-src 'self' * data: mediastream: blob: filesystem:; script-src 'self' 'unsafe-eval' 'unsafe-inline' *">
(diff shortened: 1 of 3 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3098/commits/3262c837e140f5e3aa3c2b2a2e93381d6df883e2 is at https://pr3098.build.csb.dev/s/new.

AndyOGo:
Something is misisng:

```sh
Refused to load the script 'https://ajax.googleapis.com/ajax/libs/webfont/1.6.26/webfont.js' because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-eval'". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.

new:1 Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-eval'". Either the 'unsafe-inline' keyword, a hash ('sha256-F8S4lQpr9oEYTIUgykvYcJLRXuZpuY5apEYv2Q7rrbc='), or a nonce ('nonce-...') is required to enable inline execution.

new:1 Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-eval'". Either the 'unsafe-inline' keyword, a hash ('sha256-dwjtN/BH0nVflw1NQvE7EEJB9GUgS4B6ACePIFUCnk4='), or a nonce ('nonce-...') is required to enable inline execution.

new:1 Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-eval'". Either the 'unsafe-inline' keyword, a hash ('sha256-JVEvEpunbRUCHlL1DaxutF4fTTwTC5UfOCb9GsJxqH4='), or a nonce ('nonce-...') is required to enable inline execution.
```

AndyOGo:
@lbogdan 
@CompuIves 

The react cross-origin error is now gone.
[Edit vibrant-dawn-ylpwy](https://pr3098.build.csb.dev/s/vibrant-dawn-ylpwy?fontsize=14&hidenavigation=1&theme=dark)

Here is a screenshot:
Screenshot 2019-11-29 at 12 23 22

packages/app/src/app/index.html
```diff
@@ -4,6 +4,8 @@
 <head>
 	<meta charset="utf-8">
 	<meta name="viewport" content="width=device-width, initial-scale=1">
+  <!-- fix cross-origin error, by allowing to run JavaScript from eval or Function constructor source -->
+  <meta http-equiv="Content-Security-Policy" content="script-src 'self' 'unsafe-eval' 'unsafe-inline' https://*.googleapis.com">
```
AndyOGo: @lbogdan 
@CompuIves 
Are there any other domain, which are need to load scripts?
Or should be any domain allowed? In that case just `*` would be best.

I have omitted to write a [`default-src` CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/default-src), which [should be always provided](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP#Writing_a_policy)...

jdolearydl:
This is very necessary, especially when using React Hooks.  Good work Andy

CompuIves:
Hey! Thanks for taking the time to fix this, this is an important bug. I'm currently taking a look at this sandbox: https://codesandbox.io/s/reactjs-popup-issue-template-f8ibg, but it doesn't seem to be fixed for me in the pr: https://pr3098.build.csb.dev/s/f8ibg. Is there anything specific I need to check?

AndyOGo:
@CompuIves 
Thanks for your answer.

Though I get a proper stack trace for your example without any cross-origin error.
It just says `AudioCalibrator` at line 45 is `undefined` at runtime.

Screenshot 2019-12-06 at 00 10 54

I just moved the class `AudioCalibrator` and `MovingAvg` up before `App`.
https://codesandbox.io/s/reactjs-popup-issue-template-fm17c

**But:** if I only move the class `AudioCalibrator` up and leave the `MovingAvg` below `App`, then I still get a cross-origin error:
https://pr3098.build.csb.dev/s/reactjs-popup-issue-template-q6yw3

Meanwhile I will allow any origin `*` and add `default-src`.

packages/app/src/app/index.html
```diff
@@ -4,6 +4,8 @@
 <head>
 	<meta charset="utf-8">
 	<meta name="viewport" content="width=device-width, initial-scale=1">
+  <!-- fix cross-origin error, by allowing to run JavaScript from eval or Function constructor source -->
+  <meta http-equiv="Content-Security-Policy" content="default-src: 'self' data: mediastream: blob: filesystem: *; script-src 'self' 'unsafe-eval' 'unsafe-inline' *">
```
AndyOGo: @CompuIves 
@lbogdan 
Alright, I enabled any schema including `data:`, `blob:`, `meadiastream:` and `filestream:` and any domain for **default-src**.

Maybe other rules are still necessary for other types of content.
https://content-security-policy.com/

AndyOGo:
> Content Security Policy: Ignoring source ‘frame-ancestors’ (Not supported when delivered via meta element).

AndyOGo:
@CompuIves 
@lbogdan 

Your reproduction still hasn't access to errors 🤔 
- https://pr3098.build.csb.dev/s/reactjs-popup-issue-template-q6yw3
- https://codesandbox.io/s/reactjs-popup-issue-template-fm17c

E.g.: this one also try to access a undefined variable inside `ReactDOM.reander()`, but works now:
- https://pr3098.build.csb.dev/s/react-q1xc8
- https://codesandbox.io/s/react-88o0s

I hope I did not not missed any necessary CSP rule.
Anyway the resort which comes to my mind is `Access-Control-Allow-Origin: *` and `crossorigin` attribute as described here:
https://reactjs.org/docs/cross-origin-errors.html

AndyOGo:
Also broken:
https://pr3098.build.csb.dev/s/pedantic-mountain-ndspp

PR #3106, opened 2019-12-01
WIP work on making live joining while typing more smooth

CompuIves:
Main problem:

When someone joins a live session while the other user is typing, the first few characters are not processed because VSCode hasn't initialized yet. 

My "fixes"

1. Introduce a queue for live messages and ensure that the queue is not processed in parallel (should this logic be in `live` instead maybe?
2. Make `live:module_state` more aggressive. This is our repair mechanism for when there is a state mismatch, it wasn't useful before, but now that we can call `vscode.setCode` it actually is more useful so I do more checks and make sure to call the module state when there is a mismatch. This works, but I had to add `this.applyingOperation = true`, which doesn't feel clean to do in a generic function that's called `setCode`, maybe you know what to do here? @christianalfoni 

So the interesting thing is, fix number 1 doesn't even work. I tried, but it seems to apply the operations on the saved code instead of module state code? My hunch is that the module state hasn't returned yet, and so the operations are applied on saved code instead of module state code. We maybe need to have an initialized state or something on live to fix that.

Fix number 2 fixes everything of course, but we should not resort to it that often.

This is WIP, just a couple of suggestions on fixes. But it's very quickly done, and ugly.

### DIFF
--- a/packages/app/src/app/overmind/effects/vscode/ModelsHandler.ts
+++ b/packages/app/src/app/overmind/effects/vscode/ModelsHandler.ts
@@ -123,6 +123,17 @@ export class ModelsHandler {
     return moduleModel.model;
   };
 
+  public async setModelCode(module: Module, code: string) {
+    const moduleModel = this.getModuleModel(module);
+
+    if (moduleModel && moduleModel.model) {
+      const model = await moduleModel.model;
+      this.isApplyingOperation = true;
+      model.setValue(code);
+      this.isApplyingOperation = false;
+    }
+  }
+
   public async updateTabsPath(oldPath: string, newPath: string) {
     const oldModelPath = '/sandbox' + oldPath;
     const newModelPath = '/sandbox' + newPath;
@@ -161,7 +172,7 @@ export class ModelsHandler {
     // promise, because there might be multiple operations fired before
     // the model is actually resolved. This creates a "natural" queue
     if (!moduleModel.model) {
-      if (modelEditor) {
+      if (modelEditor && modelEditor.textModelReference) {
         moduleModel.model = modelEditor.textModelReference.then(
           ref => ref.object.textEditorModel
         );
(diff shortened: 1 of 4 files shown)


packages/app/src/app/overmind/effects/vscode/ModelsHandler.ts
```diff
@@ -161,7 +172,7 @@ export class ModelsHandler {
     // promise, because there might be multiple operations fired before
```
CompuIves: I'm wondering with this comment, wouldn't the operations be applied all on the same version of code? Not 100% sure.

packages/app/src/app/overmind/effects/vscode/ModelsHandler.ts
```diff
@@ -384,15 +395,19 @@ export class ModelsHandler {
     let index = 0;
     const currentEOLLength = model.getEOL().length;
     let eolChanged = false;
+    const modelCode = model.getValue();
+    const modelLines = modelCode.split(/\n/) || [];
+
+    if (operation.baseLength !== modelCode.length) {
```
CompuIves: This is our more aggressive check.

packages/app/src/app/overmind/effects/vscode/index.ts
```diff
@@ -528,139 +571,143 @@ export class VSCodeEffect {
   }
 
   private async loadEditor(monaco: any, container: HTMLElement) {
-    this.monaco = monaco;
-    this.workbench = new Workbench(monaco, this.controller, this.runCommand);
+    return new Promise(async vscodeResolved => {
```
CompuIves: Oh yes, I wrapped `loadEditor` in a promise and only let it resolve when the editor layout is mounted. I'm not sure if this is better?

packages/app/src/app/overmind/namespaces/live/actions.ts
```diff
@@ -90,11 +90,15 @@ export const liveMessageReceived: Operator<LiveMessage> = pipe(
...
+export const applyTransformation: AsyncAction<{
   operation: any;
   moduleShortid: string;
-}> = ({ effects }, { operation, moduleShortid }) => {
-  effects.vscode.applyOperation(moduleShortid, operation);
+}> = async ({ effects }, { operation, moduleShortid }) => {
+  try {
+    await effects.vscode.applyOperation(moduleShortid, operation);
+  } catch (e) {
+    effects.live.send('live:module_state', {});
```
CompuIves: This is the "repair" call.

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3106/commits/8e149474dc80b663c0e1d8909a7311abbd5a3b56 is at https://pr3106.build.csb.dev/s/new.

PR #3107, opened 2019-12-01
also write module state changes

christianalfoni:


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3107/commits/6a6966f802a247993b718426b3765e73f6222729 is at https://pr3107.build.csb.dev/s/new.

packages/app/src/app/overmind/effects/vscode/ModelsHandler.ts
```diff
@@ -185,6 +185,19 @@ export class ModelsHandler {
...
 
+  public async setModuleCode(module: Module) {
+    const moduleModel = this.getModuleModel(module);
+    const model = await moduleModel.model;
+
+    if (!model) {
+      return;
+    }
+
+    this.isApplyingOperation = true;
```
CompuIves: We might need to make this an option in the future, I can think of some use cases where we want to set the code and send the changes. But not needed for now!

packages/app/src/app/overmind/effects/vscode/index.ts
```diff
@@ -216,7 +216,11 @@ export class VSCodeEffect {
     moduleShortid: string,
     operation: (string | number)[]
   ) {
-    return this.modelsHandler.applyOperation(moduleShortid, operation);
+    if (!this.modelsHandler) {
```
CompuIves: Poh, so the operation gets dropped? I mean, I understand it because we sync state, maybe we need to put a comment here. I can imagine someone seeing this as a bug 😂

packages/app/src/app/overmind/effects/vscode/ModelsHandler.ts
```diff
@@ -384,6 +397,12 @@ export class ModelsHandler {
     let index = 0;
     const currentEOLLength = model.getEOL().length;
     let eolChanged = false;
+    const modelCode = model.getValue();
+
+    if (operation.baseLength !== modelCode.length) {
+      throw new Error('Length mismatch');
```
CompuIves:
```suggestion
      throw new Error("The base length of the operation doesn't match the length of the code");
```

packages/app/src/app/overmind/namespaces/live/internalActions.ts
```diff
@@ -103,30 +97,35 @@ export const initializeModuleState: Action<any> = (
...
       // If we are getting code that has not been saved, we want to update the actual saved code
       // so that we properly detect "changed files" in the explorer. But we only do this if the
       // savedCode is not set, as we might get multiple initializeModuleState
-      if (
-        !moduleInfo.synced &&
-        state.editor.currentSandbox.modules[index].savedCode == null
-      ) {
-        state.editor.currentSandbox.modules[index].savedCode =
-          state.editor.currentSandbox.modules[index].code;
+      if (!moduleInfo.synced && module.savedCode !== moduleInfo) {
```
CompuIves: Isn't `module.savedCode !== moduleInfo` always false?

christianalfoni:
```suggestion
      if (!moduleInfo.synced && !module.savedCode) {
```

Oh, it should be this. Not sure what happened there.

christianalfoni:
```suggestion
      if (!moduleInfo.synced && module.savedCode === null) {
```

PR #3108, opened 2019-12-01
QOL improvements

CompuIves:
Some extra fixes:

1. Change the syncing logic to only sync on change
2. We didn't sync if you join a sandbox with a live session as owner (`/s/:id`), I moved the syncing to live initialize to make sure that it always happens
3. We now debounce the selection updates when typing.

### DIFF
--- a/packages/app/src/app/overmind/effects/live/index.ts
+++ b/packages/app/src/app/overmind/effects/live/index.ts
@@ -259,9 +259,6 @@ export default {
       message,
     });
   },
-  sendModuleState() {
-    return this.send('live:module_state', {});
-  },
   sendModuleSaved(module: Module) {
     return this.send('module:saved', {
       type: 'module',
@@ -272,7 +269,7 @@ export default {
   sendChatEnabled(enabled: boolean) {
     return this.send('live:chat_enabled', { enabled });
   },
-  sendModuleUpdateRequest() {
+  sendModuleStateSyncRequest() {
     return this.send('live:module_state', {});
   },
   sendUserSelection(moduleShortid: string, liveUserId: string, selection: any) {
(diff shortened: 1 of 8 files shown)


packages/app/src/app/overmind/effects/live/index.ts
```diff
@@ -272,7 +269,7 @@ export default {
   sendChatEnabled(enabled: boolean) {
     return this.send('live:chat_enabled', { enabled });
   },
-  sendModuleUpdateRequest() {
+  sendModuleStateSyncRequest() {
```
christianalfoni: ❤️

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3108/commits/d8da95060d7b6cb22114eae64474bfe3d69c4aef is at https://pr3108.build.csb.dev/s/new.

packages/app/src/app/overmind/namespaces/live/internalActions.ts
```diff
@@ -106,16 +110,28 @@ export const initializeModuleState: Action<any> = (
...
+      const moduleChanged =
+        moduleInfo.code !== module.code ||
+        moduleInfo.saved_code !== module.savedCode;
+
+      if (moduleChanged) {
+        module.savedCode = moduleInfo.saved_code;
+        module.code = moduleInfo.code;
+
+        if (savedCodeChanged) {
+          effects.vscode.sandboxFsSync.writeFile(
```
christianalfoni: This writes the `module.code` to the filesystem. We only want to do this if the `module.code` is the saved code, AKA the module is `synced`, AKA the `savedCode` is `null`?

CompuIves: Hmm, you're right. But what if savedCode has changed but the module is dirty? Maybe we need to change the writeFS call to use getSavedCode as well...

CompuIves:
Great @yeion7!! I can take a look today. I think keeping the folder open sounds good, your reasoning is good.

PR #3109, opened 2019-12-01
Fix parsedConfigurations type

MichaelDeBoey:


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3109/commits/d0f6dfd9f4acf594b8d0b3abc521aeade188b770 failed.

packages/app/src/app/pages/Sandbox/Editor/Workspace/Dependencies/index.tsx
```diff
@@ -32,26 +32,27 @@ export const Dependencies: FunctionComponent = () => {
     resource => !resource.includes('fonts.googleapis.com/css')
   );
 
-  if (!parsedConfigurations.package) {
+  if (!parsedConfigurations?.package) {
     return <ErrorMessage>Unable to find package.json</ErrorMessage>;
   }
 
-  const { parsed, error } = parsedConfigurations.package;
+  const { error, parsed } = parsedConfigurations.package;
```
arthurdenner: I think we need to add a default value to `parsed` here. Otherwise, it'll continue to break because it may be `undefined` when destructuring on line 46.

MichaelDeBoey: No we don't need to, since we now for sure `error` is undefined on line 46, since if it wasn't we would've returned on line 41.

MichaelDeBoey:
If codesandbox/codesandbox-importers#54 gets merged in, we can revert the changes on line 35 again 🙂

packages/app/babel.config.js
```diff
@@ -22,4 +22,5 @@ module.exports = {
       ],
     },
   },
+  plugins: ['@babel/plugin-proposal-optional-chaining'],
```
CompuIves: This is for making tests work?

packages/app/src/app/overmind/namespaces/editor/state.ts
```diff
@@ -56,7 +59,7 @@ type State = {
   mainModule: Derive<State, Module>;
   currentPackageJSON: Derive<State, Module>;
   currentPackageJSONCode: Derive<State, string>;
-  parsedConfigurations: Derive<State, any>;
+  parsedConfigurations: Derive<State, ParsedConfigurationFiles> | null;
```
CompuIves: Is it possible for them to be null?

packages/app/babel.config.js
MichaelDeBoey: It's for making the `build` script work

packages/app/src/app/overmind/namespaces/editor/state.ts
MichaelDeBoey: https://github.com/codesandbox/codesandbox-client/blob/1aca6b669358bc42a3487f2b0e3b5cb33c8ca5be/packages/app/src/app/overmind/namespaces/editor/state.ts#L145-L146

MichaelDeBoey:
@CompuIves All green now 🙂

If we can get codesandbox/codesandbox-importers#54 to be merged, I can revert some type changes (L35) in this one and then this one's good to be merged too 🙂

CompuIves:
Nice! I've published those changes as `2.1.11`

packages/app/babel.config.js
CompuIves: Hmm, we don't use this file in `app`, that's what confuses me. We only use this file for running tests. The other babel configs (`babel.dev.js` and `babel.prod.js`) already have the optional chaining libs.

packages/app/src/app/overmind/namespaces/editor/state.ts
CompuIves: Hm, but at the same time, `currentSandbox: Derive<State, Sandbox>` doesn't have ` | null`. What's your opinion on this @christianalfoni? Should we make this `| null`?

The tricky thing is that when we're in `/s/:id` we can assume that `currentSandbox` is set, and thus that `parsedConfigurations` is true. A perfect solution would be to make this nullable and say that for a specific path/actions we can assume that the value is set, but that seems impossible to do.

packages/app/babel.config.js
MichaelDeBoey: Running the `build` script sets the `env` to `production`, so I can't include it in `test` environment

https://github.com/codesandbox/codesandbox-client/blob/a98b774783bd183539996f181b3ec5ed00ced84e/packages/app/package.json#L12

packages/app/src/app/overmind/namespaces/editor/state.ts
MichaelDeBoey: I don't think that's possible to type tbh.

In our code we can force it to be set using the bang operator (`!`) in specific parts or the code tho.

packages/app/babel.config.js
CompuIves: So removing this entry breaks the prod build? I thought we used optional chaining already, this babel config is not used in prod.

MichaelDeBoey: Yeah this breaks build indeed.

We only use optional chaining in `common`, not in `app` atm.

MichaelDeBoey:
@CompuIves Hmmm only build on `drone` seems to fail, but the error-message doesn't make sense.

Maybe I need to update the cache to a new version? 🤔

MichaelDeBoey:
Seems like merging after your beaf88f commit was causing the drone error, so rebased and fixed it 🙂

packages/app/babel.config.js
CompuIves: Ah, right, I also did this in the `follow-templates` branch. It would be best to add the plugin to `babel.dev.js` and `babel.prod.js`, as we set all plugins there.

MichaelDeBoey: Moved the plugin to `babel.dev.js` & `babel.prod.js`

MichaelDeBoey:
@CompuIves All green now 🙂

packages/app/src/app/overmind/namespaces/editor/state.ts
christianalfoni: Even if we are on `/s/:id` the Sandbox needs to be downloaded, so it is not necessarily there :)

MichaelDeBoey: So I can just leave the type nullable right?

packages/app/src/app/overmind/namespaces/editor/internalActions.ts
```diff
@@ -189,7 +191,7 @@ export const updateCurrentTemplate: AsyncAction = async ({
       templateDefinition.isServer ||
       state.editor.parsedConfigurations.sandbox.parsed.template
     ) {
-      const { parsed } = state.editor.parsedConfigurations.package;
+      const { parsed = {} } = state.editor.parsedConfigurations!.package!;
```
christianalfoni:
```suggestion
      const parsed = state.editor.parsedConfigurations?.package?.parsed || {};
```

packages/app/src/app/overmind/namespaces/editor/internalActions.ts
```diff
@@ -189,7 +191,7 @@ export const updateCurrentTemplate: AsyncAction = async ({
       templateDefinition.isServer ||
       state.editor.parsedConfigurations.sandbox.parsed.template
```
christianalfoni:
```suggestion
      state.editor.parsedConfigurations?.sandbox?.parsed?.template
```

packages/app/src/app/overmind/namespaces/editor/state.ts
christianalfoni: Jup!

christianalfoni:
I have tested changing template of a sandbox, looks good! 👍 😄

PR #3114, opened 2019-12-02
fix(svelte): add fallback toUrl method to generate sourcemaps

### LINKED ISSUE #3094: Svelte 3.12+ stuck transpiling modules
# 🐛 bug report

## Description of the problem
Any version of Svelte 3.12 and above hangs on transpiling modules and never loads.

Unable to use most recent versions of Svelte

### Link to sandbox: [link](https://codesandbox.io/s/modest-wind-fpzv0)

arthurdenner:
Fix. Closes #3094. 

Svelte replaced `magic-string` with `code-red` on version `3.13.0` and with this, CodeSandbox can't compile Svelte components anymore because it relies on `toUrl` to [generate inline sourcemaps](https://github.com/codesandbox/codesandbox-client/blob/master/packages/app/src/sandbox/eval/transpilers/svelte/svelte-worker.js#L143).

## What is the new behavior?
I've added a fallback function to generate the source maps like [`magic-string` does](https://github.com/Rich-Harris/magic-string/blob/master/src/SourceMap.js#L27).

I've opened [a PR](https://github.com/Rich-Harris/code-red/pull/22) to add this method to `code-red` but I'm not sure if it's the right approach or will even get merged. However, if we want to output sourcemaps, we'd need this because since 3.13.0, Svelte is "broken" on this matter.

1. Open [this sandbox](https://codesandbox.io/s/modest-wind-fpzv0);
2. Verify that it compiles without the fallback if you change `svelte`s version to 3.12.1;
3. Verify that it compiles with the fallback if you change `svelte`s version to 3.13.0 or more.

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      
## Notes
What are your thoughts on this approach?
Should we **not** generate inline sourcemaps?

### DIFF
--- a/packages/app/src/sandbox/eval/transpilers/svelte/svelte-worker.js
+++ b/packages/app/src/sandbox/eval/transpilers/svelte/svelte-worker.js
@@ -36,6 +36,12 @@ function getV3Code(code, version, path) {
       });
     });
 
+    if (!js.map.toUrl) {
+      js.map.toUrl = () =>
+        'data:application/json;charset=utf-8;base64,' +
+        btoa(JSON.stringify(js.map));
+    }
+
     return js;
   } catch (e) {
     return self.postMessage({
--- a/packages/app/src/sandbox/eval/transpilers/svelte/svelte-worker.js
+++ b/packages/app/src/sandbox/eval/transpilers/svelte/svelte-worker.js
@@ -36,6 +36,8 @@ function getV3Code(code, version, path) {
       });
     });
 
+    // Fallback to generate sourcemaps on Svelte 3.13+
+    // See https://github.com/codesandbox/codesandbox-client/pull/3114
     if (!js.map.toUrl) {
       js.map.toUrl = () =>
         'data:application/json;charset=utf-8;base64,' +


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3114/commits/5c7202d67fecc38cdc961780c42350cf47c4fe0e is at https://pr3114.build.csb.dev/s/new.

packages/app/src/sandbox/eval/transpilers/svelte/svelte-worker.js
```diff
@@ -36,6 +36,14 @@ function getV3Code(code, version, path) {
       });
     });
 
+    // Fallback to generate sourcemaps on Svelte 3.13+
+    // See https://github.com/codesandbox/codesandbox-client/pull/3114
+    if (!js.map.toUrl) {
```
siddharthkp: this is beautiful, thanks!

CompuIves:
Very nice!! Thank you!

PR #3118, opened 2019-12-03
🔨 Switch VSCodePlaceholder to use useOvermind

MichaelDeBoey:
I also removed all `children` where `VSCodePlaceholder` is used, since `settings.experimentVSCode` is always `true`.

Since `KeyMapping` is now just `VSCodePlaceholder`, this component can maybe be deleted too? 🤔


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3118/commits/51971a8b5fb3ddcad3c937b67b9a32933701aaf8 failed.

PR #3120, opened 2019-12-03
🔨 Switch AliasDeploymentButton to use useOvermind

MichaelDeBoey:
Resubmission of #2844

Follow-up of #2742

Things I did extra:
- Change `aliasDeployment`'s signature to accept a `string` instead of `{ id: string }`, because it only has 1 argument


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3120/commits/de0c1d3d0abd9e626177c6ce07b47db855bbf1a1 failed.

SaraVieira:
Looks good!

PR #2623, opened 2019-10-03
🔨 Refactored app/pages/CLI/index.tsx

HTMLGhozt:
## What kind of change does this PR introduce?

This is part of the requested refactor for hacktoberfest mentioned in #2621.
@Saeris @christianalfoni 

## What is the current behavior?
Previously it was utilizing `inject` and `hooksObserver` from `app/componentConnectors`.

#2621 

## What is the new behavior?
Prefers `useOvermind` over `inject` and other small stylistic choices mentioned in #2621 

## What steps did you take to test this?

- `yarn:lint`: There were several linting errors, but none relevant to the scope of this PR
- `yarn:test`: Ran successfully

## Checklist

- [ ] Documentation - N/A
- [x] Testing 
- [x] Ready to be merged 
- [ ] Added myself to contributors table 

I was unable to add the 🔨 Refactor, 🧠 Overmind, or Hacktoberfest labels.

PR #3121, opened 2019-12-03
🔨 Switch CLI to use useOvermind

MichaelDeBoey:
Resubmission of #2848

Follow-up of #2623

Things I did extra:
- Use a `ref` for `TokenInput`'s `onClick` instead of using the `MouseEvent`'s `target`
- Export `CLI` by a named export  instead of a `default export`

### DIFF
--- a/packages/app/src/app/pages/CLI/Prompt/index.tsx
+++ b/packages/app/src/app/pages/CLI/Prompt/index.tsx
@@ -1,26 +1,19 @@
-import React from 'react';
 import { Button } from '@codesandbox/common/lib/components/Button';
-import { Title } from 'app/components/Title';
+import React, { FunctionComponent, useRef } from 'react';
+
 import { SubTitle } from 'app/components/SubTitle';
-import { Container, Buttons, TokenContainer } from './elements';
+import { Title } from 'app/components/Title';
+import { useOvermind } from 'app/overmind';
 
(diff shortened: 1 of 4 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3121/commits/da67f70b8fe34b931f4ef33fed44bc4cf58e6edf failed.

packages/app/src/app/pages/CLI/Prompt/index.tsx
```diff
@@ -45,13 +38,15 @@ export const Prompt: React.FC<IPromptProps> = ({
...
+          <Button onClick={() => signInCliClicked()}>
+            Sign in with GitHub
+          </Button>
         </Buttons>
       </Container>
     );
   }
 
-  if (loading) {
+  if (isLoadingCLI) {
```
SaraVieira: Where did this come from?

Was it changed in the state?

MichaelDeBoey: This was first passed to `Prompt` as a prop by `CLI`, but can just be taken from the `useOvermind` hook

SaraVieira: OHH

Cool! Will merge :)

Issue #3124, opened 2019-12-04
Can't move sandbox from "My Sandboxes" to a folder


mtorcellini:
<h1>💥 Crash Report</h1>

  <h2>What were you trying to accomplish when the crash occurred?</h2>

  <p>Clicking and attempting to drag a sandbox from within the My Sandboxes directory into one of my folders within My Sandboxes causes the sandbox to disappear from view (as soon as I begin dragging) and I cannot move it. This occurs in both Firefox and Safari. If I release the mouse button, the sandbox does NOT reappear as it should, but remains hidden from view until I refresh the page.</p>

  <h3>Link to sandbox: <a href="https://codesandbox.io/s/l60-build-an-unordered-list-using-a-data-model-bz4qt" target="_blank">https://codesandbox.io/s/l60-build-an-unordered-list-using-a-data-model-bz4qt</a></h3>

  <h3>Crash Details</h3>

  <details>
  <summary>Environment</summary>

  | Browser |  Version  | Operating System | CodeSandbox Version |
  | ------- | --------- | ---------------- | ------------------- |
  | firefox | 70.0.0 | OS X 10.15           | PROD-1575388352-84201088e          |

  **Route:**
  https://codesandbox.io/dashboard/recent

  </details>

  <details>
  <summary>Error Message</summary>

  ```bash
  TypeError: this is undefined
  ../../node_modules/react-dnd/lib/createTargetMonitor.js/i</e.prototype.getItem@https://codesandbox.io/static/js/common.8156f37c.chunk.js:1:1602240
canDrop@https://codesandbox.io/static/js/app.247e9aec.js:449:178960
../../node_modules/react-dnd/lib/createTargetFactory.js/t.default/t</t.prototype.canDrop@https://codesandbox.io/static/js/common.8156f37c.chunk.js:1:1601187
../../node_modules/dnd-core/lib/DragDropMonitorImpl.js/s</e.prototype.canDropOnTarget@https://codesandbox.io/static/js/common.8156f37c.chunk.js:1:619307
../../node_modules/react-dnd/lib/createTargetMonitor.js/i</e.prototype.canDrop@https://codesandbox.io/static/js/common.8156f37c.chunk.js:1:1601995
./src/app/index.js/Jd<@https://codesandbox.io/static/js/app.247e9aec.js:449:179160
[...]

PR #3127, opened 2019-12-04
🔨 Switch TemplateItem to use useOvermind

MichaelDeBoey:
Resubmission of #3070 after 37223ea, with a fixed `entryTarget` (see #3124)

Follow-up of #2737

Things I did extra:


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3127/commits/b271a13c05f8851dbf6396d1378625d685e927b7 is at https://pr3127.build.csb.dev/s/new.

PR #3128, opened 2019-12-04
Improve experience of enabling VIM

christianalfoni:
It is easy to miss the comment that a browser refresh is necessary for VIM to be enabled/disabled. Also it is not really VIM being enabled, it is the extension itself being enabled or not. You control its enabled/disabled state from the command palette


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3128/commits/c798e00f7532040e53a8b394672664585b164937 is at https://pr3128.build.csb.dev/s/new.

PR #3132, opened 2019-12-05
fix preview reload on history changes

christianalfoni:
When changing urls with "pushState" / "history" it does a refresh now. This PR fixes that. Tested using history and location. Location does full refresh, as it should, while history just changes the url.

This issue might have been introduced realted to `vscodeeffects` as some changes was necessary related to "smooth" forking.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3132/commits/68acb4f6d339be98736557bf39035eaa525c23b8 is at https://pr3132.build.csb.dev/s/new.

PR #3161, opened 2019-12-10
Fall back to image if hardware acceleration is disabled

CompuIves:
We now fall back to an image instead on an animation in the hero if we see
3 frames of less than 5fps consecutively.


packages/homepage/src/screens/home/hero/BoxAnimation/index.js
```diff
@@ -129,6 +150,32 @@ export default function App({ boxes, showPlane }) {
     }, 300);
   }, [boxes, set]);
 
+  if (animationDisabled) {
+    return (
+      <div
+        style={{
+          position: 'absolute',
```
SaraVieira: please create elements for this instead of adding the css like this :(

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3161/commits/1c69a9d4929ad586aab427b16c702918cb9f6775 is at https://pr3161.build.csb.dev/s/new.

packages/homepage/src/screens/home/hero/BoxAnimation/index.js
CompuIves: Changed it! Also changed another one, this whole file is a mess because of how quickly it was built.

PR #3163, opened 2019-12-10
Change API response typing of sandbox

CompuIves:
I want to remove these fields from the API response, since we don't use it anyways (and never did). This will make the API respond faster. We were using `SmallSandbox` before, which uses `viewCount`, `forkCount` and `likeCount`. The SQL queries to this are expensive, and if we can get rid of it it's only better :D


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3163/commits/40be4341e9cdb2202b86c81d10fa47799d461a3d is at https://pr3163.build.csb.dev/s/new.

PR #3170, opened 2019-12-10
do smooth forking, but restart executor and refresh preview

christianalfoni:
Now we still do a smooth fork, but we fire up the executor on the forked sandbox and we refresh the preview so that we see the container fires up :)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3170/commits/a3e21adf0b9f7eb95536976dfcd2612e7f37402e is at https://pr3170.build.csb.dev/s/new.

PR #3176, opened 2019-12-11
Fix renaming forked sandboxes

### LINKED ISSUE #3175: No longer able to successfully rename sandbox via app bar
# 🐛 bug report

## Description of the problem

When I try to rename a sandbox from the auto-generated name, I get the following error message:

> alias: has already been taken

To reproduce:

* Need to be logged in. The behavior does not seem to be present when using it as an anonymous user. I have a patron account and I sign in using GitHub.
* Start a new React sandbox via https://codesandbox.io/s/new
* Make a change to index.js and save so that the title portion of the app bar changes from "React <kbd>Template</kbd>" to "My Sandboxes / generated-name-as6df"
* Click on the generated name in the app bar and edit it (it doesn't seem to matter what I change it to)

If instead I edit the name via clicking the pencil icon in the "SANDBOX INFO" side bar, I am able to successfully change the name, and after changing the name that way I am then able to make further changes to the name successfully using the app bar.

I rename my sandboxes so that I can tell the purpose of each one when looking through my sandboxes later. Over the past week I have been unable to do this (only when writing up this bug report did I discover that it was possible to successfully rename via the "SANDBOX INFO" side bar).

### Your Environment

| Сodesandbox      |
| Browser          | Chrome 78
| Operating System |

CompuIves:
Fixes #3175

When renaming a sandbox users would get an error saying "alias has already been taken", which is thrown because we send an alias that has been taken in the payload.

I thought to resolve this by conditionally adding title, description and alias to the payload based on whether they were changed, but there was another issue that arose from this.

After a fork we don't set the `project` fields to the forked sandbox, so we always think that the alias changed when changing the name of a sandbox. I updated this too.

### DIFF
--- a/packages/app/src/app/overmind/namespaces/workspace/actions.ts
+++ b/packages/app/src/app/overmind/namespaces/workspace/actions.ts
@@ -106,31 +106,15 @@ export const sandboxInfoUpdated: AsyncAction = withOwnedSandbox(
 
       effects.analytics.track(`Sandbox - Update ${event}`);
 
-      const payload: {
-        title?: string;
-        description?: string;
-        alias?: string;
-      } = {};
+      sandbox.title = project.title;
+      sandbox.description = project.description;
+      sandbox.alias = project.alias;
 
-      if (hasChangedTitle) {
-        payload.title = project.title;
-        sandbox.title = project.title;
-      }
-
-      if (hasChangedDescription) {
-        payload.description = project.description;
-        sandbox.description = project.description;
-      }
-
-      if (hasChangedAlias) {
-        payload.alias = project.alias;
-        sandbox.alias = project.alias;
-      }
-
-      const updatedSandbox = await effects.api.updateSandbox(
-        sandbox.id,
-        payload
-      );
+      const updatedSandbox = await effects.api.updateSandbox(sandbox.id, {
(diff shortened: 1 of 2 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3176/commits/fd71dde441c9be78794498dc68fc5d3a3605dddb is at https://pr3176.build.csb.dev/s/new.

packages/app/src/app/overmind/namespaces/workspace/actions.ts
```diff
@@ -106,15 +106,31 @@ export const sandboxInfoUpdated: AsyncAction = withOwnedSandbox(
...
+      }
+
+      if (hasChangedAlias) {
+        payload.alias = project.alias;
+        sandbox.alias = project.alias;
+      }
+
+      const updatedSandbox = await effects.api.updateSandbox(
+        sandbox.id,
+        payload
```
christianalfoni: Hm, could you not just update the sandbox with the project? Then update the sandbox? Does not matter if it has changed or not?

CompuIves: So you mean sending the sandbox as payload?

PR #3181, opened 2019-12-11
SandboxCard: refactor and remove observing

christianalfoni:
Do not see why we used "observer" here, so removed it and updated types etc. This is the card on the Curator Dashboard. I tested by opening it and clicking "pick sandbox", seems to be good!


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3181/commits/cd779c45cf8b8283d29e58b70db2b7928f191ed4 is at https://pr3181.build.csb.dev/s/new.

PR #3192, opened 2019-12-12
Privacy popup

siddharthkp:
Add a way to change privacy of a sandbox - which is a pro feature right now

Design: https://www.figma.com/file/CB73VJsBOHZYiqFYV4NxBS/Go-Pro


packages/app/src/app/pages/Sandbox/Editor/Header/elements.ts
```diff
@@ -47,6 +47,7 @@ export const Centered = styled.div`
   display: flex;
   justify-content: center;
   z-index: 0; /* So pointer events on left and right continue */
+  z-index: 10; /* override: this probably breaks something, not sure what */
```
siddharthkp: @CompuIves @christianalfoni Overriding this, what should I be testing for this?

packages/app/src/app/pages/Sandbox/Editor/index.tsx
```diff
@@ -94,7 +94,7 @@ const ContentSplit: React.FC = () => {
               right: 0,
               bottom: bottomOffset,
               height: statusBar ? 'auto' : 'calc(100% - 3.5rem)',
-              zIndex: 999999,
+              zIndex: 9,
```
siddharthkp: @CompuIves @christianalfoni I'm sure this breaks something, any tips on where should I start looking?

packages/app/src/app/pages/Sandbox/Editor/Header/PrivacyTooltip/elements.ts
```diff
@@ -0,0 +1,72 @@
+import styled from 'styled-components';
+import css from '@styled-system/css';
+
+export const Container = styled.div(
+  css({
+    '.tippy-content': {
```
siddharthkp: Override tippy design inside this popup. Would be nice when we can move this to common

packages/app/src/app/pages/Sandbox/Editor/Header/PrivacyTooltip/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+                    You can change privacy of a sandbox as a Pro.
+                    <br />
+                    <Link href="/pricing">Upgrade to Pro</Link>
+                  </>
+                )}
+              </Text>
+
+              <Select
+                marginBottom={2}
+                value={privacy}
```
siddharthkp: @christianalfoni Could use some help here. 

```
Property 'value' does not exist on type 'IntrinsicAttributes'
```

I'm not using value inside the styled component, but we need to type this inside the element?

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3192/commits/ade67388a81d3b4cde7dab3966bdb3b822b6a684 is at https://pr3192.build.csb.dev/s/new.

packages/app/src/app/pages/Sandbox/Editor/Header/PrivacyTooltip/index.tsx
SaraVieira: I think you do because the value tag does not exist in the select tag, its a react invention to keep us crying

siddharthkp: ohhh, the semantic version is have a `<option selected>`

should we use that instead?

SaraVieira: if we have that react complains so it's all about who you want to hear complain ahah

siddharthkp: oh lol, didn't know that. okay then i'll prefer typescript complaining

SaraVieira: Yeah, just add the value to the styled-component

CompuIves: Ah, `as` support in TypeScript strikes again! I've pushed a commit which forces the component to be seen as a `select`, but this is really not ideal. @christianalfoni what do you think?

CompuIves:
When I'm not signed in I see this, with the disabled select: 
image

Would it make sense to let the user know to sign in or something? Or refer to Pro anyway?

CompuIves:
What do you think @DannyRuchtie ?

CompuIves:
When seeing the sandbox you don't own there seems to be a margin missing:

image
image

CompuIves:
I can change the privacy as a patron, but there is still a message for upgrading to pro, would it make sense to remove it in that case?

image

packages/app/src/app/pages/Sandbox/Editor/Header/PrivacyTooltip/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  return (
+    <ThemeProvider theme={theme}>
+      <Container>
+        <Tooltip
+          delay={0}
+          interactive
+          content={
+            <>
+              <Text size="3" marginBottom={4}>
+                {!user || !user.subscription ? (
```
CompuIves:
```suggestion
                {user?.subscription ? (
```

This might fix it?

siddharthkp:
whoa, these are all bugs, not how it was intended

siddharthkp:
@CompuIves Added check for owner which takes care of logged out case as well

siddharthkp:
> I can change the privacy as a patron, but there is still a message for upgrading to pro, would it make sense to remove it in that case?

You shouldn't be seeing this as a patron 🤔 

The logic is on `user.subscription` which both pro and patron have

packages/app/src/app/pages/Sandbox/Editor/Header/PrivacyTooltip/index.tsx
siddharthkp: Took this in

siddharthkp: Whoop back to: `Cannot read property 'subscription' of null`

DannyRuchtie:
Screenshot 2019-12-13 at 16 26 12

The selectbox is not rendering correctly on safari. 
I know people have opinions about this. But for an application, I would like to strive for an as holistic experience as much as possible. Let me know if you need help

PR #3193, opened 2019-12-12
Add ui and docs for port

### LINKED ISSUE #3088: Document the `container` option
# 🌈 Document the `container` option

## The feature
Expand Sandbox Configuration documentation to explain the `container` option of `sandbox.config.json`.  (Assuming a good set of sub-properties, like `port`, are ready for use.)

## How it could work
The [Codesandbox Documentation on Configuration](https://codesandbox.io/docs/configuration) should be updated to show the `container` option currently exists.  Since it is an object option and applicable only to containers, it should probably have a special section after the Sandbox Configuration section. The layout can be very similar to the Sandbox Configuration section. Every sub-option of `container` could be listed with option, description, default value, and possible values columns.

## An example of what the documentation might look like

**Container Configuration**
If your sandbox is a container environment, then you also have access to the `container` option in  `sandbox.config.json`. We support these properties:

Property | Description | Possible Values | Default Value
--- | --- | --- | ---
`port` | The port which the IDE browser window listens to | `0 - 65535` | `random*`

Currently accessing the `container` option requires manually editing the `sandbox.config.json`. As an example:
```json
{
  "template": "node",
  "container": {
    "port": 1234
  }
}
```

*I'm not actually sure what the default config value for `port` is 😅 

## An example of why it is useful
If you manually edit `sandbox.config.json` you can currently  declare the `container` property. This allowed me to achieve [existing client-side parcel environment behavior in a node container](https://codesandbox.io/s/elated-dew-je4cj). [The default port parcel serves pages under is 1234](https://parceljs.org/cli.html#port), so the trick was to tell the container to listen to that port.

```json
{
  "template": "node",
  "container": {
    "port": 1234
  }
}
```
This example illustrates no `codesandbox-client` code needs to be updated to use the `container` option.  Documenting that the option exists, and that it can be set via raw editing, should prove helpful to some users.

SaraVieira:

Docs Update

#3088 

## What is the new behavior?

Add docs and UI to allow to change the port if you are in a server template

closes #3088

1. Open server template
2. change port
3. open non server one to maker sure its not an option


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3193/commits/cd26c4b7b8d42ce998e8f03b213d6e908c8099a5 is at https://pr3193.build.csb.dev/s/new.

packages/homepage/content/docs/3-configuration.md
```diff
@@ -35,3 +35,5 @@ support these options:
 | `infiniteLoopProtection` | Whether we should throw an error if we detect an infinite loop                                    | `true`/`false`                                                                                                     | `true`                                             |
 | `hardReloadOnChange`     | Whether we should refresh the sandbox page on every change, good for sandboxes with global state. | `true`/`false`                                                                                                     | `false`                                            |
 | `template`               | Which sandbox template to use                                                                     | [see here](https://github.com/codesandbox-app/codesandbox-importers/blob/master/packages/types/index.d.ts#L24-L39) | smart detection, w/ fallback to `create-react-app` |
+| `container`              | Options inside the container object to                                                            |                                                                                                                    |                                                    |
```
garethx: 'Options inside the container object to' - to what? I don't understand this description, can we re-word?

SaraVieira: That was terrible english

SaraVieira: What I meant was these options are inside the container object

PR #3208, opened 2019-12-13
LivePage: refactor LivePage and make changedShortIds a derived

christianalfoni:
Okay, so the `LivePage` has been tested here and should work now. Related to looking into this I saw that we should rather derive `changedShortIds`. It is not being used much now, but it allowed us to remove a bunch of code. This change is not very well tested, not sure about all scenarios.

I also noticed an issue with unsaved code and changing sandbox. The code is saved nevertheless, and in LIVE mode it causes an error. We need to fix this, but should be a different PR 😄


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3208/commits/35fb6158deb00fea09b5cfeb689ed4922f35e0ea is at https://pr3208.build.csb.dev/s/new.

packages/app/src/app/overmind/namespaces/editor/actions.ts
```diff
@@ -70,7 +70,7 @@ export const sandboxChanged: AsyncAction<{ id: string }> = withLoadApp<{
     return;
   }
 
-  await effects.vscode.closeAllTabs();
+  // await effects.vscode.closeAllTabs();
```
CompuIves: Why this close?

christianalfoni: Oh, it should actually be there. The reason it closes is because the tabs references models that references files of the previous Sandbox. 

I just needed to test something and it sneaked in

christianalfoni:
Okay, so I changed how we manage signing into a Live session, when not signed in with a profile. The `when` concept introduced is not really a good concept. The reason is that it lives beyond the context of the page. Which means we might wait for a signin, but the user might move to a different page and then sign in, causing the Live logic to continue its execution.

Instead I created a new action for signing in to a Live session. This logs the user in first.

packages/app/src/app/overmind/namespaces/live/actions.ts
```diff
@@ -36,7 +61,7 @@ export const roomJoined: AsyncAction<{
   effects.live.sendModuleStateSyncRequest();
   effects.vscode.openModule(state.editor.currentModule);
   effects.preview.executeCodeImmediately({ initialRender: true });
-  state.live.isLoading = false;
```
CompuIves: Did we remove `state.live.isLoading`?

packages/app/src/app/pages/Live/BlinkingDot.tsx
```diff
@@ -12,7 +12,8 @@ const DotContainer = styled.div`
   }
 `;
 
-export class BlinkingDot extends React.PureComponent {
+export class BlinkingDot extends React.PureComponent<{}, { showing: boolean }> {
+  timer: NodeJS.Timer;
```
CompuIves: I believe in the browser case we need `number` here. Then use `this.timer = window.setInterval`.

CompuIves: Not blocking from merging, but for reference later.

CompuIves:
During testing: when joining a live session I initially see the sign in button, then the loading live. Is there a way to not show the button? I think we can use `state.hasLogIn`, since that property just checks if there's a `jwt`.

packages/app/src/app/overmind/namespaces/live/actions.ts
christianalfoni: Yeah, this is being set inside `live.initialize`, no reason to do it twice :-)

christianalfoni:
> During testing: when joining a live session I initially see the sign in button, then the loading live. Is there a way to not show the button? I think we can use `state.hasLogIn`, since that property just checks if there's a `jwt`.

Okay, now it is flipped. It shows "joining..." and then if not authenticated it shows the button instead of continuing into the Live session :)

packages/app/src/app/pages/Live/BlinkingDot.tsx
christianalfoni: Ah, yeah, hm... weird typing behaviour. Was not aware that `window.setInterval` and `setInterval` was typed differently. Fixed

PR #3209, opened 2019-12-14
Update preview header

SaraVieira:
This PR updates the preview header to the new design you can see here:

https://www.figma.com/file/Pn3qsrxzqFN0s6tB2eTbe6SC/IDE?node-id=1175%3A4920

New:

Screenshot 2019-12-14 at 15 13 24
Screenshot 2019-12-14 at 15 10 22

Colors will never match because of vscode themes.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3209/commits/394d57e947aa810a541961ee3344b6e1529cc793 is at https://pr3209.build.csb.dev/s/new.

siddharthkp:
> Colors will never match because of vscode themes.

Not even matching for codesandbox black?

SaraVieira:
@siddharthkp Those are pretty close but all is done with fancy draken and lightens because of the lack opf inconsistencies in themes :(

packages/app/src/app/components/Preview/DevTools/Tabs/Tab/elements.ts
```diff
@@ -33,9 +33,9 @@ export const Tab = styled.div<{ active: boolean; isOver: boolean }>`
   ${props =>
     props.active
       ? css`
-          background-color: ${props.theme['tab.activeBackground'] ||
-            props.theme['editor.background'] ||
-            'transparent'};
+          background-color: ${props.theme['input.background'] ||
```
siddharthkp: input background instead of tab background? america explain

packages/common/src/components/Preview/Navigator/elements.ts
```diff
@@ -0,0 +1,86 @@
...
+  line-height: 0.5;
+  margin: 0 0.25rem;
+  vertical-align: middle;
+  text-align: center;
+  padding: 0;
+  outline: none;
+  cursor: pointer;
+
+${({ bg, theme }) =>
+  bg &&
```
siddharthkp: bg: boolean? what does this do?

packages/common/src/components/icons/Back.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
```
siddharthkp: ooooh so that's where we keep our icons. I had no idea, sorry fam 😅

packages/app/src/app/components/Preview/DevTools/Tabs/Tab/elements.ts
SaraVieira: A lot of them don't have tab background so that was ruining danny's intentions :(

packages/common/src/components/Preview/Navigator/elements.ts
SaraVieira: Its for these icons 

Screenshot 2019-12-16 at 15 55 41

packages/common/src/components/icons/Back.tsx
SaraVieira: Honestly found that now too ahah

packages/common/src/components/Preview/Navigator/elements.ts
SaraVieira: Its for thse icons
Screenshot 2019-12-16 at 15 55 41

siddharthkp: should we call it a more descriptive name. I couldn't have guessed

packages/app/src/app/components/Preview/DevTools/Tabs/Tab/elements.ts
siddharthkp: should we pick tab.activeBackground and then fallback to input.background?

SaraVieira: But the colors aren't the same, it looks off

packages/common/src/components/Preview/Navigator/elements.ts
SaraVieira: Named it `IconWithBackground` and made a new styled component

packages/app/src/app/components/Preview/DevTools/Tabs/Tab/elements.ts
CompuIves: Hmm, what about these tabs? The browser looks nicer, but the other tabs are now inconsistent. I'm thinking this might be something the theme should solve (different colors).

image

CompuIves: Maybe we could use the old colors for the current theme still? Then when we switch to CodeSandbox Black everything should look like the current design cc @DannyRuchtie

CompuIves: I also like @siddharthkp suggestion of defaulting to `tab.activeBackground` and fallback to `input.background`, but with mixed feelings. Remember that these themes probably decided to no set the `tab.activeBackground` on purpose.

packages/common/src/components/icons/Back.tsx
CompuIves: Haha, I added it a couple days ago 😄.

DannyRuchtie:
screenshot_2019-12-13_at_19 33 02
@CompuIves i'm not entirely sure what the impact is.
The problem that we had with the current setup was a mismatch with the navbar and the tab (see image above) does your solution solve that?

packages/app/src/app/components/Preview/DevTools/Tabs/Tab/elements.ts
siddharthkp: @CompuIves we moved the conversation to slack, so you're missing some context now :)

siddharthkp: > Remember that these themes probably decided to no set the tab.activeBackground on purpose.

oh p.s. vscode falls back for you  if you don't set `tab.activeBackground`, based on its own dark/light rules. So we can also do the same based on what default looks best on codesandbox

CompuIves: > oh p.s. vscode falls back for you if you don't set tab.activeBackground, based on its own dark/light rules. So we can also do the same based on what default looks best on codesandbox

Yes, agree. We apply this too since we use the theme generated by vscode AFAIK. Need to double check, I've been wrong before.

CompuIves:
My suggestion would be to still use the old colors of the current VSCode theme for the navigation bar. Then when we'll move to CodeSandbox Black we will get the design of what it is right now automatically. It's pretty hard to show two different themes in the same editor, and understandably a workaround is by using different color tokens from the theme for different parts, but I think that's not clean and will, in the end, look worse in other places.

SaraVieira:
The changes asked were merged on Sid's PR to mine

PR #3211, opened 2019-12-16
Only track opening the tooltip when it actually opens

CompuIves:
Tested to confirm that this works, it works!


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3211/commits/96e44fc749989d1e5acdeda1d18c801fe04f8c64 is at https://pr3211.build.csb.dev/s/new.

PR #3212, opened 2019-12-16
Update user type, increase API perf

CompuIves:
After doing some profiling on our API calls (especially forking and showing sandboxes) I found out that the majority of the time is spent on rendering the author of a sandbox. This is because for the author we do some expensive computations, like counting the total fork count and like count, but also rendering up to 5 sandboxes for the top sandboxes.

I think this is unnecessary for the `author` field, so I propose with this PR to remove a bunch of fields from `sandbox.author` so that we can have much faster API calls. I think we will really shave ~50% response time when doing this change.

### DIFF
--- a/packages/common/src/types/index.ts
+++ b/packages/common/src/types/index.ts
@@ -188,15 +188,13 @@ export type User = {
   sandboxCount: number;
   givenLikeCount: number;
   receivedLikeCount: number;
-  currentModuleShortid: string;
   viewCount: number;
   forkedCount: number;
   sandboxes: PaginatedSandboxes;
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3212/commits/d5b06293d155418ab0246e4f87af1d6044f53aea is at https://pr3212.build.csb.dev/s/new.

SaraVieira:
Is the backend work done?

CompuIves:
Yes, tested on staging as well. We used a normal User when we were still using Redux Entities. Now it's not needed anymore.

PR #3216, opened 2019-12-16
🔨 Switch FilterOptions to use useOvermind

MichaelDeBoey:
Resubmission of #2857

Follow-up of #2650

Things I did extra:
- Change `blacklistedTemplateAdded` & `blacklistedTemplateRemoved`'s signatures to accept a `string` instead of `{ template: string }`, because they only have 1 argument
- Change `blacklistedTemplatesChanged`'s signature to accept a `string[]` instead of `{ templates: string[] }`, because it only has 1 argument
- Export `FilterOptions` by a named export  instead of a `default export`
- Derive `Filters`' props type from child components
- Derive `Content`'s props type from child components
- Delete `types.ts` and make `Template` a private type from `FilterOptions`


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3216/commits/1df01f24f1d4f500e1cdc4ede42a3aa506e0f199 is at https://pr3216.build.csb.dev/s/new.

PR #3217, opened 2019-12-16
Remove sandbox service worker

CompuIves:
After handling a bug report in sandboxes that was related to another service worker that didn't update I decided that it would be worth looking into removing the service worker, for sandboxes in this case.

I don't think that we need service workers in a sandbox because:

1. They often don't need to be offline available
2. The performance improvements are not there, because:
  a. Every request (unpkg) is cached to disk anyway
  b. The cache of the sandbox is on the html, which is not loaded because of the service worker if it's cached. We could argue that the sandbox would speed up without service worker

This will also be less obnoxious for developers, when they look in their App tab, they will see a massive amount of service workers registered for sandboxes, since every sandbox registers its own service worker.

The advantage is that:

1. Sandboxes might speed up
2. Less bugs because of version differences/no cache availability
3. Sandboxes take less cache storage

Disadvantages:

1. You can't install a sandbox as an app (but people can add this in user code)
2. When there's no network connection the sandbox won't load

Still want to do a bit of testing on this, see if the performance gets affected.

### DIFF
--- a/packages/app/config/webpack.prod.js
+++ b/packages/app/config/webpack.prod.js
@@ -12,6 +12,7 @@ const ManifestPlugin = require('webpack-manifest-plugin');
 const CopyWebpackPlugin = require('copy-webpack-plugin');
 const VERSION = require('@codesandbox/common/lib/version').default;
 // const childProcess = require('child_process');
+const RemoveServiceWorkerPlugin = require('webpack-remove-serviceworker-plugin');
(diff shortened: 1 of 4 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3217/commits/d3a414b285ec33467c6326209b17de01fa42531e is at https://pr3217.build.csb.dev/s/new.

PR #3232, opened 2019-12-20
Sandbox downloader

SaraVieira:

Code Refactor

Right now all the zipping and downloading of sandboxes is being done in the client so we have to save all of these zip files on the client

## What is the new behaviour?

A new microservice: https://github.com/codesandbox/sandbox-downloader 

This means we get the zip code from there and then all we need to do is transform it into a blob and give it to the user

1. Download public sandbox
2. Download private sandbox
3. Deploy to Netlify
4. Deploy to zeit


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3232/commits/a7572a9420ab551abc46839775d52783bfbc58f0 is at https://pr3232.build.csb.dev/s/new.

PR #3279, opened 2020-01-07
Improve embed for mobile view

CompuIves:
Changes:

1. Show "View Source"/Preview Icon for opening/closing the preview. This only happens for small touch screens.
2. Change drag handler to not float, but be separate. This is to remove some confusion with it as a scroll bar.
3. Let the editor resize immediately after dragging. This way the editor shows immediately after dragging.
4. Change like icon to be codesandbox icon, which opens the embed in a separate tab
5. Make the default opened file based on template rather than `sandbox.entry` as first choice


packages/app/src/embed/components/App/elements.js
```diff
@@ -33,7 +33,7 @@ export const Moving = styled.div`
   position: absolute;
   top: 0;
   bottom: 0;
-  left: props.sidebarOpen ? 250 : 0;
+  left: ${props => (props.sidebarOpen ? 250 : 0)}px;
```
siddharthkp: how did this not break completely until now 😱

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3279/commits/7bf3cfdc96021a68a28496e39be848683ab84278 is at https://pr3279.build.csb.dev/s/new.

packages/app/src/embed/components/App/elements.js
SaraVieira: It did 😂

PR #3183, opened 2019-12-11
SortOptions: refactored to overmind

christianalfoni:
This is the sorting mechanism on the Dashboard. Refactored and tested the sorting. Seems to work as it should 👍

PR #3281, opened 2020-01-07
🔨 Switch SortOptions to use useOvermind

MichaelDeBoey:
Follow-up of #3183

Things I did extra:
- Change `orderByChanged`'s signature to accept a `OrderBy` instead of `{ orderBy: OrderBy }`, because it only has 1 argument
- Extract `Overlay` into it's own component (separate file)
- Derive `SortOptions`' Props type
- Type `SortOptions`' `toggleSort` function


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3281/commits/1d6a3d2b732d0d781f7987fb5526707eb8a7a1bb is at https://pr3281.build.csb.dev/s/new.

MichaelDeBoey:
Ping @christianalfoni @SaraVieira

MichaelDeBoey:
Ping @christianalfoni @SaraVieira

MichaelDeBoey:
Gerne 🙂

PR #3284, opened 2020-01-07
🔨 Switch Preview to use useOvermind

MichaelDeBoey:
Follow-up of #2638

Things I did extra:
- Change `previewActionReceived`'s signature to accept `any ` instead of `{ action: any }`, because it only has 1 argument
- use `ServerContainerStatus` enum instead of string values


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3284/commits/a635b65cce8c97aeb705daea407573d6665a29d2 is at https://pr3284.build.csb.dev/s/new.

PR #3290, opened 2020-01-08
Add stream option flag

CompuIves:
This PR changes stream to default API. This is just for dev purposes, the goal is to make this switchable with a flag, but we need to make some more changes before that will be possible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "start:dev_api": "concurrently --raw \"cd packages/app && yarn start:dev_api\" \"cd packages/common && yarn start\" \"cd packages/components && yarn start\"",
     "start:dynamic": "lerna run dev --scope dynamic-pages --stream",
     "start:fast": "concurrently --raw \"cd packages/app && yarn start\" \"cd packages/common && yarn start\" \"cd packages/components && yarn start\"",
-    "start:fast:stream": "cross-env STAGING_STREAM=1 concurrently --raw \"cd packages/app && yarn start\" \"cd packages/common && yarn start\" \"cd packages/components && yarn start\"",
+    "start:fast:stream": "cross-env STAGING_API=1 concurrently --raw \"cd packages/app && yarn start\" \"cd packages/common && yarn start\" \"cd packages/components && yarn start\"",
     "start:home": "yarn build:deps && yarn start:home:fast",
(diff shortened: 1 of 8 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3290/commits/2fd3907a194e303691ce8ec525bc8dd233990022 is at https://pr3290.build.csb.dev/s/new.

package.json
```diff
@@ -50,6 +50,7 @@
     "start:dev_api": "concurrently --raw \"cd packages/app && yarn start:dev_api\" \"cd packages/common && yarn start\" \"cd packages/components && yarn start\"",
     "start:dynamic": "lerna run dev --scope dynamic-pages --stream",
     "start:fast": "concurrently --raw \"cd packages/app && yarn start\" \"cd packages/common && yarn start\" \"cd packages/components && yarn start\"",
+    "start:fast:stream": "cross-env STAGING_STREAM=1 concurrently --raw \"cd packages/app && yarn start\" \"cd packages/common && yarn start\" \"cd packages/components && yarn start\"",
```
lbogdan: I think `STAGING_STREAM` is kind of confusing, because `STAGING` and `STREAM` mean pretty much the same thing. 🙂 Maybe `DEV_STREAM` is better, i.e. "develop on stream"?

CompuIves: Hmmmmm, that's right. But I'd interpret `DEV_STREAM` as the development server on stream. Pretty hard to get a good name here. I think we should still include `STAGING` in the name, since quite some people don't know that `STREAM` is actually staging.

lbogdan: `DEV_STAGING`, maybe? Or turn it around and say `DEV_PROD=1` and make stream default?

christianalfoni: I think we should stick with one term... if `STREAM` is our staging environment, let us just stick with that all over. Does not really matter what we call it, as long as it is consistent. Or are we referencing `STAGING` somewhere else?

CompuIves: Yeah we have another env var called `STAGING`, otherwise I would've chosen that env var.

CompuIves: What if we would change the var name to `STAGING_API`?

lbogdan: `USE_STAGING_API`

PR #3291, opened 2020-01-08
components: use Text in Collapsible

siddharthkp:
Using Text inside Collapsible instead of creating new elements.

The stories are especially interesting.

### DIFF
--- a/packages/components/src/components/Collapsible/index.tsx
+++ b/packages/components/src/components/Collapsible/index.tsx
@@ -14,7 +14,7 @@ export const Header = styled.div(
   css({
     display: 'flex',
     alignItems: 'center',
-    height: 32, // replace with 6
+    height: 6,
     paddingX: 3,
     borderBottom: '1px solid',
     // Note: sideBarSectionHeader exists but we dont use it because it is rarely implemented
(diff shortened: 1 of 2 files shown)


packages/components/src/components/Collapsible/index.tsx
```diff
@@ -82,12 +81,12 @@ export const Collapsible: React.FC<ICollapsibleProps> = ({
   const toggle = () => setOpen(!open);
 
   return (
-    <Element as="section" {...props}>
+    <Section {...props}>
```
siddharthkp: I'm curious, how would it be to write the styles in place - 

```jsx
<Element as="section" css={{fontSize: 3}}>
```

vs 

```js
const Section = styled(Element).attrs({ as: 'section' })(
  css({
    fontSize: 3,
  })
);
```

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3291/commits/aff16cf1b4fc78b82d382f85c3a19ac320d9076f is at https://pr3291.build.csb.dev/s/new.

packages/components/src/components/Collapsible/index.tsx
SaraVieira: I perfer the top one but I think we should use the bottom one as it's more js like and I know drake also perfers that one

PR #3292, opened 2020-01-08
Add Textarea To Components

SaraVieira:
* Adds textarea
* Adds max length logic
* changes labels to extend Text component


packages/components/.storybook/config.tsx
```diff
@@ -32,18 +32,24 @@ const viewports = {
 // using sidebar as the styles for body for now 🤷
 const GlobalStyle = createGlobalStyle`
   ${global};
-  body {
+  html body {
```
SaraVieira: Storybook was overriding the box sizing

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3292/commits/b85043f04dc00fa47494aa72638effc7a025533a is at https://pr3292.build.csb.dev/s/new.

packages/components/src/components/Input/index.tsx
```diff
@@ -25,11 +26,9 @@ export const InputComponent = styled.input(
   })
 );
 
-const Label = styled.label(
+const Label = styled(Text)(
   css({
-    fontSize: 2,
-    paddingBottom: 2,
-    color: 'sidebar.foreground',
+    marginBottom: 2,
```
siddharthkp: The responsibility of laying out the elements falls on the container (based on the use case) . So components should not have any margin on them inside.

as each of our component is extended from Element - you can now use - <Label marginBottom={2}>

packages/components/src/components/Input/index.tsx
```diff
@@ -52,7 +51,9 @@ export const Input: React.FC<IInputProps> = ({
           <label htmlFor={id}>{props.placeholder}</label>
         </VisuallyHidden>
       ) : null}
-      {label ? <Label htmlFor={id}>{label}</Label> : null}
+      <Label size={2} as="label" htmlFor={id}>
```
siddharthkp: This is a mixture of styles :)

It would be nicer as 

a) 
```jsx
<Text as="label" size={2} marginBottom={2} htmlFor={id} css={{display: 'block' }}><Text>
```
* it seems we'll need to combine margin with display: block/inline-block quite often. should we add the option to make Text block - <Text block /> or <Text display="block"/>. This property could also be moved all the way up the Element.

or 

b) 
```js
const Label = styled(Text).attrs({as: 'label', size: 2})(css({
  marginBottom: 2,
  display: block
}))

<Label htmlFor={id}></Label>
```

packages/components/src/components/Textarea/index.tsx
```diff
@@ -0,0 +1,120 @@
+import React, { useState, useCallback } from 'react';
+import styled from 'styled-components';
+import css from '@styled-system/css';
+import { VisuallyHidden } from 'reakit/VisuallyHidden';
```
siddharthkp: this should be same as Input

packages/components/src/components/Textarea/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { uniqueId } from 'lodash-es';
+import { Text } from '../Text';
+import { Stack } from '../Stack';
+
+const placeholderStyles = {
+  color: 'input.placeholderForeground',
+  fontSize: 13,
```
siddharthkp:
```suggestion
  fontSize: 3,
```

packages/components/src/components/Textarea/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+};
+
+export const TextareaComponent = styled.textarea(
```
siddharthkp: Should we reuse styles from Input with `styled(Input)`

packages/components/src/components/Textarea/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+    marginBottom: 2,
+    display: 'block',
+  })
+);
+
+const Count = styled.div<{ limit: boolean }>(({ limit }) =>
+  css({
+    fontSize: 11,
+    paddingTop: 1,
+    color: limit ? 'errorForeground' : 'input.placeholderForeground',
```
siddharthkp: ooh nice touch!

packages/components/src/components/Textarea/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const [wordCount, setWordCount] = useState(0);
+  const [value, setValue] = useState('');
+
+  const Wrapper = useCallback(
+    ({ children }) =>
+      maxLength ? <Stack direction="vertical">{children}</Stack> : children,
+    [maxLength]
+  );
+
+  // eslint-disable-next-line consistent-return
```
siddharthkp: freaking hate that rule

packages/components/src/components/Textarea/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+    display: 'block',
+  })
+);
+
+const Count = styled.div<{ limit: boolean }>(({ limit }) =>
+  css({
+    fontSize: 11,
+    paddingTop: 1,
+    color: limit ? 'errorForeground' : 'input.placeholderForeground',
+    alignSelf: 'flex-end',
```
siddharthkp: interesting way to use this, hadn't thought of that!

packages/components/src/components/Textarea/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+  css({
+    marginBottom: 2,
+    display: 'block',
+  })
+);
+
+const Count = styled.div<{ limit: boolean }>(({ limit }) =>
+  css({
+    fontSize: 11,
+    paddingTop: 1,
```
siddharthkp: if you want you can also use <Stack gap={1}> for this. either way is cool

packages/components/src/components/Textarea/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+  label?: string;
+  maxLength?: number;
+}
+
+export const Textarea: React.FC<ITextareaProps> = ({
+  label,
+  maxLength,
+  ...props
+}) => {
+  const id = props.id || uniqueId('form_');
```
siddharthkp: forgot to say this on the other PR, this is really nifty, good stuff!

packages/components/src/components/Input/index.tsx
SaraVieira: Done! Used the first one

packages/components/src/components/Input/index.tsx
SaraVieira: Used the first one in the other comment

packages/components/src/components/Textarea/index.tsx
SaraVieira: Done

packages/components/src/components/Textarea/index.tsx
SaraVieira: Done!

packages/components/src/components/Textarea/index.tsx
SaraVieira: danke

packages/components/src/components/Textarea/index.tsx
SaraVieira: We can turn it off in this package, what do you think?

packages/components/src/components/Textarea/index.tsx
SaraVieira: changed to the theme :)

packages/components/src/components/Textarea/index.tsx
SaraVieira: Flex is amazing! ahah

packages/components/src/components/Textarea/index.tsx
SaraVieira: Devs are lazy ahah

packages/components/src/components/Textarea/index.tsx
siddharthkp: yaaas!

PR #3296, opened 2020-01-08
components: Switch

siddharthkp:
Added a switch component

also fixed 2 unrelated bugs
1. theme in story/config
2. grid column types


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3296/commits/32a5322d0e338a7371250efbcd75757f81e151b9 failed.

packages/components/src/components/Switch/switch.stories.tsx
```diff
@@ -0,0 +1,15 @@
...
+  title: 'components/Switch',
+  component: Switch,
+};
+
+export const Basic = () => <Switch />;
+
+export const defaultOn = () => <Switch defaultOn />;
+
+/* eslint-disable no-console */
+export const onChange = () => <Switch onChange={action('changed')} />;
```
SaraVieira: This makes the console log be on the addons tab

packages/components/src/components/Switch/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+interface ISwitchProps {
+  on?: boolean;
+  defaultOn?: boolean;
+  onChange?: (event: any) => void;
+}
+
+export const Switch: React.FC<ISwitchProps> = ({ on, defaultOn, ...props }) => (
+  <SwitchContainer as="label">
+    <SwitchInput
```
SaraVieira: This input doesn't have a label maybe we need something here?

siddharthkp: The label is given by the label on the left

Screenshot 2020-01-09 at 8 38 19 AM

How do we suggest folks to attach them together with for and id, one way could be to make `id` compulsory on the switch and throw a warning if it's not given?

packages/components/src/components/Switch/switch.stories.tsx
siddharthkp: amazing,thanks

packages/components/src/components/Switch/index.tsx
SaraVieira: Sounds good!

PR #3297, opened 2020-01-08
Add Search Input

SaraVieira:
* Adds search input
* Adds more onChange test cases


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3297/commits/56827a04ff921cb7978a76fb477719b917013d0e is at https://pr3297.build.csb.dev/s/new.

packages/common/src/design-language/theme.ts
```diff
@@ -7,6 +7,16 @@ const theme = {
   fontWeights,
 
   // we use a 4 point grid
+  // 0 - 0
+  // 1 - 4
+  // 2 - 8
+  // 3 - 12
+  // 4 - 16
+  // 5 - 24
```
siddharthkp: this isn't correct. 5 * 4 = 20

siddharthkp:
Do you think we should use <Input type="search"/> instead of a new component?

packages/common/src/design-language/theme.ts
SaraVieira: Not on my watch!

siddharthkp: but but?

PR #3298, opened 2020-01-08
Select component

SaraVieira:
Adds select component


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3298/commits/d52b27b0976c8969c63784e6701480190f48db9b is at https://pr3298.build.csb.dev/s/new.

packages/components/src/components/Select/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+import css from '@styled-system/css';
+import VisuallyHidden from '@reach/visually-hidden';
+import { uniqueId } from 'lodash-es';
+import { Text } from '../Text';
+import { Element } from '../Element';
+
+// Svg used for the icon
+const svg = (hover: boolean) =>
+  `"data:image/svg+xml,%3Csvg width='8' height='24' viewBox='0 0 8 24' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M4.00006 17L1 13L7 13L4.00006 17Z' fill='${
+    hover ? 'white' : '%23757575'
```
siddharthkp: we can't hard code colors 🤷‍♀ Let's replace this with `currentcolor` and style it in the Select component

As to what colors should we use - it should either be the same as text or a muted version, cc @DannyRuchtie

packages/components/src/components/Select/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+      backgroundImage: `url(${svg(true)})`,
+      color: 'input.foreground',
+    },
+  })
+);
+
+const IconWrapper = styled(Element)(
+  css({
+    position: 'absolute',
+    left: '8px',
```
siddharthkp:
```suggestion
    left: 2,
```

packages/components/src/components/Select/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+        htmlFor={id}
+        style={{ display: 'block' }}
+      >
+        {label}
+      </Text>
+      <Wrapper>
+        {icon ? <IconWrapper>{icon()}</IconWrapper> : null}
+        <SelectComponent icon={Boolean(icon)} id={id} {...props}>
+          {props.placeholder ? (
+            <option value="" disabled selected>
```
siddharthkp: not sure if this should be disabled, this will not allow you to unselect your value

packages/components/src/components/Select/index.tsx
SaraVieira: You are right, fixed :)

packages/components/src/components/Select/index.tsx
SaraVieira: Done :)

packages/components/src/components/Select/index.tsx
SaraVieira: Honestly, either one is fine by me, I saw this being more used but I guess the value is 0 so if we check required we will be good


I will remove it

PR #3319, opened 2020-01-12
Always favor 'module' over 'main'

### LINKED ISSUE #3305: Feature request: prefer module over main
# 🐛 bug report

## Description of the problem

I've developed a npm package with [vue-sfc-rollup](https://github.com/team-innovation/vue-sfc-rollup) and the build in codesandbox use the ssr file not the esm.
You can see the [package.json](https://github.com/rtransat/vhue/blob/master/package.json#L22-L23) of my project.

It's works with a vue cli project in local.

It seems that codesandbox use the "main" entry point and I have the following error:

```
[Vue warn]: Error in render: "TypeError: _vm._ssrNode is not a function"
__vue_render__@https://zq0n5.csb.app/node_modules/vhue/dist/vhue.ssr.js:4440:66
```

### Your Environment

| Сodesandbox      | Online
| Browser          | Firefox
| Operating System | Linux Mint 18.3

lbogdan:
This PR matches [`webpack` resolving behavior](https://webpack.js.org/configuration/resolve/#resolvemainfields), which always favors `module` over `main`.

Fixes #3305 .

### DIFF
--- a/packages/app/src/sandbox/eval/utils/resolve-utils.js
+++ b/packages/app/src/sandbox/eval/utils/resolve-utils.js
@@ -1,6 +1,5 @@
-export function packageFilter(p: Object) {
-  if (!p.main && p.module) {
-    // eslint-disable-next-line
+export function packageFilter(p) {
+  if (p.module) {
     p.main = p.module;
   }
 


packages/app/src/sandbox/eval/utils/resolve-utils.js
```diff
@@ -1,6 +1,5 @@
-export function packageFilter(p: Object) {
-  if (!p.main && p.module) {
-    // eslint-disable-next-line
+export function packageFilter(p) {
+  if (p.module) {
```
lbogdan: @CompuIves I guess we should also account for `browser` here, maybe something like

```diff
-   if (p.module)
+   if(!p.browser && p.module)
```

packages/app/src/sandbox/eval/utils/resolve-utils.js
```diff
@@ -1,6 +1,5 @@
-export function packageFilter(p: Object) {
```
lbogdan: @CompuIves How was this even working?! 🙂

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3319/commits/63f0c99a86177ca0a506d4fc696a460b9e89274e is at https://pr3319.build.csb.dev/s/new.

packages/app/src/sandbox/eval/utils/resolve-utils.js
CompuIves: Well, how was it not working? Or do you mean the type definitions?

packages/app/src/sandbox/eval/utils/resolve-utils.js
CompuIves: Hmm yes, I think the resolver prefers browser always over the main/module.

packages/app/src/sandbox/eval/utils/resolve-utils.js
lbogdan: > Or do you mean the type definitions?

Yes, looks like that was left from the `flow` days?

tajo:
It would be great to use the same logic as webpack since this can lead to some very unexpected bugs when you load part of the files from dist/commonjs and other from dist/es. That would normally not be a big issue unless you rely on things like React context where there needs to be a single instance. 

(We had this problem with `baseui` in CodeSandbox, could give more details.).

PR #3334, opened 2020-01-14
Remove # from tags

CompuIves:
Adding a tag with `#` doesn't really make sense, and it breaks removing tags. So this removes the `#` from the tag when adding it.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3334/commits/ce464229a3914480d520b30c45fe6116bdc46473 is at https://pr3334.build.csb.dev/s/new.

PR #3340, opened 2020-01-15
Debounce the resizer of codemirror

CompuIves:
This is a performance fix for codemirror, if we spam the resize operation of codemirror (which is expensive because of `getClientBoundingRect`) the embed will be slow. With this we debounce the call with 30ms.

### DIFF
--- a/packages/app/src/app/components/CodeEditor/CodeMirror/index.js
+++ b/packages/app/src/app/components/CodeEditor/CodeMirror/index.js
@@ -12,6 +12,7 @@ import { getCodeMirror } from 'app/utils/codemirror';
 import CodeMirror from 'codemirror';
 import { listen } from 'codesandbox-api';
 import * as React from 'react';
+import { debounce } from 'lodash-es';
 import { withTheme } from 'styled-components';
 // eslint-disable-next-line
 import LinterWorker from 'worker-loader?publicPath=/&name=monaco-linter.[hash:8].worker.js!app/overmind/effects/vscode/LinterWorker/index';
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3340/commits/ffa8a4003d241ae83d029e8904a5c45b77b483c2 is at https://pr3340.build.csb.dev/s/new.

siddharthkp:
Tested on my phone, it's a lot smoother ✨

packages/app/src/app/components/CodeEditor/CodeMirror/index.js
```diff
@@ -54,6 +55,7 @@ class CodemirrorEditor extends React.Component<Props, State> implements Editor {
     this.settings = props.settings;
 
     this.codeSandboxListener = this.setupCodeSandboxListener();
+    this.resizeEditor = debounce(this.resizeEditor, 30);
```
siddharthkp: curious: what's the significance of this exact number?

packages/app/src/app/components/CodeEditor/CodeMirror/index.js
```diff
@@ -111,6 +112,11 @@ class CodemirrorEditor extends React.Component<Props, State> implements Editor {
     });
   }
 
+  resizeEditor() {
+    // eslint-disable-next-line no-unused-expressions
+    this.codemirror?.refresh();
```
siddharthkp: wouldn't this always exist on codemirror 🤔

CompuIves: There may be the case where `codemirror` is null, so this is to be on the safe side there.

packages/app/src/app/components/CodeEditor/CodeMirror/index.js
CompuIves: It's a random number really, but I've seen this number in other codebases as well when debouncing a resize. I think makes sure that it happens as fast as possible after an animation.

PR #3347, opened 2020-01-16
Many improvements to component composition

siddharthkp:
Sorry, big PR - one thing led to another and next thing I know, 25 files changed

Bonus: look ma, no typescript errors

New components: 
- [x] FormField

  Form field takes a label and a child element. It creates a label and attaches them together. It supports 2 layouts and invisible labels when you only a label for screen readers. [See stories](https://www.chromaticqa.com/component?appId=5d6e8335ddc81f0020dbed33&name=components%2FFormField&mode=snapshot&buildNumber=344&specName=Horizontal&viewport=1200&browser=chrome)

  This component also simplifies all our input components - Input, Select, Switch, Textarea
   
- [x] Label

   Every time you need to use <Text as="label">, use <Label> instead. Inherits all properties of Text.

- [x] SidebarRow

  Literally a shortcut for `<Stack align="center" css={{height: 8}} />` because it gets repeated so many times in the sidebar. Danny likes to use it as a layout block. Would help in creating a better vertical flow.

Enhancements: 

- [x] Animations for Collapsible + fixed the height, should have been 8 all along

- [x] Hovering anywhere on a Select triggers the hover state for all elements (ditto for focus)

- [x] Text now supports `block` prop

-- 

Reviewed all the chromatic changes. All of them are intention 👍


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3347/commits/e0e1c1c846c4e40684a18fdec53a4e9e9ca61186 is at https://pr3347.build.csb.dev/s/new.

packages/components/.storybook/config.tsx
```diff
@@ -54,7 +54,7 @@ const vsCodeThemes = allThemes.map(b => makeTheme(b, b.name));
 const blackCodesandbox = vsCodeThemes.find(
   theme => theme.name === 'CodeSandbox Black'
 );
-console.log(isChromatic());
```
SaraVieira: Danke! ahah

packages/components/src/components/FormField/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+  id?: string;
+  // always ask for a label
+  label: string;
+  // sometimes you don't need to show the label to a visual user
+  // because the section header and description give enough context.
+  // this however is not true for a user that's using a screen reader
+  // they would still require a description label to be read out.
+  // for this purpose, we visually hide the label but still  keep in the
+  // elements tree.
+  hideLabel?: boolean;
```
SaraVieira: That's a really good API!

SaraVieira:
It may be a good idea to add a `relative` prop to the element, what do you think?

siddharthkp:
> It may be a good idea to add a relative prop to the element, what do you think?

which element?

SaraVieira:
> which element?

The Actual `element` component

siddharthkp:
> The Actual element component

Ah got it.

I like the idea. We can add the upgrade `position` to a prop. Same treatment as `margin`. I'm not sure what is a good set of common props and where does it stop 🤔 

The _recommended_ method right now is to use `css={{position: 'relative'}}`. All components support it now ✨

A good way would be to - If you find yourself using Element only when you need to add position:relative, then it should be a _commonProp`, so that you can add it to any element.

PR #3352, opened 2020-01-16
🔨 Switch Profile to use useOvermind

MichaelDeBoey:
Follow-up of #2729

Things I did extra:
- Change `profileMounted`'s signature to accept a `string` instead of `{ username: string }`, because it only has 1 argument
- Fix `Profile`'s props by using `react-router-dom`'s `RouteComponentProps` generic instead of hardcoding it
- Export `Profile` by a named export  instead of a `default export`


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3352/commits/78e85ee5a012f98c2d81c91beefe95ac0223f5e7 is at https://pr3352.build.csb.dev/s/new.

SaraVieira:
Looks good! Thank you!

MichaelDeBoey:
With pleasure! 🙂

PR #3361, opened 2020-01-18
🔨 Switch Dependencies to use useOvermind

MichaelDeBoey:
Follow-up of #3066

Things I did extra:
- Change `npmDependencyRemoved`'s signature to accept a `string` instead of `{ name: string }`, because it only has 1 argument
- Make `AddResource` a `FunctionComponent`


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3361/commits/edda54bde018ab51cb7c1436ac1b527028b8dfe6 failed.

MichaelDeBoey:
Ping @christianalfoni @SaraVieira

SaraVieira:
Closed for the same reason as the sidebar ones :/

PR #3364, opened 2020-01-19
Split out forkCount so we get faster API response times

CompuIves:
Loading fork count takes often 80% of the response time, sometimes building up to 2 seconds. With this change we load the fork count in parallel.


CompuIves:
Idea: maybe we can add a param to ignore the fork count.

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3364/commits/b477c4522a15010ab2db8771253e9ca81d1ef8d9 is at https://pr3364.build.csb.dev/s/new.

PR #2774, opened 2019-10-14
Refactored app/pages/common/Modals/ForkServerModal

alanhoskins:
## What kind of change does this PR introduce?

This is part of the requested refactor for hacktoberfest mentioned in #2621.
@Saeris @christianalfoni

## What is the current behavior?

Previously it was utilizing inject and hooksObserver from app/componentConnectors.

## What is the new behavior?

Uses `useOvermind` and `useEffect` to access state and perform checks on `isLoggedIn`.

## What steps did you take to test this? This is required before we can merge.

1. `yarn lint`
2. `yarn test`
3. Ran locally, but not sure where this is used.

## Checklist

- [ ] Documentation - N/A
- [X] Testing 
- [X] Ready to be merged
      
- [ ] Added myself to contributors table

I was unable to run `yarn contributors:add`

PR #3367, opened 2020-01-19
🔨 Switch ForkServerModal to use useOvermind

MichaelDeBoey:
Follow-up of #2774 

Things I did extra:


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3367/commits/146b6c1e328289233a4ff6037b3592755a3722de is at https://pr3367.build.csb.dev/s/new.

SaraVieira:
Nice! Thank you!

MichaelDeBoey:
With pleasure! 🙂

PR #3370, opened 2020-01-20
Improve esmodules loading by writing a custom converter

CompuIves:


tajo:
I'm debugging a problem with esm loading in CodeSandbox that we start seeing just recently. Can you tell when this was released and also what the intention? Thanks!

Is this related to https://github.com/codesandbox/codesandbox-client/pull/3319 ?

PR #3393, opened 2020-01-24
Fix checking if preview is standalone

lbogdan:
In `app` we were only checking for `!window.opener`, which is not enough, while in `sse-hooks` we weren't checking at all.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -26,7 +26,7 @@ steps:
       - yarn build
 
   - name: deploy
-    image: lbogdan/codesandbox-deploy-plugin:3
+    image: lbogdan/codesandbox-deploy-plugin:4
     settings:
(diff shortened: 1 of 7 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3393/commits/0d83b236885c84a55c15381970b1836f6e12a65c is at https://pr3393.build.csb.dev/s/new.

lbogdan:
This also bumps the Drone `deploy-plugin` version, which includes a change to how CORS headers are set for PR builds: https://github.com/codesandbox/drone-plugins/commit/b922ada64ed290ee9f5a5873d630e82b5c08a19a .

Should be ready for review and merge.

PR #3404, opened 2020-01-28
Fix uploading empty files

### LINKED ISSUE #3087: Can't import blank files
# 🐛 bug report

## Description of the problem
If you drag and drop and empty file into the editor, you get the error: "Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded."

That error isn't accurate, and it would be useful if this case could be specifically handled to allow importing of projects with blank files for better UX.

Trying to import the Angular getting started project but couldn't do so without manual intervention.

lbogdan:
Fixes #3087 .

### DIFF
--- a/packages/app/src/app/overmind/namespaces/files/internalActions.ts
+++ b/packages/app/src/app/overmind/namespaces/files/internalActions.ts
@@ -99,7 +99,10 @@ export const uploadFiles: AsyncAction<
             file.type === 'application/json') &&
           dataURI.length < MAX_FILE_SIZE
         ) {
-          const text = dataURI ? atob(dataURI.replace(/^.*base64,/, '')) : '';
+          const text =
+            dataURI && dataURI !== 'data:'
+              ? atob(dataURI.replace(/^.*base64,/, ''))
+              : '';
           parsedFiles[filePath] = {
             content: text,
             isBinary: false,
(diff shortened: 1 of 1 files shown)


packages/app/src/app/overmind/namespaces/files/internalActions.ts
```diff
@@ -99,7 +99,10 @@ export const uploadFiles: AsyncAction<
             file.type === 'application/json') &&
           dataURI.length < MAX_FILE_SIZE
         ) {
-          const text = dataURI ? atob(dataURI.replace(/^.*base64,/, '')) : '';
+          const text =
+            dataURI && dataURI !== 'data:'
```
lbogdan: When the file is empty, `dataURI` is set to `data:`. Not exactly sure what the initial `dataURI` was supposed to check for, does it make sense to leave it in, or can we only do `dataURI !== 'data:'`? @CompuIves

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3404/commits/036d28bf1b2ee8f1984b21b6e48ebfa25a4256b5 is at https://pr3404.build.csb.dev/s/new.

packages/app/src/app/overmind/namespaces/files/internalActions.ts
CompuIves: We can remove the `dataURI` check then I think, especially because there is a check on length before that already. Good find!

lbogdan: Oh, I didn't see the suggestion! 🙂 Updated.

PR #3408, opened 2020-01-29
🔨 Switch Profile/Navigation to use useOvermind

MichaelDeBoey:
Things I did:
- Export `Navigation` by a named export  instead of a `default export`


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3408/commits/b790ae001fd9c5d7683777975af310860331cb0d is at https://pr3408.build.csb.dev/s/new.

MichaelDeBoey:
With pleasure 🙂

PR #3410, opened 2020-01-29
Move sentry to a new, clean project

CompuIves:
We still get 100s of thousands of errors from users that are running an old version of CodeSandbox. This makes sure that we only get errors from latest versions, and that we ignore all the errors from previous versions.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3410/commits/32e4adabb8e7b0a3ddacabbe8f28211fa0690a9c is at https://pr3410.build.csb.dev/s/new.

Issue #3359, opened 2020-01-17
Latency between switching files and split windows.


rmarquardt1:
# 🐛 bug report

## Description of the problem
When I switch to another file or move between split windows in my project, it can take up to 20 seconds before I can start editing the file.

## How has this issue affected you? What are you trying to accomplish?
I love using this site. It makes working on my project from multiple environments very easy. But it's taking me much longer to work on my project because of the latency between switching files.

### Your Environment

| Software         | Name/Version |
| ---------------- | ------------ |
| Сodesandbox      |
| Chrome         |
| Windows XP / 10 |

CompuIves:
That is very interesting. Does this happen in every sandbox? If it only happens in one sandbox, would you mind sharing that sandbox?

rmarquardt1:
It seems to be happening in all my sandboxes. The sandbox is https://codesandbox.io/s/cool-night-9o58q. Thanks!

garethx:
I haven't been able to reproduce in the linked sandbox. Can you confirm whether the issue continues when you're in an incognito window?

CompuIves:
Hmm, I'm able to reproduce now. It seems like one of our components is rerendering excessively, diving in!

PR #3413, opened 2020-01-29
Improve performance by not rendering children in explorer

### LINKED ISSUE #3389: IDE has slowed to crawl (Explorer to blame, and maybe something else too)
# 🐛 bug report
IDE has slowed to crawl. 

I'm using a `create-react-app` template with some dependencies.

## Description of the problem
When `Explorer` is open:
- Clicking (in Explorer) to open an existing file takes 3 seconds before a new tab is opened
- Clicking to switch tabs takes 9 seconds (!) before the tab is switched. That's too slow
- Closing a tab takes ~7 seconds (during that time the IDE is unresponsive to user input)
- Is `Explorer` the cause?
When turning off File Explorer or switching to different sidebar options (ie. such as switching from Explorer to Sandbox Info / Configuration Files /Deployment / Github /Live) makes switching tabs take 2 seconds, instead of 9 seconds. This leads me to believe that perhaps Explorer is coded with some kind of subscriptions on every folder/file. If true, because my sandbox has many folders and files (it's a React SaaS app after all), then that would cause everything to slow down, and could explain the why Codesandbox is becoming slower as my app grows.

`Transpiles` performance
Another performance issue is that Transpiles are frequent and slow. For example, if the app preview is fully rendered, then I press `enter` in the IDE, then the browser preview begins reloading. Going from => "downloads dependencies" => to => "transpiling modules". After that, a white screen is rendered (the preview didnt work). I'm forced to manually refresh the browser after in order to re-render the preview. All of that occurs when pressing enter on the keyboard. This seems to be a recent issue, as I've been using codesandbox for ~1.5 years on this project without these issues.

Transpiling takes ~12 seconds (sometimes >20 seconds). Even though Codesandbox has the setting "Preview without saving" enabled, sometimes transpiling occurs every time I type in a character (even though I have that feature disabled in the Candsandbox => Preview settings). The net effect of continous re-rendering on every key stroke (even though the option is disabled) is that it makes the IDE feel unresponsive, which is visible in the half-minute lag between what I type and what I see in the preview window. Furthermore, transpiling feels like it slows down the IDE (service-worker is effecting the main window?).

I'm forced to develop without the File Explorer window open, since its slowing things to a crawl.

The IDE is moving so slow, I may have to switch. There's gotta be a solution, so I can keep using Codesandbox...what would you do if switching tabs took 9 seconds and you're constantly waiting a half minute before you see changes? Meanwhile, I can isolate the issue to Explorer and perhaps something with transpiling, resolving these issues would be good for me, other patrons, and the community at large.

<3 

### Link to sandbox: [link](https://codesandbox.io/s/frontend-c9jwk) (optional)
https://codesandbox.io/s/frontend-c9jwk

### Your Environment

| Сodesandbox      | PROD-1579516524-d2848195c |
| Browser          | Chrome 79 (latest) |
| Operating System | Windows 10 |

CompuIves:
We used to do this a long time for animations, but it's not needed anymore.

cc @siddharthkp this might need to be reflected in the new design.

Fixes #3389 #3359


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3413/commits/91b44bd0959c0ccb6375c7bc3d94d353b0bf006d is at https://pr3413.build.csb.dev/s/new.

siddharthkp:
Don't think we're using the Opener in the new design :)

PR #3416, opened 2020-01-29
Load model directly from vscode

### LINKED ISSUE #3417: TypeError: Cannot read property 'getLinesContent' of null
Sentry Issue: [CLIENT-2](https://sentry.io/organizations/codesandbox/issues/1483023821/?referrer=github_integration)

```
TypeError: Cannot read property 'getLinesContent' of null
  at call (./src/app/overmind/effects/vscode/ModelsHandler.ts:243:25)
...
(1 additional frame(s) were not displayed)
```

CompuIves:
It turns out that the `getModel()` call can actually return a null value. This will make sure that we always set the model and load it directly from VSCode.

Something for the future: we could possibly now always set the model using this function when asked for it, right?

FIxes #3417


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3416/commits/3e0518ac404d808e2d797c7d1b796080045485d3 is at https://pr3416.build.csb.dev/s/new.

PR #3424, opened 2020-01-30
Themessss

siddharthkp:
Our interface does not map 1-1 with vscode. To make codesandbox look good and still keep it themeable with vscode themes, we tweak the theme on runtime to add additional tokens that are used for our interface but make no different to vscode


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3424/commits/9e2e821851421d4a3d5403eb38c1ae6027b67f51 failed.

PR #3425, opened 2020-01-30
Don't show error response HTML in notification when uploaded file is …

lbogdan:
This PR changes this error notification when uploading a file that's too large:

image

into this:

image


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3425/commits/88d9868d5fb2ae85b292d5cf3fe11f03ede11974 is at https://pr3425.build.csb.dev/s/new.

packages/app/src/app/overmind/internalActions.ts
```diff
@@ -481,6 +481,10 @@ export const handleError: Action<{
     });
   }
 
+  if (message.startsWith('The uploaded file is bigger')) {
```
lbogdan: This is quite hacky, but looking at the code, I couldn't figure out a good place where to properly handle this in a layer above.

CompuIves: Hm, perhaps we can make this function call: https://github.com/codesandbox/codesandbox-client/blob/a707e2d61fd1bf7e6adedc9660379330b7a3eb0a/packages/app/src/app/overmind/namespaces/files/actions.ts#L441-L444 accept error as optional, and then we don't pass it in?

lbogdan: Hm, seems a bit tedious, we use `error` a lot in that function, so we either have to check it everywhere, or wrap everything in an `if (error) { ... }`.

CompuIves: Hmm, we could also add the option to `hideErrorMessage`. Then add this at the start of `handleError`:

```js
  if (hideErrorMessage) {
    effects.analytics.logError(error);
    effects.notificationToast.add({
      message,
      status: NotificationStatus.ERROR,
    });
  }```

lbogdan: Updated.

PR #3427, opened 2020-01-30
🔨 Switch Profile/Sandboxes to use useOvermind

MichaelDeBoey:
Things I did:
- Just use `index.tsx` for `SandboxList` instead of the extra `SandboxList.tsx` file
- Change `sandboxesPageChanged`'s signature to accept a `number` instead of `{ page: number }`, because it only has 1 argument (`force` is never passed)
- Change `likedSandboxesPageChanged`'s signature to accept a `number` instead of `{ page: number }`, because it only has 1 argument
- Change `deleteSandboxClicked`'s signature to accept a `string` instead of `{ id: string }`, because it only has 1 argument
- Extract `NoSandboxes` out of `elements.ts` into it's own component


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3427/commits/b35dd073b424fb47c56b9dbbfa254793e556dd47 is at https://pr3427.build.csb.dev/s/new.

PR #2769, opened 2019-10-14
🔨 Refactor, 🧠 Overmind, refactored OpenedTabs to use Overmind

fullstackzach:
## What kind of change does this PR introduce?
This PR Refactors `/app/pages/Sandbox/Editor/Workspace/OpenedTabs/index.js` to use Overmind

#2621 

## What steps did you take to test this? This is required before we can merge.

1. Ran `yarn test` successfully
2. Ran `yarn lint` successfully
3. Ran `yarn start` and confirmed that Tabs functionality in the workspace still works
4. Ran `yarn typecheck` successfully

## Checklist

- [x] Documentation
- [x] Testing 
- [x] Ready to be merged
      
- [x] Added myself to contributors table

PR #3428, opened 2020-01-31
🔨 Switch OpenedTabs to use useOvermind

MichaelDeBoey:
Follow-up of #2769

Things I did extra:
- Simplified `moduleSelected` a bit + make types correcter
- Remove `saveAllModules.ts` since it's not used anymore
- Type + simplify `getIconURL`


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3428/commits/f2d4b2bc6f81a534ff0776ccf19bee2474ab489e failed.

SaraVieira:
Thank you!

MichaelDeBoey:
Always a pleasure 🙂

PR #2803, opened 2019-10-15
🔨 Refactor, 🧠 Overmind, Hacktoberfest - /app/pages/common/Modals/PRModal/index.js: convert to tsx and uses overmind

ftonato:
## What kind of change does this PR introduce?

Refactors code as a part of hacktoberfest mentioned in #2621.
@Saeris @christianalfoni

## What is the current behavior?

/app/pages/common/Modals/PRModal/index.js was using inject from app/componentConnectors and written in JS.

## What is the new behavior?

uses useOvermind hook from app/overmind and converted to tsx.

## Checklist
* [ ]  Documentation
* [ ]  Testing
* [x]  Ready to be merged
* [ ]  Added myself to contributors table

PR #2755, opened 2019-10-12
🔨Refactored 🧠 Overmind Hacktober | /app/pages/common/Modals/PRModal/index.js : refactor to replace Cerebral with Overmind 

hetpatel33:
## What kind of change does this PR introduce?
Refactors code as a part of hacktoberfest mentioned in  #2621 
@Saeris @christianalfoni @SaraVieira

## What is the current behavior?
/app/pages/common/Modals/PRModal/index.js was using inject and hooksObserver from app/componentConnectors and written in JS.

## What is the new behavior?
uses useOvermind hook from app/overmind and converted to tsx.

## What steps did you take to test this? This is required before we can merge.

1. Step A
2. Step B
3. Step C

## Checklist

- [ ] Documentation
- [x] Testing 
- [x] Ready to be merged
      
- [ ] Added myself to contributors table

PR #3431, opened 2020-01-31
🔨 Switch PRModal to use useOvermind

MichaelDeBoey:
Follow-up of #2755 & #2803

Things I did extra:
- Fix `GitProgress`' `result` prop type
- Inline `PRModal`'s `result` instead of having an extra variable
- Export `PRModal` by a named export  instead of a `default export`

### DIFF
--- a/packages/app/src/app/components/GitProgress/index.ts
+++ b/packages/app/src/app/components/GitProgress/index.ts
@@ -1,1 +1,0 @@
-export { GitProgress } from './GitProgress';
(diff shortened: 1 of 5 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3431/commits/179374145124c8248f70c89f508bd5b0f9b6c392 is at https://pr3431.build.csb.dev/s/new.

MichaelDeBoey:
Always a pleasure to help 🙂

PR #3440, opened 2020-01-31
🔨 Switch PickSandboxModal to use useOvermind

MichaelDeBoey:
Follow-up of #2752

Things I did extra:
- Change `pickSandbox`'s signature to get its argument types directly from `PickedSandboxDetails` instead of hardcoding them
- Type `form`'s `onSubmit` prop


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3440/commits/ecb4bb759a26119fef30caf0003adb0a7158b69c is at https://pr3440.build.csb.dev/s/new.

MichaelDeBoey:
With pleasure 🙂

PR #2878, opened 2019-10-20
Refactored liveinfo packages /app/src/app/pages/Sandbox/Editor/Workspace/items/Live/LiveInfo.tsx

lusan:
## What kind of change does this PR introduce?

Refactors code as a part of hacktoberfest mentioned in #2621.
@Saeris @christianalfoni

## What is the current behavior?

packages /app/src/app/pages/Sandbox/Editor/Workspace/items/Live/LiveInfo.tsx was using Cerebral.

## What is the new behavior?

Removed Cerebral and the component didn't need to use overmind, hence main work has been done to convert it to TS.

/app/src/app/pages/Sandbox/Editor/Workspace/items/Live/LiveInfo.tsx

## What steps did you take to test this? This is required before we can merge, make sure to test the flow you've updated.

1. Removed inject and hooksObserver.
2. yarn test
3. yarn lint
4. yarn typecheck

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      
- [ ] Added myself to contributors table

PR #3444, opened 2020-02-01
🔨 Switch LiveInfo to use useOvermind

MichaelDeBoey:
Follow-up of #2878

Things I did extra:
- Change `onModeChanged`, `onLiveMode` & `sendLiveMode`'s signature to get its argument types directly from `RoomInfo` instead of hardcoding them
- Change `onAddEditorClicked` & `onFollow`'s signature to accept a `string` instead of `{ liveUserId: string }`, because it only has 1 argument
- Change `onRemoveEditorClicked`'s signature to accept a `string` instead of `{ liveUserId: string }`, because it only has 1 argument (`data` is never passed)
- Extract `FollowButton`, `LiveMode`, `MakeEditorButton`, `MakeSpectatorButton` & `Preferences` from `LiveInfo`


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3444/commits/f902fbad9039aa120d4c8403ac150a9dcdd28dc4 is at https://pr3444.build.csb.dev/s/new.

SaraVieira:
Same idea as https://github.com/codesandbox/codesandbox-client/pull/3461

PR #3452, opened 2020-02-03
Move esmodule export to be after the declaration

### LINKED ISSUE #3448: convert-esmodule exports default class before initialisation
# 🐛 bug report

## Description of the problem
The logic in the `convert-esmodule` transpiler treats default class and function exports equally: 
https://github.com/codesandbox/codesandbox-client/blob/304a43823fe4ac1e63bdebdc8eb7129876d8ed38/packages/app/src/sandbox/eval/transpilers/babel/convert-esmodule/index.ts#L162-L165

However, this results in classes being exported in the temporal dead zone, before they are initialised:
https://github.com/codesandbox/codesandbox-client/blob/304a43823fe4ac1e63bdebdc8eb7129876d8ed38/packages/app/src/sandbox/eval/transpilers/babel/convert-esmodule/__snapshots__/index.test.ts.snap#L3-L10

This will give a `Cannot access x' before initialization` error for any imports that use `export default Class`.

Import any es6 modules with `export default Class` statements.

### Your Environment

| Сodesandbox      |  any |
| Browser          | Chrome |
| Operating System | Mac |

CompuIves:
Potentially fixes #3448.

We need to check if this breaks anything else.

### DIFF
--- a/packages/app/src/sandbox/eval/transpilers/babel/convert-esmodule/__snapshots__/index.test.ts.snap
+++ b/packages/app/src/sandbox/eval/transpilers/babel/convert-esmodule/__snapshots__/index.test.ts.snap
@@ -4,8 +4,8 @@ exports[`convert-esmodule can convert class default exports 1`] = `
 "Object.defineProperty(exports, \\"__esModule\\", {
   value: true
 });
-exports.default = A;
 class A {}
+exports.default = A;
 "
 `;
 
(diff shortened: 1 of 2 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3452/commits/890e4b26ccfedab3a8e99bd645fe129b5af6ab6e is at https://pr3452.build.csb.dev/s/new.

CompuIves:
Tested with multiple sandboxes, and seems to work fine. Also the tests are green, so I think we're safe to merge!

PR #3460, opened 2020-02-04
Add support for terminal messages from sse in embed

### LINKED ISSUE #3356: Cannot get terminal to work in CodeSandbox embed
Hi,

I have a CodeSandbox which is a basic Express app.
You can find it here: https://codesandbox.io/s/cold-butterfly-nl30s
As you can see, when you visit the above URL, "Hello World!" is displayed in the preview pane and "It Works!" is logged to the terminal.

My problem is that when I embed the sandbox on our site and set the terminal to be open in the embed,  nothing is logged to the terminal.

You can see the problem demonstrated here: https://www.sitepoint.com/?p=174674&preview=1&_ppp=efe3e7b8a1

"Hello World!" is displayed in the preview pane, as expected. However the terminal remains blank (I would like to see "It Works!" logged).

Here is my embed code:

```html
<iframe src="https://codesandbox.io/embed/cold-butterfly-nl30s?codemirror=1&expanddevtools=1&fontsize=14&hidenavigation=1&moduleview=1&theme=dark"
     style="width:100%; height:500px; border:0; border-radius: 4px; overflow:hidden;"
     title="cold-butterfly-nl30s"
     allow="geolocation; microphone; camera; midi; vr; accelerometer; gyroscope; payment; ambient-light-sensor; encrypted-media; usb"
     sandbox="allow-modals allow-forms allow-popups allow-scripts allow-same-origin"
   ></iframe>
```

What am I doing wrong?

CompuIves:
Fixes #3356 .

### DIFF
--- a/packages/app/src/embed/components/Content/index.tsx
+++ b/packages/app/src/embed/components/Content/index.tsx
@@ -120,16 +120,16 @@ export default class Content extends React.PureComponent<Props, State> {
    * Initialize the interface responsible for talking with the server, managing
    * the terminal output
    */
-  initializeExecutor = () => {
+  initializeExecutor = async () => {
     const templateDefinition = getTemplate(this.props.sandbox.template);
     if (templateDefinition.isServer) {
-      this.executor.initialize({
+      await this.executor.initialize({
         host: 'https://codesandbox.io',
         files: {},
         sandboxId: this.props.sandbox.id,
       });
 
-      this.executor.setup();
+      await this.executor.setup();
 
       this.executor.on('sandbox:log', message => {
         dispatch({ type: 'terminal:message', data: message.data });
--- a/packages/app/src/embed/components/Content/index.tsx
+++ b/packages/app/src/embed/components/Content/index.tsx
@@ -5,6 +5,7 @@ import {
   resolveModule,
 } from '@codesandbox/common/lib/sandbox/modules';
 import getTemplate from '@codesandbox/common/lib/templates';
+import { ServerExecutor, IExecutor } from '@codesandbox/executors';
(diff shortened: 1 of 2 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3460/commits/624b9c90f77ce2e0b9ded3f28afc3aec281ae166 is at https://pr3460.build.csb.dev/s/new.

lbogdan:
Unfortunately it doesn't seem to work, https://pr3460.build.csb.dev/embed/nl30s should show "It works!" for every preview request (see https://pr3460.build.csb.dev/s/nl30s ).

CompuIves:
Yeah I noticed this too, seems like the handshake doesn't fire?

lbogdan:
Yes, it doesn't send the initial `sandbox:start` message.

CompuIves:
Ah, missing `await`...

lbogdan:
While testing, I've noticed a small visual inconsistency: each tab has a different background color:

image

image

image

Also, I'm thinking if we should remove the "Open terminal" button? Or keep it for discoverability?

CompuIves:
> While testing, I've noticed a small visual inconsistency: each tab has a different background color:
Mhm, this is related to the embed redesign (cc @DannyRuchtie @siddharthkp), I think we can keep it like this until we've applied the new theme?

> Also, I'm thinking if we should remove the "Open terminal" button? Or keep it for discoverability?
Yeah, would be better to remove it I think!

lbogdan:
This looks ready to merge, after removing the "Open terminal" button!

PR #2818, opened 2019-10-16
🔨 Refactored 🧠 Overmind Hacktober | Refactor app/pages/Sandbox/Editor/Workspace/items/Server/EnvVars/index.js to use Overmind

deinakings:
## What kind of change does this PR introduce?
This is a refactor related to https://github.com/codesandbox/codesandbox-client/issues/2621
@Saeris @christianalfoni

## What is the current behavior?
**EnvironmentVariables** component was using `inject` and `hooksObserver` from `app/componentConnectors`.

## What is the new behavior?
- Refactored to use `useOvermind` from `app/overmind`
- Changed extension of the file from `.js` to `.tsx`

## What steps did you take to test this? This is required before we can merge, make sure to test the flow you've updated.

1. 
```
yarn typecheck
yarn lint
yarn test
```
>I got some warnings for `yarn lint` but no errors.
2. Then added the `jwt` cookie
3. Created a new sandbox container
4. Opened "Server Control Panel"
5. Added a test environment variable:
Screen Shot 2019-10-16 at 9 52 33 AM

## Checklist

- [ ] Documentation N/A
- [x] Testing 
- [x] Ready to be merged
      
- [ ] Added myself to contributors table

PR #3461, opened 2020-02-04
🔨 Switch Server to use useOvermind

MichaelDeBoey:
Follow-up of #2535 & #2818

Things I did extra:
- Change `deleteEnvironmentVariable`'s signature to accept a `string` instead of `{ name: string }`, because it only has 1 argument
- Change `onBrowserFromPortOpened`'s signature to accept a `ServerPort` instead of `{ port: ServerPort }`, because it only has 1 argument
- Extract `ControlContainer`, `OpenPorts`, `RunScripts`, `SecretKeys` & `Status` out of `Server`
- Make `EnvEntry`, `EnvModal`, `RunScripts` (was called `Tasks`) & `Status` a `FunctionComponent`


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3461/commits/cfbddc709b613a3c0ccca7410965bc016d8b4e06 is at https://pr3461.build.csb.dev/s/new.

SaraVieira:
Hey!

I am so sorry for the work but right now we can't accept changes to the sidebar since we will be releasing a new version of the sidebar next week and all this code will be deleted

I know you made some changes also to the actions but the work it would be to make it work in the new one two days before deployment doesn't make sense right now

I'm sorry we didn't make this public :/

PR #3463, opened 2020-02-04
Only send updates to non-optimistic modules in live

CompuIves:
We currently send updates for optimistic modules to the server, which the server cannot process. With this change we don't send updates for optimistic modules, and once we get the confirmation that the module has been created we send the whole module.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3463/commits/84270981ebd680092a7b4aaf8d8a1cd97fa23f69 is at https://pr3463.build.csb.dev/s/new.

PR #3475, opened 2020-02-05
Add new sidebar experiment toggle

SaraVieira:


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3475/commits/04d59b7e9e399712e7dffc8d9bd3da4147ea0e35 is at https://pr3475.build.csb.dev/s/new.

CompuIves:
We used to have a different way to apply settings and experiments, I think. But this seems fine to me as long as it's only temporary! Also interesting for @christianalfoni, since we're going to do this for the LSP as well 😄 .

SaraVieira:
Yeah, I placed it this way so it's easier to remove since it will only stay for a week :)

I tagged Christian if he wants to take a look

PR #3490, opened 2020-02-09
Add ownership to GitHub sandboxes

CompuIves:
This allows users to change properties of a sandbox if they're the owner.
They can change the title, description, things like env variables. Everything
except the code, in which we'll autofork.

The goal of this is to answer to the multiple feature requests related to using
a git sandbox. While the best solution would be to invent a permission model,
we're not there yet and I'm not sure when we'll get it. So this seems like a good
intermediate solution instead.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3490/commits/5553e82c620540468d883678da9ab821bd6db32a is at https://pr3490.build.csb.dev/s/new.

PR #3497, opened 2020-02-11
fix pick sandbox issue

christianalfoni:


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3497/commits/d75e2458d535d63126bea5785ef0b1ee9688c7fc is at https://pr3497.build.csb.dev/s/new.

packages/app/src/app/overmind/namespaces/explore/actions.ts
```diff
@@ -23,30 +23,21 @@ export const pickSandbox: AsyncAction<{
...
-      effects.notificationToast.error(
-        'Could not pick Sandbox, please refresh and try again'
-      );
-      return;
+    if (popularSandbox) {
+      popularSandbox.picks = [
+        {
+          ...data,
+          // Why are we doing this?
+          id: Math.random().toString(),
```
SaraVieira: I did that at the time but honestly I don't remeber why I think it was for all picks to have an id

christianalfoni: Ah, cool, thanks @SaraVieira :) 

Btw, I added a fix now so that it pushes to that array, not replacing it. To avoid it overriding existing picks... not sure how this works, but yeah, seems more correct :)

PR #3517, opened 2020-02-14
WIP: Redesign header

siddharthkp:
Header redesign: https://www.figma.com/file/Pn3qsrxzqFN0s6tB2eTbe6SC/IDE?node-id=2762%3A11753


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3517/commits/72eb9a2cea073e17dad4ad522fb226846da8f508 is at https://pr3517.build.csb.dev/s/new.

SaraVieira:
A couple of issues I saw:

The cog and text are not aligned:
Screenshot 2020-02-14 at 16 34 41


The icon is too close to the name of the sandbox
Screenshot 2020-02-14 at 16 34 44

It also breaks when you log in, being logged in is fine but logging in breaks
Screenshot 2020-02-14 at 16 37 08

CompuIves:
Very nice! Compared to the design, the text seems quite big for the button itself I think. It also looks a bit better in my opinion with font-weight 500. What confuses me endlessly is that the dev implementation is exactly implemented as the design (except for the font size, but that's not the main reason I think), but it looks worse in the browser. I don't understand how that could happen 🤔 .

siddharthkp:
@SaraVieira Whoops, didn't check logged out state! BRB!

siddharthkp:
@SaraVieira done!

siddharthkp:
> Very nice! Compared to the design, the text seems quite big for the button itself I think. 

yep, the designs _had_ 10px which doesn't render nicely in browsers, danny changed it to 12 in figma too :D

packages/app/src/app/pages/Sandbox/Editor/Header/SandboxName/index.tsx
```diff
@@ -90,66 +81,72 @@ export const SandboxName: FunctionComponent = () => {
       (currentSandbox.team ? currentSandbox.team.name : 'My Sandboxes')
     : 'My Sandboxes';
 
-  const spring = useSpring({
```
CompuIves: Hmm, why is the animation removed in this case?

CompuIves: Ohh, I see, the current animation is quite bad 😅

packages/app/src/app/pages/Sandbox/Editor/Header/Actions.tsx
```diff
@@ -0,0 +1,134 @@
...
+  };
+
+  let primaryAction;
+  if (!hasLogIn) primaryAction = 'Sign in';
+  else primaryAction = owned ? 'Embed' : 'Fork';
+
+  return (
+    <Stack align="center" gap={1} css={{ button: { width: 'auto' } }}>
+      {updateStatus === 'available' && (
+        <TooltipButton
```
SaraVieira: This should be green or something like it s now to call out attentio

packages/app/src/app/pages/Sandbox/Editor/Header/SandboxName/index.tsx
```diff
@@ -90,66 +81,72 @@ export const SandboxName: FunctionComponent = () => {
...
-            owned={owned}
-            aria-label={
-              owned ? `${sandboxName}, change sandbox name` : sandboxName
-            }
-          >
-            {sandboxName}
-          </Name>
+          <>
+            {owned ? (
+              <Button
```
SaraVieira: Can you please copy over the aria aswell?

packages/app/src/app/pages/Sandbox/Editor/Header/SandboxName/index.tsx
siddharthkp: Yeah, it didnt contribute much to the UX. Would be cool to work on a consistent _motion_ design for our interface with danny

packages/app/src/app/pages/Sandbox/Editor/Header/SandboxName/index.tsx
siddharthkp: yep! my bad

packages/app/src/app/pages/Sandbox/Editor/Header/Actions.tsx
siddharthkp: The idea was not to call too much attention to it 🤷‍♀ because it isn't the core thing you're here for

SaraVieira: The issue is that if people don't click they may get unrelated bugs :/

siddharthkp:
> the text seems quite big for the button itself I think. It also looks a bit better in my opinion with font-weight 500. 

will look at it, we might want to bump up the weight for primary and secondary buttons to distinguish them a little

packages/app/src/app/pages/Sandbox/Editor/Header/Actions.tsx
siddharthkp: important problem, bad solution?

PR #2880, opened 2019-10-20
refactored /app/src/app/pages/Dashboard/Sidebar/index.tsx dashboard sandbox index to overmind

lusan:
## What kind of change does this PR introduce?

Refactors code as a part of hacktoberfest mentioned in #2621.
@Saeris @christianalfoni

## What is the current behavior?

packages packages/app/src/app/pages/Dashboard/Sidebar/index.js was using Cerebral

## What is the new behavior?
packages /app/src/app/pages/Dashboard/Sidebar/index.tsx is using Overmind

## What steps did you take to test this? This is required before we can merge, make sure to test the flow you've updated.

1. Replaced inject and hooksObserver with useOvermind
2. yarn test
3. yarn lint
4. yarn typecheck

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      
- [ ] Added myself to contributors table

PR #3518, opened 2020-02-14
🔨 Switch Dashboard/Sidebar to use useOvermind

MichaelDeBoey:
Follow-up of #2859 & #2880 

Things I did extra:
- `Item`
  - Make `Item` a `FunctionComponent`
  - Use `react-router-dom`'s `useRouteMatch` instead of `Route` component so we have less indented code
  - Extract `Container` into it's own component

- `TemplateItem`
  - Redo #3127 (since it was reverted by #2225 for some reason)
    > Resubmission of #3070 after 37223ea, with a fixed `entryTarget` (see #3124)
    > 
    > Follow-up of #2737
    > 
    > Things I did extra:
    > - Just use a single file to not pollute the project files unnecessarily
    > - Remove all `react-router-dom` references, since it's not used in this component
    > - Type `entryTarget` & `collectTarget`
    > - Simplify `entryTarget`'s `canDrop`, since `!props.removedAt` will always be `true` as `removedAt` isn't a prop and isn't ever passed through & `monitor` is never `null`

- `TrashItem`


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3518/commits/391d2319ba667c9fd5cdf98f4b052b396aed0191 is at https://pr3518.build.csb.dev/s/new.

SaraVieira:
Hey!

Sorry for the delay but we are now writing a new dashboard so it doesn't really make sense to merge this and create conflicts

PR #3540, opened 2020-02-18
Publish @codesandbox/components package

siddharthkp:
yay?

not sure if i broke the app


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3540/commits/900e486b0221c3bcb73700e2f453e61a53235823 is at https://pr3540.build.csb.dev/s/new.

packages/components/src/design-language/index.ts
```diff
@@ -0,0 +1,3 @@
+import designLanguage from '@codesandbox/common/lib/design-language/theme';
+
+export default designLanguage;
```
MichaelDeBoey: Any specific reason you just re-export `@codesandbox/common/lib/design-language/theme`? 🤔

siddharthkp: Yep

We replace that file during the build step to "bundle in" the design language. It's a quick way of creating a bundle that can be published to npm without publishing the latest version of common. 

A nicer way of doing this would be to use rollup to resolve internal dependencies.

MichaelDeBoey: Seems like a good trade-off then indeed 🙂

PR #3555, opened 2020-02-20
fix expired token experience

christianalfoni:
When the token is expired we have a bit of weird behaviour.

Now we ensure that the `NewSandboxPage` does not load until we have verified the token. Also not having duplicate messages etc. The same for `Dashboard`, it checked if we have a token, but it has no idea if that token is actually valid. The `Sandbox` handles this already as it is waiting for a "bigger loading state".

To test this add a silly token to your local storage: "123" or something. Then load the frontpage, the sandbox page and the dashboard page. Also try signing in and signing out on each page to see if it works correctly.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3555/commits/192c3cf956659f18218ead5f9ef5dbc7682044a0 is at https://pr3555.build.csb.dev/s/new.

SaraVieira:
we have moved on to cookies lol

PR #3177, opened 2019-12-11
Jyotiarora2610 refactor liveindex

SaraVieira:
closes #2879

PR #3558, opened 2020-02-20
🔨 Switch Live to use useOvermind

MichaelDeBoey:
Follow-up of #3177 & #3208

Things I did extra:
- Change `signInToRoom` & `roomJoined`'s signatures to accept a `string` instead of `{ roomId: string }`, because they only have 1 argument
- Extract `Error`, `Loading` & `NotAuthenticated` from `Live`
- Remove `BlinkingDot`


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3558/commits/2e017e337613434788180aad81b179df85a9ffc3 failed.

MichaelDeBoey:
@SaraVieira @christianalfoni I've rebased onto `upstream/master`, so this could be merged now 🙂

MichaelDeBoey:
@SaraVieira @christianalfoni I've rebased onto `upstream/master`, so this could be merged now 🙂

MichaelDeBoey:
@SaraVieira @christianalfoni I've rebased onto `upstream/master`, so this could be merged now 🙂

MichaelDeBoey:
@SaraVieira @christianalfoni I've rebased onto `upstream/master`, so this could be merged now 🙂

SaraVieira:
Thank you!

PR #3571, opened 2020-02-24
Sidebar enhacements

siddharthkp:
changes in polyfill settings:

1. File tree and other lists in sidebar ensure list.hover and list.background (sidebar.background) are different colors by falling back to list.inactiveSelectionBackground
2. Activity bar: ensure enough contrast between selected foreground and inactive foreground

PR #3577, opened 2020-02-25
Fix feature flag for container lsp

CompuIves:
For the feature flag for the container we don't use the localStorage, but a database property. We need to do this to determine which sandboxes have the experimental feature enabled. It turns out that our implementation of localStorage vars is half done, which is quite bad.

This PR changes the LSP feature flag to use the user, but there is still one API call missing. We need to do a POST `/api/v1/users/current` with body:

```
{
  "user": {
    "experiments": {
      "container_lsp" boolean
    }
  }
}
```

To update the setting. Preferably we'd have an overmind action for this. Can anyone @codesandbox/frontend pick this up and finish this PR? I can't do it until tomorrow.

### DIFF
--- a/packages/app/src/app/pages/common/Modals/PreferencesModal/Experiments/ContainerLSP.tsx
+++ b/packages/app/src/app/pages/common/Modals/PreferencesModal/Experiments/ContainerLSP.tsx
@@ -17,6 +17,8 @@ export const ContainerLSP: React.FunctionComponent = () => {
     setContainerLSP(val);
     window.localStorage.setItem('CONTAINER_LSP', val.toString());
     location.reload();
+
+    // TODO(@someone): call the API call on POST /api/v1/users/current with body `{ experiments: { container_lsp: true } }`
   };
 
   return state.user ? (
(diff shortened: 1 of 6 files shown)


packages/app/src/app/pages/common/Modals/PreferencesModal/Experiments/ContainerLSP.tsx
```diff
@@ -17,6 +17,8 @@ export const ContainerLSP: React.FunctionComponent = () => {
     setContainerLSP(val);
     window.localStorage.setItem('CONTAINER_LSP', val.toString());
     location.reload();
+
+    // TODO(@someone): call the API call on POST /api/v1/users/current with body `{ experiments: { container_lsp: true } }`
```
MichaelDeBoey: I think we should revert @SaraVieira's changes from https://github.com/codesandbox/codesandbox-client/pull/3571/commits/6e8d7e68a732230fa755bb2259e3863b77b46406 and use the action again

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3577/commits/5fa7318c8b9a16e0412ea5433136fc8d16d63d1f is at https://pr3577.build.csb.dev/s/new.

PR #3578, opened 2020-02-25
fix experiments tab when logged out

SaraVieira:
with the implementation in #3577 if the user is logged out the experiments tab breaks


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3578/commits/12023e2542ff38761cf699587247c06dacc02b4d is at https://pr3578.build.csb.dev/s/new.

PR #3587, opened 2020-02-26
sync styled version in notifications

siddharthkp:
Notifications uses a different major version of styled-components which means there are 2 version on the page, which is not really useful (theme clashes, bundlesize, etc.)

Downgraded to 4.x in notifications

⚠️ Not sure if we depend on any v5 features in notifications 🤔

### DIFF
--- a/packages/notifications/package.json
+++ b/packages/notifications/package.json
@@ -16,7 +16,7 @@
   },
   "dependencies": {
     "react-spring": "^8.0.25",
-    "styled-components": "^5.0.0-rc.2",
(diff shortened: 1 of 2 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3587/commits/0a29fb9d4ff53ff3e8f3928c84a7add2923ec82c is at https://pr3587.build.csb.dev/s/new.

CompuIves:
Nice!!!

PR #3594, opened 2020-02-28
Dynamically send operations to server

christianalfoni:
Dynamically sends operation to server based on number of connections.

**NOTE!**

We should remove the console log here before merging with `master`


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3594/commits/dd93dc8af59bbc685d21a7846d33aacd736cece3 is at https://pr3594.build.csb.dev/s/new.

PR #3607, opened 2020-03-04
Make black theme the default

siddharthkp:
- [x] set black as default for folks that haven't explicitly chosen a theme
- [x] update raw theme cache to black
- [x] remove duplicate black theme

Folks are allowed to switch back to codesandbox theme if they don't like codesandbox black. Drawback: If in the past, you have changed themes and then switched to codesandbox back, we don't change your theme.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3607/commits/3d0af8d578473d6ba4438e8509875d498192c460 is at https://pr3607.build.csb.dev/s/new.

PR #3621, opened 2020-03-05
improves sourcemaps

christianalfoni:
sourcemaps has been broken in dev. Now they work and is scoped to `src/app`, meaning it looks like this:

Screenshot 2020-03-05 at 13 07 20

So all references that are **not** `app.js`, is actual references to the app. In this case `factories.ts` and `index.tsx`:

Screenshot 2020-03-05 at 13 07 28

The rest is just libraries and stuff where we do not really care about the sourcemaps. Which should also make the dev flow build faster.

### DIFF
--- a/packages/app/config/webpack.common.js
+++ b/packages/app/config/webpack.common.js
@@ -163,31 +163,6 @@ module.exports = {
 
   module: {
     rules: [
-      {
-        test: /\.(graphql|gql)$/,
-        exclude: /node_modules/,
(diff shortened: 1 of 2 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3621/commits/fbab58ad859185d1c16db8f28a1a351dcd3ddbeb is at https://pr3621.build.csb.dev/s/new.

PR #3661, opened 2020-03-09
Fix dependency picker hidden behind text

siddharthkp:
this bug was introduced last week in a different bug fix :)

Screenshot 2020-03-09 at 2 55 07 PMScreenshot 2020-03-09 at 2 54 25 PM


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3661/commits/0aa43515c082439153b24e25bcbe769ebce16040 is at https://pr3661.build.csb.dev/s/new.

SaraVieira:
Good catch

PR #3689, opened 2020-03-13
Fix: Icons cant be on scale

siddharthkp:
Discussed this one with Danny and realised that our icons can't be on the 4pt scale

Example:

Screenshot 2020-03-13 at 11 21 36 AM

The button that contains both these icons is the same size - 26x26, but the icons inside them have to be adjusted based how they look next to each other

Because of that, they can't always be 16. or even be on the 8/12/16 scale we have. We need more granular control of them

The ones in the image are 14x14 and 10x10 to look _optically aligned._ (both inside a 26x26 button)

- [x] Changed the icon component
- [x] updated all the components that use icon (backward compat lol)

Notes for future: It would be cool to optically align icons in the library so that there are less scenarios in which we need to optically align them. This can't be abstracted away fully because optical alignment depends on the context/neighbouring icons, so we'll always need this escape hatch 🙃

### DIFF
--- a/packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/Comments/Comment.tsx
+++ b/packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/Comments/Comment.tsx
@@ -70,7 +70,7 @@ export const Comment = React.memo(({ comment }: any) => {
             <Icon name="check" title="Resolved" color="green" />
           )}
           <Menu>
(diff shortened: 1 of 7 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3689/commits/95374a0204b4c9e093971056b67bc5312ec866b2 is at https://pr3689.build.csb.dev/s/new.

PR #3690, opened 2020-03-13
experiment: version in status bar

siddharthkp:
- [x] add version to status bar
- [x] remove version from sidebar
- [x] add status bar to skeleton
- [x] add click to copy

Warning: To make it look and feel like vscode's status bar, the code is absolute garbage. It's not even completely semantic 🤷 But it feels *integrated* to the status bar


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3690/commits/b5a41549be8e2e442cc06b62f6c5bf316dc4ed59 is at https://pr3690.build.csb.dev/s/new.

CompuIves:
I like this very much! Would love to get it in somewhere in the coming week.

I prefer it to show on the left side I think, right side gives information about the current document, left info is global. It would also be nice if you could click on it to copy the version, and another very nice thing would be if we could indicate that there's an update (if there's an error) there.

siddharthkp:
@CompuIves ready for re-review

DannyRuchtie:
Can we have the same padding left and right?
Now the version is closer to the edge than the Filetype (right)

siddharthkp:
> Now the version is closer to the edge than the Filetype (right)

oh damn, will do

siddharthkp:
This is fixed now 🎉

PR #3706, opened 2020-03-17
Accessible dialog component v1

siddharthkp:
- [x] added component
- [x] added stories/tests

supporting changes: IconButton supports innerRef now

Pending decisions:
- [ ] should _close when clicked outside_ be optional (thinking comments)
- [ ] fix upstream bug https://github.com/reach/reach-ui/issues/390

Probably not in this PR:
- [ ] add option for animate from trigger position (thinking comments)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3706/commits/e9f544e174a07cb71ae182e6eab7a148d151e346 failed.

packages/components/src/components/Dialog/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+    dialogVisible,
+    setDialogVisibility,
+    triggerRect,
+    label,
+  } = React.useContext(context);
+
+  const [overlayVisible, setOverlayVisiblity] = React.useState(false);
+
+  React.useEffect(() => {
+    if (dialogVisible) setOverlayVisiblity(true);
```
MichaelDeBoey: Put curly braces around body to make it less error prone

```suggestion
    if (dialogVisible) {
      setOverlayVisiblity(true);
    }
```

packages/components/src/components/Dialog/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+            )}
+          </AnimatePresence>
+        )}
+      </Rect>
+    </DialogOverlay>
+  );
+};
+
+const centered = (triggerRect, dialogRect) => {
+  if (!dialogRect || !triggerRect) return { left: 0, top: 0 };
```
MichaelDeBoey: Same

```suggestion
  if (!dialogRect || !triggerRect) {
    return { left: 0, top: 0 };
  }
```

packages/components/src/components/Dialog/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+const context = React.createContext({
+  dialogVisible: false,
+  setDialogVisibility: null,
+  triggerRef: null,
+  triggerRect: null,
+  label: null,
+});
+
+type DialogTypes = React.FC<{
```
MichaelDeBoey:
```suggestion
type DialogTypes = FunctionComponent<{
```

packages/components/src/components/Dialog/index.tsx
```diff
@@ -0,0 +1,179 @@
+import React from 'react';
```
MichaelDeBoey:
```suggestion
import React, { FunctionComponent } from 'react';
```

packages/components/src/components/Dialog/index.tsx
siddharthkp: Not a fan of this, sorry!

PR #3713, opened 2020-03-18
Explicitly set Same-Site:None on preview secret cookies

### LINKED ISSUE #3617: Chrome 80 SameSite=Lax change broke private sandboxes
# 🐛 bug report

## Description of the problem

It looks like the SameSite=Lax change in Chrome 80 has broken private sandboxes. When accessing a private sandbox, the app never loads and the page continues to reload tons of network assets (network tab goes nuts).

If you set [chrome://flags/#same-site-by-default-cookies](chrome://flags/#same-site-by-default-cookies) to `Disabled`, everything goes back to normal. I've tried this with every scale of app, from the most basic, `document.write('test')` app to a complex React CRA-based SPA and get the same symptoms.

The only cookie I see that's affected by this change is `_cfduid`, but I'm not 100% sure.

It's broken my heart. I'm not sure I'll ever recover 😭 

For real, though -- we leverage CodeSandbox for our organizations FE tech screens. We're able to work around this by instructing our candidates to use Firefox, but we are really averse to any added friction at the start of a tech screen.

### Link to sandbox: [link]() (optional)

Useless since you can only see the issue in private sandboxes. Should be easy enough to repro, though.

### Your Environment

| Сodesandbox      | Current |
| Browser          | Chrome 80+ |
| Operating System | Mac OS X Catalina |

CompuIves:
Fixes #3617

### DIFF
--- a/packages/sandbox-hooks/preview-secret.js
+++ b/packages/sandbox-hooks/preview-secret.js
@@ -22,7 +22,7 @@ export const setPreviewSecret = secret => {
     cookieValue !== secret
   ) {
     if (secret) {
-      document.cookie = `${PREVIEW_SECRET_COOKIE_NAME}=${secret}`;
+      document.cookie = `${PREVIEW_SECRET_COOKIE_NAME}=${secret};samesite=none;secure;`;
       location.reload();
     } else {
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3713/commits/42db61b9278b37473ed5ad44824a508ae3bca96e is at https://pr3713.build.csb.dev/s/new.

PR #3716, opened 2020-03-18
Module recover fixes

christianalfoni:
When recovering files it will give you a sticky notification with three choices:

1. Apply - It just applies the code from the recovered files
2. Compare - It fires up VSCode diff views for each recovered file, where you manually decide to merge in our not
3. Discard - You just delete it and move on

If you are to exit the notification, the recovered files are kept in local storage, meaning if you refresh again you get the notification again.

This is also tested with **Collaboration mode**.

This PR includes a fix to `Toast`, which now actually allows having multiple primary and secondary buttons.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3716/commits/285053d2a46fd464dd73d96de16ef9414c8bcd61 is at https://pr3716.build.csb.dev/s/new.

packages/notifications/src/component/Toast.tsx
```diff
@@ -138,40 +138,40 @@ export function Toast({ toast, removeToast, getRef, colors, Button }: Props) {
                 }}
               >
                 {toast.notification.actions.primary &&
-                  toast.notification.actions.primary[0] && (
+                  toast.notification.actions.primary.map(primary => (
```
CompuIves: I'm not 100% sure if this works as intended as I was trying to simulate how VSCode does this. I will take a second look at that now, I think they put primary buttons behind a caret or something?

packages/app/src/app/overmind/effects/vscode/index.ts
```diff
@@ -496,6 +496,29 @@ export class VSCodeEffect {
...
+    const recoverPath = `/recover/${sandboxId}/recover-${module.title}`;
+    const filePath = `/sandbox${module.path}`;
+    const fileSystem = window.BrowserFS.BFSRequire('fs');
+
+    if (!fileSystem.existsSync(`/recover/${sandboxId}`)) {
+      fileSystem.mkdirSync(`/recover/${sandboxId}`);
+    }
+    fileSystem.writeFileSync(recoverPath, oldCode);
+
+    this.editorApi.editorService.openEditor({
```
CompuIves: Let's add a comment on what the heck is happening here 😂. This is some really tricky logic, we're dancing around VSCode here.

packages/notifications/src/component/Toast.tsx
CompuIves: Seems to work fine!! Maybe we should make one button primary, instead of 3 primary buttons?

CompuIves:
When running "Apply Changes" the changes do get applied, but the VSCode file will stay dirty (the dot in the tab) if we change the file in such a way that it's the same as saved. We probably have to run `model.revert()` if `savedCode === code` to fix that?

christianalfoni:
Hm, not sure what you mean? When you do "apply changes" it means that you apply your unsaved changes, which means the file is dirty? We do not "apply and save", we just apply.

If you change the content to what was originally there, the saved content, it should become "undirty" again. 

In my mind this behaves correctly? 

Fixed the typing :)

PR #3719, opened 2020-03-19
Handle long names in the sidebar

siddharthkp:
Use maxWidth liberally

- [x] long file names
- [x] long sandbox names
- [x] long environment names

Haven't solved dependencies yet, that needs a change in structure to work (different PR)

### DIFF
--- a/packages/components/src/components/Avatar/index.tsx
+++ b/packages/components/src/components/Avatar/index.tsx
@@ -21,6 +21,7 @@ export const AvatarContainer = styled(Element).attrs({ as: 'span' })(
     height: 8,
     width: 8,
     position: 'relative',
+    flexShrink: 0, // avatars should never shrink inside flex
   })
 );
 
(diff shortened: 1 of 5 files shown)


packages/components/src/components/Avatar/index.tsx
```diff
@@ -21,6 +21,7 @@ export const AvatarContainer = styled(Element).attrs({ as: 'span' })(
     height: 8,
     width: 8,
     position: 'relative',
+    flexShrink: 0, // avatars should never shrink inside flex
```
siddharthkp: flex makes its children shrink when there's not enough space, an avatar should never shrink though

i'm slightly conflicted about the location of this rule. Should it be in a Stack? The stack has rules for elements it will shrink and not; Or should it be in the Avatar which controls whether it will shrink no matter where it is put 🤷‍♂ 

@SaraVieira thoughts?

SaraVieira: I think it makes sense to be in here since this a very specific case, I am good with having it in here

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3719/commits/d5e072e7a06f122dc5338a81e209439a8a78315e is at https://pr3719.build.csb.dev/s/new.

packages/components/src/components/Avatar/index.tsx
siddharthkp: here - Stack or Avatar?

MichaelDeBoey: I think it should be in `Avatar`.
You also can't fix this inside the parent I'm afraid, it's just how `flex` works

siddharthkp: I meant nested inside the Stack

pseudo-syntax: 

'> Avatar' {
  flex-shrink: 0
}

PR #3720, opened 2020-03-19
Don't use list.hover for file tree

siddharthkp:
To differentiate between hover and selected state in file tree - we only change foreground color to signal hover


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3720/commits/3d76ab00f77ffd5d29974cf668b075e15e4db581 is at https://pr3720.build.csb.dev/s/new.

DannyRuchtie:
He @siddharthkp  looking good,
Might be nice to have a transition like this on hover:   transition: color 150ms ease-in-out;
But I vaguely remember a discusion that it may not be desirable (for snappiness) ?

CompuIves:
> Might be nice to have a transition like this on hover: transition: color 150ms ease-in-out;

Yes, at least from my experience I think that a list of items feels slower when individual items have an animation. If you hover over all items quickly then the items "lag" behind with their animation.

PR #3726, opened 2020-03-19
WIP: Comment glyphs

christianalfoni:


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3726/commits/d90016ef9869a5a4649020bb02e5b8688c6ba6d8 is at https://pr3726.build.csb.dev/s/new.

christianalfoni:
I did some changes here:

1. I refactored `comments` into its own namespace of Overmind
2. Though the implementation of active glpyhs made perfect sense I moved it way back to where the glyphs are created, so it rather builds everything at the same spot
3. Now you can select code and write a comment and you it actually creates a comment 😮

packages/@types/ot/index.d.ts
```diff
@@ -21,5 +21,12 @@ declare module 'ot' {
     toJSON(): SerializedTextOperation;
   }
 
-  export { TextOperation };
+  class Selection {
```
siddharthkp: t4st comment

PR #3744, opened 2020-03-24
add permalinks

SaraVieira:
Adds permalinks for comments

Linking to:
http://localhost:3000/s/twilight-platform-fqnmj?comment=0ab2c879-de32-4227-9a79-fc5899f9d72f

Opens the comments sidebar and sets that comment as the selected one opening it


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3744/commits/53f434d692150ef0eac5f98ecb658253d57619de failed.

packages/app/src/app/pages/Sandbox/index.tsx
```diff
@@ -47,6 +72,7 @@ export const Sandbox: React.FC<Props> = ({ match }) => {
     actions.preferences,
     match.params,
     match.params.id,
+    query,
```
CompuIves: I think this would break the collaborator invite token thingie. If we include this in the `useEffect` we will reload the sandbox page every time the URL changes, which is not intended I think. We should probably split this `useEffect` up.

This is what I did to fix it for collaborator invite tokens:

https://github.com/codesandbox/codesandbox-client/commit/9484194377b5f35229596c9883d084cc9c74d9be

SaraVieira: That makes sense, maybe the best idea is to disable eslint for those two effects because I don't want that in the effect I made too

christianalfoni:
This looks great!

❤️  the clipboard effect, would be nice to have in our current `browser` effect (overmind/effects/browser.ts).

To get more insight into what is happening when we load a Sandbox I want to suggest an alternative approach, but you can evaluate if you want to do that refactor or not @SaraVieira 😄 

Instead of using a `React.useEffect` we can rather execute this stuff exactly after we load the comments. The `sandboxChanged` action is where it loads the comments of the sandbox.

So by adding a new effect `effects.router.getCommentId()`, using your query logic, we can use the logic you have to run the actions in the `React.useEffect`. This way there are no checks "if it should run", it just runs naturally after loading the comments initially.

This just generally gives us more control of what and when things happens when we load a Sandbox and we avoid the risk of the React effect running the logic when it should not... which would be difficult to debug.

Again, I do not really see any risk with the implementation so it is totally up to you if you want to do this change 😄

packages/app/src/app/overmind/namespaces/comments/actions.ts
```diff
@@ -155,8 +155,8 @@ export const selectComment: AsyncAction<{
   } else {
     state.comments.currentCommentId = commentId;
     state.comments.currentCommentPositions = {
-      trigger: bounds,
-      dialog: null,
+      trigger: { right: window.innerWidth / 3, top: 0 },
+      dialog: { left: 0, top: 40 },
```
christianalfoni:
```suggestion
      trigger: bounds,
      dialog: null
```

To avoid having duplicate defaults we can rather pass in the bounds from the outside (see below) and then depend on the defaults set by the component 😄

packages/app/src/app/overmind/namespaces/editor/actions.ts
```diff
@@ -220,6 +220,13 @@ export const sandboxChanged: AsyncAction<{ id: string }> = withLoadApp<{
 
         return aggr;
       }, {});
+
+      if (effects.router.getCommentId()) {
+        actions.workspace.setWorkspaceItem({ item: 'comments' });
+        actions.comments.selectComment({
+          commentId: effects.router.getCommentId(),
+        });
```
christianalfoni:
```suggestion
        actions.comments.selectComment({
          commentId: effects.router.getCommentId(),
          bounds: {
            left: effects.browser.getWidth() / 2,
            top: 0,
            right: effects.browser.getWidth() / 3,
            bottom: 0
          }
        });
```

By passing in the trigger bounds here we are more explicit about when this scenario actually occurs, which is when we grab a comment id from the url 😄

packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/Comments/Comment.tsx
```diff
@@ -83,7 +84,18 @@ export const Comment = React.memo<{
               >
                 Mark as {comment.isResolved ? 'Unr' : 'r'}esolved
               </Menu.Item>
-              <Menu.Item onSelect={() => {}}>Share Comment</Menu.Item>
+              <Menu.Item
+                onSelect={() => {
+                  effects.browser.copyToClipboard(
+                    `${window.location.origin}${window.location.pathname}?comment=${comment.id}`
```
christianalfoni: I would put this behind a:

```ts
effects.router.createCommentUrl(comment.id)
```

Just move as much generic logic as possible behind more domain specific methods 😄

SaraVieira: Done :D

packages/app/src/app/overmind/effects/router.ts
```diff
@@ -49,4 +48,10 @@ export default {
   getSandboxOptions() {
     return getSandboxOptions(decodeURIComponent(document.location.href));
   },
+  getCommentId() {
+    return new URL(location.href).searchParams.get('comment');
+  },
+  createCommentUrl(id) {
```
christianalfoni:
```suggestion
  createCommentUrl(id: string) {
```

packages/app/src/app/overmind/namespaces/comments/actions.ts
```diff
@@ -102,7 +102,7 @@ export const closeComment: Action = ({ state }) => {
 
 export const selectComment: AsyncAction<{
   commentId: string;
-  bounds: {
+  bounds?: {
```
christianalfoni:
```suggestion
  bounds: {
```

Since we now always pass it 😄

packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/Comments/Dialog/index.tsx
```diff
@@ -162,6 +162,18 @@ export const Dialog: React.FC = () => {
                         <Menu.Item onSelect={() => setEditing(true)}>
                           Edit Comment
                         </Menu.Item>
+                        <Menu.Item
+                          onSelect={() => {
+                            effects.browser.copyToClipboard(
```
christianalfoni: Okay, I have a suggestion here.

You can create an action in `comments` called like: `copyPermalinkToClipboard`, where you pass the `commentId`. Inside that action you do what you do here, run the two effects.

The great thing now is that both `Dialog` and `Comment` will just call this action and we avoid duplicate logic 😄

packages/app/src/app/pages/Sandbox/index.tsx
```diff
@@ -41,6 +41,9 @@ export const Sandbox: React.FC<Props> = ({ match }) => {
     actions.live.onNavigateAway();
 
     actions.editor.sandboxChanged({ id: match.params.id });
+
+    // I don't want a run if the url changes
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
christianalfoni:
```suggestion
```

PR #3760, opened 2020-03-27
fix: disable self inviting/collaboration

jyash97:
Bug fix

User can invite themselves. 

https://www.loom.com/share/bddade5e001246c3a81e7b1705a5d2bb

## What is the new behavior?
Let's stop self invitation :)

**Open Question**: Do we want to show a notification specifying the error?

https://www.loom.com/share/7a9e617bfcff45348e1ee0c294c09b61

1. tested locally by inviting myself

## Checklist

- [x] Testing 
- [x] Ready to be merged
      
### DIFF
--- a/packages/app/src/app/pages/Sandbox/Editor/Header/Collaborators/AddCollaboratorForm.tsx
+++ b/packages/app/src/app/pages/Sandbox/Editor/Header/Collaborators/AddCollaboratorForm.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import { motion, useAnimation } from 'framer-motion';
 import { Stack, Button } from '@codesandbox/components';
+import { get } from 'lodash';
+import { hasPermission } from '@codesandbox/common/lib/utils/permission';
 import css from '@styled-system/css';
 
 import { Authorization } from 'app/graphql/types';
@@ -10,6 +12,11 @@ import { PermissionSelect, MENU_WIDTH } from './PermissionSelect';
 
 export const AddCollaboratorForm = () => {
   const { actions, state } = useOvermind();
+  const isOwner = hasPermission(
+    state.editor.currentSandbox.authorization,
+    'owner'
+  );
+  const user = get(state, 'user');
 
   const controls = useAnimation();
   const [inputValue, setInputValue] = React.useState<string>('');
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3760/commits/8f548937a8c4e26ff45633102cc1b0fc204728ac is at https://pr3760.build.csb.dev/s/new.

jyash97:
Just checking here is it the intentional behavior? @CompuIves

CompuIves:
Hah, yes this is an interesting one. I decided to allow a user to invite themselves, but give them a warning. That's why you see a warning next to their name: 
image

This way people can still play and test without involving others. But I can see that it can be confusing as well, do you think this would create confusion @jyash97?

jyash97:
Hey @CompuIves 

Thanks for clarifying, I like the idea behind this. And here is my take on it: Initially, I didn't notice the warning icon and thought it as a bug. But soon after noticing this seems confusing as even on view I was able to edit ( because I was the owner ). 

In my opinion, I think this can lead to confusion as on view permission also user can edit their sandbox. We can add a video demonstrating the feature. 

As of now, we can keep it and let us see how the users respond to the same.

jyash97:
Hey @CompuIves 

Another feedback, Check this modal: 

image

On a public sandbox, we show an option to share with permission `can view` but at the top, we say `Everyone can view` so is that option necessary in a public sandbox? What are your thoughts on this ?

jyash97:
Ping @CompuIves

CompuIves:
> On a public sandbox, we show an option to share with permission can view but at the top, we say Everyone can view so is that option necessary in a public sandbox? What are your thoughts on this ?

Ah yes, that's true! I wanted to keep this in here as well, since I wanted everyone to know that the option is available instead of hiding it. In the future we'll have more permission types, and I think it would be confusing for people to see the list change seemingly randomly if another input changes. This would also allow you to add collaborators with view permission, and change the sandbox view type later.

jyash97:
Hey 👋 @CompuIves 

I am impressed with the reason behind the choices and the decision that has been made on the UI and the above issues that I had are clear to me. 

Closing this PR :)

Thanks for the info 💯

PR #3785, opened 2020-03-30
move components and fix adding comment on existing

christianalfoni:
Can now open comments without having to be on comments workspace item.

This also changes the behaviour of adding comment in gutter. You can NOT do that now when existing comments, have to right click


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3785/commits/0d827c224ee538fe5fef879d2e191a82f749b2ba failed.

PR #3792, opened 2020-03-30
Fix code

SaraVieira:
- [x]  Click on avatar, remove the link @Sara Vieira
- [x]  Pre and Code don't have text overflow scroll (do we need to have max-height)? @Sara Vieira
- [x] Make code 13px instead 16

Also adds break word and fixes when you delete a comment in a modal


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3792/commits/fc3066a64ee5dca9b83a513f05bdbe3b96136949 is at https://pr3792.build.csb.dev/s/new.

PR #3813, opened 2020-04-01
Reply animation refactor

siddharthkp:
motivation: animations are imperative, the reply animations has multiple steps in it. with our current implementations, there were multiple bugs that kept cropping up because of the declarative/imperative translation.

Rewrote the animations with a better (more explicit) transition for logic.

TODO: check for state changes in replies and decide animations for them
 - [x] new reply
 - [x] delete reply
 - [x] edit reply
 - [x] switch comments

### DIFF
--- a/packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/Comments/Dialog/index.tsx
+++ b/packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/Comments/Dialog/index.tsx
@@ -416,12 +416,12 @@ const Replies = ({ replies, repliesRenderedCallback }) => {
    * If replies are loaded, do nothing and wait for next animation
    * */
   React.useEffect(() => {
-    if (!replies.length || T > 0) {
+    if (!replies.length) {
       // If the dialog is already open without any replies,
       // just skip all of the animations for opening transitions
       repliesController.set({ opacity: 1, height: 'auto' });
       setStepInTimeline(2);
-    } else if (!repliesAlreadyLoadedOnFirstRender.current) {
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3813/commits/3311486d27222d82aad72fa8a4c61139ba34173a is at https://pr3813.build.csb.dev/s/new.

packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/Comments/Dialog/index.tsx
```diff
@@ -135,12 +135,11 @@ export const Dialog: React.FC = () => {
                 setEditing={setEditing}
                 hasReplies={replies.length}
               />
-              {replies.length ? (
```
SaraVieira, line 138: I set this because of `[null]` that the replies would break when you deleted the last one

siddharthkp: no worries, this is now handled inside the component (so it doesn't break): https://github.com/codesandbox/codesandbox-client/pull/3813/files#diff-6b9465e97e2db14cb6ef5c5076e3b42bR419

The reason for moving it in is to handle the initial state of "no replies" inside the component as the first step to the subsequent states of loading replies / transitions

SaraVieira: Goood

PR #3821, opened 2020-04-02
update styles for comments

SaraVieira:
- [x]  Theme primary color for active comment

    

- [x]  Should be able to open Dialog with sidebar close @Sara Vieira

This is a refactor of icons.css and I moved it to the editor as a global style in styled-components so we have acess to the theme


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3821/commits/a5b58fc465cc62965201a28584fa1ab6ce68dc93 is at https://pr3821.build.csb.dev/s/new.

packages/app/src/app/pages/Sandbox/Editor/index.tsx
```diff
@@ -19,6 +19,9 @@ import { Navigation as NavigationOld } from './NavigationOld';
 import { ContentSkeleton } from './Skeleton';
 import getVSCodeTheme from './utils/get-vscode-theme';
 import { Workspace } from './Workspace';
+import { CommentsGlobalStyles } from './Workspace/screens/Comments/components/GlobalStyles';
+import { MultiComment } from './Workspace/screens/Comments/components/MultiComment';
+import { CommentDialog } from './Workspace/screens/Comments/Dialog';
```
siddharthkp, line 24: totally optional enhancement: I would wrap these 3 in the same file /comments/index and treat it as the _external api_ for comments while all of the other components are internal implementation detail

SaraVieira: The issue is that we already have an index.tsx that is the actual entry point for comments

maybe an API.tsx would make sense and its a good middle ground?

siddharthkp: oooh, i like it!

SaraVieira: Done! :D

PR #3833, opened 2020-04-03
more polish

SaraVieira:
- [x]  Line hight of in text-area is to small
- [x]  Border-radius of Image Place-holder container
- [x]  Hover action reads 'resolve comment' when it's already been resolved
- [x]  Edit textarea linehight (not sure it's a different component)
- [x]  When Linking to a comment in an other comment  it currently does a full refresh (not sure how tricky it would not be to just  jump to the other comment - no refresh)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3833/commits/0ce361f7c78d84baf4b16870e01820f44384bcfa is at https://pr3833.build.csb.dev/s/new.

PR #3836, opened 2020-04-04
fix: hide delete button from liked sandboxes

jyash97:
## What kind of change does this PR introduce?
Bug Fix

## What is the current behavior?
Non-functional delete button in Liked sandboxes in Profile view

## What is the new behavior?
Hide delete button in Liked Sandboxes page

## What steps did you take to test this? This is required before we can merge, make sure to test the flow you've updated.

- tested locally by traversing sandboxes page & Like sandbox page

## Checklist

- [x] Testing 
- [x] Ready to be merged
      
- [x] Added myself to contributors table

PR #3838, opened 2020-04-04
Cleanup SandboxList

MichaelDeBoey:
Follow-up of #3836

Things I did extra:
- Move `DeleteSandboxButton` into the `SandboxList`, since it's only used there
- Don't show the tableHeader for `DeleteSandboxButton` when it's not shown


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3838/commits/9ccbdc2486e86a71bdd37965988a986213124c1f is at https://pr3838.build.csb.dev/s/new.

MichaelDeBoey:
Always a pleasure 🙂

PR #3844, opened 2020-04-06
better deriving

christianalfoni:
- In this PR we are rather deriving replies by looking at what comments has the current comment as its `parentComment`. Meaning we simplify optimistic updates
- I noticed the `waitAtLeast` for replies loading used values as seconds, now making sure they are represented as milliseconds, giving the correct delay
- I have introduced an issue, which I am not sure how to solve. On subscriptions we are now only updating if the user id differs, this is to avoid duplicate renders (flickers). But now if you have two tabs open as the same user, one tab will not be updated when you write comment in the other... not sure if it is an actual issue

Btw... I saw the skeleton for replies is removed? Was that because of the wrong timer?


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3844/commits/87fbabe4ff46fa3f6e7fbd3fb480b9771513dc7d is at https://pr3844.build.csb.dev/s/new.

packages/app/src/app/overmind/namespaces/comments/actions.ts
```diff
@@ -523,22 +499,34 @@ export const onCommentAdded: Action<CommentAddedSubscription> = (
   { state },
   { commentAdded: comment }
 ) => {
-  state.comments.comments[comment.sandbox.id][comment.id] = comment;
+  // We only add it if it was NOT created by the user, cause then we
+  // have already set it, avoiding UI flickers
+  if (comment.user.id !== state.user?.id) {
```
CompuIves, line 504: Maybe it would be better to do an `Object.assign` or a nullcheck. I can see that this breaks if someone has the sandbox open with two tabs, and the second tab doesn't receive the comment.

packages/app/src/app/overmind/namespaces/comments/actions.ts
```diff
@@ -523,22 +499,34 @@ export const onCommentAdded: Action<CommentAddedSubscription> = (
...
   { state },
   { commentChanged: comment }
 ) => {
-  Object.assign(
-    state.comments.comments[comment.sandbox.id][comment.id],
-    comment
-  );
+  // We only change it if it was NOT created by the user, cause then we
+  // have already set it, avoiding UI flickers
+  if (comment.user.id !== state.user?.id) {
```
CompuIves, line 515: Same here!

packages/app/src/app/overmind/namespaces/comments/actions.ts
```diff
@@ -523,22 +499,34 @@ export const onCommentAdded: Action<CommentAddedSubscription> = (
...
 };
 
 export const onCommentRemoved: Action<CommentRemovedSubscription> = (
   { state },
   { commentRemoved: comment }
 ) => {
-  delete state.comments.comments[comment.sandbox.id][comment.id];
+  // We only delete it if it was NOT created by the user, cause then we
+  // have already set it, avoiding UI flickers
+  if (comment.user.id !== state.user?.id) {
```
CompuIves, line 529: Same here! Haha, maybe in case some admin removes a comment for example.

packages/app/src/app/overmind/namespaces/comments/state.ts
```diff
@@ -108,8 +108,17 @@ export const state: State = {
 
     return {
       ...comments[currentSandbox.id][currentCommentId],
-      comments: comments[currentSandbox.id][currentCommentId].comments
-        .map(commentId => comments[currentSandbox.id][commentId.id] || null)
+      comments: Object.keys(comments[currentSandbox.id])
```
CompuIves, line 111: Cleaner, nice!

packages/app/src/app/overmind/namespaces/comments/actions.ts
```diff
@@ -325,18 +325,17 @@ export const addComment: AsyncAction<{
...
 
-  if (optimisticComment.parentComment) {
-    comments[sandboxId][optimisticComment.parentComment.id].comments.push({
-      id,
-    });
-  }
   state.comments.selectedCommentsFilter = CommentsFilterOption.OPEN;
 
+  if (parentCommentId) {
+    comments[sandboxId][parentCommentId].replyCount++;
```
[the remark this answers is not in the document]
christianalfoni: We do not always have a `parentCommentId` passed to action, we only have that on replies 😄

packages/app/src/app/overmind/namespaces/comments/actions.ts
```diff
@@ -373,6 +360,9 @@ export const addComment: AsyncAction<{
       'Unable to create your comment, please try again'
     );
     delete comments[sandboxId][id];
+    if (parentCommentId) {
+      comments[sandboxId][parentCommentId].replyCount--;
```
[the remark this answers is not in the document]
christianalfoni: Same here! ;)

packages/app/src/app/overmind/namespaces/comments/actions.ts
christianalfoni: I think this can be improved when we now define the ID in the client, will update from master and look at this after @SaraVieira polish is merged in

packages/app/src/app/overmind/namespaces/comments/actions.ts
christianalfoni: Jup, should be okay here as well after merge from master (like, removing it)

PR #3846, opened 2020-04-06
Comments dialog: create an overlay when dragging

siddharthkp:
When you drag the comment dialog around, it tends to select elements in the background that you can drop things in. Example: the url bar in preview.

To prevent this, we create an overlay that is shown only when the user is dragging.

We track "dragging" by hooking on to the `onPanEnd` on `DragHandle`


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3846/commits/125c3882fb5e63e01095c413405df2dceebb44c4 is at https://pr3846.build.csb.dev/s/new.

PR #3850, opened 2020-04-07
More polish

SaraVieira:
- [x]  The commentCount and text are outside the hierarchy of the comment itemprop, so aren’t linked to the comment metadata like name, dateCreated etc.
- [x]  commentCount takes an integer, we're currently passing a string e.g. 5 replies
- [x]  Spacing underneath text input
- [x]  Filter active not showing
- [x]  Why is permalinking not an action in Floating Comments?
- [x]  Scroll bar when expanding replies from Comments History. (On FireFox)
- [x]  Clicking add comment icon in margin twice, breaks.
- [x]  When typing shift+return in floating comments text gets a bit wobbly (or better: quick and single shift right after return).


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3850/commits/87a00ebfafff9f6508d49e3c015d7e1f7d5b56fb is at https://pr3850.build.csb.dev/s/new.

packages/app/src/app/overmind/namespaces/comments/state.ts
```diff
@@ -2,8 +2,9 @@ import { CommentsFilterOption } from '@codesandbox/common/lib/types';
 import { CommentFragment, CommentWithRepliesFragment } from 'app/graphql/types';
 import { Derive } from 'app/overmind';
 import isToday from 'date-fns/isToday';
+import { v4 as uuidv4 } from 'uuid';
 
-export const OPTIMISTIC_COMMENT_ID = 'OptimisticCommentId';
+export const OPTIMISTIC_COMMENT_ID = uuidv4();
```
christianalfoni, line 7: Ops, this needs to be a factory or we will have the same ID on every new comment :-)

`export const OPTIMISTIC_COMMENT_ID = () => uuidv4();`

But! I would rather remove this and use `uuid` directly wherever we need to create an id.

That way we can also remove any logic related to `OPTIMISTIC_COMMENT_ID` in components, cause it is actually not needed anymore as we always have an ID to reference now 😄

CompuIves:
Server is live!

PR #3853, opened 2020-04-07
Component stories typecheck

siddharthkp:
Ugh, want to do a refactor, but don't have tests for types.

Fixing types in stories first so that we can check them.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3853/commits/fc0491156cac25acfab0a6602de0376f38d6c5e1 failed.

PR #3869, opened 2020-04-09
Comments: Now the reply skeleton is actually themed

siddharthkp:
Oh boy, the replies skeleton wasn't actually themed until now. It was always black.

```diff
- const themeType = props.theme.type;
+ const themeType = props.theme.vscodeTheme.type;
```

### DIFF
--- a/packages/components/src/components/SkeletonText/index.tsx
+++ b/packages/components/src/components/SkeletonText/index.tsx
@@ -11,7 +11,7 @@ const pulse = keyframes`
 
 export const SkeletonText = styled(Element)(props => {
   const color = props.theme.colors.sideBar.border;
-  const themeType = props.theme.vscodeTheme.type;
+  const themeType = props.theme.type || props.theme.vscodeTheme.type;
 
   /**
    * This is fun,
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3869/commits/b03e23e764fc89e938f9b18ef37679deaae88a01 is at https://pr3869.build.csb.dev/s/new.

packages/components/src/components/SkeletonText/index.tsx
```diff
@@ -11,7 +11,7 @@ const pulse = keyframes`
 
 export const SkeletonText = styled(Element)(props => {
   const color = props.theme.colors.sideBar.border;
-  const themeType = props.theme.type;
+  const themeType = props.theme.type || props.theme.vscodeTheme.type;
```
siddharthkp, line 14: hate doing this, but this based on if you use <ThemeProvider> or `withTheme` inside the component.

`withTheme` use cases would vanish over time, but right there are a few 🙃

SaraVieira: why this so sad

PR #3872, opened 2020-04-09
Improve login and logout

christianalfoni:
This improves the general login/logout flow:

- When you logout there are no state changes, causing this weird middle state... you log out and the browser refreshes
- Now we track what page you are on. When you log in on a page it will rerun the mounting action for that page

I think we might have had a few subtle bugs related to this as some pages did not rerun the "appLoad" logic. Meaning you might have had a user and jwt, but missing settings etc.

I have tested logging in and out on all pages affected.

### DIFF
--- a/packages/app/src/app/overmind/actions.ts
+++ b/packages/app/src/app/overmind/actions.ts
@@ -3,6 +3,7 @@ import {
   convertTypeToStatus,
 } from '@codesandbox/common/lib/utils/notifications';
 
+import { Page } from '@codesandbox/common/lib/types';
 import { withLoadApp } from './factories';
 import * as internalActions from './internalActions';
 import { Action, AsyncAction } from '.';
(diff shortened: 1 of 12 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3872/commits/f23f72a2da74af56ad4d141ce5f539a59035491a is at https://pr3872.build.csb.dev/s/new.

CompuIves:
Very nice! Much better than what we have, I was testing this flow, and it's behaving a bit weird (or it behaves correctly?). It seems to fork, which is okay I think, I'm not sure. okay this is the flow:

1. Open a sandbox that you don't own
2. Sign in
3. See what happens

SaraVieira:
closing this as the flow for login as changed, feel free to open @christianalfoni if you see there are parts you want to use

PR #3873, opened 2020-04-09
fix: skeleton with color having opacity

jyash97:
Bug Fix

When we select theme `red` and load the editor, the app breaks because skeleton is using `Color` package and `Color('#ff000033')` throws an error. 

## What is the new behavior?
Fixes the color access with opacity.

Screen Recording: https://www.loom.com/share/c56684ce70834e91a8ec822b80061e8d 

1. Locally tested with red theme

## Checklist

- [x] Testing 
- [x] Ready to be merged
      

jyash97:
We may need to find other places as well where we use the same package, we can create a `util` and use it everywhere. Let me know if we need it in this PR or a separate PR ?

CompuIves:
Great work @jyash97 !! Yes, it would be great if we could also have a separate PR with a util!

PR #3881, opened 2020-04-12
Fix refresh for hash based URL in Preview

### LINKED ISSUE #3757: Refresh-button doesn't work for URL with hash (#)
# 🐛 bug report

The manual refresh-button (in the preview browser panel) doesn't work if the page URL has a hash `#` in it.

I'm testing this sandbox, which uses a hash-based router:

https://codesandbox.io/s/sinuous-router-g2eud

Try pressing refresh-button, it doesn't work.

Making changes to the source-code and triggering a refresh with CTRL+S also doesn't work.

It seems anything with a hash in the URL can't be refreshed at all?

(Using Chrome 80.0.3987.149 on 64-bit Windows 10 Pro.)

tanmoyopenroot:

Fix #3757 

`hash` change in the `url` doesn't trigger a page reload in the Iframe

## What is the new behavior?

dispatch `refresh` to manually to trigger `reload` in Iframe

1. Fork -> https://codesandbox.io/s/sinuous-router-g2eud
2. Click on any route Eg: `Bob's Profile` which changes the url in the address bar to `/#users/bob`
3. Then click on `Reload` icon in the `Browser` Tab.
4. It should reload the Iframe.

## Checklist

- [ ] Documentation
- [x] Testing 
- [x] Ready to be merged
      
### DIFF
--- a/packages/common/src/components/Preview/index.tsx
+++ b/packages/common/src/components/Preview/index.tsx
@@ -487,6 +487,8 @@ class BasePreview extends React.Component<Props, State> {
         back: false,
         forward: false,
       });
+
+      this.handleHashedURL(urlInAddressBar);
     }
   };
 
@@ -497,7 +499,10 @@ class BasePreview extends React.Component<Props, State> {
     const urlToSet = urlInAddressBar || url;
 
     if (this.element) {
-      this.element.src = urlToSet || this.currentUrl();
+      const iframeSRC = urlToSet || this.currentUrl();
+      this.element.src = iframeSRC;
+
+      this.handleHashedURL(iframeSRC);
     }
 
     this.setState({
@@ -505,6 +510,14 @@ class BasePreview extends React.Component<Props, State> {
       back: false,
       forward: false,
     });
+  };
+
+  handleHashedURL = url => {
+    if (url.includes('#')) {
+      dispatch({
+        type: 'refresh',
+      });
+    }
   };
 
   handleBack = () => {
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3881/commits/fe5e4ba0a6a2ad1d1618a9f165e979202e29d618 is at https://pr3881.build.csb.dev/s/new.

packages/common/src/components/Preview/index.tsx
```diff
@@ -507,6 +512,14 @@ class BasePreview extends React.Component<Props, State> {
     });
   };
 
+  handleHashedURL = url => {
```
CompuIves, line 515: Would be better if this name would be a bit more descriptive, maybe something like `refreshHashedUrl`.

tanmoyopenroot: Done

packages/common/src/components/Preview/index.tsx
```diff
@@ -507,6 +512,14 @@ class BasePreview extends React.Component<Props, State> {
     });
   };
 
+  refreshHashedUrl = url => {
+    if (url.includes('#')) {
+      dispatch({
+        type: 'refresh',
+      });
+    }
```
MichaelDeBoey, line 520: You could use an early return

```suggestion
    if (!url.includes('#') {
      return;
    }
    
    dispatch({ type: 'refresh' });
```

tanmoyopenroot: Done

CompuIves:
Thanks for the contribution!!

PR #3897, opened 2020-04-14
Don't reuse cache of a module if there are new globals

### LINKED ISSUE #3505: Tests fail when run more than once
# 🐛 bug report

## Description of the problem
When I run some tests the first time then the tests pass. When I run (using the run button) the tests twice or more it seems that the rendered html does not start with a fresh html page and the tests fail. Same happens with refresh button.

### Link to sandbox: [link](https://codesandbox.io/s/5z6x4r7n0p)

CompuIves:
When we change the globals of a file, we need to throw away the cache of that file when we re-evaluate. We didn't do this, and because of that the global of `document` and `window` was re-used between two test runs, even though the globals had changed.

Fixes #3505

### DIFF
--- a/packages/app/src/sandbox/eval/manager.ts
+++ b/packages/app/src/sandbox/eval/manager.ts
@@ -326,7 +326,7 @@ export default class Manager {
 
   evaluateModule(
     module: Module,
-    { force = false, globals = {} }: { force?: boolean; globals?: object } = {}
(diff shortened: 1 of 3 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3897/commits/9a6b08d567f3c28a68761bc2695905a6ee652f35 is at https://pr3897.build.csb.dev/s/new.

PR #3900, opened 2020-04-14
Show actual errors in sandbox error

SaraVieira:

Shows the actual error from the backend when we can't open a sandbox
Also removes the login option because we know the error is that the backend cannot process the repo

Always shows error 422

## What is the new behavior?

Errors, amazing 

1. Open a github repo with a lot of modules or without a package.json

Screenshot 2020-04-14 at 21 33 14

### DIFF
--- a/packages/app/src/app/overmind/namespaces/editor/actions.ts
+++ b/packages/app/src/app/overmind/namespaces/editor/actions.ts
@@ -211,7 +211,9 @@ export const sandboxChanged: AsyncAction<{ id: string }> = withLoadApp<{
 
     actions.internal.setCurrentSandbox(sandbox);
     actions.workspace.openDefaultItem();
+    state.editor.unprocessableEntityError = false;
   } catch (error) {
+    state.editor.unprocessableEntityError = false;
     state.editor.notFound = true;
     const errors = error.response?.data?.errors;
     let detail = errors?.detail;
--- a/packages/app/src/app/overmind/namespaces/editor/actions.ts
+++ b/packages/app/src/app/overmind/namespaces/editor/actions.ts
@@ -213,10 +213,20 @@ export const sandboxChanged: AsyncAction<{ id: string }> = withLoadApp<{
     actions.workspace.openDefaultItem();
   } catch (error) {
     state.editor.notFound = true;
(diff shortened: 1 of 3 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3900/commits/b3754c8233b394953de1f4a61f78770552b4e5bd is at https://pr3900.build.csb.dev/s/new.

packages/app/src/app/overmind/namespaces/editor/actions.ts
```diff
@@ -213,10 +213,20 @@ export const sandboxChanged: AsyncAction<{ id: string }> = withLoadApp<{
     actions.workspace.openDefaultItem();
   } catch (error) {
     state.editor.notFound = true;
-    let detail = error.response?.data?.errors?.detail;
+    const errors = error.response?.data?.errors;
+    let detail = errors?.detail;
+    if (error.response.status === 422) {
+      state.editor.unprocessableEntityError = true;
```
CompuIves, line 219: Will we put this one to false at any point later? Maybe @christianalfoni has a good suggestion for this.

SaraVieira: Oh true, should set it to false on enter that function

packages/app/src/app/overmind/namespaces/editor/actions.ts
```diff
@@ -211,12 +211,24 @@ export const sandboxChanged: AsyncAction<{ id: string }> = withLoadApp<{
 
     actions.internal.setCurrentSandbox(sandbox);
     actions.workspace.openDefaultItem();
+    state.editor.unprocessableEntityError = false;
```
christianalfoni, line 214: I would move this way up to the top of the action, so that every time we start running this logic we "reset" the error. Mr David Xstate would scream if he saw this though. We should ideally have an ENUM of `SandboxError`, being  "notFound" or "unprocessableEntity"... as it can not be both at the same time. 

So if you have time, make the `state.editor.error` to be an object of: `{ message: string, reason: SandboxErrorReason }`. 

But noooo biggie, just if you want to. Call me if you want help 😄

PR #3907, opened 2020-04-15
Add watermark button for non-pro sandboxes

CompuIves:
We now show an "Open Sandbox" button for sandboxes that are shared externally. This only happens for non-pro users.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3907/commits/5a92281a1b24e69a193a7d67f86f322226852345 is at https://pr3907.build.csb.dev/s/new.

PR #3815, opened 2020-04-01
New Privacy Policy

oskarabc:
(test)

Fixes CSB-41

PR #3908, opened 2020-04-15
fix(CI): Only use resource-class:xlarge on master

MichaelDeBoey:
After upgrading containers in 4d3ef6e, a73c2b3 & 5af1e63, CI keeps failing

https://app.circleci.com/pipelines/github/MichaelDeBoey/codesandbox-client/1321/workflows/1069edcc-f24d-463b-9a71-520112f3f5c5/jobs/9259

Schermafbeelding 2020-04-15 om 17 12 46

PR #3924, opened 2020-04-17
Privacy policy followup

MichaelDeBoey:
Follow-up of #3815 

- Use early return
- Cleanup Privacy toast

### DIFF
--- a/packages/app/src/app/overmind/onInitialize.ts
+++ b/packages/app/src/app/overmind/onInitialize.ts
@@ -108,7 +108,11 @@ export const onInitialize: OnInitialize = async (
   effects.preview.initialize(overmindInstance.reaction);
 
   // show terms message on first visit since new terms
-  if (!effects.browser.storage.get(seenTermsKey) && !state.isFirstVisit) {
+  if (effects.browser.storage.get(seenTermsKey)) {
+    return;
+  }
+
+  if (!state.isFirstVisit) {
     effects.analytics.track('Saw Privacy Policy Notification');
     effects.notificationToast.add({
       message:
--- a/packages/app/src/app/overmind/onInitialize.ts
+++ b/packages/app/src/app/overmind/onInitialize.ts
@@ -1,4 +1,3 @@
-import { NotificationStatus } from '@codesandbox/notifications';
 import { OnInitialize } from '.';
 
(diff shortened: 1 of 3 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3924/commits/6d419b3bec2ace14f866731b573e6ff97e410bcd failed.

packages/app/src/app/overmind/onInitialize.ts
```diff
@@ -108,25 +108,27 @@ export const onInitialize: OnInitialize = async (
   effects.preview.initialize(overmindInstance.reaction);
 
   // show terms message on first visit since new terms
-  if (!effects.browser.storage.get(seenTermsKey) && !state.isFirstVisit) {
```
CompuIves, line 111: I wouldn't do this change in this function, since it's now very easy for people to add something unrelated after this body which will not get executed. This function itself is a function that initializes everything, adding an early return could cause confusion. It might be better to extract this logic into a separate function instead and add an early return there.

MichaelDeBoey: I can create an `effect` instead if you want.
What namespace should I put it?

Or do you prefer just having the function inside the same file and passing `effects` & `state`?
```ts
const showPrivacyPolicyNotification = ({ effects, state }) => {
  if (effects.browser.storage.get(seenTermsKey) || state.isFirstVisit) {
    return;
  }
  
  // ...
}

// ...

showPrivacyPolicyNotification({ effects, state });
```

I prefer having an `effect`, since that will make typing a lot easier

CompuIves: Yes, agree an `effect` or `action` would be best indeed! Maybe we can put it here? `packages/app/src/app/overmind/internalActions.ts`, then it's an internal action on the root.

MichaelDeBoey:
Build fails due to #3908

SaraVieira:
Hey!

There has been a conflict and I don't know how to fix it :(

MichaelDeBoey:
@SaraVieira Rebased onto `upstream/master` 🙂

CompuIves:
Great!

PR #3939, opened 2020-04-18
chore: add defaultProps in custom ThemeProvider

jyash97:
default prop in ThemeProvider

We have to pass theme prop

## What is the new behavior?
Don't need to pass a theme prop every time

1. Tested GitHub page
2. Tested editor + dashboard

Haven't tested patron + pro pages ( as I have not subscribed yet 😕)

## Checklist

- [x] Testing 
- [ ] Ready to be merged


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3939/commits/c59f59aeea1cbb9d7fdbeed55a3eba1cb3e32373 is at https://pr3939.build.csb.dev/s/new.

SaraVieira:
Approved! I am just gonna ping @siddharthkp since he was the one building this provider

packages/components/src/components/ThemeProvider/index.tsx
```diff
@@ -80,3 +81,7 @@ export const ThemeProvider = ({ theme, children }) => {
     </>
   );
 };
+
+ThemeProvider.defaultProps = {
+  theme: codesandboxBlack,
```
MichaelDeBoey, line 86: What's the point of setting a `defaultProp` if there aren't any props passed? 🤔

SaraVieira: What do you mean? You pass the theme, we used to have to pass it always and now with this you only need to pass it when it changes and not in the static pages

MichaelDeBoey: @jyash97 You can just use it in the function arguments btw 🙂

```tsx
export const ThemeProvider = ({ children, theme = codesandboxBlack }) => {
  // ...
};
```

MichaelDeBoey: @SaraVieira But passing the `theme` prop is removed everywhere in this PR, no? 🤔

SaraVieira: Noo, just in those pages, we also pass it in places in the editor where it changes based on the user theme choice

MichaelDeBoey: My bad, I thought it was removed everywhere. 😓

Since we're using function components, we can just pass the default prop like we do in functions instead of defining `ThemeProvider.defaultProps`, so I would change that one tho

jyash97: Done @MichaelDeBoey  👋

SaraVieira:
Intesting, seems like this breaks ts :(

MichaelDeBoey:
Hmmm that's strange 🤔

jyash97:
Any idea on how to fix the TS error ?

PR #3942, opened 2020-04-19
fix: update view after deleting folder in dashboard

jyash97:
Bug Fix

After deleting the folder in the dashboard, the view is not updated.

Ref: https://www.loom.com/share/e4c2b3be5fa54e38ac7e3b17cc5210f1

## What is the new behavior?
Update view after deleting

Ref: https://www.loom.com/share/33fc94ee6bdf49bab1a733938f5e9f81

1. Locally tested by deleting folders

## Checklist

- [x] Testing 
- [x] Ready to be merged

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Sidebar/SandboxesItem/FolderEntry/index.tsx
+++ b/packages/app/src/app/pages/Dashboard/Sidebar/SandboxesItem/FolderEntry/index.tsx
@@ -203,6 +203,14 @@ class FolderEntry extends React.Component<Props, State> {
               });
             },
           });
+
+          const subPath = path
+            .split('/')
+            .slice(0, -1)
+            .join('/');
+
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3942/commits/297f1f8b7ea5d7508cd8935d00464393e409c16d is at https://pr3942.build.csb.dev/s/new.

MichaelDeBoey:
I'm not sure if the provided solution is the best one, since that' relying on the URL structure, rather then the folder structure.
I think getting the parent and going to its link is a more correct and more sustainable way tbh.

CompuIves:
Yep, though currently in terms of UI we also don't keep a parent (we just keep a string which is a path). We're revamping the dashboard more, so I think this is a good intermediate fix that we can put live!

PR #3949, opened 2020-04-20
fix: version copied from editor status bar

jyash97:
Enhancement

When we copy the text from the status bar present in Editor we also get a comma in between version and commitSha. Like this 
`Version: ,ac5b1e45c`

## What is the new behavior?
Avoid Comma

1. Locally copied version
2. Tested the react behavior here: https://codesandbox.io/s/goofy-solomon-nq9bm?file=/src/App.js

## Checklist

- [x] Testing 
- [x] Ready to be merged
      
      

`Question` Can anyone explain to me how this is happening. Is React tracking explicitly if a string is passed it keeps as a string ( this is expected ) but when we declare a variable it tracks as an array ( Not able to understand this :/ )

### DIFF
--- a/packages/app/src/app/pages/Sandbox/Editor/index.tsx
+++ b/packages/app/src/app/pages/Sandbox/Editor/index.tsx
@@ -213,7 +213,7 @@ const Editor = () => {
               })}
             >
               <FakeStatusBarText>
-                Version: {VERSION.split('-').pop()}
+                {`Version: ${VERSION.split('-').pop()}`}
               </FakeStatusBarText>
               <StatusBar
                 className="monaco-workbench mac nopanel"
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3949/commits/0667ac7724d56a37e1f467a9e07f23ee618de70d is at https://pr3949.build.csb.dev/s/new.

lbogdan:
TBH, I would remove `Version: ` entirely, and just copy the short SHA. @CompuIves what do you say?

CompuIves:
Agree!

lbogdan:
(thanks for fixing this, btw, it also annoyed me a bit! 🙂)

jyash97:
@lbogdan  Fixed :)

PR #3950, opened 2020-04-20
fix typos

SimonSiefke:
Succesfully -> Successfully


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3950/commits/85f10d268e6ac5c98bac1eebc68e3f13dfb116e4 is at https://pr3950.build.csb.dev/s/new.

SimonSiefke:
Maybe it would be a good idea to explicitly set `arrowParens: 'avoid'` because Prettier v2 adds arrowParens automatically

MichaelDeBoey:
@SimonSiefke We're using `prettier@1.19.2`, so there should be no problem.

PR #3955, opened 2020-04-21
Disable visual tests for the month

siddharthkp:
We crossed our open source limit of 35,000 snapshots per month

Short term fix: Commenting out the test from circleCI so that we can uncomment it back at the start of new billing cycle in 9 days

Suggestion for long term fix: We should only run visual tests after checking if there are changes to packages/components. Right now, we're running them for every change in the app which running out of our quota faster


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3955/commits/a82da5d09b04a9583bc963b6039940623c97db27 is at https://pr3955.build.csb.dev/s/new.

PR #3956, opened 2020-04-21
chore: Only run Chromatic when files are changed

MichaelDeBoey:
As suggested by @siddharthkp in #3955:
> Suggestion for long term fix: We should only run visual tests after checking if there are changes to `packages/components`. Right now, we're running them for every change in the app which running out of our quota faster

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -24,7 +24,7 @@
     "chromatic": "(if-env CIRCLE_BRANCH=master && yarn chromatic:master) || yarn chromatic:branch",
     "chromatic:base": "./node_modules/.bin/chromatic --build-script-name=build:storybook --exit-zero-on-changes --project-token=nffds42ndde",
     "chromatic:master": "yarn chromatic:base --auto-accept-changes",
-    "chromatic:branch": "(!(git diff master...HEAD --quiet ./src) && yarn chromatic:base) || echo 'Not running Chromatic...'",
+    "chromatic:branch": "(!(git diff master...HEAD --exit-code ./src) && echo 'RUN yarn chromatic:base') || echo 'Not running Chromatic...'",
     "prepublish": "yarn build"
   },
   "dependencies": {
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -24,7 +24,7 @@
     "chromatic": "(if-env CIRCLE_BRANCH=master && yarn chromatic:master) || yarn chromatic:branch",
     "chromatic:base": "./node_modules/.bin/chromatic --build-script-name=build:storybook --exit-zero-on-changes --project-token=nffds42ndde",
     "chromatic:master": "yarn chromatic:base --auto-accept-changes",
-    "chromatic:branch": "(!(git diff master...HEAD --exit-code ./src) && echo 'RUN yarn chromatic:base') || echo 'Not running Chromatic...'",
+    "chromatic:branch": "(!(git diff master...HEAD --quiet ./src) && yarn chromatic:base) || echo 'Not running Chromatic...'",
     "prepublish": "yarn build"
   },
   "dependencies": {
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -22,8 +22,9 @@
     "start:storybook": "start-storybook",
     "typecheck": "tsc --noEmit --pretty",
     "chromatic": "!(git diff --quiet ./src) && (if-env CIRCLE_BRANCH=master && yarn chromatic:master || yarn chromatic:branch)",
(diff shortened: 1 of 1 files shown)


CompuIves:
Wow, creative fix!

MichaelDeBoey:
Maybe I should also revert #3955 in this one? 🤔

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3956/commits/d82d64a781b63ea4bf25767ba8cfa88d80c11bc1 is at https://pr3956.build.csb.dev/s/new.

siddharthkp:
TIL!

> Maybe I should also revert #3955 in this one? 🤔

We still need that till the end of the month :( because we already exceeded the limit

packages/components/package.json
```diff
@@ -21,7 +21,7 @@
     "start": "yarn tsc --watch",
     "start:storybook": "start-storybook",
     "typecheck": "tsc --noEmit --pretty",
-    "chromatic": "if-env CIRCLE_BRANCH=master && yarn chromatic:master || yarn chromatic:branch",
+    "chromatic": "!(git diff --quiet ./src) && (if-env CIRCLE_BRANCH=master && yarn chromatic:master || yarn chromatic:branch)",
```
siddharthkp, line 24: we would probably always want to run master to keep the latest baseline. Should move this to chromatic:branch

siddharthkp:
Couple of questions here just to make sure:

1. Does running diff give you changes that are in commits as well? 🤔 

2. Would this command run with exit code 1? Because CI would assume that's a fail instead of skip

```yarn run v1.12.1
warning package.json: No license field
$ !(git diff --quiet ./src) && echo hi
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

MichaelDeBoey:
Couple of questions here just to make sure:

> 1. Does running diff give you changes that are in commits as well? 🤔 (curious how this works on CI environments)

Because I use `--quiet`, there won't be any messages shown.
https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---quiet

I could use `--exit-code` instead to have the same behavior, but show the diff.
I decided to not do that, since that _could_ log a lot of diff messages.


> 2. Would this command run with exit code 1? Because CI would assume that's a fail instead of skip
>     ```
>     yarn run v1.12.1
>     $ !(git diff --quiet ./src) && echo hi
>     error Command failed with exit code 1.
>     info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this 
>     command.
>     ```

It indeed exits with code of `1` if it fails.
If this will fail CI too, I'll look into something to fix this.

siddharthkp:
> Because I use --quiet, there won't be any messages shown.

I think I didn't ask the question properly :)

What I mean is does running `git diff` already give you changes on branch as compared to master? as compared to something like `git diff master...branch`

> If this will fail CI too, I'll look into something to fix this.

👍

MichaelDeBoey:
> What I mean is does running `git diff` already give you changes on branch as compared to master? as compared to something like `git diff master...branch`

Thanks for bringing that up, didn't think about that before! 👊
Fixed now 🙂

MichaelDeBoey:
I'm not 100% happy with how it turned out, but I think creating a separate script file is overkill so leaving it like this is the best thing to do imo.

MichaelDeBoey:
@CompuIves @siddharthkp I've reverted #3955 too, since we're in a new month now 🙂

MichaelDeBoey:
@CompuIves Although I've rebased onto master, I still get a git diff
https://app.circleci.com/pipelines/github/codesandbox/codesandbox-client/6643/workflows/292e3390-a870-47e2-94e1-37b822b6f8f1/jobs/61445

Could this be a caching problem? 🤔

MichaelDeBoey:
Seems like it was a caching problem 😓
https://app.circleci.com/pipelines/github/codesandbox/codesandbox-client/6645/workflows/11155a6b-c46a-4f8d-be29-0e324d1f15fd/jobs/61461

MichaelDeBoey:
@CompuIves @siddharthkp Rebased onto `upstream/master` so I think this one can be merged 🙂

siddharthkp:
It's almost there! There's a () mismatch which crashes the build

MichaelDeBoey:
@siddharthkp Should be fixed now 😅

MichaelDeBoey:
@siddharthkp All green now 😉

SaraVieira:
Danke!!

MichaelDeBoey:
Always a pleasure to help out 🙂

PR #3967, opened 2020-04-21
Create Skeleton for recent page

SaraVieira:
This also moves all the queries to the overmind to clean the view

Also adds the idea of a currentTeam on the state so we can toggle what is shown to the user at any time when they change teams

Completly screws over the drafts because I had to put react virtualized there and EVERYTHING is absolute when you use it

closes CSB-97


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/3967/commits/5c930145d8467016b5a0a45595037205afe95d6f is at https://pr3967.build.csb.dev/s/new.

PR #4018, opened 2020-04-29
Improve render performance

CompuIves:
Replaces `@styled-system/css` with a version that doesn't use Object.assign, since that's very expensive for us. It improves render time 3 times for rendering components that use `css`.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4018/commits/865c8c030c3a1e28bf0b470ad6d057a4cc655bd3 is at https://pr4018.build.csb.dev/s/new.

MichaelDeBoey:
Could this also be fixed upstream, so we can remove our own fork over time? 🤔

CompuIves:
Here: https://pr4018.build.csb.dev/s/github/excalidraw/excalidraw?file=/src/index.tsx

Try opening and closing the components folder. vs codesandbox.io

CompuIves:
Yes, I think it can be fixed upstream, but it's a bit tricky since we can do this since we don't rely on having `defaultTheme`. I'm not sure what breaks if it's fixed upstream, but I would recommend doing it, or at least memoizing the function.

siddharthkp:
I'd want to wait on this for another couple of days so that we can run chromatic on this.

CompuIves:
Sounds good!

CompuIves:
To give a better sense of the diff. I changed this in `@styled-system/css`:

```diff
-  var theme = _extends({}, defaultTheme, {}, props.theme || props);
+  var theme = props.theme || props;
```

After seeing that our `props` already has all values defined in `defaultTheme`, which are:

```
var defaultTheme = {
  space: [0, 4, 8, 16, 32, 64, 128, 256, 512],
  fontSizes: [12, 14, 16, 20, 24, 32, 48, 64, 72]
};
```

CompuIves:
Another thing we can look at is maybe reducing the size of our `theme` object. Not sure if we're using everything on it.

siddharthkp:
> To give a better sense of the diff. I changed this in @styled-system/css

Oh yeah that's cool. We don't use the alternative API, so this is should be safe

> Another thing we can look at is maybe reducing the size of our theme object. Not sure if we're using everything on it.

For sure, would that help with faster interpolations?

> Another thing is moving the get call to happen during compile-time

It's theme based, that's why we do it in runtime

CompuIves:
>  For sure, would that help with faster interpolations?

Not 100% sure, I think we'd need to test it a bit and see if there are any other calls where there is a loop through all keys.

> It's theme based, that's why we do it in runtime

Maybe we could do something like this: for every time we see eg. `"blues.500"` in a `css` object we replace it with `blues[500]`. So something like this:

```
css({
  backgroundColor: 'blues.500'
})
```

Turns into this with a babel plugin:

```
css({
  backgroundColor: props => props.theme.colors.blue[500]
})
```

siddharthkp:
> Maybe we could do something like this: for every time we see eg. "blues.500" in a css object we replace it with blues[500]. So something like this

Ah, got it. Yep, that could work.

siddharthkp:
Just checked, [our billing cycle](https://www.chromatic.com/billing?accountId=5d6e832ff79d9e00205168dc) is 3rd of everything month, not 1st 😅

CompuIves:
This one should be ready to review now! 😄

CompuIves:
Okay! Let me know once it's merged, I'm looking forward to profiling codesandbox.io 😄

siddharthkp:
The `Complete setup.` should start working once it's on master :)

MichaelDeBoey:
@CompuIves @siddharthkp Like @jxnblk said in https://github.com/styled-system/styled-system/issues/1306#issuecomment-627381911, we should have a look into [`theme-ui`](https://github.com/system-ui/theme-ui), which possibly has a fix for this one too.

If so, we can delete our own fork of `styled-system` again 🙂

siddharthkp:
This is the part that we changed and is shared in both styled-system/css and theme-ui/css

https://github.com/system-ui/theme-ui/blob/master/packages/css/src/index.ts#L258-L261

```js
  const theme: Theme = {
    ...defaultTheme,
    ...('theme' in props ? props.theme : props),
  }
```

PR #4029, opened 2020-04-30
Cleanup GitHub screen

MichaelDeBoey:
Things I did:
- Change `repoTitleChanged`'s signature to accept a `string` instead of `{ title: string }`, because it only has 1 argument
- Change `subjectChanged`'s signature to accept a `string` instead of `{ subject: string }`, because it only has 1 argument
- Change `descriptionChanged`'s signature to accept a `string` instead of `{ description: string }`, because it only has 1 argument
- Make `Changes`, `CommitForm`, `CreateRepo`, `GitHub`, `GithubLogin`, `NotLoggedIn` & `NotOwner` a `FunctionComponent`
- Type the icons

### DIFF
--- a/packages/app/src/app/overmind/namespaces/git/actions.ts
+++ b/packages/app/src/app/overmind/namespaces/git/actions.ts
@@ -4,9 +4,7 @@ import * as internalActions from './internalActions';
 
 export const internal = internalActions;
 
-export const repoTitleChanged: Action<{
-  title: string;
-}> = ({ state }, { title }) => {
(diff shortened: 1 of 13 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4029/commits/cb689a6cc0a2ea2428383c5fa9112c9b1e7d2956 is at https://pr4029.build.csb.dev/s/new.

SaraVieira:
Hey!

This screen is under refactor for the new Git Flow so it wouldn't be a good idea to merge right now

PR #4070, opened 2020-05-03
🔨 Cleanup Avatar component

MichaelDeBoey:
This PR adds `@codesandbox/common` as a dependency of `@codesandbox/components`, just because I need the `User` type. (see c11f8596)
I can imagine we will have some other components that will need other types too.

So what I'm suggesting is to extract all types from `@codesandbox/common/lib/types` into a separate `@codesandbox/types` package (see #4862).
Thoughts @SaraVieira @siddharthkp @CompuIves? 🤔


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4070/commits/e2ee220f1487cbc5292832e002229b5e1617ed3a is at https://pr4070.build.csb.dev/s/new.

siddharthkp:
> This PR adds @codesandbox/common as a dependency of @codesandbox/components, just because I need the User type. 

need to confirm with @SaraVieira on this one, from what I remember, this led to a circular dependency

MichaelDeBoey:
@siddharthkp It does, so that’s why I’m suggesting the extraction of types into a new `@codesandbox/types` package.

SaraVieira:
Has a bunch of conclicts if you pull from master

We also can't do the circular dependency :(

MichaelDeBoey:
I'll rebase 🙂

PR #4072, opened 2020-05-03
Extract @codesandbox/types package

MichaelDeBoey:
Like I suggested in #4070, this is the extraction of the `@codesandbox/types` package, so it can be used in all other packages without circular dependencies.

At this moment, I just copied over all files, but I didn't remove it from `@codesandbox/common` and I also didn't change all imports yet.
This because I came to the conclusion that `types` are dependent on types defined in `themes`.
```ts
import { TemplateType } from '../templates';
```
Not sure if we should move `TemplateType` to `@codesandbox/types` too tho.
We could re-export the type in `@codesandbox/common/lib/themes` to keep the same functionality _or_ we could just change all imports to the new `@codesandbox/types`.

Thoughts are welcome 🙂


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4072/commits/592da0e165d002d1ead8ddec4a6290c04c226f35 is at https://pr4072.build.csb.dev/s/new.

SaraVieira:
Hey!

As an effort to clean GitHub and only have mergable PR's open I am gonna close this if you want to keep working on it please open a PR when it's more done 

Thank you

MichaelDeBoey:
@SaraVieira I can't work on this until there's made a descision
> At this moment, I just copied over all files, but I didn't remove it from `@codesandbox/common` and I also didn't change all imports yet.
> This because I came to the conclusion that `types` are dependent on types defined in `themes`.
> ```ts
> import { TemplateType } from '../templates';
> ```
> Not sure if we should move `TemplateType` to `@codesandbox/types` too tho.
> We could re-export the type in `@codesandbox/common/lib/themes` to keep the same functionality _or_ we could just change all imports to the new `@codesandbox/types`.
> 
> Thoughts are welcome 🙂

christianalfoni:
Hi there!

If you can make this PR expose a `types` package where all the types of current `common/types` is moved into, where `common/types` exports all the types from the `types` package. That would be really great. Cause then we are not bound to refactoring anything yet. It is basically just preparing us for it.

And now it becomes even more explicit where the Codesandbox types should be. 

Hope that made sense, just ask if you have any questions 😄

MichaelDeBoey:
I'll re-expose `@codesandbox/types` in `@codesandbox/common/types` and resubmit the PR 🙂

PR #4076, opened 2020-05-04
Greatly improve file switching performance

CompuIves:
Switching files was noticably slow, with this change it's almost instant. The reason for this was a `React.useEffect` that would execute for every time we switch a file, and this would cause VSCode to do a layout update. With this change file switching is instant.

image

### DIFF
--- a/packages/app/src/app/components/CodeEditor/VSCode/index.tsx
+++ b/packages/app/src/app/components/CodeEditor/VSCode/index.tsx
@@ -17,9 +17,10 @@ export const VSCode: React.FunctionComponent = () => {
   const { state, actions, effects } = useOvermind();
   const containerEl = useRef(null);
 
-  const getCurrentModule = React.useCallback(() => state.editor.currentModule, [
-    state,
-  ]);
+  const getCurrentModule = React.useCallback(
+    () => state.editor.currentModule,
+    [] // eslint-disable-line
+  );
 
   useEffect(() => {
     const rootEl = containerEl.current;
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4076/commits/4f7a700dc51341668f86047b76b7907e75babcb7 is at https://pr4076.build.csb.dev/s/new.

MichaelDeBoey:
Really nice and noticeable improvement! 👊

But I think we can fix this without using `useCallback`, since it's mostly [not necessary](https://kentcdodds.com/blog/usememo-and-usecallback).
I'm not 100% sure, but since everything in `useOvermind` is a proxy, I think that performance will be better if we listen to the change of all states separately

```ts
const {
  actions,
  effects,
  state: { editor: { currentModule } },
} = useOvermind();
```

instead of listening to the whole state change:

```ts
const { state, actions, effects } = useOvermind();
```

Since listening to the whole state will give a new object every time something changes, so every value you get will be a new value, will will cause a recall of the `useEffect`.

Like I said: I'm not 100% sure, so your solution needs to be tested against what I'm suggesting to be 100% sure.

CompuIves:
I believe that Overmind only listens to the fields of the state objects (so only to eg. `state.editor.currentModule`, not to `state.editor` or `state`). Ideally, we'd like to update no component when `currentModule` changes, to prevent any rerender since that bogs down the interaction. That's why I like to not depend on any component on `currentModule`, only the ones that we really need.

@christianalfoni Does that make sense?

PR #4078, opened 2020-05-04
Remove unnecessary calculations for rendering files

CompuIves:
Spreads are expensive, and it seems like we were doing it for all our folders. This removes the spreads.

It seems like we weren't using siblings at all anymore, even not in spreads and checks.

I tested this by opening and closing files/directories.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4078/commits/ed2e9e6ed154ef54983cb504cfb893ff7321726f is at https://pr4078.build.csb.dev/s/new.

PR #4089, opened 2020-05-06
fix(create): react to keyboard events when search input is focused

### LINKED ISSUE #4073: cursor in input search box is not movable in explore template page
# 🐛 bug report

## Description of the problem
- Open `codesandbox` in browser
- click on create Sanbox
- search in templates / explore templates
- try to move the cursors, it in search box it's not moving, instead it moves / iterates over the templates item

Ideally it should move to template items when press tab or down arrow key at least once.

Nothing much, trying to fix typo error.

### Your Environment

| Сodesandbox      | web
| Browser          | Chrome Version 80.0.3987.149 (Official Build) (64-bit)
| Operating System | Mac Osx Mojave

arthurdenner:
Bug fix. Closes #4073.

Keyboard events when the search input is focused are "swallowed" by the `useKey` hooks.

## What is the new behavior?
Keyboard events fired from the search input are not handled by the `useKey` hooks.

1. Type something in the search input on the create modal;
2. Press the arrow keys and verify that the interaction is kept inside the input;
3. Press `Esc` to blur from the search input;
4. Press the arrow keys and verify that it navigates through the list.

## Checklist

- [x] Documentation
- [x] Testing 
- [x] Ready to be merged
      
### DIFF
--- a/packages/app/src/app/components/CreateNewSandbox/CreateSandbox/SearchBox/SearchBox.tsx
+++ b/packages/app/src/app/components/CreateNewSandbox/CreateSandbox/SearchBox/SearchBox.tsx
@@ -42,6 +42,7 @@ export const SearchBox = ({
     >
       <InputWrapper>
         <SearchElement
+          id="filter-templates"
           placeholder={placeholder}
           ref={inputEl}
           value={value}
--- a/packages/app/src/app/components/CreateNewSandbox/CreateSandbox/TemplateList/TemplateList.tsx
+++ b/packages/app/src/app/components/CreateNewSandbox/CreateSandbox/TemplateList/TemplateList.tsx
@@ -154,6 +154,9 @@ export const TemplateList = ({
   useKey(
     'ArrowRight',
     evt => {
+      if ((evt.target as HTMLInputElement).id === 'filter-templates') {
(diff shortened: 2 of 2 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4089/commits/7ccd586a7f684dcf26fa3be2c6e662664fcd587b is at https://pr4089.build.csb.dev/s/new.

CompuIves:
Ah, in my case I was intentionally adding that you can arrow key through the list while typing in the filter. But your case probably fixes the fact that that you can't change selection of characters, right?

arthurdenner:
@CompuIves, yes, exactly. The current implementation prevents the user from navigating in the input. It's possible to allow both interactions by removing the `preventDefault` but it feels weird.

I can add a video showing the old and current behavior if you prefer.

ankibalyan:
@CompuIves With this PR user will still be able to navigate through up/ down arrows while typing the filter, and once the user reaches to templates section. s/he/they can move around using any arrow key.

CompuIves:
Ah yep, that makes a lot of sense. Thanks for this contribution!

PR #4119, opened 2020-05-12
SSE Edge

lbogdan:
Before SSE Edge™ we had a single cluster, a single `sse-manager`, hence a single SSE websocket URL. With SSE Edge we have multiple of those, so for any one sandbox we first have to query the load-balancer for the cluster it's scheduled on, and from that derive the websocket URL.

This was partially solved in #4228, by adding the sandbox id to the websocket query, so that the CF proxy can read it and query the load-balancer. This PR fixes it completely, by moving that logic to the client, so that it can connect the websocket directly, without going through the CF proxy, which will only be used for previews.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4119/commits/5589e79b00681611e841619efa7d529a712179ff is at https://pr4119.build.csb.dev/s/new.

lbogdan:
This is ready to be merged, I tested it on staging, and the starter is already deployed in `sse-manager`.

packages/executors/src/serverExecutor.ts
```diff
@@ -71,20 +72,18 @@ export class ServerExecutor implements IExecutor {
...
-    const sseHost = usedHost.replace('https://', 'https://sse.');
+    const sseLbHost = usedHost.replace('https://', 'https://sse-lb.');
+    const res = await axios.get(`${sseLbHost}/api/cluster/${this.sandboxId}`);
+    const sseHost = res.data.hostname;
 
     this.socket = io(sseHost, {
       autoConnect: false,
       transports: ['websocket', 'polling'],
-      query: {
-        sandboxid: this.sandboxId,
```
CompuIves, line 86: Do we lose any data in the manager with this? Might there be a reason in the future that we want to infer the sandbox id from the WS connection? If that's the case we might want to keep it in the URL, but I'm not sure if you already have this data.

packages/executors/src/serverExecutor.ts
```diff
@@ -71,20 +72,18 @@ export class ServerExecutor implements IExecutor {
...
-  private initializeSocket() {
+  private async initializeSocket() {
     if (!this.sandboxId) {
       throw new Error('initializeSocket: sandboxId is not defined');
     }
-
     const usedHost = this.host || 'https://codesandbox.io';
-    const sseHost = usedHost.replace('https://', 'https://sse.');
+    const sseLbHost = usedHost.replace('https://', 'https://sse-lb.');
+    const res = await axios.get(`${sseLbHost}/api/cluster/${this.sandboxId}`);
```
CompuIves, line 81: Maybe we should put this in a function that retries up to 3 times, in case of flaky connections or when the LB has an error.

packages/executors/src/serverExecutor.ts
lbogdan: The client already sends the sandbox id to `sse-manager` in the websocket protocol.

packages/executors/src/serverExecutor.ts
lbogdan: That's a good point, do we use any retry library anywhere that we could reuse here? Also I think some error handling would be nice, in case all retries fail, but I'm not sure what the UI/UX pattern for this is.

CompuIves: We don't use anything generic right now, I have written a poor mans retry function. Would be good to introduce some common helper function in `@codesandbox/common` or a library I think.

Regarding the notice, we can send a notification from here. You can import
```
import {
  notificationState,
  NotificationStatus
} from '@codesandbox/common/lib/utils/notifications';
```

and do

```
notificationState.addNotification({ message: 'blah', status: NotificationStatus.Error })
```

Issue #2031, opened 2019-06-06
Add font: Victor Mono


rubjo:
# 🌈 Feature

Hi, and thanks for making an amazing code editor!

Over the last few months, I have designed a programming font available for free here: [rubjo.github.io/victor-mono](https://rubjo.github.io/victor-mono/). It's light and crisp, and features cursive italics and some ligatures. I think it looks particularly good in small-to-medium sizes on high-DPI displays.

If you want to include it CodeSandBox, go ahead. It will look like this:

![codesandbox-victormono](https://user-images.githubusercontent.com/42270947/60435579-3ee03d80-9c0a-11e9-9ce1-d238d174738f.png)

![specimen-dark](https://github.com/rubjo/victor-mono/raw/master/src/assets/img/specimen-dark.png)

PR #4131, opened 2020-05-14
[RFC] Change Font from Dank Mono to Mono Lisa

SaraVieira:
Font Link: https://monolisa.dev/

Why?

Dank Mono is great but it doesn't support Cyrillic languages or Greek languages giving an inconsistent experience for people and also Ukraine is huge, damn so many people live there, its also not getting updated :/

Mono Lisa was made by developers I know and it's maintained and being updated since it's a product with the intent to be sold.

Screenshots attached:

DM with no fallback on the right and MonoLisa on the left:
CleanShot 2020-05-14 at 19 06 33

DM with fallback to Menlo  on the right and MonoLisa on the left:
CleanShot 2020-05-14 at 19 07 20


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4131/commits/e780fdecfb59789dc0f8245b4204b8a8737e12d9 is at https://pr4131.build.csb.dev/s/new.

MichaelDeBoey:
@SaraVieira What about Jetbrains Mono? 🤔
https://www.jetbrains.com/lp/mono

SaraVieira:
Mostly because this one was made by developers and I am sure it will be maintained.

Also, like the whole thing of helping the little shops instead of the big ones

The last point is that you can always change to Jetbrains Mono if you have it installed but not the other way arround.

MichaelDeBoey:
I don't really have an opinion on the default one, just wanted to consider Jetbrains Mono too, since it's open source and is now widely used by all WebStorm/PhpStorm/... users by default so has a large userbase

rubjo:
I guess going for this font closes #2031.

Are there any plans for making it possible to set and on-demand load a preferred code font from a selection of alternatives, rather than always bundling all possible font files? (From the ~500KB of fonts loaded on page load I see Dank Mono, Source Code Pro, maybe others meant as code font options as well?)

okonet:
I wonder if I could somehow help with this one.

siddharthkp:
@okonet Nah, we got busy with some other things. Sara's picking this up again 😇

okonet:
@SaraVieira make sure to update the font to the latest v1.600. It has tons of improvements and fixes.

SaraVieira:
Hey

I copied the last version from the vendor link today :)

rubjo:
MonoLisa looks like a very safe and traditional font, but I am still a bit surprised by the decision to swap out one commercial font for another commercial one. And even more surprised by the adding of >1MB of font CSS (!) to the mountain of data being transferred on page load. It looks like this brings the list of fonts loaded (~3 MB) to this:
- Inter (Regular, Medium, SemiBold and Bold) (Regular loaded twice?)
- MonoLisa (seven weights * two styles) (Is it added twice, BTW, both in monolisa.css and fonts.css?)
- Source Code Pro (500)
- Roboto (300,400,500,700)
- Poppins (300-700)

I love CodeSandbox for its ease of use and quick-to-setup dev envs. I wouldn't say that the app feels outright sluggish, but it's been getting closer over time – to the point where it is tempting to use faster and slimmer alternatives.

siddharthkp:
@rubjo good looking out! 

Separate pull request - I'd say we can swap out the other 2 code fonts for MonoLisa as well reducing the bandwidth used by fonts

SaraVieira:
Also good check on the weights, we should only load regular and italics so I will fix that tomorrow aswell

packages/app/src/app/index.html
```diff
@@ -43,6 +43,7 @@
   <link crossOrigin="anonymous" rel="preload" href="/static/fonts/inter/Inter-Bold.woff2" as="font">
   </link>
   <link href="/static/fonts/inter/inter.css" rel="stylesheet">
+  <link href="https://www.monolisa.dev/static/fonts.css" rel="stylesheet">
```
SaraVieira, line 46: Mistake, remove tomorrow sara

okonet: We have a smaller CSS file. @bebraw what's the name?

bebraw:
https://www.monolisa.dev/static/initial-fonts.css is the smaller one with only basic weights.

CompuIves (approved):
Looks clean! Only thing to consider is that people who want Dank Mono as font can't configure it anymore I think? Because we switch then, anyway, might not be a lot of people.

SaraVieira:
I meaan maybe but if you downloaded dank mono it's usually not called DM on your machine ahah

okonet:
Huraa! Thanks @SaraVieira to pulling this off!!!

One thing I noticed is that the home page still uses Dank Mono. Maybe it's worth updating?

CleanShot 2020-12-11 at 10 35 02@2x

SaraVieira:
Oh, those are images, let me ask @DannyRuchtie if he thinks we should update them

DannyRuchtie:
ah, unfortunately this is a still from the video. 
Updating is not as straightforward.

PR #4150, opened 2020-05-18
Fix Skeleton Flicker when view mode is List

SaraVieira:
I was setting the view mode after the app loaded and now I set it on the init of the application and no more flicker


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4150/commits/3ec05899afc0d700a432ddb36e8cc930fedeb135 is at https://pr4150.build.csb.dev/s/new.

PR #4163, opened 2020-05-20
updated TS, deps and overmind

christianalfoni:
This PR updates to latest TS, which is WAY faster. It needed some deps updates and also Overmind got a circular dependency issue which is solved with the new `derived` API introduced in previous version.

Still think there are some issues here, probably related to styled-components.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4163/commits/86173d804dd83cb28033a9e70af182d5f28cdd15 failed.

packages/app/package.json
```diff
@@ -276,7 +276,7 @@
     "@types/semver": "6.2.0",
     "@types/socket.io-client": "^1.4.32",
     "@types/stripe-v3": "^3.1.7",
-    "@types/styled-components": "^4.4.2",
+    "@types/styled-components": "^5.1.0",
```
MichaelDeBoey, line 279: You only updated `styled-component` types, so these can be reverted I think? 🤔

```suggestion
    "@types/styled-components": "^4.4.2",
```

packages/common/package.json
```diff
@@ -98,7 +98,7 @@
     "@types/react": "^16.9.17",
     "@types/react-icons": "^2.2.7",
     "@types/storybook__addon-actions": "^3.4.3",
-    "@types/styled-components": "^4.1.13",
+    "@types/styled-components": "^5.1.0",
```
MichaelDeBoey, line 101:
```suggestion
    "@types/styled-components": "^4.4.2",
```

packages/app/package.json
christianalfoni: The 4.4.2 version is not compatible with TS 3.9 :-( But we updated to styled-components 5.1.0 instead 😄

packages/common/package.json
christianalfoni: Same as above 😄

PR #4168, opened 2020-05-20
Support private binary git files

CompuIves:
We have to give a jwt to BrowserFS, not ideal. Ideally we'd wait for the new auth to be live before merging, alternatively we can put this live first and undo it once it's live.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4168/commits/5d51d836d25b21153f85a41143483a56ebbf3689 is at https://pr4168.build.csb.dev/s/new.

Issue #4175, opened 2020-05-21
Editor crashes for everyone when someone leaves live session after they sent a chat


CompuIves:
# 🐛 bug report

## Description of the problem

When someone in a live session has a chat message in the live session, and then leaves, everyone will crash. That's because we refer to `roomInfo.users` for getting the Avatar, but in this case the user is not here.

This is the code that causes it to crash:
https://github.com/codesandbox/codesandbox-client/blob/master/packages/app/src/app/pages/Sandbox/Editor/Workspace/Chat/index.tsx#L115-L119

We should assume that `messageData(message);` can return `null`, so I would add a nullcheck on `Avatar` to fix it.

PR #4176, opened 2020-05-21
Add null check on avatar on chat page (#4175)

sunnyssr:

This PR introduces a bug fix for chat which results in a crash

Currently, when one user who has sent a message on chats quit the chat, it results in  crash for all other users who are still active.
https://github.com/codesandbox/codesandbox-client/issues/4175

## What is the new behavior?

This is not a feature change. In new behaviour if users quit the app after sending a message in chat, it should continue working for other users.

1. Step A
2. Step B
3. Step C
Yes, I tested it by doing the chat. I used three devices(different browsers) to join the chat. Sent a message by one and quit.
## Checklist

- [ ] Documentation
- [x] Testing 
- [x] Ready to be merged
      

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4176/commits/45784fa26e85cf31330a028f59c7d5f541d19334 is at https://pr4176.build.csb.dev/s/new.

CompuIves:
Thanks a lot for the timely fix!! I've just taken a look, and it works well! It just looks a bit quirky: 

image

Maybe we should hide the Avatar altogether to make it look well? And I think we should also add ` (left)` to the name here: https://github.com/codesandbox/codesandbox-client/pull/4176/files#diff-7aad850f6f72d32dc15f29bcbfc7f34cR121. So we could do something like
`{name}{metadata ? '' : ' (left)'`.

sunnyssr:
I think we should instead provide a fallback image which shows blank user image with a text saying left?

CompuIves:
That's even better!! Together with the left text, I think that would be the best solution. @DannyRuchtie do you have an image for something like this?

sunnyssr:
@CompuIves 
If we don't have image of the user, can we show CSB logo there?

CompuIves:
I think it would be best if we show a user icon here if possible. @DannyRuchtie do you have an SVG to show here?

DannyRuchtie:
User icon when there is no avatar and 20% opacity when they not there? 
The icon in the zip 

[chat.zip](https://github.com/codesandbox/codesandbox-client/files/4710269/chat.zip)

Screenshot 2020-06-01 at 10 30 48

CompuIves:
Yeah that sounds great. We don't have the `avatarUrl` if the user isn't there right now, but I can add it in the future. For now I think we should show the user icon with 20% opacity.

SaraVieira:
Design added 
CleanShot 2020-06-09 at 18 26 55


@CompuIves

packages/app/src/app/pages/Sandbox/Editor/Workspace/Chat/index.tsx
```diff
@@ -111,22 +126,30 @@ export const Chat: React.FC = () => {
...
-                      <Avatar
-                        color={color}
-                        alt={metadata.username}
-                        src={metadata.avatarUrl}
-                      />
+                      {metadata ? (
+                        <Avatar
+                          color={color}
+                          alt={metadata && metadata.username}
+                          src={metadata && metadata.avatarUrl}
```
CompuIves, line 137:
```suggestion
                          alt={metadata.username}
                          src={metadata.avatarUrl}
```

SaraVieira:
Thank you so much @sunnyssr !

Issue #3493, opened 2020-02-10
Hiding console by default can lead to missed warnings


eps1lon:
# 🌈 Feature

In React (and 3rd party libraries) warnings are oftentimes logged to the console that show actual issues with your code (without crashing the app).

When landing on codesandbox the console is collapsed by default which makes it easy to miss these issues. It would be nice if codesandbox would be opinionated about the default state of console. Either show it by default for React sandboxes or open it as soon as warnings/errors are logged and the user didn't explicitly collapse it.

flickz:
@garethx Can you assign me to this issue?

flickz:
Looking at this issue again, does it not matter we have badge indicating number of unread errors and warning? 

<img width="488" alt="Screenshot 2020-05-19 at 12 04 24 AM" src="https://user-images.githubusercontent.com/15707013/82267679-f71eed80-9964-11ea-83e8-71c0f25661e2.png">

eps1lon:
It does matter but it's easier to miss. IMO if you open a codesandbox or cause a warning/error you'd always want to see the error.

CompuIves:
We have an option for the devtools to be open by default (like we do with the container sandboxes with the terminal), but it does take up quite some screen space. Maybe there's a middle way? Maybe we can open the console once there are warnings, or we should maybe show a better indication?

eps1lon:
> Maybe we can open the console once there are warnings, or we should maybe show a better indication?

That's what I suggested and prefer. IMO I don't see a reason to have the console collapsed if there are errors. Maybe other ecosystems treat them differently but React puts a lot of effort into only issuing them if there is an actual problem that can be fixed.

> We have an option for the devtools to be open by default (like we do with the container sandboxes with the terminal)

My concern here are non-regular or first-time users. An option that requires knowledge and multiple steps to setup does not improve the situation for this group. Even I didn't know this option existed and it's still unclear to me where this option is and whether it is a user or sandbox option.

CompuIves:
> My concern here are non-regular or first-time users. An option that requires knowledge and multiple steps to setup does not improve the situation for this group. Even I didn't know this option existed and it's still unclear to me where this option is and whether it is a user or sandbox option.

Ah yes that makes sense. I was thinking of enabling this option by default, right now it's not exposed.

[...]

PR #4198, opened 2020-05-25
feat(devTools): open dev tools panel on error

flickz:
Ref https://github.com/codesandbox/codesandbox-client/issues/3493

This PR opens the DevTools panel on error, this is to prevent the users from missing these errors. 

Todo: 

- [ ] Close the devtools when the errors are fixed.

### DIFF
--- a/packages/app/src/app/components/Preview/DevTools/Console/index.tsx
+++ b/packages/app/src/app/components/Preview/DevTools/Console/index.tsx
@@ -158,6 +158,10 @@ class ConsoleComponent extends React.Component<StyledProps> {
         },
       })
     );
+
+    if (method === 'error') {
+      this.props.openDevTools();
+    }
   }
 
   list;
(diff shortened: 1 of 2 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4198/commits/a0946d9a22df3abfa902f61fd002a018dd64ab71 is at https://pr4198.build.csb.dev/s/new.

CompuIves:
Oh wow, thanks for another contribution! I don't think that this is the best way to show the errors, because things suddenly happen for the user without any interactions. I think that could be confusing, so I think we should find another way that we can show that there are errors without a sudden movement.

flickz:
@CompuIves I agree with you, definitely need design team input on that. Let me know their input on this or anyone else's suggestions that might work better, I'm ready to implement as soon as there's a clear way. In the meantime, I'll close this PR and continue the conversation in the issue.

Btw I'm learning a lot from contributing but there are lots of issues, difficult to pick which might be of value to the team currently, if you have time and don't mind please assign me 1 or 2 issues you think are important.

PR #4224, opened 2020-05-27
Redesign of Docs

SaraVieira:
This part starts with the redesign of the doc pages themselves so then we can move to the next ones


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4224/commits/d0e26abfd6326efe0b82f71ee09e4b467be7fb7a is at https://pr4224.build.csb.dev/s/new.

DannyRuchtie:
At @SaraVieira Sara not sure this is because of "staging" but there appears to be no signing state for the navbar.
Is this just staging or do we need to add more things?

Screenshot 2020-06-24 at 14 48 49

Issue #3507, opened 2020-02-13
ModuleNotFoundError Could not find module in path


moroshko:
Locally, running with a fresh new Create React App everything works as expected.

In CodeSandbox, however, I get:

> ModuleNotFoundError
Could not find module in path: 'basis/dist/icons/arrow-back' relative to '/node_modules/basis/dist/components/Icon.js'

https://codesandbox.io/s/compassionate-surf-orpco

Any ideas?

(btw, is there a way in CodeSandbox to browse `node_modules`?)

CompuIves:
Ah, yes, a tricky one. This line:

`  var IconComponent = require("../icons/".concat(name))["default"];`

Confuses our dependency resolver, since it's a dynamic require. I think we did support glob requires, but something might've gone wrong. Checking now...

> (btw, is there a way in CodeSandbox to browse node_modules?)

Yes, in the devtools you can find all executed files. Another way is to check `manager.transpiledModules` on the sandbox window!

moroshko:
Thanks @CompuIves for looking into this!

Is there is any other way to write the dynamic require that CodeSandbox can understand (as a temporary workaround until this gets fixed)?

Issue #3985, opened 2020-04-24
React project that works locally does not work in codesandbox


kentcdodds:
# 🐛 bug report

## Description of the problem

I have [this repo](https://github.com/kentcdodds/react-fundamentals) that works locally with regular react-scripts, but does not work in codesandbox.

The specific error is: `Loader 'mdx-loader' could not be found.`

The way it works is I have a package `@kentcdodds/react-workshop-app` which reads the filesystem to know which files are available and generates code for the entry file. Markdown files are loaded via mdx-loader using inline loader syntax.

I've attempted adding `mdx-loader` as a dep explicitly and that didn't fix the issue. Same error either way.

## How has this issue affected you? What are you trying to accomplish?

I have 8 of these repositories for teaching people React and I'd love to be able to let people work through this stuff directly in codesandbox without having to pull things down locally.

### Your Environment

| Software         | Name/Version |
| ---------------- | ------------ |
| Сodesandbox      | Codesandbox.io
| Browser          | Chrome 81
| Operating System | macOS (latest)

Thanks!

CompuIves:
Hmm, this is an interesting one. We don't have `mdx-loader` in our bundler, and we can't really use the webpack loader since we're not running Webpack itself.

I think the best solution would be to add `mdx-loader`, but it's quite an involved loader (https://unpkg.com/browse/mdx-loader@3.0.2/index.js), using 12 dependencies by itself. 

Ideally we'd support loading webpack loaders from dependencies. I think that would work with 90% of the loaders if we write a converter that can make webpack loaders behave like sandpack loaders. In the meantime, we should probably make a custom `mdx-loader` for sandpack to make this case work. I can work on that after React Europe/.

kentcdodds:
Ah! That's new information to me. I didn't realize that webpack loaders didn't work out of the box.

Getting this update would be spectacular!

kentcdodds:
What can I do to get this prioritized?

kentcdodds:
Awesome to see some progress on this issue @CompuIves! Thanks.

Unfortunately, the linked project still does not work on codesandbox 😬

CompuIves:
Yep, that's true! There are some other things I need to resolve as well, which are in PR #4191. Aiming to merge that today.

CompuIves:
It works now if you add `mdx-loader` to the dependencies!

kentcdodds:
Sweet! Thanks a bunch!

[...]

PR #4269, opened 2020-06-01
Dynamically transpile dynamic imports

CompuIves:
Fixes https://github.com/elastic/eui/issues/3157

When seeing an 

```import('test')```

we eagerly transpile this file. From now on we will only transpile it when it's called (so not at compile time).

Might help with #3985


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4269/commits/52b90222df39e2062bde87991bc1d4f6271a8786 is at https://pr4269.build.csb.dev/s/new.

miukimiu:
I think it might also fix: https://github.com/codesandbox/codesandbox-client/issues/3507

CompuIves:
Hmm, just tested that one but it still doesn't work. I think it's because they use `require` instead of `import`, so we can only resolve the file synchronously in that case (which doesn't always work because we need to download).

PR #4289, opened 2020-06-02
make deployments be whatever the person sends

SaraVieira:
Issue: Now we set the public flag in vercel deployments always to true because back in the day that was a requirement for free accounts and the default

that is long gone

Fixes email support issue


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4289/commits/72d7eaf275406d8e62ef1bcca54b415be945951d is at https://pr4289.build.csb.dev/s/new.

PR #4298, opened 2020-06-03
Fix how we transpile esmodule default

CompuIves:
We used to do

```
const a = interopRequireDefault(require('a')).default

a()
```

Now it will be

```
const $csb_a = require('a');
const a = interopRequireDefault($csb_a);

a.default()
```

In case `default` has been changed (often in case of cyclic dependencies).


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4298/commits/cc15b0aecd3e7b54f1fcb79e9c0bda17c3dd96bd is at https://pr4298.build.csb.dev/s/new.

PR #4308, opened 2020-06-03
Use new, faster oniguruma wasm library

CompuIves:
Thanks to the work from https://github.com/microsoft/vscode/pull/95958, we can use a new oniguruma library that results in 2-3x faster syntax highlighting.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4308/commits/bbbb2de74e7bf14b55bc03fefaea9164b121d894 failed.

PR #4320, opened 2020-06-07
new overmind version

christianalfoni:
In this version of Overmind there is no more automatic transform functions into derived on the state tree. You have to explicitly use `derived`, which we have already refactored. Though I found one spot where we overlooked it, on `modals`.

Should quickly move through the app to check if something else pops up, but very unlikely as this only happens related to `state.ts` files and if overlooked they become functions instead of the value they represent.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4320/commits/fb45088fe7c9b8ccde72ecf01e1789be70f71eaf is at https://pr4320.build.csb.dev/s/new.

PR #4329, opened 2020-06-08
Add Sandbox Skeleton as Background of template universe

SaraVieira:
closes CSB-155


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4329/commits/6b096fd20c5a746c9d63b7b5e6bbe7205cd2ee67 is at https://pr4329.build.csb.dev/s/new.

CompuIves:
Ohh this already feels much better! I was wondering if there's a way to make the editor itself load, like we load the page `Sandbox` without passing a sandbox, but passing a prop like `showTemplateModal`. And then it opens the modal `NewSandbox` if that prop is given. The big advantage of that is that we even load VSCode on the page in the background already, so that when someone chooses a sandbox, they will only have to download the sandbox and everything should work.

packages/app/src/app/pages/NewSandbox/index.tsx
```diff
@@ -7,8 +7,11 @@ import {
   COLUMN_MEDIA_THRESHOLD,
```
CompuIves, line 7: Do we still need this component? We can now use the editor completely I think?

packages/app/src/app/pages/Sandbox/Editor/index.tsx
```diff
@@ -177,6 +178,46 @@ const Editor = () => {
...
+                            css={css({
+                              backgroundColor: 'sideBar.background',
+                              maxWidth: '100%',
+                              width: 1200,
+                              position: 'relative',
+                              zIndex: 100,
+                            })}
+                            marginTop={8}
+                          >
+                            <CreateSandbox />
```
CompuIves, line 214: One feature request from me: I'd love to start using this in more places, but nowadays I often need to specify where the sandbox will be created in the dashboard by passing the `collectionId` prop. Would it be possible to allow setting this in the URL (something like `/s/?collectionId=blaaah`) and that it can be passed as a prop in here?

PR #4333, opened 2020-06-09
Allow markdown in description, external links and fix linebreaks

christianalfoni:
So this started with me wanting to implement markdown support for the description, but ended up in:

1. Well... markdown support for description
2. I do not see why we prevent external links in markdown? In comments and description you really want to be able to link external docs etc? And we did not give any indication that we prevent it, so it was very confusing
3. Now your linebreaks are properly represented in the output 

😄


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4333/commits/043d5891018e9fa8bb8da591330b7c37d7d2ac8c is at https://pr4333.build.csb.dev/s/new.

packages/app/src/app/components/Markdown/Link.tsx
```diff
@@ -28,13 +28,9 @@ export const LinkElement = ({ href, children, ...props }) => {
...
-  if (!href.includes('codesandbox')) {
-    return (
-      <Link target="_blank" rel="noopener noreferrer ugc" {...props}>
-        {children}
-      </Link>
-    );
-  }
-
-  return <Link {...props}>{children}</Link>;
+  return (
```
SaraVieira, line 31: This wasn't to allow links that are outside links but just to add the rel="noopener noreferrer ugc"  to only links for outside codesanbdox

packages/app/src/app/components/Markdown/index.tsx
```diff
@@ -35,7 +36,7 @@ export const Markdown = ({ source }) => (
     })}
   >
     <ReactMarkdown
-      source={source}
+      source={source.replace(/\n/gi, '&nbsp;\n\n')}
```
SaraVieira, line 39: wat

SaraVieira: OHH, looks good :D

PR #4341, opened 2020-06-09
Add a collection id to creating a sandbox

CompuIves:
When you create a sandbox from a folder, it should appear in that folder after creation.

This did work before, but regressed with the new create modal. I want to bring it back in, but am not sure how to pass props into the modal.

### DIFF
--- a/packages/app/src/app/components/CreateNewSandbox/index.tsx
+++ b/packages/app/src/app/components/CreateNewSandbox/index.tsx
@@ -41,7 +41,7 @@ export const CreateNewSandboxButton: FunctionComponent<Props> = ({
   };
 
   const handleClick = () => {
-    modalOpened({ modal: 'newSandbox' });
+    modalOpened({ modal: 'newSandbox', props: { collectionId } });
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4341/commits/3ab7a80ed04d96645f1d0e53c52de5359cfd9da0 is at https://pr4341.build.csb.dev/s/new.

CompuIves:
Oh! I found the new modal logic. I'll port it over to there then!

christianalfoni:
I see there are several places we use `newSandbox` modal, even in old `getSignals` code. But yeah, it would be better to use the new approach.

Go to `overmind/modals.ts` and add:

```ts
export const newSandbox = {
  state: {
    collectionId: -1
  }
}
```

This will type the props you can now pass to:

```ts
actions.modals.newSandbox.open({
  collectionId: 0
})
```

And access through:

```ts
state.modals.newSandbox.collectionId
```

PR #4359, opened 2020-06-12
Fix how we call imported esmodules

### LINKED ISSUE #4355: Illegal invocation error - ESM bundle without export default/module.exports interop
I'm getting an `illegal invocation` error with some library in csb.
Try clicking a button in this sandbox for example useing the `wana` package https://codesandbox.io/s/wana-global-undoredo-2eyk0?file=/src/App.js

I'm told it's because csb is compiling the ESM bundle without `export default`/`module.exports` interoperability. I'm not having issues using the library in other environments.

CompuIves:
Fixes #4355

```js
import a from './b';

a()
```

will become

```js
const a = interopRequireDefault(require('b'));

(0, a.default)()
```

The tricky thing is that we also do it for this case, will it break things?

```js
import a from './b';

const c = a;
```

```js
const a = interopRequireDefault(require('b'));

const c = (0, a.default)
```


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4359/commits/48403399a3c9c4d2d5de601d118d5c4a966ba845 is at https://pr4359.build.csb.dev/s/new.

PR #4362, opened 2020-06-12
Increase test threshold

siddharthkp:
The test isn't consistent in the time it takes. So, sometimes it fails in unrelated PRs

Increasing the threshold from 5ms to 10ms so that we can still catch regressions but it has more room in CI


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4362/commits/7ff1b7502fdb664ef54fe2b6cb14b8fae3ad6993 is at https://pr4362.build.csb.dev/s/new.

PR #4380, opened 2020-06-15
Fix live timeouts

CompuIves:
If there was a timeout (10s) of a message the promise would never resolve or reject. This makes sure that it does that. I also raised the timeout of sending an operation to 45s.

Finally I made the reconnect logic for the socket more robust, will now only stop trying if we're getting 401 (unauthorized) responses.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4380/commits/74599a383acc9817f122266879283e7c3ad40277 is at https://pr4380.build.csb.dev/s/new.

PR #4384, opened 2020-06-16
Send sentry error code with a crash report

CompuIves:


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4384/commits/3467469469f881a280a7bcd895505694f49e4748 is at https://pr4384.build.csb.dev/s/new.

lbogdan:
It would be nice to have a simple, reproducible way to generate a crash (if we don't already) so that we can test this?

CompuIves:
Right now I can go to codesandbox.io/new-dashboard to get a crash, but Sid is working on a fix.

CompuIves:
We can also remove this one: https://github.com/codesandbox/codesandbox-client/blob/sentry-report/packages/app/src/app/pages/index.tsx#L146-L148.

But I did test it on dev by removing the dev guards on Sentry first.

lbogdan:
Maybe have a route like `https://codesandbox.io/crash` that always crashes, à la `chrome://crash`?

CompuIves:
Yeah we have codesadbox, so we could remove the dev guard there.

PR #4393, opened 2020-06-16
do full reconnect on 30 minute timeout

christianalfoni:


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4393/commits/5bbdadac1adfdc6fa14102c073d85cea76694c86 is at https://pr4393.build.csb.dev/s/new.

CompuIves:
Hmm, I think that Phoenix handles this for us as well, but we need to listen to the reconnect event maybe? In this case I'm afraid that it won't work in other cases, even though Phoenix calls a callback where we can know when we reconnect.

SaraVieira:
Spoke to christian, closing as it was implemented in elixir

PR #4398, opened 2020-06-16
add back 'bluetooth' to iframe allowed features

FWeinb:
Allow to use the [Web Bluetooth](https://webbluetoothcg.github.io/web-bluetooth/)

As per this [commit](https://chromium-review.googlesource.com/c/chromium/src/+/657572) on chromium it is mandatory to add `bluetooth` to the allowed list for cross origin iframes to use bluetooth. 

Currently the Web Bluetooth API can't be used anymore it isn't in the allowed list of on the iframe. 

## What is the new behavior?
The Web Bluetooth API would work inside a sandbox.

### DIFF
--- a/packages/common/src/components/Preview/index.tsx
+++ b/packages/common/src/components/Preview/index.tsx
@@ -608,7 +608,7 @@ class BasePreview extends React.Component<Props, State> {
           {(style: { opacity: number }) => (
             <>
               <StyledFrame
-                allow="accelerometer; ambient-light-sensor; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
+                allow="accelerometer; ambient-light-sensor; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking; bluetooth"
                 sandbox="allow-autoplay allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
                 src={this.state.url}
                 ref={this.setIframeElement}


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4398/commits/86509d036deaf58fe2794f728755b6d2a85d64b0 is at https://pr4398.build.csb.dev/s/new.

FWeinb:
This seams to be not working as expected. Adding this to the iframe has no effect in `Chrome v83.0.4103.106` anymore. There might have been some changes regarding allowing bluetooth in cross-origin iframes.

SaraVieira:
Hey @FWeinb We also use it here:

https://github.com/codesandbox/codesandbox-client/blob/master/packages/app/src/app/pages/common/Modals/ShareModal/getCode.ts#L75
https://github.com/codesandbox/codesandbox-client/blob/master/packages/homepage/src/screens/home/explore/index.js#L99
https://github.com/codesandbox/codesandbox-client/blob/master/standalone-packages/react-sandpack/src/components/SandpackProvider/SandpackProvider.tsx#L276

I think the issue may be that it's not on the first link I sent and that is where we build the embebeds

FWeinb:
@SaraVieira Thanks for the other files this was used. I did confirm that `bluetooth` was added to the iframe's allow attribute in the sandbox. 

[Here](https://pr4398.build.csb.dev/s/sleepy-cohen-vq2hl) is a minimal test case that confirms that it still ends up with a `requestDevice() called from cross-origin iframe.` on `Chrome 83.0.4103.106` even with `bluetooth` added to the iframe.

SaraVieira:
Hey!

Yeah this seems like something that chrome no longer supports, I will try to contact someone at google and figure this out :(

PR #4399, opened 2020-06-17
only download definitelytype if no types included

christianalfoni:
This fixes issue where we grab `@types/` even though the package includes types. Tested with `react` and `react-textarea-autosize`... React grabs it from @types and even though react-textarea-autosize has a @types package it uses its internal typing, cause it is there

### DIFF
--- a/packages/app/src/app/overmind/effects/vscode/SandboxFsSync/index.ts
+++ b/packages/app/src/app/overmind/effects/vscode/SandboxFsSync/index.ts
@@ -11,8 +11,8 @@ import {
   Sandbox,
   SandboxFs,
 } from '@codesandbox/common/lib/types';
-import { getGlobal } from '@codesandbox/common/lib/utils/global';
 import delay from '@codesandbox/common/lib/utils/delay';
+import { getGlobal } from '@codesandbox/common/lib/utils/global';
 import { protocolAndHost } from '@codesandbox/common/lib/utils/url-generator';
 import { getSavedCode } from 'app/overmind/utils/sandbox';
 import { json } from 'overmind';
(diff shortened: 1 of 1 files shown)


packages/app/src/app/overmind/effects/vscode/SandboxFsSync/index.ts
```diff
@@ -404,30 +404,33 @@ class SandboxFsSync {
...
-
       try {
         const files = await this.fetchDependencyTypingFiles(
           dep.name,
           dep.version
         );
-
-        this.setAndSendPackageTypes(dep.name, files);
+        const hasTypes =
+          Object.keys(files).filter(key => key.endsWith('.d.ts')).length > 0;
```
CompuIves, line 413: One check, maybe we need to check for the dependency path when checking (so something like `key.startsWith(dep.name) && key.endsWith('.d.ts')` because we tend to also include `.d.ts` files of transient dependencies sometimes, even if the dependency itself doesn't have the types.

We could also use `.find` or `.some` here, then it would stop traversing the files as soon as it found one match!

christianalfoni: OH, yes! Very nice feedback, updated 😄

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4399/commits/4b481d795078f1d940509ba3095bb9defda1193e is at https://pr4399.build.csb.dev/s/new.

packages/app/src/app/overmind/effects/vscode/SandboxFsSync/index.ts
```diff
@@ -404,30 +404,36 @@ class SandboxFsSync {
...
-      }
-
       try {
         const files = await this.fetchDependencyTypingFiles(
           dep.name,
           dep.version
         );
+        const hasTypes = Boolean(
+          Object.keys(files).some(
+            key => key.startsWith(dep.name) && key.endsWith('.d.ts')
```
CompuIves, line 414: Laast nitpick, apparently it needs be this (checked https://ata.codesandbox.io/api/v8/react%40latest)

```suggestion
            key => key.startsWith(`/${dep.name}`) && key.endsWith('.d.ts')
```

PR #4413, opened 2020-06-18
Improve execution time of custom babel plugins

CompuIves:
This really improves execution time of custom babel presets and plugins. Because:

1. We sync the FS in bulk, not file by file
2. We preload `@babel/types`, which is a really expensive lib to download, but needed by custom presets and plugins

Because of this we got transpilation time of Vue down from 18s to 2s.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4413/commits/c926e266352e243d6d63c40f99fde1e8ee4b54c0 is at https://pr4413.build.csb.dev/s/new.

PR #4416, opened 2020-06-19
install specific version when using tags

christianalfoni:
This handles two things:

1. When you use a TAG we add the actual version that tag represents (The same behaviour as when we do not add any specific version, we add the specific latest version)

2. A dependency can no longer override the types of the main package, if it is installed as an explicit package

### DIFF
--- a/packages/app/src/app/overmind/namespaces/editor/actions.ts
+++ b/packages/app/src/app/overmind/namespaces/editor/actions.ts
@@ -112,10 +112,10 @@ export const addNpmDependency: AsyncAction<{
   async ({ actions, effects, state }, { name, version, isDev }) => {
     effects.analytics.track('Add NPM Dependency');
     state.currentModal = null;
-    let newVersion = version;
-
-    if (!newVersion) {
-      const dependency = await effects.api.getDependency(name);
+    let newVersion = version || 'latest';
+    const isTag = !newVersion.includes('.');
+    if (isTag) {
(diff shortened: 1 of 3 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4416/commits/abee2041b0e24f7860e597058b271ce1101f8403 is at https://pr4416.build.csb.dev/s/new.

packages/app/src/app/overmind/namespaces/editor/actions.ts
```diff
@@ -112,10 +112,10 @@ export const addNpmDependency: AsyncAction<{
   async ({ actions, effects, state }, { name, version, isDev }) => {
     effects.analytics.track('Add NPM Dependency');
     state.currentModal = null;
-    let newVersion = version;
-
-    if (!newVersion) {
-      const dependency = await effects.api.getDependency(name);
+    let newVersion = version || 'latest';
+    const isTag = !newVersion.includes('.');
```
CompuIves, line 116: Maybe we should rather check here whether the version is absolute or not. I have a function for this: https://github.com/codesandbox/codesandbox-client/blob/master/packages/common/src/utils/dependencies.ts#L35-L39, maybe we could use this for it?

christianalfoni: Fixed 👍 😄

PR #4418, opened 2020-06-19
auto flip preview

christianalfoni:


CompuIves:
I'm not sure if this is the best solution to the flip, because this might be confusing for the user I think. That it would automatically switch when changing viewport, especially when people have specified that they want to have the vertical layout and then it switches back to horizontal layout on media change. I wouldn't expect something like this to happen in eg. VSCode or Unity either.

I was thinking, maybe we should make the option to flip more apparent? Maybe an explicit button or a menu in the devtool tabs (on the right) where you can change orientation. That would make it more intentional from the side of the user if they search for it, and it would make the feature of flipping more discoverable.

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4418/commits/95fadf360f7d5bb35c170ebfedd067a217c1395a is at https://pr4418.build.csb.dev/s/new.

PR #4419, opened 2020-06-19
Fix Weird behaviour or environment filters (new dahsboard)

SaraVieira:
The issue with using defaultChecked is that it doesn't update checkboxes when select all was clicked

Adding checked fixes it


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4419/commits/217b997c96418c947a779d1c17879af2dacbdac2 is at https://pr4419.build.csb.dev/s/new.

siddharthkp:
Can you please also fix all the warnings that using checked without onChange creates?

PR #4459, opened 2020-06-24
fix caching of typings

christianalfoni:
This ensure we get the latest version looking at ETAG, instead of keeping it cached:

```ts
  // Download a resource with cache busting when dealing with a
  // properly configured server that will send the correct ETag
  // and Date headers and properly handle If-Modified-Since and
  // If-None-Match request headers, therefore we can rely on the
  // validation to guarantee a fresh response.
  fetch("some.json", {cache: "no-cache"})
    .then(function(response) { /* consume the response */ });
```

### DIFF
--- a/packages/app/src/app/overmind/effects/vscode/SandboxFsSync/index.ts
+++ b/packages/app/src/app/overmind/effects/vscode/SandboxFsSync/index.ts
@@ -377,7 +377,12 @@ class SandboxFsSync {
       return this.typesInfo;
     }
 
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4459/commits/deee8dfcdba79ed7163b796db1a6bbf4ffbd9f75 is at https://pr4459.build.csb.dev/s/new.

CompuIves:
Ohhh, this makes a ton of sense! Good point

PR #4461, opened 2020-06-24
Repos in the dashboard

SaraVieira:
Adds repos to the dashboard
CleanShot 2020-06-30 at 15 45 40
CleanShot 2020-06-30 at 15 45 31
CleanShot 2020-06-30 at 15 46 37


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4461/commits/ce55d23814b4e72c54c4180f6c4f6b48917a7bfc failed.

packages/app/src/app/overmind/actions.ts
```diff
@@ -66,6 +66,7 @@ type ModalName =
   | 'deleteSandbox'
   | 'feedback'
   | 'forkServerModal'
+  | 'newSandbox'
```
CompuIves, line 69: We have a new way to render the "New Sandbox" modal, which allows for props. I can show you with a later suggestion.

packages/app/src/app/overmind/effects/gql/dashboard/queries.ts
```diff
@@ -127,6 +131,33 @@ export const getCollections: Query<
   ${sidebarCollectionDashboard}
 `;
 
+export const getRepos: Query<GetReposQuery, GetReposQueryVariables> = gql`
```
CompuIves, line 134: Would be good to call this `getPersonalRepos`, makes it more clear that there are two (also a team one).

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -398,6 +400,75 @@ export const getSandboxesByPath: AsyncAction<string> = async (
   }
 };
 
+export const getReposByPath: AsyncAction | Action<string> = async (
```
CompuIves, line 403:
```suggestion
export const getReposByPath: AsyncAction<string> = async (
```

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -398,6 +400,75 @@ export const getSandboxesByPath: AsyncAction<string> = async (
...
+  { state, effects },
+  path
+  // eslint-disable-next-line consistent-return
+) => {
+  const { dashboard } = state;
```
CompuIves, line 408: Nice implementation!

packages/app/src/app/overmind/namespaces/dashboard/utils.ts
```diff
@@ -1,6 +1,10 @@
 import { SidebarCollectionDashboardFragment as Collection } from 'app/graphql/types';
 import { DELETE_ME_COLLECTION } from './state';
 
+export const isRepoPage = location.pathname.includes('/repositories');
```
CompuIves, line 4: This doesn't work when you switch pages, and I think it's in general not a good practice to rely on URL string operations to determine what should be done. For example, if there's a directory in the dashboard called "repositories" this could break, and there are probably more examples where it doesn't work.

It would be better if we'd pass in the actions in overmind what the page is that we're dealing with, so we can base the logic on that instead.

packages/app/src/app/overmind/namespaces/dashboard/utils.ts
```diff
@@ -1,6 +1,10 @@
...
+
+export const repoName = location.pathname.split('/repositories/')[1];
```
CompuIves, line 6: Same for this, I would pass the `repoName` in to an action instead of gathering it from the URL.

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -571,14 +642,34 @@ export const deleteSandbox: AsyncAction<string[]> = async (
   const { user } = state;
   if (!user) return;
   const oldSandboxes = state.dashboard.sandboxes;
-  actions.dashboard.deleteSandboxFromState(ids);
+  if (!isRepoPage) {
```
CompuIves, line 645: Would it maybe be better that we also delete sandboxes from `state.dashboard.sandboxes.REPOS` in the `actions.dashboard.deleteSandboxFromState`? If you delete a sandbox tied to the GH repo from Drafts, it should also not be visible from Repositories I reckon.

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -683,25 +774,70 @@ export const renameFolderInState: Action<{ path: string; newPath: string }> = (
...
   oldTitle: string;
 }> = async ({ effects, actions }, { id, title, oldTitle }) => {
-  actions.dashboard.renameSandboxInState({
-    id,
-    title,
-  });
-  try {
-    await effects.gql.mutations.renameSandbox({
+  if (isRepoPage) {
+    actions.dashboard.renameRepoSandboxInState({
```
CompuIves, line 813: Same for here, shouldn't we handle renaming repo sandboxxes in `renameSandboxInState`. I'd like to know your opinion on this as well, @christianalfoni.

An idea that I had before would be that we actually keep all sandboxes in one array, and create `derived` functions for ordering them in `Recents`, `AllSandboxes`, `Repos`, etc... That would allow us to eg. only rename one instance of a sandbox and it's reflected everywhere immediately, also would save a lot of work when we're going to add live updates. I realize that this should be a PR on its own, but I'm thinking that if we're going to plan to do this anyway, then it would be better if we already put the renaming logic in a single function (`renameRepoSandboxInState`) instead of calling it conditionally.

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -776,13 +912,18 @@ export const makeTemplates: AsyncAction<{ sandboxIds: string[] }> = async (
   { sandboxIds: ids }
 ) => {
   const oldSandboxes = state.dashboard.sandboxes;
-  actions.dashboard.deleteSandboxFromState(ids);
+  if (!isRepoPage) {
```
CompuIves, line 915: I would also make this conditional.

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -1026,14 +1167,51 @@ export const changeSandboxPrivacy: AsyncAction<{
...
+        }),
+      },
+    };
+  };
   // optimistic update
-  actions.dashboard.changeSandboxPrivacyInState({ id, privacy });
+  if (isRepoPage) {
+    repoChangePrivacy(privacy);
+  } else {
+    actions.dashboard.changeSandboxPrivacyInState({ id, privacy });
```
CompuIves, line 1200: I would also put the repo privacy change in this function instead of 2 entry points.

packages/app/src/app/pages/NewDashboard/Components/Breadcrumbs/index.tsx
```diff
@@ -1,43 +1,45 @@
...
-    >
-      All Sandboxes {path && ' / '}
-    </Link>
-    {path
-      ? path.split('/').map((p, i) => {
-          const partPath = path
-            .split('/')
-            .slice(0, i + 1)
-            .join('/');
+  const makeLink = (p: string, link: string) => {
```
CompuIves, line 8: There was a bug in breadcrumbs for 2 folders deep, which is why I refactored this component. I think it would be better to base the changes in `Breadcrumbs` on the new version instead, I can definitely help with that!

packages/app/src/app/pages/NewDashboard/Components/EmptyScreen/index.tsx
```diff
@@ -39,6 +40,29 @@ export const EmptyScreen: React.FC<EmptyScreenProps> = ({
     );
   }
 
+  if (page === 'repos') {
```
CompuIves, line 43: Nice!!

packages/app/src/app/pages/NewDashboard/Components/Folder/FolderCard.tsx
```diff
@@ -102,7 +102,7 @@ export const FolderCard: React.FC<FolderItemComponentProps> = ({
         <IconButton
           name="more"
           size={9}
-          title="Sandbox actions"
+          title="Folder actions"
```
CompuIves, line 105: Good find!

packages/app/src/app/pages/NewDashboard/Components/Header/index.tsx
```diff
@@ -52,7 +54,7 @@ export const Header = ({
           {title}
         </Text>
       ) : (
-        <Breadcrumbs activeTeam={activeTeam} path={path} />
+        <Breadcrumbs repos={repos} param={path} />
```
CompuIves, line 62: `activeTeam` needs to be passed in the crumbs, because it's used to determine the URL (`teamId` is passed in the URL).

packages/app/src/app/pages/NewDashboard/Components/Repo/ImportRepo.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { useOvermind } from 'app/overmind';
+import { Stack, Text, Button, Icon } from '@codesandbox/components';
+import css from '@styled-system/css';
+
+export const ImportRepo = () => {
+  const { actions } = useOvermind();
+
+  const onClick = () => actions.modalOpened({ modal: 'newSandbox' });
```
CompuIves, line 9: There's a new one, I think it's 
```suggestion
  const onClick = () => actions.modals.newSandboxModal.open({});
```

packages/app/src/app/pages/NewDashboard/Components/Repo/RepoCard.tsx
```diff
@@ -0,0 +1,86 @@
...
+            '-webkit-box-orient': 'vertical',
+          })}
+        >
+          {props.owner}/{name}
+        </Text>
+        <Text size={3} variant="muted" weight="medium">
+          {props.branch}
+        </Text>
+      </Stack>
+      <IconButton
```
CompuIves, line 78: `IconButton` is really expensive to render for some reason, it takes over 3ms to render at times, which makes the page super slow when scrolling. I have a trick on the `SandboxCard` component where we can prevent the slowdown by just rendering an icon when we're scrolling. It would be good if we could use that here as well!

packages/app/src/app/pages/NewDashboard/Components/Repo/RepoListItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+      <Column span={[0, 4, 4]} as={Stack} align="center">
+        <Text size={3} weight="medium" variant="muted">
+          {props.owner}
+        </Text>
+      </Column>
+      <Column span={[0, 3, 3]} as={Stack} align="center">
+        {/* empty column to align with sandbox list items */}
+      </Column>
+    </Grid>
+    <IconButton
```
CompuIves, line 72: Same one here, would be good to render something different while scrolling. `SandboxCard` has an example for it.

packages/app/src/app/pages/NewDashboard/Components/Sandbox/index.tsx
```diff
@@ -239,6 +239,8 @@ const GenericSandbox = ({ isScrolling, item }: GenericSandboxProps) => {
   };
 
   const sandboxProps = {
+    prNumber: item.sandbox.prNumber,
+    originalGit: item.sandbox.originalGit,
```
CompuIves, line 243: Aren't these props not on `sandbox` itself? (on line 249)

packages/app/src/app/pages/NewDashboard/Components/Selection/ContextMenu.tsx
```diff
@@ -76,6 +100,10 @@ export const ContextMenu: React.FC<IContextMenuProps> = ({
     menu = <SandboxMenu item={selectedItems[0]} setRenaming={setRenaming} />;
   } else if (selectedItems[0].type === 'folder') {
     menu = <FolderMenu folder={selectedItems[0]} setRenaming={setRenaming} />;
+  } else if (selectedItems[0].type === 'repo') {
```
CompuIves, line 103: Very nice!

packages/app/src/app/pages/NewDashboard/Components/Selection/ContextMenu.tsx
```diff
@@ -342,6 +370,35 @@ const SandboxMenu: React.FC<SandboxMenuProps> = ({ item, setRenaming }) => {
   );
 };
 
+const RepoMenu = ({ repo }) => {
+  const { effects } = useOvermind();
+  const { visible, setVisibility, position } = React.useContext(Context);
+  const link = `https://github.com/${repo.owner}/${repo.name}/tree/${repo.branch}`;
+  const csbLink = `https://codesandbox.io/s/github/${repo.owner}/${repo.name}/tree/${repo.branch}`;
```
CompuIves, line 377: For generating links it would be better to use the helper function. This wouldn't work for example for sandboxes that are linked to a commit. in `url-generators.ts` there is a `sandboxLink` (or similar name) function that handles the `repo` info.

packages/app/src/app/pages/NewDashboard/Components/Selection/ContextMenu.tsx
```diff
@@ -41,16 +44,38 @@ export const ContextMenu: React.FC<IContextMenuProps> = ({
...
+    | DashboardRepo
+    | DashboardNewMasterBranch
   > = selectedIds.map(id => {
     if (id.startsWith('/')) {
+      if (repos.length) {
+        const repo = repos.find(f => '/' + f.name === id);
+        return { type: 'repo', ...repo };
+      }
+
+      if (id.includes('/github')) {
```
CompuIves, line 67: Is there a more explicit way we can check here? To make sure nothing goes wrong? Like checking for `id.startsWith(` instead maybe?

packages/app/src/app/pages/NewDashboard/Components/Selection/ContextMenu.tsx
```diff
@@ -496,6 +559,48 @@ const MultiMenu = ({ selectedItems }: IMultiMenuProps) => {
...
+    <Menu.ContextMenu
+      visible={visible}
+      setVisibility={setVisibility}
+      position={position}
+      style={{ width: 120 }}
+    >
+      <MenuItem
+        onSelect={() =>
+          window.open(
+            `https://codesandbox.io/s/github/${repo.owner}/${repo.name}`
```
CompuIves, line 584: Same here, is there a better way that we can generate the URL?

CompuIves: I think it would be better to use relative URLs as well (so it works on staging for example).

packages/app/src/app/pages/NewDashboard/Components/Selection/index.tsx
```diff
@@ -562,9 +570,50 @@ export const SelectionProvider: React.FC<SelectionProviderProps> = ({
     else if (new Date().getTime() - callbackCalledAt.current > 60) callback();
   };
 
-  const onContainerMouseUp = event => {
+  const onContainerMouseUp = () => {
     if (drawingRect) resetSelectionRect();
   };
+  if (page === 'repos') {
```
CompuIves, line 576: Do we need a completely different one for the `'repos'` page?

packages/app/src/app/pages/NewDashboard/Components/VariableGrid/index.tsx
```diff
@@ -71,8 +80,12 @@ const ComponentForTypes: ComponentForTypes = {
     <Sandbox item={props.item} isScrolling={props.isScrolling} />
   )),
   folder: props => <Folder {...props.item} />,
+  repo: props => <Repo {...props.item} />,
+  // @ts-ignore TODO: find a better way to type this
```
CompuIves, line 84:
```suggestion
```

packages/app/src/app/pages/NewDashboard/Content/routes/Repositories/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+  React.useEffect(() => {
+    const p = home ? null : param;
+    actions.dashboard.getReposByPath(p);
+  }, [param, actions.dashboard, activeTeam, home]);
+
+  const activeSandboxes =
+    (sandboxes.REPOS && Object.values(sandboxes.REPOS)) || [];
+
+  const itemsToShow = (): DashboardGridItem[] => {
+    if (!sandboxes.REPOS) {
```
CompuIves, line 34: We need to check explicitly with `== null` here, because it now will also show the skeleton in case `sandboxes.REPOS = []` I believe.

packages/app/src/app/pages/NewDashboard/Content/routes/Repositories/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+          ]
+        : items;
+    }
+
+    return [{ type: 'skeleton-row' }, { type: 'skeleton-row' }];
+  };
+
+  const possibleTemplates = itemsToShow()
+    // @ts-ignore
+    .filter(s => s.sandbox)
```
CompuIves, line 62: Hmm, why do we do this? And what is TS complaining about?

packages/app/src/app/pages/NewDashboard/Content/routes/Repositories/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+    },
+  } = useOvermind();
+
+  React.useEffect(() => {
+    const p = home ? null : param;
+    actions.dashboard.getReposByPath(p);
+  }, [param, actions.dashboard, activeTeam, home]);
+
+  const activeSandboxes =
+    (sandboxes.REPOS && Object.values(sandboxes.REPOS)) || [];
```
CompuIves, line 31: `Object.values` is quite expensive. It maps over all the values and puts them in an array. Maybe we could check explicitly for `null` instead here?

packages/app/src/app/pages/NewDashboard/Content/routes/Repositories/useFilteredItems.ts
```diff
@@ -0,0 +1,60 @@
...
+      },
+    },
+  } = useOvermind();
+  const [items, setItems] = useState<Array<DashboardGridItem>>([]);
+  const sandboxesForPath =
+    sandboxes.REPOS && param
+      ? getFilteredSandboxes(sandboxes.REPOS[param].sandboxes || [])
+      : [];
+
+  const repos = (sandboxes.REPOS && Object.values(sandboxes.REPOS)) || [];
```
CompuIves, line 28: Same here, I think it's better to not use `Object.values` but rather something like `== null`.

packages/app/src/app/pages/NewDashboard/Sidebar/index.tsx
```diff
@@ -172,6 +172,13 @@ export const Sidebar: React.FC<SidebarProps> = ({
             ]}
           />
 
+          <RowItem
+            name="Repositories"
+            page="repos"
+            path={dashboardUrls.repos(activeTeam)}
```
CompuIves, line 178: This is super nice!

packages/app/src/app/pages/NewDashboard/types.ts
```diff
@@ -12,14 +13,20 @@ export type DashboardBaseFolder = {
 
 export type DashboardSandbox = {
   type: 'sandbox';
-  sandbox: SandboxFragmentDashboardFragment;
+  sandbox: SandboxFragmentDashboardFragment & {
```
CompuIves, line 16: Maybe we should add the properties to the `SandboxFragmentDashboardFragment` instead? Since it seems like we're going to use it in other pages besides `Repositories` as well (from what I've seen).

packages/common/src/global.css
```diff
@@ -1,6 +1,6 @@
 html,
 body {
-  font-family: 'Roboto', sans-serif;
+  font-family: 'Inter', sans-serif;
```
CompuIves, line 3: This might break some styling in other places, we definitely need to test all pages to make sure that everything still works with this font.

Alternatively, I think we should do this in a separate PR. Then we are not bringing this PR out of scope.

SaraVieira: removed, it was a request from danny

packages/app/src/app/pages/NewDashboard/types.ts
SaraVieira: hm?

packages/app/src/app/pages/NewDashboard/Content/routes/Repositories/useFilteredItems.ts
SaraVieira: fixed

SaraVieira: Ohhh, no because I actually want the values of each key in the object

packages/app/src/app/pages/NewDashboard/Content/routes/Repositories/index.tsx
SaraVieira: That's the thing I need that, to map the values and put them in an array ahah

packages/app/src/app/overmind/namespaces/dashboard/utils.ts
SaraVieira: All done :)

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
SaraVieira: because of line 421 It can also be a normal actions

packages/app/src/app/overmind/effects/gql/dashboard/queries.ts
SaraVieira: good point! Done :)

packages/app/src/app/pages/NewDashboard/Components/Selection/ContextMenu.tsx
SaraVieira: changed

packages/app/src/app/pages/NewDashboard/Components/Selection/index.tsx
SaraVieira: omg yes, there is noe fancy stuff, I don't think the contextMenu should be there tbh

packages/app/src/app/pages/NewDashboard/Components/VariableGrid/index.tsx
SaraVieira: Removed :)

packages/app/src/app/pages/NewDashboard/Content/routes/Repositories/index.tsx
SaraVieira: changed :)

packages/app/src/app/pages/NewDashboard/Content/routes/Repositories/index.tsx
SaraVieira: it says there is no sandbox in the type 🤷‍♀️

packages/app/src/app/pages/NewDashboard/Components/Selection/ContextMenu.tsx
SaraVieira: Done :)

packages/app/src/app/pages/NewDashboard/Components/Selection/ContextMenu.tsx
SaraVieira: done :)

packages/app/src/app/overmind/namespaces/dashboard/utils.ts
SaraVieira: I honestly think its better to get it from the URL, in the cases where we are in the repo page this will never break and we already send the page

I think this makes it easier to not screw up

packages/app/src/app/pages/NewDashboard/Content/routes/Repositories/index.tsx
CompuIves: Ahhh ohhh, yep this makes sense. I initially thought you were only using it to find whether the array was empty or not. Sorry! 😂

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
SaraVieira: Moved the condition to the function itself

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
SaraVieira: done

packages/app/src/app/pages/NewDashboard/Components/Header/index.tsx
SaraVieira: done

packages/app/src/app/pages/NewDashboard/Components/Breadcrumbs/index.tsx
SaraVieira: updated :)

packages/app/src/app/pages/NewDashboard/Components/Repo/ImportRepo.tsx
SaraVieira: updated

packages/app/src/app/overmind/actions.ts
SaraVieira: removed :)

packages/app/src/app/pages/NewDashboard/Components/Repo/RepoCard.tsx
SaraVieira: done :)

packages/app/src/app/pages/NewDashboard/Components/Repo/RepoListItem.tsx
SaraVieira: done

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
SaraVieira: nvm, done

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
SaraVieira: done

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
SaraVieira: this is done, I don't know what the rest is ahah

packages/app/src/app/pages/NewDashboard/Components/Sandbox/index.tsx
SaraVieira: But the sandbox isn't passed in totality down

CompuIves: It is, on line 251 the `item.sandbox` is passed down in totality. Or is there another reason to send down these props separately?

packages/app/src/app/pages/NewDashboard/Components/Selection/ContextMenu.tsx
```diff
@@ -41,16 +47,39 @@ export const ContextMenu: React.FC<IContextMenuProps> = ({
...
   const selectedItems: Array<
-    DashboardFolder | DashboardSandbox | DashboardTemplate
+    | DashboardFolder
+    | DashboardSandbox
+    | DashboardTemplate
+    | DashboardRepo
+    | DashboardNewMasterBranch
   > = selectedIds.map(id => {
     if (id.startsWith('/')) {
+      if (repos.length) {
```
CompuIves, line 66: Should we maybe change `repos?: ` to `repos: ` in the props? Since we're assuming here that `repos` cannot be `undefined`.

packages/app/src/app/pages/NewDashboard/Components/Selection/index.tsx
CompuIves: Do we still need it here in that case?

packages/app/src/app/pages/NewDashboard/types.ts
CompuIves: I'm wondering, do we also show specific properties or menu items for repo sandboxes in other places than the repo page? If that's the case we could also change the fragment query of sandbox itself to include these properties.

packages/common/src/global.css
```diff
@@ -6,14 +6,14 @@ body {
   -moz-osx-font-smoothing: grayscale;
   font-smoothing: auto;
   text-rendering: optimizeLegibility;
-  font-smooth: always;
+  font-smooth: alays;
```
CompuIves, line 9:
```suggestion
  font-smooth: always;
```

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -6,10 +7,21 @@ import {
...
   TemplateFragmentDashboardFragment,
   SandboxFragmentDashboardFragment,
+  RepoFragmentDashboardFragment,
 } from 'app/graphql/types';
 import { TEAM_ID_LOCAL_STORAGE } from 'app/overmind/utils/team';
 import { OrderBy, sandboxesTypes } from './state';
-import { getDecoratedCollection } from './utilts';
+import { getDecoratedCollection, repoName } from './utils';
+
+type PageTypes =
```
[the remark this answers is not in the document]
SaraVieira: done, moved a bunch of stuff there

packages/app/src/app/overmind/namespaces/dashboard/utils.ts
SaraVieira: this does work when you move pages, I tested this a lot to make sure it wouldn't break but sure I can pass the repo name into the action

SaraVieira: done

christianalfoni:
Hi there! 😄 

Okay, first of all! This is an insanely complicated implementation where @CompuIves has a TON of existing knowledge having implemented the dashboard once before. So what I think is important to remember here is that the goal was not only to implement a new dashboard, but also transfer the ownership from @CompuIves. It is really great we get this review and get some laste minute knowledge sharing and improvements, but also... this is not the last time we will touch the dashboard 😄  

So I am super pleased that the code is now also manageable for me personally, as it is consistent with our other implementations as well. And browsing through the comments and fixes here, I believe we are in a shipable state after the last ones are resolved. And also in a really good state to fix any issues popping up, and also continue extending its features, as we are now 3 people who feels comfortable changing the code, instead of just one 😄 

I do have some suggestions on how we can structure things a little bit differently to improve consistency and reduce some boilerplate code, but that would be for the next time we look at the dashboard. In advance, sorry if any of these suggestions are misleading as there are certainly parts of the implementation I have not grasped skimming through code 😄 : 

1. I see that Sandboxes are now organized explicitly by page and thinking about this, I think it makes total sense. Though... instead of storing the data for each page, I would rather split storing data and deriving what data to show for each page:

```ts
{
  sandboxes: { [id: string]: Sandbox },
  templates: { [id: string]: Template },
  repos: { [path: string]: Repo }, // This is the full type, including `sandboxes`
  currentPage: Page
  currentSandboxes: {
    [Page.DRAFTS]: Sandbox[], // derive sandboxes based on page
    [Page.TEMPLATES]: Template[], // derive templates based on page
    ...
  }
}
```

This will also remove things like `recentSandboxesByTime` and `deletedSandboxesByTime`, as those are now the derived for the specific page. It would make all the derived dirty when moving between pages, but since you can move sandboxes between pages it makes sense to recompute them when you open a page as things could have changed. Also whenever a page is opened a new query should run and update the data source, ex `sandboxes`, if the ID is not already there. 

Anyways... this ensures that the UI is always consistent by recomputing, instead of manually updating the data sources.

What this also now allows is having a single action for opening a page:

```ts
export const openPage = ({ state, effects }, page) => {
  state.dashboard.currentPage = page
  try {
    switch (page) {
      case Page.DRAFTS: {
        ... fetch some data
      }
      case Page.RECENT: {
        ... fetch some data
      }
    }
  } catch () {}
}
```

This also simplifies things like deleting:

```ts
export const deleteSandboxes: AsyncAction<{
  ids: string[];
> = async ({ state, effects, actions }, { ids, page }) => {
  const { user } = state;
  if (!user) return;
  
  const sandboxesToDelete = ids.reduce((id) => aggr.concat(state.dashboard.sandboxes[id]), [])

  sandboxesToDelete.forEach(sandbox => {
    delete state.dashboard.sandboxes[sandbox.id]
  }) 

  try {
    await effects.gql.mutations.deleteSandboxes({
      sandboxIds: ids,
    });
  } catch (error) {
    sandboxesToDelete.forEach((sandbox) => {
      state.dashboard.sandboxes[sandbox.id] = sandbox
    })

      'There was a problem deleting your Sandbox'
    );
  }
};
```

Okay, those where the things I think could clean up a little bit and also ensure some more consistency. 

Anyways, awesome work getting this in and also hammering through feedback... and types... just one more run and it is shipable! 😄 ❤️

packages/app/src/app/pages/NewDashboard/Components/Selection/ContextMenu.tsx
SaraVieira: my bad, also checked if repos exist now

packages/app/src/app/pages/NewDashboard/Components/Sandbox/index.tsx
SaraVieira: ohh, my bad, used that instead

packages/app/src/app/pages/NewDashboard/Components/Selection/index.tsx
SaraVieira: spoke in slack

packages/app/src/app/pages/NewDashboard/types.ts
SaraVieira: No we won't but that's the issue, if I try to get them TS will complain even thought I know they will exist since I am on the repos page

CompuIves:
Nice!!! :shipit:

PR #4028, opened 2020-04-30
Cleanup Explorer screen

MichaelDeBoey:
Things I did:
- Add `algoliasearch` dependency
- Make `BundleSizes`, `Dependency`, `Explorer` & `Files` a `FunctionComponent`
- Move `overmind` subscriptions as close as possible to the components itself instead of passing it through
- Type the icons

PR #4505, opened 2020-07-02
Cleanup Explorer screen

MichaelDeBoey:
Resubmission of #4028

Things I did:
- Add `algoliasearch` dependency
- Make `BundleSizes`, `Dependency`, `Explorer` & `Files` a `FunctionComponent`
- Type the icons

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -87,7 +87,6 @@
     "@types/socket.io-client": "^1.4.32",
     "@vue/babel-preset-app": "^3.2.0",
     "airtable": "^0.5.8",
-    "algoliasearch": "^3.35.1",
     "apollo-boost": "^0.4.4",
     "apollo-link-batch-http": "^1.2.12",
     "apollo-link-context": "^1.0.18",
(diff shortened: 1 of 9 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4505/commits/f7238473824c72e55b55cbbdb3980a411d038551 is at https://pr4505.build.csb.dev/s/new.

MichaelDeBoey:
@SaraVieira I've rebased onto `upstream/master`, so this could be merged now 🙂

MichaelDeBoey:
@SaraVieira Why did you delete the `algoliasearch` dependency? 🤔
It's a dependency of this module and it wasn't listed yet, so it should be.

Also the static check will make the mental model clearer. Without my change, you have to think "OK so if the condition is true, now take the opposite", while with my change you just have to think "if the condition is true".

SaraVieira:
1. I removed the algolia because it's already in the yarn.lock so there is no need to duplicate since the version your installed was different than the one installed and it was actually going to install two versions instead of one we have now

2. I changed the check because that is completely subjective and since it was a different and not a `!something` I found the way it was changed to way more confusing

Again thank you!

PR #4509, opened 2020-07-06
add avatar ulpload ability

SaraVieira:
Alow team members to upload a team avatar


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4509/commits/b3a1368d87246793b41af436e40546da038e0c3c is at https://pr4509.build.csb.dev/s/new.

CompuIves:
Nice!! Doing a usability check now

image
Could we change the max-width here maybe to make it a bit larger? I think it's now 994px, which seems a bit crammed, could we get it to 1280px?

image
Such a small lonely C  😂 . I think we should bump the font-size a bit here.

Okay, but other than that, this flow was incredibly smooth! Big props, so nice: 
image.

Going into code review now.

packages/app/src/app/components/TeamAvatar/TeamAvatar.tsx
```diff
@@ -15,29 +15,51 @@ export const backgrounds = [
 
 interface TeamAvatarProps {
   name: string;
-  size?: 'small' | 'big';
+  size?: 'small' | 'big' | 'bigger';
   className?: string;
+  avatar?: string;
+  style?: any;
```
CompuIves, line 21: Nice helper function for this:

```suggestion
  style?: React.CSSProperties;
```

packages/app/src/app/components/TeamAvatar/TeamAvatar.tsx
```diff
@@ -15,29 +15,51 @@ export const backgrounds = [
...
 }
 
 export const TeamAvatar = ({
   name,
   size = 'big',
   className,
+  avatar,
+  ...props
```
CompuIves, line 29: Is there an explicit need for the use of `...props` here? I would prefer to not use it too much because it adds a bit of ambiguity (you don't know what can be passed to this component anymore) and causes a small performance hit, because it's transpiled to this code: https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBAhjAvDAFAbwEYBoYDp8AOATiARAL4CUSAfKgFAwwA8AJgJYBuMGim5MNPlzFSFGAHoa9SkA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-2%2Cenv&prettier=false&targets=&version=7.10.4&externalPlugins=.

If we can explicitly name the props I would always go for that approach, but yeah, only if possible.

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -1127,3 +1138,24 @@ export const changeSandboxPrivacyInState: Action<{
...
 };
+
+export const updateTeamAvatar: AsyncAction<{
+  name: string;
+  url: string;
+  teamId: string;
+}> = async ({ actions, effects, state }, { name, url, teamId }) => {
+  if (!state.activeTeamInfo) return;
+  const oldAvatar = state.activeTeamInfo.avatarUrl;
+  state.activeTeamInfo.avatarUrl = url;
```
CompuIves, line 1149: Nice handling with optimistic updates!

packages/app/src/app/components/TeamAvatar/TeamAvatar.tsx
SaraVieira: I do that mostly because of style props, so we can always style everything :/

Changed to passing style

SaraVieira:
All comments fixed, changed the font to 6 in our scale and the width of the team and user settings to 1100 since 1280 was too big in average monitors

packages/app/src/app/components/TeamAvatar/TeamAvatar.tsx
```diff
@@ -15,29 +15,57 @@ export const backgrounds = [
...
 interface TeamAvatarProps {
   name: string;
-  size?: 'small' | 'big';
+  size?: 'small' | 'big' | 'bigger';
   className?: string;
+  avatar?: string;
+  style?: React.CSSProperties;
 }
 
+const textSizes = {
```
CompuIves, line 24: Very nice!

PR #4528, opened 2020-07-09
Add sync step in gitflow

SaraVieira:
This PR adds a step where the user says it wants to contribute back and then we will sync with Github

### DIFF
--- a/packages/app/src/app/overmind/namespaces/git/actions.ts
+++ b/packages/app/src/app/overmind/namespaces/git/actions.ts
@@ -872,18 +872,26 @@ export const _tryResolveConflict: AsyncAction = async ({
   }
 };
 
-export const enableGitSync: AsyncAction<string> = async (
+export const linkToGitSandbox: AsyncAction<string> = async (
   { state, effects, actions },
   sandboxId
 ) => {
   if (!state.editor.currentSandbox) return;
-  state.git.isEnablingSync = true;
-  const newGitData = await effects.api.makeGitSandbox(sandboxId);
-  state.editor.currentSandbox = {
-    ...state.editor.currentSandbox,
-    originalGitCommitSha: newGitData.originalGitCommitSha,
-    originalGit: newGitData.originalGit,
-  };
-  state.git.isEnablingSync = false;
-  await actions.git.loadGitSource();
+  try {
+    state.git.isLinkingToGitSandbox = true;
+    const newGitData = await effects.api.makeGitSandbox(sandboxId);
+    state.editor.currentSandbox = {
+      ...state.editor.currentSandbox,
+      originalGitCommitSha: newGitData.originalGitCommitSha,
+      originalGit: newGitData.originalGit,
+    };
+    state.git.isLinkingToGitSandbox = false;
+    await actions.git.loadGitSource();
+  } catch (error) {
+    actions.internal.handleError({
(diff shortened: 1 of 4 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4528/commits/f2a91390f286a9bce8bd47a699abe4e9270a8966 is at https://pr4528.build.csb.dev/s/new.

packages/app/src/app/overmind/namespaces/git/actions.ts
```diff
@@ -871,3 +871,19 @@ export const _tryResolveConflict: AsyncAction = async ({
...
+) => {
+  if (!state.editor.currentSandbox) return;
+  state.git.isEnablingSync = true;
+  const newGitData = await effects.api.makeGitSandbox(sandboxId);
+  state.editor.currentSandbox = {
+    ...state.editor.currentSandbox,
+    originalGitCommitSha: newGitData.originalGitCommitSha,
+    originalGit: newGitData.originalGit,
+  };
+  state.git.isEnablingSync = false;
```
CompuIves, line 887: Maybe we should wrap this in a `try...catch` and put `state.git.isEnablingSync = false`  in the `finally`. Not much could go wrong here, but it's a good way to ensure that we can't get stuck.

packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/GitHub/index.tsx
```diff
@@ -191,6 +196,32 @@ export const GitHub = () => {
     );
   }
 
+  if (!originalGit && forkedTemplateSandbox?.git) {
```
CompuIves, line 199: Great!! Could you also maybe check for `forkedFromSandbox`? So something like:

```
const forkedSandbox = sandbox.forkedTemplateSandbox || sandbox.forkedFromSandbox;
```

If we're ever going to remove the default template from GitHub sandboxes, this will still then work in the future.

packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/GitHub/index.tsx
```diff
@@ -53,20 +54,24 @@ export const GitHub = () => {
         isFetching,
         isExported,
         pr,
+        isEnablingSync,
```
CompuIves, line 57: Maybe this is a bit ambiguous for other readers, we could maybe name this to `isLinkingToGitSandbox`?

packages/app/src/app/overmind/namespaces/git/actions.ts
SaraVieira: done :)

packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/GitHub/index.tsx
SaraVieira: Done!

packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/GitHub/index.tsx
SaraVieira: true, I made this when I was calling git sync

PR #4539, opened 2020-07-11
Update JSDOM and create script to automate it

### LINKED ISSUE #4530: JSDOM is out of date
# 🐛 bug report

## Description of the problem

It seems like Code sandbox has an old version of JSDOM (4.0.0) hard coded into the bundle: https://github.com/codesandbox/codesandbox-client/pull/1812/files#diff-3fcdeb2de813a974b29d8610b234c7f2R51

Testing Library users have been having issues like testing-library/user-event#391 which are caused by bugs in JSDOM that have since been fixed in newer versions. This could theoretically affect any project running Jest tests that rely on JSDOM though (including create-react-app's default configuration).

## Suggested fix

For reference, it seems like Jest is currently using JSDOM 12, and in the future will be using JSDOM 16. It would be nice if we could either automatically detect Jest's version of JSDOM or use the latest dynamically via Unpkg/npm. If that's not possible, just upgrading the hard coded version to 12 or 16 should help a lot. I would have tried to do this myself, but JSDOM doesn't seem to release a single bundle file any more.

### Link to sandbox: [link](https://codesandbox.io/s/code-sandbox-bug-with-old-jsdom-version-9gcnp?file=/index.test.js)

If you download this and run the `test` script locally, it will pass because Jest is using a newer version of JSDOM that fixes the bug with the initial values of `selectionStart` and `selectionEnd`. This is also the same behavior if you run similar code in a browser.

### Your Environment

| Сodesandbox      | latest at the time of issue |
| Browser          | N/A |
| Operating System | N/A |

CompuIves:
Fixes #4530


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4539/commits/930c0e334e797dc04357b5a5b39fa9aa43da1c7a is at https://pr4539.build.csb.dev/s/new.

nickmccurdy:
Unfortunately my test from #4530 is still failing with the same issue, seems like old JSDOM is still used in this sandbox: https://pr4539.build.csb.dev/s/code-sandbox-bug-with-old-jsdom-version-9gcnp?file=/index.test.js

CompuIves:
> It would probably be best to rebase this PR to only add the .min.js file (so the 5MB git object for .js isn't written to the repository) and remove the old JSDOM 4 files that aren't in use any more.

I've decided to not remove JSDOM 4 yet, since some service workers need to request the file still. I'm planning to remove it in a month though! For the `.min.js` vs `.js`, that makes sense since we can generate one ourselves now. I'll remove the `.js` one!

> Unfortunately my test from #4530 is still failing with the same issue, seems like old JSDOM is still used in this sandbox

Hmm, that is weird. I see the template is `static`, so the bundler shouldn't be running any tests there.

nickmccurdy:
That's still strange, if it's running tests why is it still using JSDOM (and still 4 not 12)? I could tell if it was using the real DOM because the tests would be passing. Do we need to test this without the static template and disable the test runs separately?

marcosvega91:
Just for info I have created a repo with the code that is broken on our side

https://pr4539.build.csb.dev/s/cold-currying-p7r3g?file=/package.json

nickmccurdy:
FYI: My sandbox was based on this so should reproduce the same issue. The bug in `user-event` is caused by a combination of a regression (which we decided not to fix) and a bug in JSDOM (which is in Code Sandbox's version, JSDOM 4). My sandbox shows how JSDOM incorrectly marks `selectionStart` and `selectionEnd`, which causes that regression in `user-event`.

CompuIves:
Picking this one up again! Apologies for the time, I lost track of this one.

I now updated JSDOM to 16.3.0, and only committed the `min` version, but still, getting the same error. It's really werid.

CompuIves:
Actually... The log does work...

image

CompuIves:
Weird, this returns false:

```
Object.prototype.hasOwnProperty.call(element, 'selectionStart')
```

CompuIves:
Ah, it turns out this was a bug in `jest-matchers` that have since been fixed. I've just updated and the report now seems to work correctly on my side!

marcosvega91:
[this](https://pr4539.build.csb.dev/s/cold-currying-p7r3g?file=/src/__tests__/index.js) test is working now, so on our side we are good too 😄

Issue #1543, opened 2019-02-14
Angular Cli not keeping ES6 JavaScript code


AmyScript:
# 🐛 bug report

## Description of the problem
When using the Angular CLI sandbox and a tsconfig.app.json file to configure the "target" to es6, the compiler still transpiles .js files to ES5.  In particular, if you look at the lit-element.js file which is a dependency, all the ES6 classes gets transpiled into ES5 causing the error:

updating-element.js? [sm]:75 Uncaught TypeError: Failed to construct 'HTMLElement': Please use the 'new' operator, this DOM object constructor cannot be called as a function.

## How has this issue affected you? What are you trying to accomplish?
Trying to use web components in Angular CLI in codesandbox without the es5-adapter.  When I have this in my local enivonroment, using the same Angular CLI, I do not have this issue.
###  Link to sandbox: [link]() (optional)
Link to sandbox not working without es5-adapter:
https://codesandbox.io/s/nrxlzl8754
Link to sandbox working with es5-adapter, but the adapter shouldn't be needed is the typescript transpiler is working properly:
https://codesandbox.io/s/538491qzmx
### Your Environment
| Software         | Name/Version|
Angular CLI
| Сodesandbox  | 
| Browser | 
| Operating System |

waterplea:
Any news on this? I also need to use ES6 in my code, but cannot configure it to compile to ES6

waterplea:
Tried all sorts of tsconfig and browserslist placement and configurations. Please take a look, this issue is over a year old and not being able to use es2015 in 2020 is a darn shame. I'm limited to codesandbox because of all online Angular sandboxes it is the only one allowing uploading assets.

HristoP96:
We also have problems with the lack of configuration for es2015 - [https://codesandbox.io/s/igx-grid-custom-summaries-jvpsf](https://codesandbox.io/s/igx-grid-custom-summaries-jvpsf).
And now, given the fact that with the [Angular 10](https://blog.angular.io/version-10-of-angular-now-available-78960babd41) release "The Angular Package Format no longer includes ESM5 or FESM5 bundles", the severity of this issue should be escalated.

CompuIves:
Apologies for the delay, I'm looking at this today and I'm aiming to fix it today as well!

HristoP96:
[...]

PR #4544, opened 2020-07-13
Optimize resolving of sass files

CompuIves:
Related to #1543.

Our sass-loader requires a round-trip from the worker thread to the main thread to check if a file exists, because the file should live on the main thread. Because of this it can take long to resolve a single file, since we need to use the node resolver (which checks ~20 paths) before finding the right result. Because 99% of the time, the file is relative to the requester, we now check that first before using the node resolver. This results in 99% of the cases that we only need to do 1 roundtrip, instead of 20, and thus in faster Sass compiling.

### DIFF
--- a/packages/app/src/sandbox/eval/transpilers/sass/index.ts
+++ b/packages/app/src/sandbox/eval/transpilers/sass/index.ts
@@ -9,7 +9,7 @@ class SassTranspiler extends WorkerTranspiler {
   worker: Worker;
 
   constructor() {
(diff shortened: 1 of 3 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4544/commits/d4d25e2d98d2f317da2680971a0f274809936701 is at https://pr4544.build.csb.dev/s/new.

PR #4549, opened 2020-07-13
remove "no-loop-func" rule

SaraVieira:
Removes this rule from eslint, discussed on twitter here 

https://twitter.com/AdamRackis/status/1282758047807438851?s=20


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4549/commits/fc45d4a2caf5da2653c9f14681e5a4cfb55009b1 is at https://pr4549.build.csb.dev/s/new.

CompuIves:
Yep, I agree with this change. I think this rule is enabled because I copied the recommended eslint rules while implementing it, which was several years ago so probably outdated by now.

I'll merge this now, let's make another follow up PR that syncs this list with the current recommended rules!

PR #4559, opened 2020-07-16
add sass and css support to svelte

SaraVieira:
Adds support for (scss) in svelte templates

To test use this example: https://codesandbox.io/s/agitated-snow-u0y0z?file=/App.svelte

### DIFF
--- a/packages/app/src/sandbox/eval/presets/svelte/index.js
+++ b/packages/app/src/sandbox/eval/presets/svelte/index.js
@@ -2,6 +2,9 @@ import babelTranspiler from '../../transpilers/babel';
 import jsonTranspiler from '../../transpilers/json';
 import rawTranspiler from '../../transpilers/raw';
 import svelteTranspiler from '../../transpilers/svelte';
+import sassTranspiler from '../../transpilers/sass';
+import styleProcessor from '../../transpilers/postcss';
+import stylesTranspiler from '../../transpilers/style';
 
 import Preset from '..';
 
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4559/commits/663b4a11e231f413de3e92c004ddbf125a7a65e8 is at https://pr4559.build.csb.dev/s/new.

packages/app/src/sandbox/eval/presets/svelte/index.js
```diff
@@ -23,6 +26,23 @@ export default function initialize() {
...
+    { transpiler: styleProcessor },
+    {
+      transpiler: stylesTranspiler,
+      options: { hmrEnabled: false },
+    },
+  ]);
+
+  sveltePreset.registerTranspiler(module => /\.s[c|a]ss$/.test(module.path), [
+    { transpiler: sassTranspiler },
+    { transpiler: styleProcessor },
```
CompuIves, line 39: Mini comment: this transpiler can be slow at times, if we don't necessarily need it we should maybe consider not using it, but it's just something to keep in mind.

SaraVieira: I mean I added it because we need it :/

I tried without it

CompuIves: Ah, okay, good to know!

PR #4571, opened 2020-07-17
Fix removeChild error in Explorer

CompuIves:
React throws a removeChild error when this is rendered:

```
<>
{false && <div />
</>
```

because the contents are empty, as I understood from here: https://stackoverflow.com/questions/54880669/react-domexception-failed-to-execute-removechild-on-node-the-node-to-be-re.

This seems to happen sometimes in the explorer, and this change fixes that.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4571/commits/515e3a516adc61d0bd9527dc337b5c29d6b91ff7 is at https://pr4571.build.csb.dev/s/new.

PR #4582, opened 2020-07-21
Dashboard - only show counter for multiple items

siddharthkp:
showing a counter of 1 doesn't convey any information. We should remove it to keep it minimal

Screenshot 2020-07-21 at 11 30 10 AM

on the flip side, it does give a hint that you can select multiple. Leaving this for @DannyRuchtie and @CompuIves's input

### DIFF
--- a/packages/app/src/app/pages/NewDashboard/Components/Selection/DragPreview.tsx
+++ b/packages/app/src/app/pages/NewDashboard/Components/Selection/DragPreview.tsx
@@ -151,24 +151,26 @@ export const DragPreview: React.FC<DragPreviewProps> = React.memo(
               stiffness: 1000,
             }}
           >
-            <div
-              css={css({
-                position: 'fixed',
-                top: '-1rem',
-                right: '-1rem',
-                zIndex: 20,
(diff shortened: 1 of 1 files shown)


DannyRuchtie:
👍

CompuIves:
Ah, an interesting one. I do like the idea that it implies that you can select multiple. Also, I think there's some information still given with the 1: maybe some people don't realize that they're dragging more when you see the stack.

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4582/commits/ad27b890e8ac9b8a6f10d07a4cff8aa25cb520ce is at https://pr4582.build.csb.dev/s/new.

siddharthkp:
If it isn't intentional, let's go for no-counter then

DannyRuchtie:
@siddharthkp dragging on the home tab. 
This problem is on production as well and is happening when you expediently select more than just the cards (in this I also selected the title).

- Webkit
  

Screenshot 2020-07-27 at 10 29 09

siddharthkp:
Approved

PR #4585, opened 2020-07-21
Create Sandbox Move Modal 

CompuIves:
This adds a sandbox move modal, which works cross workspaces:

image

Some details:
* Whenever you move the sandbox between teams, we refetch current sandbox info and change current active team to this sandbox if you're in the editor
* This uses the new modal logic in Overmind, which allows us to pass params. Because of this we can give a list of sandbox ids
* I've also added this to the dashboard (context menu item)
image

Now the dirty part. I've re-used some components from the old dashboard for this. I copied them over, so we can still delete the old dashboard, but using these components gave me some flexibility that now makes it possible to:

* Create folders
* Move folders
* Rename folders
* Delete folders

within this modal as well. I found myself often doing this as I was using the modal (for example, creating a "Debugging" folder while moving the sandbox), so I thought it would be valuable to still have this. The current state of overmind isn't really suited for this though, that's why I opted for re-using logic. I think that in the future we can fully move it. 

I've also:

* Moved the team switcher to a generic component, so we can use it in the modal and in the dashboard


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4585/commits/64d052ff6dbece4ac68161a89c3052b4e5a4faf4 is at https://pr4585.build.csb.dev/s/new.

CompuIves:
Now with passing ts 😄 . Forgot to regenerate the GQL typings.

PR #4605, opened 2020-07-27
add forked indication to name of sandbox

christianalfoni:
During the stream I noticed I was working on a fork of my original project... but was really no indication. I asked on the stream and got immediate response from an other person having the same issue. With this fix we add `(FORKED)` to the title of the sandbox you fork. This has two benefits:

1. You know by looking at the name if it is a forked sandbox
2. When you now fork a template the name if your sandbox will be the name of the template + `(FORKED)`

For example when you now fork the React template you will see this, instead of a cryptic name:

Screenshot 2020-07-27 at 11 20 02

### DIFF
--- a/packages/app/src/app/overmind/namespaces/editor/internalActions.ts
+++ b/packages/app/src/app/overmind/namespaces/editor/internalActions.ts
@@ -428,7 +428,7 @@ export const forkSandbox: AsyncAction<{
 
     const forkedSandbox = await effects.api.forkSandbox(id, {
       ...(sandbox.title && !sandbox.title.includes('(Forked)')
-        ? { title: sandbox.title + ' (Forked)' }
+        ? { title: `${sandbox.title} (Forked)` }
         : {}),
       ...usedBody,
     });
--- a/packages/app/src/app/overmind/namespaces/editor/internalActions.ts
+++ b/packages/app/src/app/overmind/namespaces/editor/internalActions.ts
@@ -427,7 +427,7 @@ export const forkSandbox: AsyncAction<{
     }
 
     const forkedSandbox = await effects.api.forkSandbox(id, {
-      ...(sandbox.title && !sandbox.title.includes('(FORKED)')
+      ...(sandbox.title && !sandbox.title.includes('(Forked)')
         ? { title: sandbox.title + ' (Forked)' }
         : {}),
       ...usedBody,
--- a/packages/app/src/app/overmind/namespaces/editor/internalActions.ts
+++ b/packages/app/src/app/overmind/namespaces/editor/internalActions.ts
@@ -428,7 +428,7 @@ export const forkSandbox: AsyncAction<{
 
     const forkedSandbox = await effects.api.forkSandbox(id, {
       ...(sandbox.title && !sandbox.title.includes('(FORKED)')
-        ? { title: sandbox.title + ' (FORKED)' }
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4605/commits/5af955af3b442d1e0a4e8e580a2ab26b720418ca is at https://pr4605.build.csb.dev/s/new.

packages/app/src/app/overmind/namespaces/editor/internalActions.ts
```diff
@@ -426,7 +426,12 @@ export const forkSandbox: AsyncAction<{
       }
     }
 
-    const forkedSandbox = await effects.api.forkSandbox(id, usedBody);
+    const forkedSandbox = await effects.api.forkSandbox(id, {
+      ...(sandbox.title && !sandbox.title.includes('(FORKED)')
+        ? { title: sandbox.title + ' (FORKED)' }
```
siddharthkp, line 431: Love the idea!

siddharthkp, line 431: Love the idea! Can we make it capitalised. All caps in my head is like shouting :P

```suggestion
        ? { title: sandbox.title + ' (Forked)' }
```

packages/app/src/app/overmind/namespaces/editor/internalActions.ts
```diff
@@ -426,7 +426,12 @@ export const forkSandbox: AsyncAction<{
       }
     }
 
-    const forkedSandbox = await effects.api.forkSandbox(id, usedBody);
+    const forkedSandbox = await effects.api.forkSandbox(id, {
+      ...(sandbox.title && !sandbox.title.includes('(FORKED)')
```
MichaelDeBoey, line 430:
```suggestion
      ...(sandbox.title && !sandbox.title.includes('(Forked)')
```

packages/app/src/app/overmind/namespaces/editor/internalActions.ts
```diff
@@ -426,7 +426,12 @@ export const forkSandbox: AsyncAction<{
...
+        ? { title: sandbox.title + ' (Forked)' }
```
MichaelDeBoey, line 431:
```suggestion
        ? { title: `${sandbox.title} (Forked)` }
```

christianalfoni:
Thanks @MichaelDeBoey 😄

MichaelDeBoey:
Always happy to help out! 🙂

CompuIves:
Hah, great that you noticed this on stream!! We should probably change this logic on the server, there is already some logic there related to names. It generates a new name if it's a fork from a template, and copies the name of the fork is from a sandbox.

About the title, I was thinking as alternative that we could add a badge saying that it's forked (similar to Github, where you can click on it to go to the parent),  what do you think about that? I like both approaches.

christianalfoni:
@CompuIves I agree with your changes, but think we should add them to the backlog as a project, cause it will require design and discussions. This fix is very straight forward and is an improvement until the next iteration 😄

CompuIves:
Agree, but then I would still rather have this be done on the server than on the client. By doing it on the client we override the server logic of generating a name based on template/sandbox type.

christianalfoni:
Assigning to server side

PR #4606, opened 2020-07-27
do not show privacy buttons if not pro

SaraVieira:
dashboard fix


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4606/commits/4c5363bfdd8cc6666547a2704018b5d0ac835668 is at https://pr4606.build.csb.dev/s/new.

lbogdan:
Shouldn't we instead disable the buttons and show a tooltip that this is only available for PRO users?

christianalfoni:
Yeah, that sounds like an improvement @lbogdan 👍  what you think @DannyRuchtie ?

PR #4636, opened 2020-08-03
Enable run on click by default for Android and IOS

CompuIves:
Sometimes Safari on iPhone would crash if there were multiple embeds on the same page. I think the same could happen on Android. With this we add a default, which prevents crashes if multiple embeds are on the same page.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4636/commits/86fa20f554357da1a5dd2f930b3cd3dd5a655a32 is at https://pr4636.build.csb.dev/s/new.

PR #4657, opened 2020-08-07
Dashboard: Fix bugs reported by andras

siddharthkp:
1. Don't show `Create new sandbox` in menu on All sandboxes screen
2. Turn renaming off when dragging

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Components/Selection/ContextMenus/ContainerMenu.tsx
+++ b/packages/app/src/app/pages/Dashboard/Components/Selection/ContextMenus/ContainerMenu.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { useLocation } from 'react-router-dom';
 import { Menu } from '@codesandbox/components';
 import { Context, MenuItem } from '../ContextMenu';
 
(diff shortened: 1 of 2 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4657/commits/37025d6bd87bc58da886f2bcd20249527030ea7e is at https://pr4657.build.csb.dev/s/new.

CompuIves:
@siddharthkp I think for the removing of the "Create new sandbox" there might be a better solution, since for team workspaces you should be able to create a sandbox in "All sandboxes". If I understand correctly we did already disable the button for personal workspaces here: https://github.com/codesandbox/codesandbox-client/blob/master/packages/app/src/app/pages/Dashboard/Content/routes/All/index.tsx#L78-L86 since `currentCollection` is `null`. I think the real bug is on the `MenuItem`, we have to check if `createNewSandbox` is null here: https://github.com/codesandbox/codesandbox-client/blob/add/sandpack-tar-support/packages/app/src/app/pages/Dashboard/Components/Selection/ContextMenus/ContainerMenu.tsx#L23-L25 and not render if it's `null`. 

```tsx
      {createNewSandbox && <MenuItem onSelect={() => createNewSandbox()}>
        Create new sandbox
      </MenuItem>}
```

We also probably need to type it correctly here: https://github.com/codesandbox/codesandbox-client/blob/add/sandpack-tar-support/packages/app/src/app/pages/Dashboard/Components/Selection/ContextMenus/ContainerMenu.tsx#L5-L9 by changing it to be

```ts
interface ContainerMenuProps {
  createNewFolder: () => void;
  createNewSandbox: null | () => void;
}
```

PR #4697, opened 2020-08-17
Don't set sandbox HTML if there's an HTML module

CompuIves:
Whenever a CRA without React dependency requires an HTML file, which and that one requires a JS file, top-level event listeners don't catch on.

That's because the compiler sets `body.innerHTML` after the JS has been loaded. This PR is a quick fix, but I will investigate tomorrow if this is the best fix, and if it fixes all cases.

Refer: https://codesandbox.io/s/vibrant-herschel-puilg?file=/public/index.html

### DIFF
--- a/packages/app/src/sandbox/compile.ts
+++ b/packages/app/src/sandbox/compile.ts
@@ -597,11 +597,7 @@ async function compile({
           manager.clearCompiledCache();
         }
 
-        if (
-          !manager.preset.htmlDisabled ||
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4697/commits/60a5dacd402b733d33f68ce2ba35938934fa809f is at https://pr4697.build.csb.dev/s/new.

PR #4764, opened 2020-08-25
Don't render a tooltip if the label is null

CompuIves:
I wonder what you think of this @siddharthkp. With this change we don't render the tooltip if `label` is `null`.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4764/commits/f9764106474e52ee56e23fa5e8965647cb6c1fe3 is at https://pr4764.build.csb.dev/s/new.

siddharthkp:
Yep, seems to be fine choice in our context

siddharthkp (approved):

PR #4783, opened 2020-08-28
Fix #4750

### LINKED ISSUE #4750: POST method used for loading websites fails, cannot load non-index HTML pages
# 🐛 bug report

## Description of the problem

I have multiple static sandboxes that I store websites on. When I attempted to view my webpages, it tried and succeeded in loading a non-index site (here https://wg0pq.csb.app/creator.html), but after a second it will fail and load data from the index page instead. It appears to load the head of the desired HTML page but the body of the index page. I am currently using Chrome running Chrome OS Version 81.0.4044.141 but I also tested it on a Windows 10 PC using Edge to get the same effects. 

Checking using the developer mode, I saw that a POST fetch command failed to load, which I am guessing is the underlying issue here. Specificially, it tries to connect to https://col.csbops.io/data/sandpack. 

Because of this, my static websites that previously worked are basically completely broken now. However, this has only started recently - around August 9 it was clearly working as intended. None of my code has changed since then (except for after I noticed this and tried some debugging), so some client change must have introduced this.

### Link to sandbox: [link](https://wg0pq.csb.app/engine.html?blueDeck=&redDeck=&version=current&order=true&maxElixir=10&elixirR=1) (optional)

It will attempt to load a Processing.js program, succeed for about a second, and then load an entirely different page - the index page with incorrect styling.

### Your Environment

| Сodesandbox      | 0d15157e6 |
| Browser          | Chrome |
| Operating System | Chrome OS Ver. 81.0.4044.141 |

(Again, also fails on current Windows 10 with Edge.)

CompuIves:


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4783/commits/a518ed8dbcf8a61bd04d8c28e5666e42a9984f09 is at https://pr4783.build.csb.dev/s/new.

PR #4790, opened 2020-08-30
Merge user initialization into one function

CompuIves:
I noticed after sign in we wouldn't show all teams, and that's because we have a flow that is duplicated, and we only added it to one flow. I merged both flows into one function now.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4790/commits/1c78b7b3e049145505b38c2cc97208bfac84f4df is at https://pr4790.build.csb.dev/s/new.

Issue #4523, opened 2020-07-09
Links with src="#target" do not trigger 'click' event


cyberbiont:
# 🐛 bug report

Just faced this very strange behaviour.

A link, which src attribute is set to '#void" or "#0" or something alike (that I usually do to disable links) does not trigger click event.
A link with src="#" does.

Sample sandbox:
https://codesandbox.io/s/void-event-listener-bug-udnit?file=/src/index.js

On Codepen the example works fine
https://codepen.io/cyberbiont/pen/PoZejJr?editors=1111

oskarabc:
Hello @cyberbiont,
Not seeing errors here. Can you do a hard refresh, to upgrade to the latest version of CodeSandbox and confirm status of your issue?

cyberbiont:
Status confirmed.
Do you get message in console when you click 'doesnt work' links in my example sandbox?

CodeSandboxNotifications:
Hi,


I don't get messages in browser or CodeSandbox console upon clicking the links in https://codesandbox.io/s/void-event-listener-bug-udnit?file=/src/index.js


I've shared this with one of our devs. It's probably due to our runtime design choices that may differ from codepen.


–Oskar



--- original message ---
On July 25, 2020, 1:27 PM GMT+2 notifications@github.com wrote:



Status confirmed.

Do you get message in console when you click 'doesnt work' links in my example sandbox?

—

You are receiving this because you are subscribed to this thread.

Reply to this email directly, view it on GitHub, or unsubscribe.
--- end of original message ---

christianalfoni:
Hi there @cyberbiont and so sorry for late response!

This does seem like a bug and we have a pending PR here: https://github.com/codesandbox/codesandbox-client/pull/4809

Thanks for the examples!

PR #4809, opened 2020-09-03
should still propagate hyperlink clicks

christianalfoni:
I do not think we should `stopPropagation` here, cause this stops the event bubbling up. It is the `preventDefault` we want to prevent the browser to trigger the link.

Let me know if there is some reason we want to actually stop the click event.

Fixes: https://github.com/codesandbox/codesandbox-client/issues/4523

### DIFF
--- a/packages/sandbox-hooks/url-listeners.js
+++ b/packages/sandbox-hooks/url-listeners.js
@@ -26,8 +26,8 @@ function pathWithHash(location) {
 }
 
 function isInsideVue(el) {
-  if (el === document.body) return false
-  return el.__vue__ || el._vnode || isInsideVue(el.parentElement)
+  if (el === document.body) return false;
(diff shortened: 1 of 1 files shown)


SaraVieira (approved):
10/10

would merge again

CompuIves:
Nice!! @lbogdan does this make sense to you as well? I think you built this looong ago 😄

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4809/commits/7c06c6cee3f4912f92c3109125e3f063d7f48c3f is at https://pr4809.build.csb.dev/s/new.

lbogdan:
I'm pretty sure that if I explicitly added them, they're both needed (at least in some use-cases), I just can't remember why. But if you tested it and it still works with react / vue / etc. using both history and hash, then I guess it's fine. 🙂

PR #4838, opened 2020-09-04
remove uned text

SaraVieira:
Related to #2246


CompuIves:
A bit tricky, sometimes you do need to refresh the preview when changing this html, for example when you change things in the head tag. The reason for this is that we update it on the server, and then the preview needs to fetch it from the server by refreshing. In some cases we do update the preview ourself, we do that when the body changes by doing document.body.innerHTML =....

SaraVieira:
Makes sense! Closing it :)

PR #4854, opened 2020-09-08
Fix loading prebundled dependencies that have {error: string}

### LINKED ISSUE #4853: Cannot read property 'name' of undefined when installing dependencies
# 🐛 bug report

## Preflight Checklist

* [x] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

## Description of the problem
I get an error when installing dependencies.

### To Reproduce
See the sandbox

CompuIves:
Fixes #4853

Unfortunately there was a bug with our packager that allowed it to save `{ error: '' }` to the cache if there was an error. This bug has been fixed in the packager, but there is still an existing cache, which is why we need to fall back to our dynamic fetching if we see that the object has an error.

### DIFF
--- a/packages/app/src/sandbox/npm/index.ts
+++ b/packages/app/src/sandbox/npm/index.ts
@@ -147,15 +147,29 @@ export async function getDependenciesFromSources(
 
     const prebundledPromise = Promise.all(
       Object.keys(prebundledDependencies).map(depName =>
-        getPrebundledDependency(depName, depsWithNodeLibs[depName]).finally(
-          () => {
+        getPrebundledDependency(depName, depsWithNodeLibs[depName])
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4854/commits/98f968b5081db1683ee16e679a0aeabe19fdf1ae is at https://pr4854.build.csb.dev/s/new.

PR #4863, opened 2020-09-08
Console Improvements

CompuIves:
3 things:

1. A big devtool slowdown for console.log spam was actually updating the indicator, this debounces the update with 50ms. So it waits 50ms before committing the actual state change, if anything comes in within those 50ms the timer resets.
2. We only show the last 500 console.log messages. An improvement, later on, would be to allow lazy loading.
3. We now don't clear the console on hot reloads from the bundler, this was inconsistent behavior anyway.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4863/commits/f8389f414b5b94c29d932aad8dd3cfffe4e98a0b is at https://pr4863.build.csb.dev/s/new.

packages/app/src/app/components/Preview/DevTools/Console/index.tsx
```diff
@@ -75,13 +85,8 @@ class ConsoleComponent extends React.Component<StyledProps> {
         break;
       }
       case 'clear-console': {
-        if (this.state.initialClear) {
```
CompuIves, line 78: I realized we don't use this anymore, and prevents first clear on the console.

packages/app/src/app/components/Preview/DevTools/Console/index.tsx
```diff
@@ -104,7 +109,7 @@ class ConsoleComponent extends React.Component<StyledProps> {
           if (aggregatedResults) {
             const { summaryMessage, failedMessages } = aggregatedResults;
             this.addMessage('log', [summaryMessage]);
-            failedMessages.forEach(t => {
+            failedMessages.forEach((t) => {
```
christianalfoni, line 112: eyh... what is `t`? ;-) a message?

packages/app/src/app/components/Preview/DevTools/Tests/index.tsx
```diff
@@ -312,10 +317,10 @@ class Tests extends React.Component<DevToolProps, State> {
 
           const files = Object.keys(this.state.files);
           const failingTests = files.filter(
-            f => this.getStatus(this.state.files[f]) === 'fail'
```
christianalfoni, line 315: What is `f`?

packages/app/src/app/components/Preview/DevTools/Tests/index.tsx
```diff
@@ -186,7 +186,12 @@ class Tests extends React.Component<DevToolProps, State> {
    * Every setState call will have to go through this, otherwise we get race conditions
    * where the underlying state has changed, but the draftState didn't change.
    */
-  setStateDebounced = (setStateFunc, time = 200) => {
+  setStateDebounced = (
```
christianalfoni, line 189: Hm, are you sure you want to debounce? Cause that means you will actually skip some state changes. Debounce will drop all calls to `setState` until 200ms passed and then it runs the last call to `setState`. I think you rather want to queue up setStates? Like you push the `setState` callback to an array and then call `setStateDebounced`, this function will iterate the array of callbacks to produce the new state.

christianalfoni (commented):
Please checkout the `debounced` thing 😄 

Nagging about variable names.

Have not actually tested this, just looking at code

packages/app/src/app/components/Preview/DevTools/Console/index.tsx
CompuIves: Ahh, this turned up in the diff because of prettier 😂 . I can find a better name.

packages/app/src/app/components/Preview/DevTools/Tests/index.tsx
CompuIves: Also in the diff because of prettier 😭

packages/app/src/app/components/Preview/DevTools/Tests/index.tsx
CompuIves: Yeah in this case it does debounce it, but it keeps the result of setState as `draftState`, it only waits with committing. So you can call setState 5 times, every time on the previous version, but only after 200ms we commit that setState. Does that make sense?

CompuIves: This is the place where we do that, we save the draftState and wait debounced before committing it.

https://github.com/codesandbox/codesandbox-client/pull/4863/files#diff-250f4bac0530862bcebc35bff56abcb4R195-R201

Issue #1980, opened 2019-05-27
Support other Prettier config files such as prettier.config.js?


karlhorky:
# 🌈 Feature

What is the position on supporting other forms of Prettier config files such as `prettier.config.js`?

Other formats as per [the Prettier docs](https://prettier.io/docs/en/configuration.html):

> - A .prettierrc file, written in YAML or JSON, with optional extensions: .yaml/.yml/.json.
> - A .prettierrc.toml file, written in TOML (the .toml extension is required).
> - A prettier.config.js or .prettierrc.js file that exports an object.
> - A "prettier" key in your package.json file.

karlhorky:
I guess the position is, we don't support the others?

karlhorky:
not stale

karlhorky:
👋

karlhorky:
bot

matthewoates:
Still appears to be an issue with jest: https://github.com/codesandbox/codesandbox-client/issues/1317

PR #4874, opened 2020-09-10
Support all prettier config files

pablopunk:

Feature

package.json `"prettier"` config is ignored

## What is the new behavior?

Prettier config will be, in order of precedence:

* `"prettier"` key in `package.json`, if it exists
* `prettierrc` file in the root of the sandbox
* sandbox prettier config (via preferences)
* default prettier config

Tested all of the above scenarios

## Checklist

- [x] Documentation
- [x] Testing 
- [x] Ready to be merged
      
This addresses only part of issue #1980. I didn't implement [all the possible configuration files](https://prettier.io/docs/en/configuration.html#basic-configuration) just to keep it simple.

Another thing I wanted to mention, I notice there's another function to get the prettier config [here](https://github.com/codesandbox/codesandbox-client/blob/3c2f6ce054108032c440d28a462660ec3f2dfc10/packages/app/src/app/overmind/onInitialize.ts#L81) but I didn't manage to see where's being used. Is it being used at all?


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4874/commits/aceaa54a29ab7b1e964f347049e442c742561f3e is at https://pr4874.build.csb.dev/s/new.

pablopunk:
@SaraVieira sure, it will require to parse YAML and TOML and a few more checks for the files existing, but it is possible. I can do them all if that's what we want

SaraVieira:
We do have a toml one installed already. its called `markty-toml` and it's used for netlifly configs here: https://github.com/codesandbox/codesandbox-client/blob/b4937ea7bb9fedc7dc52b36444cff52dab018d6c/packages/app/src/app/utils/getNetlifyConfig.js#L2

We do not have a YAML one so I am okay with letting that one go and also the TOML one if it's too hard, it's mostly just for the names you can give the prettier config files :)

SaraVieira:
Also ping me if you need any help

MichaelDeBoey:
@pablopunk Maybe there's some [`cosmiconfig`](https://github.com/davidtheclark/cosmiconfig) logic we can (re-)use? 🤔
It's used by a lot of libraries to get the config of all the different supported files.

pablopunk:
@MichaelDeBoey I did take a look at that but they look for the files in the filesystem (with node)

pablopunk:
@SaraVieira I wanna resolve a module from a path (get the value of `module.exports`) [right here](https://github.com/codesandbox/codesandbox-client/pull/4874/files#diff-d8046a7427efacf13fe3e293f545981ffbadf634aac2312cc5308285bd4b79cbR86-R87), what's the convention on codesandbox? I know it uses `eval`, can I see an example?

SaraVieira:
Hey!

Gonna ping @CompuIves as he might be able to help you better there

PR #4892, opened 2020-09-15
Fix permissions on personal workspace sandboxes

CompuIves:
The server still returns `teamId === null` on sandboxes in a personal workspace for backwards compatibility reasons. That breaks our check on workspace, and makes it seem like the user has no access to the sandbox.

Here we check for `teamId === null` to also make sure that you're an owner. Because the server only responds with `teamId = null` if it's in *your* personal workspace.


siddharthkp (approved):

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4892/commits/3058c603d61728dd85f9692a11073e05f3ee0355 is at https://pr4892.build.csb.dev/s/new.

PR #4931, opened 2020-09-24
fix(dependencies): show latest version on Dependency

jyash97:
Bug fix

We have sorted versions in version selector which by defaults shows the newest release [ first value ] in most cases. But when we select and add the dependency we push the latest tag in API. Check line here: https://github.com/codesandbox/codesandbox-client/blob/master/packages/app/src/app/pages/Sandbox/SearchDependencies/index.tsx#L20

https://www.loom.com/share/fc794bba8ee2403c8cde304c806fdc23

## What is the new behavior?
Show the latest version for the dependency by default. 
https://www.loom.com/share/bc0500fa4507436da06e7698634974fd

## Checklist

- [x] Testing 
- [ ] Ready to be merged - Requires review on the new behavior

Is there any easy way to debug or visualize overmind state?  I am usually checking through hooks context value which becomes difficult to check all values in the nested object. Thanks 👋

### DIFF
--- a/packages/app/src/app/pages/Sandbox/SearchDependencies/Dependency.tsx
+++ b/packages/app/src/app/pages/Sandbox/SearchDependencies/Dependency.tsx
@@ -84,7 +84,8 @@ export const Dependency = ({ dependency }: { dependency: DependencyType }) => {
     actions,
   } = useOvermind();
 
-  const selectedVersion = hitToVersionMap[dependency.objectID];
+  const selectedVersion =
+    hitToVersionMap[dependency.objectID] || dependency.tags.latest;
 
   const versions = Object.keys(dependency.versions).sort((a, b) => {
     try {


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4931/commits/93cb1846aa5bcee54ad6b4c2e822961adc46a23f is at https://pr4931.build.csb.dev/s/new.

SaraVieira:
Thank you so much! Super nice fix!


About Overmind, it does have some nice devtools you can install: https://overmindjs.org/core/devtools, I use that every day

jyash97:
> Thank you so much! Super nice fix!

Happy to help out :)

> About Overmind, it does have some nice devtools you can install: https://overmindjs.org/core/devtools, I use that every day

Thanks will try it

PR #4952, opened 2020-09-29
Fix moving tab to top in devtools

CompuIves:
It was impossible to move bottom devtools to the top part, because of an undefined value in track. This fixes that by moving the track to where we know view is defined.

Tested moving tabs from top to bottom, other way around and in same row.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4952/commits/465b6f34c8149b89c5894e5154597472707e8efd is at https://pr4952.build.csb.dev/s/new.

christianalfoni (approved):
Looks good to me! 👍

PR #4962, opened 2020-10-01
Fix default extensions for sandpack (prefer mjs)

CompuIves:
We should prefer `mjs` over `js`. This changes that.

Fixes #4958


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4962/commits/5a90d41e32ec883350607c0f638a0c7f05ca8937 failed.

PR #4965, opened 2020-10-01
Responsive preview

SaraVieira:
Implementation done, final testing round and SHIP IT 🚀


packages/app/src/app/overmind/namespaces/preview/state.ts
```diff
@@ -0,0 +1,34 @@
...
+          Tablet: [1024, 765],
+          Desktop: [1920, 1080],
+          'Desktop HD': [1400, 800],
+        }
+      );
+    }),
+    scale: 100,
+    resolution: [320, 675],
+  },
+  mode: 'responsive',
```
SaraVieira, line 33: Change to null

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4965/commits/84dac76c6bf2dbb99ffab7716b15b0f63d053352 failed.

christianalfoni:
Comment to last optimizations and changes.

**What I did**:

1. Refactored the **Responsive Wrapper** so that it is part of the CodeSandbox application, instead of the generic **BasePreview**
2. Moved the handlers outside of the scaling container
3. Refactored how we do resizing to be a temporary state in the component
4. Removed the transition effect

**Why I did it**:
1. We had some performance issues passing a lot of state down to the **BasePreview**. Also the **BasePreview** is arguably not concerned with the responsive preview feature, it is a CodeSandbox application feature, while the **BasePreview** is a low level component that only deals with loading the preview itself
2. When scaling the preview the handlers scaled as well, becoming unmanageable as they were super tiny. Doing an "opposite transform" of the handles did not work as expected. They did not become as small, but they did not keep a consistent size. But more importantly the resizing was offset, meaning when clicking a handle and dragging the resizing did not follow the mouse
3. There was a secondary issue here not related to performance, but related to scaling. When you are in a scaled mode and try to resize the preview using the handles nothing would happen, because it would rather scale up instead of actually moving with the mouse. This is technically correct, but confusing to the user. By having an optimistic resolution in the components we are able to bypass the automatic rescaling, it rather happens when the resizing is done... creating a "popping into place" effect
4. The transition was running on the scaling container, but the handles are not part of that container and more importantly the handles are not scaled anymore. Even though we would run transitions on both containers they would not behave the same. The handles would "lag behind" as in terms of scaling they just not move the same distances as the scaled preview. I also noticed that having transitions when resizing the editor/preview split itself does not look very nice and requires additional "no-transition" logic. Chrome responsive mode does not transition either and just feel this causes way more pain than actual value

PR #4967, opened 2020-10-01
Keep imports below converted exports

CompuIves:
We hoist imports to the top, as that is how imports are treated. However, imports need to be below exports that are defined above them, this makes sure that happens.

Fixes https://github.com/tensorflow/tfjs/issues/3889


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4967/commits/38c68419c389ba7998710d050bd4fd2014c26ffb is at https://pr4967.build.csb.dev/s/new.

PR #4974, opened 2020-10-05
try to fix hash issue

### LINKED ISSUE #4857: url-listeners is blocking native behaviour
There has been several reports about clicking hyperlinks does not give the expected behaviour. We have two specific scenarios:

1. create two `.html` files and create a link to move from one to the other. When you now change the code and save it refreshes the `index.html` file instead of the one you are on. Can not see the addressbar reflecting moving between HTML files at all

2. Create hyperlinks with hashes pointing to IDs of elements on the page. The expected behaviour is that the preview window jumps to the element when you click the link. This does not happen

`app/sandbox-hooks/url-listeners.js` is the file that is responsible for intercepting URL changes and send them up to the addressbar (`sendUrlChange`). We still need this logic, but we want to re-evaluate the click listener further down in the file.

So I would suggest doing normal tracking here instead. One for `popstate` and one for `hashchange`:

https://gomakethings.com/how-to-detect-when-the-browser-url-changes-with-vanilla-js/

https://stackoverflow.com/questions/6390341/how-to-detect-if-url-has-changed-after-hash-in-javascript/37091484#37091484

christianalfoni:
Look at comment.

We have tested this and can not reproduce any issue.

closes #4857

### DIFF
--- a/packages/sandbox-hooks/url-listeners.js
+++ b/packages/sandbox-hooks/url-listeners.js
@@ -25,10 +25,12 @@ function pathWithHash(location) {
   return `${location.pathname}${location.hash}`;
 }
 
+/*
 function isInsideVue(el) {
   if (el === document.body) return false;
   return el.__vue__ || el._vnode || isInsideVue(el.parentElement);
 }
+*/
 
 export default function setupHistoryListeners() {
   function handleMessage(data, source) {
@@ -108,6 +110,11 @@ export default function setupHistoryListeners() {
     }
   });
 
+  // We do not know why this is here. It currently breaks clicking on HASH links which is expected to
+  // scroll down to element id. It seems to be related to when clicking hashes with no content
+  // "/#" or "#" or "foo/#". We are commenting it out to see what breaks and will
+  // make sure to write a comment if anything else breaks, when fixing it
+  /*
   document.addEventListener(
     'click',
     ev => {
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4974/commits/73c5a2278d94e6cd366ec3797a25a805c3009fc5 is at https://pr4974.build.csb.dev/s/new.

CompuIves:
Curious what @lbogdan thinks, he built this feature

lbogdan:
That's to make framework router libraries (React Router, `vue-router` etc.) work properly in hash mode.

christianalfoni:
Oh, I did this on the assumption you said you did not remember last time. But it was a by passing question, so I was not explicit enough about the issue 😄 

But no matter, we have to find a different approach, because this is breaking normal hash links.

But great... lets test React router, Vue router etc. again. Maybe they have changed how they work as well. This seems to be an old issue.

From the logic I understand this is only related to urls ending with `#`  (`substr(-1) === '#')`, which is a weird url to create. So we will test that explicitly

christianalfoni:
@SaraVieira I have tested react-router, and can not find any issues with normal routing or hashes there. Could you try Vue?

lbogdan:
The condition is

```js
        el.href.indexOf('#') !== -1 &&
        el.href.substr(-1) !== '#'
```

so URLs that contain `#` but **don't** end in `#`. 🙂

lbogdan:
@christianalfoni Did you test that back / forward buttons in the preview navigation bar work?

christianalfoni:
> The condition is
> 
> ```js
>         el.href.indexOf('#') !== -1 &&
>         el.href.substr(-1) !== '#'
> ```
> 
> so URLs that contain `#` but **don't** end in `#`. 🙂

DOH 😂 

But yeah, tested back and forward, works with react-router at least

lbogdan:
For more context, you can read https://github.com/codesandbox/codesandbox-client/pull/198#issuecomment-332522561 and https://github.com/codesandbox/codesandbox-client/pull/189#issuecomment-327727829 .

christianalfoni:
https://github.com/codesandbox/codesandbox-client/pull/4977

PR #4977, opened 2020-10-06
Vue router fix

christianalfoni:
Okay, 

So there is an issue with Vue not using `HashChangeEvent`, but rather `popState`: https://github.com/vuejs/vue-router/issues/1807. So this is actually not working in production now.

That means we have three approaches to solve this:

1. We detect if Vue is running
2. We detect if nothing is using a hash listener
3. We fire off both HashChange and popState in our `go` method

So my reasoning:

1. Having framework specific code in something as generic as `url-listeners` is not ideal
2. We would have to hack the browser API
3. If it works... most straightforward and least intrusive

Since router libraries will only listen to one of the events, I implemented 3. Tested with `vue-router` and `react-router-dom`. 

We do not need the intercepting click here anymore, which fixes the original issue.

### DIFF
--- a/packages/app/src/app/overmind/effects/vscode/LinterWorker/index.js
+++ b/packages/app/src/app/overmind/effects/vscode/LinterWorker/index.js
@@ -71,6 +71,8 @@ const allRules = {
     .default,
   '@typescript-eslint/no-useless-constructor': require('@typescript-eslint/eslint-plugin/dist/rules/no-useless-constructor')
     .default,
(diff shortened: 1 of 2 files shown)


SaraVieira (approved):

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4977/commits/312af5b51f211365357b5b49b186cb121e8db5d4 is at https://pr4977.build.csb.dev/s/new.

christianalfoni:
Got review approval from @lbogdan on Slack

PR #4978, opened 2020-10-06
Fix collaborator heads not updating

CompuIves:
Fix a case where the collaborator heads don't update because `liveRoomInfo` doesn't update but `liveRoomInfo.users` does.

image

### DIFF
--- a/packages/app/src/app/pages/Sandbox/Editor/Header/CollaboratorHeads/CollaboratorHeads.tsx
+++ b/packages/app/src/app/pages/Sandbox/Editor/Header/CollaboratorHeads/CollaboratorHeads.tsx
@@ -105,10 +105,7 @@ const CollaboratorHead = (props: ICollaboratorHeadProps) => (
 
 export const CollaboratorHeads: FunctionComponent = () => {
   const { state, actions } = useOvermind();
-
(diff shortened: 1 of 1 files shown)


SaraVieira (approved):
Weird bug

memo should count that as a change

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4978/commits/f50def4916b4bfaccd9faaf23dc6565b5a1f982f is at https://pr4978.build.csb.dev/s/new.

CompuIves:
Yeah hmm I'm not sure what happens there, because it seems like the users list actually changes. Do you know how this could happen @christianalfoni?

christianalfoni:
Yeah, `roomInfo` is still the same object. Also the `roomInfo.users` stays the same array as new people are added :)

But do not see any reason to use `memo` here?

But yeah... moving from the "immutable" world of Redux to Overmind takes some time to get used to. Cause whenever we add something to a list, or change a property on an object, the array/object itself does not change. This is what makes it so performant, but it does not mix well with Reacts helpers as they expect an "immutable env".

CompuIves:
Ahh, yeah that makes sense. Merci!

PR #4990, opened 2020-10-08
Don't spam the server when CodeSandbox returns 503

CompuIves:
This makes sure that we don't spam the server when CodeSandbox returns a 503. It also adds more variability in connecting so not all users reconnect exactly at the same time.

### DIFF
--- a/packages/app/src/app/overmind/effects/live/index.ts
+++ b/packages/app/src/app/overmind/effects/live/index.ts
@@ -156,13 +156,16 @@ class Live {
       // An offset between 1 and 1.7;
       const defaultReconnectOffset = 1 + Math.random() * 0.7;
       let isServerDown = false;
-      const reconnectAfterMsFunc = (tries: number) => {
+      const reconnectAfterMs = (tries: number) => {
         if (isServerDown) {
           return 3000 + Math.random() * 2000;
         }
 
+        // Based on the times tried we slowly increase the reconnect timeout, so first time
+        // we try to reconnect in 10ms, second time in 50ms, third time in 100ms, fourth time in 150ms,
+        // etc...
         return (
-          [10, 50, 100, 150, 200, 250, 500, 1e3, 2e3][tries - 1] *
+          [10, 50, 100, 150, 200, 250, 500, 1000, 2000][tries - 1] *
           defaultReconnectOffset
         );
       };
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/4990/commits/91d48989a5e0ddddb34626347b06f1a3b373d0f3 is at https://pr4990.build.csb.dev/s/new.

packages/app/src/app/overmind/effects/live/index.ts
```diff
@@ -153,8 +153,24 @@ class Live {
...
+      // An offset between 1 and 1.7;
+      const defaultReconnectOffset = 1 + Math.random() * 0.7;
+      let isServerDown = false;
+      const reconnectAfterMsFunc = (tries: number) => {
+        if (isServerDown) {
+          return 3000 + Math.random() * 2000;
+        }
+
+        return (
+          [10, 50, 100, 150, 200, 250, 500, 1e3, 2e3][tries - 1] *
```
christianalfoni, line 165: A comment on this black magic would be nice ;-)

packages/app/src/app/overmind/effects/live/index.ts
```diff
@@ -153,8 +153,24 @@ class Live {
...
+          defaultReconnectOffset
+        );
+      };
+
       this.socket = new Socket(`${protocol}://${location.host}/socket`, {
         params,
+        // @ts-expect-error Wrong typings
+        reconnectAfterMs: reconnectAfterMsFunc,
```
christianalfoni, line 173: No need to add `Func` at the end?

packages/app/src/app/overmind/effects/live/index.ts
```diff
@@ -171,6 +187,10 @@ class Live {
             // If we can't get a jwt because we're unauthorized, disconnect...
             this.socket.disconnect();
             tries = 0;
+          } else if (error.response?.status === 503) {
+            isServerDown = true;
+          } else {
+            isServerDown = false;
```
christianalfoni, line 193: I would rather set `isServerDown = false` somewhere else. Is there a "connected" event or similar? It seems risky to only say server is up again based on errors?

christianalfoni (commented):
Generally looks good, just some questions and suggestions :)

CompuIves:
Ahh, great review! I'll incorporate it today.

packages/app/src/app/overmind/effects/live/index.ts
CompuIves: Yep I think I should make it part of a separate if statement. I do think this `catch` is the place, because this reconnect happens when the server disconnects (eg. goes down), if the server returns a 503 on reconnect it means that the server is down (since we return that status code if we can't reach the server from our proxy). So this is our way to ensure that we don't flood on the next try.

PR #4998, opened 2020-10-12
fix(profile): redirect to NotFound page when profile request fails

arthurdenner:
Bug fix.

When a user goes to `/u/:username` and `username` doesn't exist, they get a blank page.

## What is the new behavior?
We show the `NotFound` component to the user when `getProfile` fails - regardless of the reason.

1. Visit `/u/aarthurdennerr` for example - [production link](https://codesandbox.io/u/aarthurdennerr);
1. Verify that we show the `NotFound` component instead of a blank page;
1. Visit `/u/arthurdenner` for example - [production link](https://codesandbox.io/u/arthurdenner);
1. Verify that we show that the user doesn't have a showcased sandbox yet.

## Checklist

- [x] Documentation
- [x] Testing 
- [x] Ready to be merged
      
## Notes

- I've added two different `try/catch` because if `getProfile` fails, we show the `NotFound`, but if the `getSandbox` fails, we want to show the profile still. Are you okay with this behaviour or should we change it?
- I've changed the text when the user doesn't have a showcased sandbox because they may have sandboxes, but none of them is showcased only. Let me know if you prefer to not change this text.

Last, if you enjoy this PR, I'd appreciate it if you can label it for Hacktoberfest with `hacktoberfest-accepted` 😄

### DIFF
--- a/packages/app/src/app/overmind/namespaces/profile/actions.ts
+++ b/packages/app/src/app/overmind/namespaces/profile/actions.ts
@@ -7,7 +7,15 @@ export const profileMounted: AsyncAction<string> = withLoadApp(
     state.profile.isLoadingProfile = true;
     state.profile.notFound = false;
 
-    const profile = await effects.api.getProfile(username);
+    let profile: Profile;
+
+    try {
+      profile = await effects.api.getProfile(username);
+    } catch (error) {
+      state.profile.isLoadingProfile = false;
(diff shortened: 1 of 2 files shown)


SaraVieira:
Hey!

Thank you so much for this PR! Super nice and sweet!

Added the label and gonna merge it 🥳 

again, thank you!

PR #5003, opened 2020-10-13
Remove outdated session expired error message

CompuIves:
This cannot be the case anymore, as the server lets the client know when a user is not signed in anymore using a cookie.

### DIFF
--- a/packages/app/src/app/overmind/internalActions.ts
+++ b/packages/app/src/app/overmind/internalActions.ts
@@ -1,4 +1,3 @@
-import { identify } from '@codesandbox/common/lib/utils/analytics';
 import {
   ModuleTab,
   NotificationButton,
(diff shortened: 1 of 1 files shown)


christianalfoni:
But does that show a notification? I would expect expired cookie to cause `401`s , where we want to notify user? 😄

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5003/commits/7c00e805878f945d0965fd03e6aeb8bb4d00e363 is at https://pr5003.build.csb.dev/s/new.

CompuIves:
Right! But the client checks first if a cookie is set or not, if it's not set, it will not try to do a request to the server. There are still some calls where the server returns a 401, but these are valid "Unauthorized" calls (for example, when a user with only read permission in a workspace is trying to fork a sandbox to that workspace). We currently mask the error message.

christianalfoni (approved):
Looks good to me! 😄

PR #5014, opened 2020-10-15
Preview comments

christianalfoni:
WIP

This is branched off of "responsive-preview", so number of commits will reduce a lot when that is merged into master.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5014/commits/b1e74661e899e34567a403704742849b4a34ce2c is at https://pr5014.build.csb.dev/s/new.

christianalfoni:
Due to imminent baby and our daughter getting a fever there is high risk I will not be able to finish this. So let me give a summary of the project state:

**Inline Image Feature**
The ability to drop images in the comment text is implemented, where the image is uploaded when the comment is sent to the server. The backend is implemented and tested for this, though there seems to be a couple of issues on the client. Most noticeably when an image is dropped we resize it. Currently it seems to be cropping instead of resizing. Logic is here:

https://github.com/codesandbox/codesandbox-client/pull/5014/files#diff-c1a54ecd9f89704c9e0eafd46f0562d7f527c97d57f67e94de1cd490792ba875

No battle testing done here yet.


**Preview Comment Feature**
We had a session, which was recorded (Check with Bas), where the general feedback supported that we want to produce an actual screenshot of the preview when creating the comment. The reason being that we want to capture as much state as possible from the preview and a screenshot is the only medium that can do that (recording explains more). We will use **html2canvas** as the default implementation, where our CodeSandbox Extension can later produce an actual screenshot. 

There are several parts to this feature where:
https://github.com/codesandbox/codesandbox-client/pull/5014/files#diff-2537a68f09a5bf66b64eed10f3d78ad6d32e6bab4794ebf33054ebe9355ed1c7 is responsible for grabbing a screenshot when CodeSandbox asks for it. This file is imported at https://github.com/codesandbox/codesandbox-client/pull/5014/files#diff-cfde6ea8b3feb90b9d05ed0e2c664f1b523583111b1fc479a736da946817e2da. **Note** that our container and static sandboxes also needs to import this file, ask @lbogdan or @CompuIves about this.

The screenshot covers the actual preview here: https://github.com/codesandbox/codesandbox-client/pull/5014/files#diff-f3c9a7dd8a4ade8fec7825864c76c7a5a21e471db5b87c148af6ac6c209a2ac2 . Where the user can click some part of the preview to produce the comment. **Note** that the full body height will probably be captured now, so need to deal with this somehow. Maybe we can capture `document` instead of `document.body` for example, as HTML2Canvas deals with scroll positions.

The screenshot is passed into the **createOptimisticPreviewComment** action which currently just adds it to the comment. Here we need to add the following logic:

- Draw the comment bubble SVG into the image by using canvas. The `x` and `y` position is also passed into the action
- Consider cropping the image based on where the comment bubble is located

This should be implemented in for example the `browser` effect.

Next the comment dialog, https://github.com/codesandbox/codesandbox-client/pull/5014/files#diff-fcb770ec114c1ad832461f8ca4b4e1758fee993b3314df981b68b73be7f5e70c, is where the screenshot is displayed. This is currently just an `img` tag, which was used for testing. This needs to work more like images in general on comments, so that it can be opened by clicking it etc.

packages/app/config/webpack.prod.js
```diff
@@ -354,7 +354,7 @@ module.exports = merge(commonConfig, {
     new CopyWebpackPlugin([
       {
         from: '../sse-hooks/dist',
-        to: 'public/sse-hooks',
+        to: 'public/[name].[hash].[ext]',
```
CompuIves, line 357: It should probably be in the sse-hooks folder?

SaraVieira: added

CleanShot 2020-11-12 at 15 22 15

CompuIves: I think it should be a single entry probably, the first one.

SaraVieira: I don't really know why he added this but I would say putting both will still work and we ask when he comes back

CompuIves: We wanted to add it to make the url unique for `sse-hooks`, I think if we have the second entry, `sse-hooks` in the file manifest here: https://codesandbox.io/file-manifest.json will maybe still link to the old one (without hash). So removing the second one should make everything work. I believe it should be something like


```
{
  from: '../sse-hooks/dist',
  to: 'public/sse-hooks/[name].[hash].[ext]'
}
```

PR #5018, opened 2020-10-15
Dashboard: Lazily update stale sandbox thumbnails

siddharthkp:
Moved Sandbox Thumbnail into a component.

Sandbox thumbnail logic:

0. Use template icon as starting point and fallback
1. Use `sandbox.screenshotUrl` if it can be successfully loaded (might not exist)
2. If `sandbox.screenshotOutdated`, lazily load latest screenshot. Switch when image is loaded.

FAQ: Why don't we directly load the latest screenshot?

Generating the latest screenshot takes a few seconds, so it's faster to use the cached `screenshotUrl` and then lazily fetch the latest one.

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Components/Sandbox/SandboxCard.tsx
+++ b/packages/app/src/app/pages/Dashboard/Components/Sandbox/SandboxCard.tsx
@@ -208,39 +208,12 @@ export const SandboxCard = ({
   ...props
 }: SandboxItemComponentProps) => {
   const [stoppedScrolling, setStoppedScrolling] = React.useState(false);
-  const [guaranteedScreenshotUrl, setGuaranteedScreenshotUrl] = React.useState<
-    string
-  >(screenshotUrl);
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5018/commits/b1a6c8e39f82c2fbc9bc4371fc93af42380b83a6 is at https://pr5018.build.csb.dev/s/new.

CompuIves (approved):
This is really great!! Code is much more understandable than what it was before, and testing shows really nice results. I think I have another improvement we can do with this.

On the `useImageLoaded` we can return in the effect with

```
return () => {
  img.src = ''
}
```
This will automatically cancel loading of the image if you scroll away, so that you can save bandwidth for loading the new images that are in view.

andrasbacsai (approved):

siddharthkp:
Good one! Added a cleanup function for effect

PR #5020, opened 2020-10-16
Make the client cache a jwt token to prevent spamming the server

CompuIves:
The client requests for every socket reconnect a new token
This happens every 100ms at the start, and sometimes
the server hasn't responded before retrying, which causes a big
spam for the server. This change makes sure that we only request
a new token when we're sure that the existing request has finished.

### DIFF
--- a/packages/app/src/app/overmind/effects/live/index.ts
+++ b/packages/app/src/app/overmind/effects/live/index.ts
@@ -143,11 +143,38 @@ class Live {
     return this.connectionPromise;
   }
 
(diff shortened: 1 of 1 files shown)


christianalfoni (approved):
Looks good to me! 👍

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5020/commits/9db70dafbe26d7819acb878025c00a4707a2f92b is at https://pr5020.build.csb.dev/s/new.

PR #5031, opened 2020-10-19
Set body.innerHTML on staging as well

CompuIves:


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5031/commits/e7c68255711e8afca9f5853fe21335381c3f9e28 is at https://pr5031.build.csb.dev/s/new.

lbogdan:
Hm, I think adding another environment variable (`PR=1`) would be better, so that we can keep the current behavior on stream.

PR #5044, opened 2020-10-22
Remove dependency on react-icons in @csb/components

CompuIves:
It was the only place where we use `react-icons` in `@codesandbox/components`, and it's not marked as a dependency. We'll have to check the test snapshot to see if it changes styling significantly.

### DIFF
--- a/packages/components/src/components/Tags/Tag.tsx
+++ b/packages/components/src/components/Tags/Tag.tsx
@@ -37,7 +37,7 @@ export function Tag({ tag, onRemove }: TagProps) {
           onClick={() => onRemove(tag)}
           marginLeft={1}
         >
-          <Icon size={8} name="cross" />
+          <Icon size={7} name="cross" />
         </Button>
       )}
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5044/commits/9cf89bdaac26578788bff6fa21e254b09dae870f is at https://pr5044.build.csb.dev/s/new.

siddharthkp:
wow this day has already arrived?!

siddharthkp (approved):

CompuIves:
I did a local test, and the icon is a bit too large. I reduced the size to 8, it was `1em` before. It feels like `1em` is better but I'm not sure, what do you think?

siddharthkp:
Yep 7 would be perfect. react-icons has a different box size than our icons (it's not edge to edge)

CompuIves:
Good one, looks much better now!

PR #5051, opened 2020-10-23
Move Inter loading to Google Fonts in Sandpack

CompuIves:
I noticed today that the bundler took a long while to load because it was downloading all Inter fonts. With this change we'll only download 2 specimen for in the bundler, and I think it will not block the loading of the loading screen.

### DIFF
--- a/packages/app/src/sandbox/status-screen/indicator-screen.html
+++ b/packages/app/src/sandbox/status-screen/indicator-screen.html
@@ -1,4 +1,3 @@
-<link href="/static/fonts/inter/inter.css" rel="stylesheet">
 <style>
   body {
     color: #fff;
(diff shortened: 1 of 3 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5051/commits/6bfc2105a9c92beb4d44dd35fc47ac66dfaf0dd8 is at https://pr5051.build.csb.dev/s/new.

DannyRuchtie (commented):
Not related to the PR but just noticed that parts of the IDE are still only loading Roboto. 
Not sure how or when to fix 

Screenshot 2020-10-27 at 08 56 15
Screenshot 2020-10-27 at 08 54 29

DannyRuchtie (approved):

PR #5085, opened 2020-11-04
Automatically resize embed on embed size change (while keeping ratio)

CompuIves:
The first commit is just converting some files to TS, the second commit is more interesting. This change introduces two things:

1. Editor and Preview automatically resize on sidebar close/open + window resize
2. The ratio between editor and preview is kept after resize

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -203,6 +203,7 @@
     "react-motion": "^0.5.0",
     "react-outside-click-handler": "^1.2.3",
     "react-refresh": "^0.7.2",
+    "react-resize-detector": "^5.2.0",
     "react-router-dom": "^5.2.0",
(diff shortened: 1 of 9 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5085/commits/98ffb6e8a9b4e405717f197db38a33ca248242a8 is at https://pr5085.build.csb.dev/s/new.

siddharthkp:
I think this PR breaks the default view logic we have based on the space available

PR on the left, production on the right

Screenshot 2020-11-04 at 6 51 47 PM

CompuIves:
Ah yep you're right. I think it snaps at the given prop of 50% and overrides default behaviour. I'll fix it tomorrow!

PR #5110, opened 2020-11-11
LSP connection maximum try attempts: 10

andrasbacsai:

**This PR is a proposal.**

Connecting to our LSP socket server could cause to crash the browser/tab. 

Why? I left a tab with codesandbox running in the background for 1-2 hours, and it caused 10000+ (connection) errors in the console. My browser was unresponsive, and had to force shutdown it.

Our `language server` tries to connect `infinite` times to our WebSocket server.

## What is the new behavior?

It only tries to connect ten times.

### DIFF
--- a/packages/app/src/app/overmind/effects/vscode/index.ts
+++ b/packages/app/src/app/overmind/effects/vscode/index.ts
@@ -776,7 +776,9 @@ export class VSCodeEffect {
   }
 
   private createWebsocketFSRequest() {
-    const socket = io(`${this.getLspEndpoint()}?type=go-to-definition`);
+    const socket = io(`${this.getLspEndpoint()}?type=go-to-definition`,{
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5110/commits/3fab39935e9ef6a2a5e450111f8e75aba45e083b is at https://pr5110.build.csb.dev/s/new.

CompuIves:
Yeap this makes sense to me, I do wonder what would happen if someone loses connectivity for eg. 5 minutes and goes back to the tab, then they probably won't get LSP messages anymore. Maybe we should create a way to check when they come online again, or when they open the tab again?

andrasbacsai:
I will check later on if it's still necessary. I guess yes.

PR #5114, opened 2020-11-12
allow access to process from within VSCode

christianalfoni:
This will rather start the client process in a tmux session, which can be accessed from within VSCode. It is inserted behind the `start:dev_api`.

It does not matter if process is already running or not. If it is running the `start:dev_api` command will open the existing session. If not, it will start the process and open the tmux session.

To exit a tmux session you use `ctrl + b`

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -7,4 +7,4 @@ ADD .git /workspace/.git
 WORKDIR /workspace
 RUN git reset --hard
 RUN apt update
-RUN apt install -y nano less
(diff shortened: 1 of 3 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5114/commits/77adde32abc7e2caf8f9f7f92c081cbd5e279594 is at https://pr5114.build.csb.dev/s/new.

CompuIves (approved):

PR #5148, opened 2020-11-18
When starting comments do not jump to responsive if comment was not made in preview mode

SaraVieira:


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5148/commits/04c9bd4674c02298242ce0b16f848faa2814d498 is at https://pr5148.build.csb.dev/s/new.

garethx:
This does resolve the specific issue in my testing - in that if I create a preview comment with responsive preview disabled, it doesn't get enabled when selecting the comment later. However, I want to note that, and likely beyond the scope of this PR, that if you then click on a comment which was created with responsive preview enabled, then it stays enabled even if you then view a comment where the responsive preview wasn't enabled when the preview comment was created.

PR #5187, opened 2020-11-26
add setting to disable browser logging

SaraVieira:
This PR adds a new setting that allows users to disable the usage of console-feed to make sure their browser does not crash in heavy logging cases


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5187/commits/a35701c5f347d8ef3c2d97ab7b1ed0fe264b0973 is at https://pr5187.build.csb.dev/s/new.

CompuIves (approved):
Looks good to me! One next step would be to also disable logging then in the bundler, I think that other people asked for that too since the log injector changes where the logs come from. I left one comment on the message, to educate people a bit more on why it is disabled, but other than that it looks good!

packages/common/src/templates/configuration/sandbox/ui.tsx
```diff
@@ -144,6 +144,20 @@ export const ConfigWizard = (props: ConfigurationUIProps) => {
...
+        <PaddedConfig>
+          <ConfigItem>
+            <PaddedPreference
+              title="Disable Console"
+              type="boolean"
+              {...bindValue(parsedFile, 'disableLogging')}
+            />
+          </ConfigItem>
+          <ConfigDescription>
+            Disable the in browser console to prevent slowing down of the page
```
CompuIves, line 157: Oh wait, one thing, this one is tricky. I have this suggestion, because in 99% of the cases it doesn't slow down.

```suggestion
            Disable the in-browser console to prevent slowing down of the page when there are many logs to the console.
```

SaraVieira:
Two very valid points have been added 

Definitely, I can take a look at that if you help me find where it is

ray-cj-huang:
Is this possible to do so for Cloud Sandbox as well?

PR #5205, opened 2020-12-01
Profile bugfix - Add pro check to avoid double watermark

siddharthkp:
We add an "Open sandbox" button on the showcased sandbox. Aaaaand we inject a "Open Sandbox" watermark/button on the server for non-pro users as an anti-phishing effort. Some users end up getting both of these 🤡 

Added a check for Pro users to avoid duplicate buttons + Moved delete to the top right so that it's not on top of the embed button


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5205/commits/29179e1fe48031961a75999dcb092217af339151 is at https://pr5205.build.csb.dev/s/new.

SaraVieira (approved):

PR #5138, opened 2020-11-16
Step 1: Do not register webpack service worker in Sandpack

souldzin:
- [POC: Sandpack Service Worker to Resolve Relative Files](https://github.com/codesandbox/codesandbox-client/pull/5137)
- **[Step 1: Do not register webpack service worker in Sandpack](https://github.com/codesandbox/codesandbox-client/pull/5138)**
- [Step 2: Setup webpack build for sandpack service worker](https://github.com/souldzin/codesandbox-client/pull/1)

## What kind of change does this PR introduce?

This will do undesirable caching, since the same host is used for various projects. Also, it will block us from using a service worker for handling relative resources.

## Context

- See the POC above for more context :smile:

## What is the current behavior?

Currently a service worker is always registered to optimize asset loading with caching. But! Since the same service worker can be used for multiple projects, this can be problematic. It will also conflict with using a service worker to communicate with the file resolver.

## What is the new behavior?

| Description | Screenshot |
|-------------|-------------|
| :tada: Sandpack asset host is being hit and still works | Screen Shot 2020-11-16 at 1 07 27 PM |
| :tada: No sandpack service worker is loaded | Screen Shot 2020-11-16 at 1 19 11 PM |

## What steps did you take to test this? This is required before we can merge, make sure to test the flow you've updated.

I've been testing this sandpack behavior in integration with GitLab. Here's how this was tested:

1. Build sandpack assets with `yarn run prepublishOnly`
2. Start a HTTPS server to host these assets
3. In GitLab project `yarn link "smooshpack"`
4. In GitLab project update [`application_setting_implementation.rb`](https://gitlab.com/gitlab-org/gitlab/blob/cf3844aff88ad3611e1d03d9d998a0135a14404b/app%2Fmodels%2Fapplication_setting_implementation.rb#L394) with path to local server hosting the sandpack assets.

## Checklist

- [X] Documentation N/A
- [ ] Testing 
- [X] Ready to be merged
      
- [ ] Added myself to contributors table

PR #5137, opened 2020-11-16
POC: Sandpack Service Worker to Resolve Relative Files

souldzin:
- **[POC: Sandpack Service Worker to Resolve Relative Files](https://github.com/codesandbox/codesandbox-client/pull/5137)**
- [Step 1: Do not register webpack service worker in Sandpack](https://github.com/codesandbox/codesandbox-client/pull/5138)
- [Step 2: Setup webpack build for sandpack service worker](https://github.com/codesandbox/codesandbox-client/pull/5218)

## DO NOTE MERGE!

This PR is just the POC for setting up the Sandpack package to use a Service Worker for resolving relative files. This will be refined and broken into smaller MR's for iteration + ease of review.

## Context

Check out the relevant:

- [GitLab issue](https://gitlab.com/gitlab-org/gitlab/-/issues/197206)
- [analysis](https://gitlab.com/gitlab-org/gitlab/-/issues/197206#note_334728792)
- [initial Sandpack Service Worker investigation](https://gitlab.com/gitlab-org/gitlab/-/issues/218438#note_378545011)

## Does it work?

Here's the [latest integration demo](https://gitlab.com/groups/gitlab-org/-/epics/3138#note_444988530) of this working with self-hosted Codesandbox assets and the GitLab Web IDE

cc @CompuIves

PR #5218, opened 2020-12-03
Step 2: Setup webpack build for sandpack service worker

souldzin:
- [POC: Sandpack Service Worker to Resolve Relative Files](https://github.com/codesandbox/codesandbox-client/pull/5137)
- [Step 1: Do not register webpack service worker in Sandpack](https://github.com/codesandbox/codesandbox-client/pull/5138)
- **[Step 2: Setup webpack build for sandpack service worker](https://github.com/codesandbox/codesandbox-client/pull/5218)**

This PR sets up webpack to have another entrypoint for building the upcoming Sandpack Service Worker. It's necessary for webpack to build this so that:

- We can have nice typesafety features.
- The service worker and sandpack client can share common types.
- Webpack can add a hash to the service worker so that we can make sure clients have the latest one. Service workers are very eagerly cached by browsers.

This PR also sets up creating a `assets.json` which is a manifest of assets created by webpack. This is important because:

- The entrypoint needs to know where the sandpack service worker lives since the file will have a hash in it to prevent eager caching.
- The Sandpack Service Worker which is going to proxy requests, **does not** proxy requests to things in assets.json (see [the POC for more details](https://github.com/codesandbox/codesandbox-client/pull/5137)).

There isn't any runtime difference here, and the production assets should be the same except now there are two new files:

- assets.json
- sandpack-service-worker.[hash].js

## Checklist

- [x] Documentation N/A
- [ ] Testing 
- [ ] Ready to be merged (waiting for [Step 1](https://github.com/codesandbox/codesandbox-client/pull/5138) to be merged and this one to be rebased and retargeted)
      
### DIFF
--- a/packages/app/config/setup-entries.js
+++ b/packages/app/config/setup-entries.js
@@ -1,0 +1,42 @@
+const AssetsPlugin = require('assets-webpack-plugin');
+const merge = require('webpack-merge');
+const path = require('path');
+const paths = require('./paths');
+
+module.exports = config => {
+  const isProd = config.mode === 'production';
+  const shared = {
+    plugins: [
+      new AssetsPlugin({
+        filename: 'assets.json',
+        path: 'www',
+        keepInMemory: true,
+        entrypoints: true,
+      }),
+    ],
+  };
+
+  return [
+    merge(config, shared),
+    merge(
+      {
+        ...config,
+        plugins: [],
+        entry: {
+          'sandpack-service-worker': path.join(
+            paths.sandboxSrc,
+            'sandpack-service-worker.ts'
+          ),
+        },
+        output: {
+          ...config.output,
+          filename: isProd ? '[name].[contenthash:9].js' : '[name].[hash].js',
+          chunkFilename: isProd
+            ? '[name].[contenthash:9].chunk.js'
+            : '[name].[hash].chunk.js',
+        },
+      },
+      shared
+    ),
+  ];
+};
--- a/packages/app/scripts/build.js
+++ b/packages/app/scripts/build.js
@@ -7,6 +7,7 @@ const filesize = require('filesize');
 const gzipSize = require('gzip-size').sync;
 const rimrafSync = require('rimraf').sync;
 const webpack = require('webpack');
+const commonConfig = require('../config/webpack.common');
 const config = require('../config/webpack.prod');
 const paths = require('../config/paths');
 const recursive = require('recursive-readdir');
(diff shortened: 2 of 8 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5218/commits/332e531aad8131c2232e983b63c0cad2ed6e6295 is at https://pr5218.build.csb.dev/s/new.

packages/app/config/setup-entries.js
```diff
@@ -0,0 +1,42 @@
+const AssetsPlugin = require('assets-webpack-plugin');
+const merge = require('webpack-merge');
+const path = require('path');
+const paths = require('./paths');
+
+module.exports = config => {
+  const isProd = config.mode === 'production';
+  const shared = {
+    plugins: [
+      new AssetsPlugin({
```
souldzin, line 10: **Why do we need `AssetsPlugin`?**

See the MR description for more details.

This is a nice plugin which provides a JSON file that can map an entrypoint name to the actual file path that webpack outputs to. So, at runtime I can query for `assets.json` and look for the `sandpack-service-worker` entry to get the hashed file path.

packages/app/scripts/build.js
```diff
@@ -140,7 +145,7 @@ function build(previousSizeMap) {
 
     if (stats.hasWarnings()) {
       console.warn(chalk.yellow('Build warnings:'));
-      stats.compilation.warnings.forEach(({ name, message }) => {
+      getStatsWarnings(stats).forEach(({ name, message }) => {
```
souldzin, line 148: **What's going on here?**

Now that there are multiple entrypoints, we actually get back a [`MultiStats`](https://webpack.js.org/api/node/#stats-object) object which has the same API, except `compilation` is nested under a `stats` array.

souldzin:
@CompuIves here's the next PR in the Sandpack-Service-Worker-Saga which updates Webpack to build the upcoming service worker (who can see how this is all done in the POC). Check out the PR description for details.

Could you please review and merge? Thanks!

CompuIves (commented):
All looks good! I have one question related to the manifest.

packages/app/config/setup-entries.js
souldzin: @CompuIves I think I remember going down this path, but finding some issue with it. But I think you're right! It **should** also do the job. I'm going to take a look 👀

Sorry for the late reply + hope you had a great new year / happy holidays!

PR #5223, opened 2020-12-04
Fix npm registry settings showing for personal workspace

CompuIves:


packages/app/src/app/pages/Dashboard/Content/routes/Settings/components/Navigation.tsx
```diff
@@ -57,10 +58,12 @@ export const SettingNavigation = (props: SettingsNavigationProps) => (
...
         label="Workspace"
       />
-      <NavigationLink
-        url={dashboardUrls.registrySettings(props.teamId)}
-        label="Npm Registry"
-      />
+      {!props.isPersonal && (
+        <NavigationLink
+          url={dashboardUrls.registrySettings(props.teamId)}
+          label="Npm Registry"
```
siddharthkp, line 64: super silly pick, but 
```suggestion
          label="npm Registry"
```

siddharthkp:
Related: We should put this in an IconButton to increase the click area

 + The tooltip text should be more informative? "Learn more about npm registry configuration"

image

siddharthkp (changes requested):
Code looks good 👍  Added 2 feedback on UX.

CompuIves:
> Related: We should put this in an IconButton to increase the click area
> 
> 
> 
> image
> 
> 
> 
> 
> 
> and The tooltip text should be more informative? "Learn more about npm registry configuration"

Good point! I'll change the content of the tooltip. For some reason the tooltip stopped working when I put it outside of it, I got a CodeSadbox and once I fixed that I couldn't see the tooltip. Pretty weird 🤨

packages/app/src/app/pages/Dashboard/Content/routes/Settings/components/Navigation.tsx
CompuIves: Oh yes I've been thinking about this! So npm needs to be written without capitals, but when it's at the start of a sentence, shouldn't then the first letter be capitalized like any other word? Not sure, I at least do like the capital more from a design perspective (consistency), but that shouldn't be the reason to do it.

garethx: That's technically right, yes, but the mixed case does look wrong here somehow. Try all upper case e.g. NPM Registry - that will visually look better and folks will still know what you mean.

siddharthkp:
> For some reason the tooltip stopped working when I put it outside of it, I got a CodeSadbox 

strange! 🤔 

If you see it again, can you pass me a repro sandbox

packages/app/src/app/pages/Dashboard/Content/routes/Settings/components/Navigation.tsx
CompuIves, line 64:
```suggestion
          label="NPM Registry"
```

CompuIves:
> > For some reason the tooltip stopped working when I put it outside of it, I got a CodeSadbox
> 
> strange! 🤔
> 
> If you see it again, can you pass me a repro sandbox

Will do!

PR #5225, opened 2020-12-04
Sandbox permissions 2

siddharthkp:
API Dependency: https://github.com/codesandbox/codesandbox-server/pull/212

What's the feature?

1. On a workspace level, you can prevent sandboxes from leaving the workspace with forks. (You can override that on the sandbox level) - _This has changes to settings, dashboard and editor_.
2. On a workspace level, you can prevent sandboxes to be exported as zip. (You can override that on the sandbox level)
3. You can set the default role new members join as - _changes scoped to settings_

All 3 of these features are only for team pro

Note: This PR constantly has merge conflicts because it touches a lot of surface. We are a couple of days away from release (api dependency). Will backmerge master again before release.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5225/commits/88475cec9bf0df0b82930bbe141badaf5840740d failed.

packages/app/src/app/pages/Sandbox/Editor/Header/ForkButton/ForkButton.tsx
```diff
@@ -153,64 +166,79 @@ export const ForkButton: React.FC<ForkButtonProps> = props => {
...
-                item={space}
-                disabled={
-                  space.userAuthorizations.find(
-                    authorization => authorization.userId === user.id
-                  )?.authorization === 'READ'
+        <Tooltip
+          label={
+            preventForksOutsideWorkspace
+              ? 'You don not have permission to fork this sandbox outside of the workspace'
+              : null
```
SaraVieira, line 173: Does nothing break when the label is null?

packages/app/src/app/pages/Sandbox/Editor/Header/ForkButton/ForkButton.tsx
```diff
@@ -153,64 +166,79 @@ export const ForkButton: React.FC<ForkButtonProps> = props => {
...
+
+                <Icon size={8} name="caret" />
+              </Menu.Button>
+
+              <Menu.List
+                css={css({
+                  fontSize: 2,
+                  lineHeight: 1,
+                })}
+                style={{
```
SaraVieira, line 206: Why is this not in the css?

christianalfoni (commented):
The code looks good to me! Though not tested the feature. We should test on stream? 😄

packages/app/src/app/pages/Sandbox/Editor/Header/ForkButton/ForkButton.tsx
siddharthkp: The opposite :)

If it's null, then the Tooltip renders its children without adding a tooltip. 

Would love to revisit some components to make this API more explicit

packages/app/src/app/pages/Sandbox/Editor/Header/ForkButton/ForkButton.tsx
siddharthkp: I'm not sure why was it not css earlier, I assumed Ives had a reason. Not going to lie, I nested it without trying to challenge what's inside.

siddharthkp:
> The code looks good to me! Though not tested the feature. We should test on stream? 😄

@christianalfoni Yep :) Sanne, Sara and Alex have all helped test on stream

PR #5234, opened 2020-12-06
detect ability to actually take preview screenshot

christianalfoni:
Since we use a shortcut now, we have to ensure when used that you are actually allowed to create the screenshot. Also include some fixes that was broken in the other broken PR which caused code not to work correctly


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5234/commits/c702186d2852081eda5465736009cdbfd417167f is at https://pr5234.build.csb.dev/s/new.

PR #5262, opened 2020-12-10
add external resources back to vanilla

SaraVieira:
Example where they are working: https://codesandbox.io/s/rktm2

### DIFF
--- a/packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/Explorer/ExternalResources/index.tsx
+++ b/packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/Explorer/ExternalResources/index.tsx
@@ -42,6 +42,7 @@ export const ExternalResources: FunctionComponent<Props> = ({ readonly }) => {
   );
 
   const { externalResourcesEnabled } = getTemplateDefinition(template);
+
   if (!externalResourcesEnabled) {
     return null;
   }
@@ -183,9 +184,6 @@ const getFontFamily = (search: string) => {
 
   return {
     name: family.split('+').join(' '),
-    id: family
-      .split('+')
-      .join('-')
-      .toLowerCase(),
+    id: family.split('+').join('-').toLowerCase(),
   };
 };
--- a/packages/common/src/templates/parcel.ts
+++ b/packages/common/src/templates/parcel.ts
@@ -56,7 +56,6 @@ export default new ParcelTemplate(
       '/.babelrc': configurations.babelrc,
       '/tsconfig.json': configurations.tsconfig,
     },
-    externalResourcesEnabled: false,
     distDir: 'dist',
     main: true,
     isTypescript: true,


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5262/commits/49c32cbfc524e3ec2f9f585ec89f63454c4015c5 is at https://pr5262.build.csb.dev/s/new.

CompuIves:
Ah, I remember now what my reasoning behind this was. At the time I wanted to slowly phase out the "External Dependencies" in the sidebar, but I never finished that. The reasoning for that is transparency, when you add an external resource with the UI, the code doesn't change, and this breaks several concepts:

1. When you download the code locally, the external resources won't be loaded
2. When you export/import to GitHub, they won't be added either
3. The resources are added in the bundler, so essentially `fetch` calls, which behaves differently from `<link />` and `<script />`

I wanted to move everything to how our `dependencies` works instead, there you can add dependencies in a UI, but in reality it changes code and you can transparently see what has changed. That's how it should ideally be with External Resources too, when you add an external resource, it should be added in `index.html` instead, because that's what should happen in reality. Right now we add it to the database, but you won't see it in the source code which will break things like deploys, imports, exports, etc.. It makes the code on CodeSandbox work differently compared to "local".

I think we should continue this, so not add this functionality to Parcel but instead look into making External Resources work with index.html instead, and once that works we enable it for Parcel. I wonder what you think about this as well.

SaraVieira:
I understand that but the breaking of deploys and exports has already been fixed and it does work now, these things have been added as people see value in it

I think ultimately it comes down to that, people see value in this and if removing it doesn't fix the problem then we should not take the value away

I hope that makes sense

CompuIves:
Could you try this with exporting to github? I'm fairly sure it breaks the sandbox.

SaraVieira:
Github does but I don't see as enough reason to remove something people find value in only in one template

What doesn't is ZIP and deploys

SaraVieira:
https://github.com/codesandbox/codesandbox-client/blob/master/packages%2Fapp%2Fsrc%2Fapp%2Futils%2Finject-resources-for-export.ts

CompuIves:
Deploys and zip will also break if you export to github. I'm trying to get rid of external resources in how it works now, so this feels like a step back. 

I think you could compare it like this:

Would you rather have:
1. People unhappy because a feature is missing?
2. People confused because their sandbox is broken on GitHub and works differently locally compared to on CodeSandbox?

In the end External Resources is a feature that I think we need, but I regret building it the way it is right now. I had the same thing with package.json, and converting "Add Dependency" to the new system turned out to be a huge undertaking. Inevitably we need to do this for "External Resources" as well, but if we keep it turned off for Vanilla the migration will at least be easier while preventing confusion for some people.

That's my perspective on it, do you still think it would be worth it to enable it?

PR #5278, opened 2020-12-14
Pin esquery to v1.0.1 to fix eslint

CompuIves:
Esquery changed their API (in 1.2 or 1.3) to become esmodules, so the
export is suddenly { default: {} }, which eslint doesn't expect. We have to
pin esquery to v1.0.1 for now, but in the long run we need to update eslint as
it seems that eslint fixed this in v7.

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,6 +105,9 @@
     "typescript": "4.0.3",
     "username": "^5.1.0"
   },
+  "resolutions": {
+    "eslint/esquery": "1.0.1"
+  },
   "husky": {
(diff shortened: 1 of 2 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5278/commits/a8543c8d21fabd14c62f8162b7278dc89fc99b58 is at https://pr5278.build.csb.dev/s/new.

christianalfoni (approved):
Great 👍

PR #5279, opened 2020-12-14
Hide fork menu item for read user

siddharthkp:
Tiny polish thing that came up in user testing. Want a review on implementation, not sure if this is a good way.

Read users don't have the ability to fork. Clicking this menu option leads to an error notification, so hiding it.

Note: Left the option in the command palette if you seek it out - it will still give you an error notification. Not sure if there's a way to show it but disable.

### DIFF
--- a/packages/app/src/app/overmind/effects/vscode/Workbench.ts
+++ b/packages/app/src/app/overmind/effects/vscode/Workbench.ts
@@ -201,15 +201,23 @@ export class Workbench {
       },
     });
 
-    this.appendMenuItem(MenuId.MenubarFileMenu, {
-      // Z to be the last item after vscode group 4
-      group: '4_zsandbox',
-      order: 1,
-      command: {
-        id: 'codesandbox.sandbox.fork',
-        title: '&&Fork Sandbox',
-      },
-    });
+    if (
+      this.controller.getState().editor?.currentSandbox?.teamId ===
+        this.controller.getState().activeTeam &&
+      this.controller.getState().activeWorkspaceAuthorization === 'READ'
+    ) {
+      // don't add the option to fork
+    } else {
+      this.appendMenuItem(MenuId.MenubarFileMenu, {
+        // Z to be the last item after vscode group 4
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5279/commits/37aa511ce57ee707bda171f1789c3c86b32b01f5 is at https://pr5279.build.csb.dev/s/new.

SaraVieira (approved):
I would still wait for a review from christian

packages/app/src/app/overmind/effects/vscode/Workbench.ts
```diff
@@ -201,15 +201,23 @@ export class Workbench {
...
-        id: 'codesandbox.sandbox.fork',
-        title: '&&Fork Sandbox',
-      },
-    });
+    if (
+      this.controller.getState().editor?.currentSandbox?.teamId ===
+        this.controller.getState().activeTeam &&
+      this.controller.getState().activeWorkspaceAuthorization === 'READ'
+    ) {
+      // don't add the option to fork
```
christianalfoni, line 209: The implementation makes sense, but I would revert the check. I typically do this trick:

```js
!(mySpecificScenarioWhereItShouldNotWork)

// so

!(this.controller.getState().editor?.currentSandbox?.teamId ===
        this.controller.getState().activeTeam &&
      this.controller.getState().activeWorkspaceAuthorization === 'READ')
```

It makes it more readable I think. Allows you to say: "this is the specific exception and I want it to work in all other scenarios" 😄

christianalfoni (changes requested):
Looks good and it needs to be implemented here, but just a tiny comment. Just a preference thing, so no biggie 😄

packages/app/src/app/overmind/effects/vscode/Workbench.ts
siddharthkp: Made the changes :)

siddharthkp:
Aw I want to revisit this soon

Issue #4905, opened 2020-09-18
Links with hash are not triggering popstate events


posva:
# 🐛 bug report

## Preflight Checklist

* [ ] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/master/CONTRIBUTING.md) for this project.
* [ ] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
* [ ] I have searched the issue tracker for an issue that matches the one I want to file, without success.

## Description of the problem

Anchors with relative links like `#foo` usually trigger the `popstate` event on window but it seems as if they were prevented so they don't propagate to any listener added by the user. They also do not trigger when navigating through history when they normally do.

## How has this issue affected you? What are you trying to accomplish?

This breaks any client side routing using the history api. Vue Router in my case.

### Link to sandbox: [link](https://codesandbox.io/s/bug-report-hash-links-d3s6r?file=/index.html)

- Click on the links in order
- Navigate back and forward with history

You will see that only the navigations made with `pushState` trigger the popstate. But all them trigger a popstate in a regular page.

### Your Environment

| Software         | Name/Version |
| ---------------- | ------------ |
| Сodesandbox      | client
| Browser          | Chrome 85
| Operating System | OSX

### Other

I don't know if it's related but noticed [this line](https://github.com/codesandbox/codesandbox-client/blob/master/packages/app/src/app/overmind/effects/router.ts#L16)

```js
window.history.replaceState({}, '', sandboxUrl({ id, alias, git }));
```

It should be

```js
window.history.replaceState(window.history.state, '', sandboxUrl({ id, alias, git }));
```

To preserve any existing state pushed by the user.

I remember there was https://github.com/codesandbox/codesandbox-client/blob/master/packages/sandbox-hooks/url-listeners.js#L117 to work in vue-router. And we could add `el.__vnode` (double underscore) to make it work with Vue 3 (and that would be great already!), but I'm opening this issue in case you seek to find a long term alternative

posva:
The reproduction no longer shows the problem

PR #5302, opened 2020-12-18
fix(router): keep existing state when replacing

posva:

This is a proposal rather than a fix. I would say that the state should be preserved when calling `history.replaceState` instead of being deleted. I came from https://github.com/codesandbox/codesandbox-client/issues/4905 but I don't remember anymore a case where the current solution would fail...

1. Step A
2. Step B
3. Step C

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      
### DIFF
--- a/packages/app/src/app/overmind/effects/router.ts
+++ b/packages/app/src/app/overmind/effects/router.ts
@@ -13,7 +13,7 @@ export default new (class RouterEffect {
     alias?: string | null;
     git?: GitInfo | null;
   }) {
-    window.history.replaceState({}, '', sandboxUrl({ id, alias, git }));
+    window.history.replaceState(window.history.state, '', sandboxUrl({ id, alias, git }));
   }
 
   updateSandboxUrl(


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5302/commits/2adeb22fa8750b5e093d9819ee6328b35e2b1044 is at https://pr5302.build.csb.dev/s/new.

posva:
Check

danilowoz:
Thanks for helping us clean up!

This PR does not seem to improve user or developer experience in a noticeable way, even fixing any bug. We are very sorry that you spent time on this, only to see us closing the PR. But we hope you understand our reasoning and do not feel discouraged to make other contributions 😄

posva:
Sure. I don't remember anymore but if this `replaceState()` was called for the window on the right of code sandbox, then it should be changed with the proposal as `replaceState()` and removing the current state will break client side routing that relies on the history state like Vue Router 4.

If it's about the codesandbox application itself, then it doesn't matter

PR #5333, opened 2021-01-06
Request account closing button

### LINKED ISSUE #5360: Allow account deletion
# 🌈 Allow account deletion
Currently there is no way to delete CodeSandbox accounts. This is a feature in almost every other major website.

This feature would (obviously) allow people to delete their own accounts. There should be account settings in the "CodeSandbox settings" menu across the website, this would have the option to delete your account, followed by a prompt to confirm (Maybe even a "Type [username here] to confirm" like in GitHub). This feature could possibly also be merged with  the "Integrations" panel, and renamed to "Account Settings"

SaraVieira:
This adds a button on the dashboard where a user can request their account to be deleted like so:

CleanShot 2021-01-06 at 14 40 23

When the user clicks to close it they will see a message saying: 

"Your account will be deleted by our support team in one working day."

closes #5360


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5333/commits/0dd227be649e5d226188b401c35ebe3b13a6cd30 is at https://pr5333.build.csb.dev/s/new.

CompuIves:
I think there might be a potential "threat" with sending directly to Airtable. People could use it to create a custom request that asks to delete a user from someone else, eg. I could use the same API call to request to delete "SaraVieira", and if we don't watch out we'll do that.

A solution for that would be if we add the Airtable entry from the server instead, because then we can make sure only the current user can be added. Or maybe there's a way to add some kind of data to the airtable entry that only the current user could know?

If you want I can create a GQL endpoint on the server to do this today, that would prevent any slowdown for you.

DannyRuchtie:
Account Managed by **Github** 
Request Account deletion 

This might be confusing. Could we come up with a better copy to make it more explicit that we remove the CSB account? 
Maybe Ceora can help with the copy. 

Or we need to move the  Account Managed by **Github**   underneath the Username.
(but this still is not explicit enough IMO)

DannyRuchtie (changes requested):
Update copy to make clear that you are removing your CSB account and not your GitHub account.

SaraVieira:
I think we should honestly remove that line and place the logo of what you used to login on the side of your photo

DannyRuchtie:
Adding a logo would solve the problem! I like the suggestion

SaraVieira:
Updated design
CleanShot 2021-01-14 at 14 26 57
CleanShot 2021-01-14 at 14 27 01

DannyRuchtie (approved):
Looking good

CompuIves (approved):
Very nice! One last nitpick on the message itself: 
image

Curious what @garethx things, but maybe we want to add a bit more time to remove it (like 3 working days). And it would be great if we add something like "If you've changed your mind, please send us an email to hello@codesandbox.io", something like that.

Everything looks good on code, so I'm going to approve it! If @garethx also approves it can go in!

garethx:
Perhaps "Thanks for your request. Your account will be deleted within 3 working days. If you've changed your mind, please email us at hello@codesandbox.io with your username."

garethx (approved):

SaraVieira:
All text changed

SaraVieira:
text changed

PR #5337, opened 2021-01-07
Update Mono Lisa Version

SaraVieira:
This PR updates Mono Lisa to the latest version to remove the windows issue where 1 and I look similar.

@lbogdan  called for testing


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5337/commits/8d25e91b761c7049d0467b67774705ab66208986 is at https://pr5337.build.csb.dev/s/new.

SaraVieira:
Screenshot 2021-01-07 171238

DannyRuchtie:
@SaraVieira is this before and after? 
The 1 and i in both windows look similar and are still hard to distinguish 

Screenshot 2021-01-07 at 17 15 48

SaraVieira:
Yeah, that's my thoughts too, waiting for the creator feedback

lbogdan:
@DannyRuchtie What zoom % are you at?

For me the new version now shows fine at 80% (which breaks on prod):

image

but it still break at <= 50%

image

At 100% they both show fine, btw.

SaraVieira:
@DannyRuchtie  had a previous version :)

bebraw:
@lbogdan How important do you consider that "but it still breaks at <= 50%" case? Is it a common use case?

I see 80%/100% cases are fixed. Do you think that's enough?

SaraVieira:
Tested on my windows and looks waaay better now 👍 

I don't think the 50% is a common case tbh

DannyRuchtie (approved):
I don't think it's realistic to optimize for a 50% scale.
Judging on the screenshot we are good to go!

bebraw:
@SaraVieira Do you know why the deployed version still seems to have the old i's?

Screen Shot 2021-01-29 at 18 01 09

That's from https://codesandbox.io/s/lucid-lumiere-bk11m?file=/src/App.js .

The deploy in this PR renders like below (https://pr5337.build.csb.dev/s/new?file=/src/App.js) if I paste the same code to it:

Screen Shot 2021-01-29 at 18 02 20

Note the increased spacing in the "i" for the dot.

There's a new version with improved f available at https://www.monolisa.dev/static/light-fonts.css so maybe you could copy that over.

I would love to understand the difference between what's deployed and what's seen in this PR, though.

bebraw:
Looking at https://github.com/codesandbox/codesandbox-client/commit/e1a24a4346c1ece5a562118bfa6af89c69dd7ff9, I see "Some checks were not successful" so is it possible the fix wasn't deployed to the public yet due to that CI failure?

SaraVieira:
The checks that failed is just because we did not deploy it to stream no worries 😀

It may be service workers again and our cache 

Mind checking in incognito?

SaraVieira:
This is assuming you do not visit CodeSadbox a lot and it may have not updated yet

bebraw:
Here's mine from incognito mode in Chrome:

Screen Shot 2021-01-29 at 18 13 17

Here's a render from Firefox (much better):

Screen Shot 2021-01-29 at 18 14 06

I think you may be right. Maybe it's service workers biting!

SaraVieira:
It can also be cache since the font is sent over by cloudflare that adds some aggressive cache but it's a matter of visiting the site twice and it updates :)

bebraw:
Yeah, it's coming from disk cache. I wonder why the last modified date is 10th of December. I bet I'm getting a version from that date.

Screen Shot 2021-01-29 at 19 04 27

PR #5340, opened 2021-01-07
Fix date of expiring for monthly users

SaraVieira:
We had an issue that no matter if you had a monthly or yearly subscription on the expiring page we would always show yearly

This fixed that by checking and if it's monthly checks the day the user subscribed, the calendar month and adds one to it.

For example, someone who subscribed on the 12th of January would see:

On Monthly: 12 of Februray
On yearly: 12 January

### DIFF
--- a/packages/app/src/app/pages/Pro/index.tsx
+++ b/packages/app/src/app/pages/Pro/index.tsx
@@ -1,4 +1,4 @@
-import { format } from 'date-fns';
+import { addMonths, format } from 'date-fns';
 import { Helmet } from 'react-helmet';
 import React, { useEffect } from 'react';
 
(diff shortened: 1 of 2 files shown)


packages/app/src/app/pages/Pro/index.tsx
```diff
@@ -237,35 +237,45 @@ const Expiring = ({
...
-      </Badge>
-      <Heading>Your subscription is expiring</Heading>
+}) => {
+  // if user has yearly subscription show the date he created the subscription and since it's passed they know it's next year
+  // if Monthly show the day he created and the next month on the calendar
+  const since = new Date(user.subscription.since);
+  const expiringDate =
+    user.subscription.duration === 'yearly'
+      ? format(since, 'do MMM')
+      : `${format(since, 'do')} ${format(addMonths(new Date(), 1), 'MMM')}`;
```
CompuIves, line 247: Would this also work when we're in February? So eg. when your `since` is `Jan 12`, and it's `Feb 3` that the user sees this message, will it say `Feb 12`?

CompuIves (commented):
Oh yep this one's nasty, really good that we're fixing it now! I've left a comment with a question on a specific scenario, not sure if that scenario works but if it does we can merge it! Might be worth extracting the logic for showing that date to a separate functon and create a small unit test for it so we can test multiple scenarios, but not required.

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5340/commits/8d0e4d78a02ea4c26601c95d8835f3c4e582bb1c is at https://pr5340.build.csb.dev/s/new.

packages/app/src/app/pages/Pro/index.tsx
SaraVieira: It wopuldn't yeah

Need to check if the day they subscribed is larger than the day we are and only then show the next month

CompuIves:
This is very clear!! 👍 LGTM

PR #5365, opened 2021-01-18
Fix redirect to /s from the dashboard when clicking the logo

CompuIves:
It turns out that we redirect within `create-react-app` but since the homepage is not in `create-react-app`, it doesn't work. We need to use an `a` tag to make the page opening go outside of the `create-react-app` framework.


SaraVieira (approved):

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5365/commits/ce431daa9e5cd9d9cbbee033e21d4579792434ec is at https://pr5365.build.csb.dev/s/new.

PR #5386, opened 2021-01-21
Sandpack refresh

alexnm:
It's a big one!


CompuIves (approved):
Tested running several projects on stream, also did a check on all code that was changed within the bundler and `codesandbox-api`, and everything looks good! Let's merge on Monday 🎉

PR #5400, opened 2021-01-25
Fix issue where Codesandbox randmly closes on login

SaraVieira:


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5400/commits/85cd34c4ec98ab51302e00090f60193efa549506 is at https://pr5400.build.csb.dev/s/new.

garethx (approved):
It looks like it works - at least accessing the csb.dev build via a link saw me as signed in. When I tested when I was logged-out initially it logged-in and didn't close the window. Will be interesting to see in prod with our repro.

CompuIves:
Would be good to test this on stream too, that flow is a bit different. And I think that we should test registration flow, that's a bit different too.

SaraVieira:
https://codesandbox.stream/ is deployed

PR #5403, opened 2021-01-26
Enable new React transform for every React version

CompuIves:
This enables the new React transform for any React version, the reason behind this is that React doesn't require v17 to utilize the transform, eg. React v16 works too. The React team uses CodeSandbox CI to test, and their version is `0.0.0-abc` so it doesn't work for them all the time.


SaraVieira (approved):

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5403/commits/1fc2ba940cee9089a24fd693d9e71fd2a875ed5a is at https://pr5403.build.csb.dev/s/new.

PR #5409, opened 2021-01-27
fix text truncate

alexnm:
Dashboard - List View
- "Always-On" column replaced with a small green circle overlapping the sandbox image
- Sandbox name column wider
- Text overflow only when there's no room, removed fixed width

Sandbox templates list
- Text overflow only when there's no room, removed fixed width
- `title` attribute added so the full name is visible on hover (via browser default tooltip)

Breakpoints
- Found a 3px difference between sidebar toggle (770px) and main content going full width (767px)

### DIFF
--- a/packages/app/src/app/pages/Dashboard/index.tsx
+++ b/packages/app/src/app/pages/Dashboard/index.tsx
@@ -88,7 +88,6 @@ export const Dashboard: FunctionComponent = () => {
                   />
                 ) : (
                   <Element
-                    as="aside"
                     id="desktop-sidebar"
                     css={css({ display: ['none', 'none', 'block'] })}
                   >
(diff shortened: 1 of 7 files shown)


packages/app/src/app/pages/Dashboard/index.tsx
```diff
@@ -88,7 +88,6 @@ export const Dashboard: FunctionComponent = () => {
                   />
                 ) : (
                   <Element
-                    as="aside"
```
alexnm, line 91: this was creating two `aside` tags nested, <Sidebar> is already adding the content as an `aside`

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5409/commits/122bc675d8ba6f7c2abce4c1d5bafb57215ee9d8 is at https://pr5409.build.csb.dev/s/new.

SaraVieira (approved):
Looks great! Good work!

DannyRuchtie (changes requested):
Hi Alex,

I discovered that we are using something similar in the activity bar. 
But I don't know if that already is a component that you can reuse 

The design is slightly different (sorry I forgot about this one). It is 10*10 plus a border of 2px with the background color as the border color.


Screenshot 2021-01-28 at 12 54 26



There also is an issue with Safari. what else :( 

Screenshot 2021-01-28 at 12 56 46

alexnm:
made the bubble 12x12 and fixed the overflow cut

DannyRuchtie (approved):

PR #5419, opened 2021-01-28
fix vertical position for context menu

alexnm:
Issue was seen when you open context menu on sandbox cards that are located in the lower part of the screen while the dashboard is open.

Still not perfect, couldn't figure out how to position the reach menu component on top of the anchor. but at least the height is now computed based on the right height for the items (31px instead of 36px)

However, `React.Children` computes the dividers as well, hence we still have a slight offset, but at least it doesn't feel broken anymore) and depends on how many dividers the menu has.

We could make a fix by "meta" looking at the children and compute the dividers separately, but I wouldn't want to get a fragile implementation in the shared component.

### DIFF
--- a/packages/components/src/components/Menu/index.tsx
+++ b/packages/components/src/components/Menu/index.tsx
@@ -146,7 +146,7 @@ const ContextMenu = ({ visible, setVisibility, position, ...props }) => {
   if (!visible) return null;
 
   const numberOfItems = React.Children.count(props.children);
-  const SUGGESTED_ITEM_HEIGHT = 36;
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5419/commits/dda5173228d870e5c6b17592d652f62750b5fbc7 is at https://pr5419.build.csb.dev/s/new.

SaraVieira (approved):
Niiice!

DannyRuchtie (approved):

Issue #5406, opened 2021-01-27
Umbrella Issue for "Cannot convert undefined or null to object"


SaraVieira:
# 🐛 bug report

## Preflight Checklist

- [x] I have read the
      [Contributing Guidelines](https://github.com/codesandbox/codesandbox-client/blob/master/CONTRIBUTING.md)
      for this project.
- [x] I agree to follow the
      [Code of Conduct](https://github.com/codesandbox/codesandbox-client/blob/master/CODE_OF_CONDUCT.md)
      that this project adheres to.
- [x] I have searched the issue tracker for an issue that matches the one I want
      to file, without success.

## Description of the problem

Sometimes when using codesandbox the page will just break with the error:

```
TypeError: Cannot convert undefined or null to object
```

This error only happens in Chrome-based browsers and it comes from the sentry SDK:
![CleanShot 2021-01-27 at 14 36 58](https://user-images.githubusercontent.com/1051509/105998656-303f5e00-60ad-11eb-8bce-fd511267e959.png) 

https://sentry.io/organizations/codesandbox/issues/1730507122/?project=2071895&query=%22Cannot+convert+undefined+or+null+to+object+%22&statsPeriod=90d

I created a related issue in the sentry javascript library

## How has this issue affected you? What are you trying to accomplish?

Not me but people's codesandbox breaks out of nowhere

### To Reproduce

No one has a clue

### Link to issues:

#5420 
 #5017
 #4992
 #5346
 #5326

### Your Environment

| Software         | Name/Version |
| ---------------- | ------------ |
| Сodesandbox      | any
| Browser          | Chromium 	
| Operating System | any

kamilogorek:
Hey there, Sentry dev here. 

> This error only happens in Chrome-based browsers and it comes from the sentry SDK:

When you open the last frame of the stack trace, you can see this note:
![image](https://user-images.githubusercontent.com/1523305/106008305-71d50680-60b7-11eb-910b-f1f8d13ff4d5.png)

The reason why it looks like it comes from Sentry, is that we are wrapping some APIs (in this case requestAnimationFrame), to extract function name and capture arguments to the function call. The code that's throwing however, is the original function that has been wrapped. You can find parsed arguments in the `Additional Data` section, and it may help finding the cause of the issue:
![image](https://user-images.githubusercontent.com/1523305/106008675-d6906100-60b7-11eb-8efc-4fa5df2c7e36.png)

Let me know if I can help you in any way, cheers! :)

[...]

PR #5423, opened 2021-01-29
Disable timers in sentry

SaraVieira:
This is related to the `Cannot convert undefined or null to object ` error and suggested here #5406 

Also removes chrome extensions disable in the code as that can be done in the UI as seen in the issue


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5423/commits/77f564b8bdfbcc46f525265d7f5130d05d9a9880 is at https://pr5423.build.csb.dev/s/new.

CompuIves:
Looks good to me! Let's keep an eye on if this fixes the error. It could be that Sentry's stacktrace is just showing because they happened to handle the error, hiding the original error.

CompuIves (approved):

PR #5430, opened 2021-02-01
Show user friendly error if we have issues fetching team token

SaraVieira:
We also seemed to have an issue where sometimes pages did not recognize login and I think it had to do with the function only being called one


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5430/commits/6f5ab3bc893b60a049bbba2d842598fdcc4f7212 is at https://pr5430.build.csb.dev/s/new.

PR #5434, opened 2021-02-01
Uses our own builder instead of using aws for building netlifly websites

SaraVieira:
To do: Add status endpoint


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5434/commits/20ae2f7bde41ed8d1ffdcce8196257709c0ed443 is at https://pr5434.build.csb.dev/s/new.

PR #5435, opened 2021-02-01
context menu position: ref + getBoundingClientRect

alexnm:
Background: Last week I solved one issue with the context menu rendering at a big offset on the dashboard, but it was still an approximation based on the number of menu items passed.

Problem with the current solution: computing based on the number of `children` does not work (eg: on profile page it still renders below the viewport) because a Fragment will be seen as a single child in the count.

Experimental solution:
- when a new menu is triggered with a change of position (x,y), the menu is set on visibility hidden and renders all its children
- on the next cycle, the ref should be set with the menu rendered and the height/width are taken from the DOM

Potential problems:
- Still a fragile solution, if for whatever reason the menu does not render on the first pass, the bounding rect computations will fail
- I get a ref warning in the console in development env when opening a context menu, couldn't find the problem, maybe the forwardRef on the MenuList is not used properly
- Wasn't able to set the ref on a `List` element which is a styled component instance, hence I used the Reach component directly, as it also forwards the ref of its div

### DIFF
--- a/packages/components/src/components/Menu/index.tsx
+++ b/packages/components/src/components/Menu/index.tsx
@@ -234,7 +234,7 @@ const MenuIconButton = props => (
 
 interface MenuListProps {
   children: any;
-  style: React.CSSProperties;
+  style?: React.CSSProperties;
 }
 
 const MenuList = React.forwardRef<HTMLDivElement, MenuListProps>(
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5435/commits/2f0aa06fe69484e68fe3b53afdfee39b11e0b90c is at https://pr5435.build.csb.dev/s/new.

DannyRuchtie (approved):
Looks good

siddharthkp:
Tested this around, works like a charm. `menuIsReadyToShow` is a bit suspicious 😅  but happy to merge this and then improve it.

siddharthkp (approved):

PR #5445, opened 2021-02-03
Only create a single babel worker when in sandpack

CompuIves:
Babel Worker is memory expensive. This will make sure that we only spawn a single worker in Sandpack, because Sandpack is usually rendered multiple times in a single window.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5445/commits/5732642cdb13d985886eb4b05dd3de579428ca24 is at https://pr5445.build.csb.dev/s/new.

PR #5458, opened 2021-02-08
[WIP] Sandbox Search

SaraVieira:
Adds a search to a sandbox using a worker

Figma file: https://www.figma.com/file/xjB6dL48lshfiIa5jIfWwe/Global-Search?node-id=1%3A2

Things to be done:

- [x]  Global Search
- [x]  Make typing faster
- [x]  Toggle ability to search case sensitive
- [x]  Allow Regex
- [x]  Add files to include
- [ ]  Profit


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5458/commits/db1a24a61f95d7df2c25e1312652df5b710ff61c failed.

CompuIves:
From Slack:

Sometimes the items they clip (eg. `clipboard.ts` goes through the result item and I can't click the result item):

image

The context is a bit small:

CompuIves:
I think on the covering items between different items, it might be that `react-window` is getting a different height than the real element height 🤔

SaraVieira:
@CompuIves All issues there are now fixed and most logic was moved to overmind

siddharthkp:
Code looks good, nothing stands out as bad

Found a tiny bug in the implementation - seems like new lines are squashed in the results:

Other than this, if we have design signoff, ship it :shipit: 

image

DannyRuchtie:
Looks good!

SaraVieira:
Let's hope for the best

PR #5459, opened 2021-02-08
Make host fallback to codesandbox.test

CompuIves:
This will ensure that the local Docker environment will work, as this env var is not set in that environment.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5459/commits/01c02ee0b5dcb342fb51b45b017e09c2df855e29 is at https://pr5459.build.csb.dev/s/new.

PR #5488, opened 2021-02-17
fix(sandpack): error logs for invalid sandbox info

jyash97:
Enhancement

If entry key is not defined the log says to add dependencies

## What is the new behavior?
Fix logs

Need help on how to test this. But imo this should work.

## Checklist

- [ ] Testing 
- [ ] Ready to be merged [ Needs testing ]

### DIFF
--- a/standalone-packages/sandpack/src/utils/generate-package-json.ts
+++ b/standalone-packages/sandpack/src/utils/generate-package-json.ts
@@ -1,4 +1,4 @@
-import { IFiles, IFile, IDependencies } from '../manager';
+import { IFiles, IDependencies } from '../manager';
 
 export function getPackageJSON(
   dependencies: IDependencies = {},
@@ -25,13 +25,13 @@ export default function createMissingPackageJSON(
   if (!newFiles['/package.json']) {
     if (!dependencies) {
       throw new Error(
-        'No dependencies specified, please specify either a package.json or dependencies.'
+        'No dependencies specified, please specify either a package.json or provide dependencies.'
       );
     }
 
     if (!entry) {
       throw new Error(
-        "No entry specified, please specify either a package.json with 'main' field or dependencies."
+        "No entry specified, please specify either a package.json with 'main' field or provide entry."
       );
     }
 
--- a/standalone-packages/sandpack/src/utils/generate-package-json.ts
+++ b/standalone-packages/sandpack/src/utils/generate-package-json.ts
@@ -25,13 +25,13 @@ export default function createMissingPackageJSON(
   if (!newFiles['/package.json']) {
     if (!dependencies) {
       throw new Error(
-        'No dependencies specified, please specify either a package.json or provide dependencies.'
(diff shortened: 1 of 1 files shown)


standalone-packages/sandpack/src/utils/generate-package-json.ts
```diff
@@ -1,4 +1,4 @@
-import { IFiles, IFile, IDependencies } from '../manager';
```
jyash97, line 1: This was throwing lint error since it was unused

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5488/commits/233daf0b95867d63e67285b72957c6f896194af3 is at https://pr5488.build.csb.dev/s/new.

jyash97:
Hey @alexnm 👋 

Can you take a look into the changes ? 

Just need a confirmation if the changes done are correct.

Thanks :)

standalone-packages/sandpack/src/utils/generate-package-json.ts
```diff
@@ -25,13 +25,13 @@ export default function createMissingPackageJSON(
   if (!newFiles['/package.json']) {
     if (!dependencies) {
       throw new Error(
-        'No dependencies specified, please specify either a package.json or dependencies.'
+        'No dependencies specified, please specify either a package.json or provide dependencies.'
```
alexnm, line 28: don't think this is needed

alexnm: Furthermore, this flow might be refactored because there could be sandpacks without dependencies in the future

standalone-packages/sandpack/src/utils/generate-package-json.ts
```diff
@@ -25,13 +25,13 @@ export default function createMissingPackageJSON(
...
       );
     }
 
     if (!entry) {
       throw new Error(
-        "No entry specified, please specify either a package.json with 'main' field or dependencies."
+        "No entry specified, please specify either a package.json with 'main' field or provide entry."
```
alexnm, line 34: I would re-formulate this to: "Missing 'entry' parameter. Either specify an entry point, or pass in a package.json with the 'main' field set.

alexnm (changes requested):

standalone-packages/sandpack/src/utils/generate-package-json.ts
jyash97: > don't think this is needed
Will revert this. 

> Furthermore, this flow might be refactored because there could be sandpacks without dependencies in the future
True currently we need to pass the dependencies for running a static html page.

standalone-packages/sandpack/src/utils/generate-package-json.ts
jyash97: Will update in sometime. Thanks for the review :)

alexnm (approved):

PR #5489, opened 2021-02-17
Do standalone check for all codesandbox.io domains

CompuIves:
It turns out that if you open CodeSandbox from an external page,
window.opener will be set as well. We need to mark it as standalone
manually in this case.

Fixes https://github.com/codesandbox/codesandboxer/issues/72

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -107,7 +107,7 @@
     "browser-resolve": "CompuIves/node-browser-resolve",
     "circular-json": "^0.4.0",
     "codemirror": "^5.27.4",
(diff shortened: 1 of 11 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5489/commits/be0367862db5c6d116760b603c9c2f3fa1ceba9f is at https://pr5489.build.csb.dev/s/new.

alexnm (approved):
Looks ok, but I'd make a new version of codesandbox-api to test with sandpack, just to be sure

alexnm (approved):

PR #5517, opened 2021-02-23
New Overmind typing

christianalfoni:
This PR has no functional changes, it is only typing changes. It is a BIG piece of work though, but theoretically it should be impossible that something is buggy, cause it is only refactoring TS files with new typing, so it would have notified about issues


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5517/commits/d0af77f77e6742776a6432f59e3102d251eafa71 is at https://pr5517.build.csb.dev/s/new.

PR #5533, opened 2021-02-26
Remove SW caching of API urls in sandbox

CompuIves:
With private npm we need to ensure that we always get the latest API
response back, right now we cache /api/v1/npm_registry/... in the SW,
which sometimes makes private npm packages return old results.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5533/commits/8b6cecce7652d5af246578ee8754f1a40a766c7c is at https://pr5533.build.csb.dev/s/new.

alexnm (approved):

PR #5550, opened 2021-03-02
Fix all lint warnings

SaraVieira:
It's impossible to see when there is a lint error since we had like 70 warnings in the whole app

This fixed all the warnings

### DIFF
--- a/packages/app/src/sandbox/compile.ts
+++ b/packages/app/src/sandbox/compile.ts
@@ -658,14 +658,12 @@ async function compile({
         const htmlEntries = templateDefinition.getHTMLEntries(configurations);
         const htmlModulePath = htmlEntries.find(p => Boolean(modules[p]));
         const htmlModule = modules[htmlModulePath];
-        let html = '<div id="root"></div>';
+        let html =
+          template === 'vue-cli'
+            ? '<div id="app"></div>'
+            : '<div id="root"></div>';
         if (htmlModule && htmlModule.code) {
-          if (htmlModule.code) {
-            html =
-              template === 'vue-cli'
-                ? '<div id="app"></div>'
-                : '<div id="root"></div>';
-          }
+          html = htmlModule.code;
         }
         const { head, body } = getHTMLParts(html);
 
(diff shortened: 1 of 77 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5550/commits/378fab5a535ca99b1da9f61e87cf6bf66751d834 is at https://pr5550.build.csb.dev/s/new.

packages/app/src/sandbox/compile.ts
```diff
@@ -657,14 +658,16 @@ async function compile({
...
-        const { head, body } = getHTMLParts(
-          htmlModule && htmlModule.code
-            ? htmlModule.code
-            : template === 'vue-cli'
-            ? '<div id="app"></div>'
-            : '<div id="root"></div>'
-        );
+        let html = '<div id="root"></div>';
+        if (htmlModule && htmlModule.code) {
+          if (htmlModule.code) {
```
CompuIves, line 663: I think here there should be html = htmlModule.code and in the if statement the Vue check.

CompuIves (changes requested):
Wooow, very impressive. This must've been a ton of work! I checked all files and I only found one thing where it mightve converted wrongly. Left a comment there.

Would be good to merge and deploy tomorrow, then we can take the day to monitor sentry to be sure that everything still works!

packages/app/src/sandbox/compile.ts
christianalfoni: Ah, yeah, so:

```ts
let html = template === 'vue-cli' ? '<div id="app"></div>' : '<div id="root"></div>';
if (htmlModule && htmlModule.code) {
  html = htmlModule.code
}
```

christianalfoni (changes requested):
This is a SOLID piece of work, thanks so much for doing this @SaraVieira !! Just one tiny logical thing that needs to be fixed 😄

packages/app/src/sandbox/compile.ts
SaraVieira: truuu, just pushed

christianalfoni (approved):
Great! 👍  😄

PR #5565, opened 2021-03-08
Add Shared with me & Liked page

SaraVieira:
Adds two new pages on dashboard:

* Sandboxes shared with you:

CleanShot 2021-03-08 at 15 38 49

* Liked sandboxes:

CleanShot 2021-03-08 at 17 16 57

Waiting on backend deploy for PR Preview

PR #5569, opened 2021-03-09
Add Basic auth as option in private npm

CompuIves:
There are also cherry picked type fixes in here, will be removed once #5565 is merged.

To test:
1. Set npm registry on "Custom"
2. Set auth on "Basic"
3. Verify that the Authentication header is Basic on the server


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5569/commits/962167e2010f397337cb9f79b9bd945c9d524b1f is at https://pr5569.build.csb.dev/s/new.

CompuIves:
@ceoraford I added an item to the documentation on private registries:

```md
### Auth Type

Npm supports two types of authentication: `Basic` and `Bearer`. More recent
implementations use `Bearer` by default. If your local configuration of the
private registry involves setting the auth token as parameter on the registry
url (like `//registry.npmjs.org/:_authToken={token}`), this should be `Bearer`.
If you're using a registry like Artifactory and configure your npm registry
using `_auth={token}` in `.npmrc`, you should use `Basic` auth and use the token
from `_auth` as auth token.
```

I'm curious what you think of this.

@DannyRuchtie The UI of configuring a custom private npm has changed a bit as well, I added the "Auth Type" field here:

image

I'm also curious what _you_ think of this 😄

ceoraford:
is there any official documentation on Bearer and Basic? It might be helpful to link to that in line if possible. 
is it possible to add a code snippet or screenshot displaying each use case? a visual might be good here too. 
also will that screenshot be in the docs? bc I think that would be great to show exactly where users would input auth type

CompuIves:
> Is there any official documentation on Bearer and Basic?

Not on any place other than our own documentation, npm doesn't elaborate on their implementation details when it comes to auth, so I had to figure out which auth to use based on testing (which is why I'm only adding it now instead of from the start, haha).

We can write good documentation on it ourselves though, I like that idea. We can maybe give example `.npmrcs` and tell which auth belongs to which configuration type!

I'll also add a screenshot. Really good points!

ceoraford:
> Not on any place other than our own documentation, npm doesn't elaborate on their implementation details when it comes to auth, so I had to figure out which auth to use based on testing (which is why I'm only adding it now instead of from the start, haha)

okay I thought so. i did a quick google search and didn't come up with anything. in that case, it would be good to include more context there since users probably won't have many other ways to learn more about it. and the rest of what you said goes right along with what I was thinking😊

ceoraford (approved):
looks good👍🏾

CompuIves:
Yee!

DannyRuchtie:
Screenshot 2021-03-16 at 09 06 12

@CompuIves I noticed a small nitpicking detail that we probably missed the first time. 
But the background color of the "card" does not match the color of the other cards (See image above).

PR #5606, opened 2021-03-22
Allow array exports

CompuIves:
Our esmodule converter for node_modules didn't take `export const [a, b] = c()` into account, with this change it does.

### DIFF
--- a/packages/app/src/sandbox/eval/transpilers/babel/convert-esmodule/__snapshots__/index.test.ts.snap
+++ b/packages/app/src/sandbox/eval/transpilers/babel/convert-esmodule/__snapshots__/index.test.ts.snap
@@ -179,6 +179,21 @@ exports.default = $csb__default;
 "
 `;
 
+exports[`convert-esmodule can do array exports 1`] = `
+"\\"use strict\\";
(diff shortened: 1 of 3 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5606/commits/962f3c96a008477ccf72844edbd1b784732838e0 is at https://pr5606.build.csb.dev/s/new.

ljosberinn:
thanks for the fast respone :) I'm afraid the same issue would still persist with object destructuring aswell, wouldn't it?

CompuIves:
Object destructuring (`export const { a, b } = c()`) works already, but I never knew that you could also export an array structure. That was completely new to me!

lbogdan:
Hey @ljosberinn ,
Any idea where exporting members of a const array as named exports is documented? I have never saw it until now, I tested that it actually works, but I can't find it anywhere. 🙄

ljosberinn:
I've searched a bit and interestingly couldn't find a single example, but examples for exporting destructured declarations from objects were already rare, so that may be why, and arguably the use case for it is probably fairly limited.

PR #5609, opened 2021-03-24
Optimize merging dependencies with no entries

CompuIves:
This allows a dependency to be merged even if the entries don't match,
with the requirement that the existing dependencies has no entries. That
way we know for sure that no other dependency is calling into it.

### DIFF
--- a/packages/sandpack-core/src/npm/merge-dependency.ts
+++ b/packages/sandpack-core/src/npm/merge-dependency.ts
@@ -180,21 +180,23 @@ export function mergeDependencies(responses: ILambdaResponse[]) {
       } else if (response.dependencyDependencies[depDepName]) {
         const exDepDep = response.dependencyDependencies[depDepName];
 
+        // Determine which version is newer, needed for some checks later.
+        const [newerVersionDepDep, olderVersionDepDep] = semver.gt(
+          newDepDep.resolved,
+          exDepDep.resolved
+        )
+          ? [newDepDep, exDepDep]
+          : [exDepDep, newDepDep];
+
         if (exDepDep.resolved === newDepDep.resolved) {
           exDepDep.parents = uniq([...exDepDep.parents, ...newDepDep.parents]);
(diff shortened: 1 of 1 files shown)


packages/sandpack-core/src/npm/merge-dependency.ts
```diff
@@ -185,7 +185,8 @@ export function mergeDependencies(responses: ILambdaResponse[]) {
           exDepDep.entries = uniq([...exDepDep.entries, ...newDepDep.entries]);
         } else if (
           intersects(exDepDep.semver, newDepDep.semver) &&
-          isEqual(exDepDep.entries, newDepDep.entries)
+          (isEqual(exDepDep.entries, newDepDep.entries) ||
+            exDepDep.entries.length === 0) // Meaning that this dependency is not called from other dependencies, so save to replace.
```
clauderic, line 189:
```suggestion
            exDepDep.entries.length === 0) // Meaning that this dependency is not called from other dependencies, so safe to replace.
```

clauderic:
You can test against both of these sandboxes to validate this fix:

https://codesandbox.io/s/dndkit-sortable-image-grid-forked-j7ent
https://codesandbox.io/s/dreamy-bhaskara-pb9sn

The problem is that `semver: "^1.0.2` is resolving to `1.0.2` instead of `1.2.0`:
Screen Shot 2021-03-24 at 5 23 18 PM

In this case, luckily there are no entries for the package that is causing the issue because the dependency is only used to import types (see https://github.com/clauderic/dnd-kit/blob/master/packages/modifiers/src/createSnapModifier.ts#L1 for example).

I'd be curious to know if this issue can also happen if a package was actually being consumed (and therefore had entries), in which case this fix wouldn't resolve the original issue.

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5609/commits/ea98aa58af88c27d259a1528f44266df2a5f90de is at https://pr5609.build.csb.dev/s/new.

PR #5623, opened 2021-03-30
bundler updates on error messages and hiding parts of the UI

alexnm:
Updates:
- codesandbox-api version set to latest (no package depends on the npm version anymore, since it's all handled by lerna)
- error screen and loading screen are shown in the bundler based on the flags received on `compile`
- for sandpack only, insert the head tags from the html entry file into the bundler html entry file
- fix subsequent errors thrown when the error screen is shown (keyEventHandler)
- send the stack frames when dispatching an error from the bundler


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5623/commits/502e79b305c0fd4843a91a0206529f0ce0dcbfb7 is at https://pr5623.build.csb.dev/s/new.

packages/app/src/sandbox/compile.ts
```diff
@@ -701,6 +709,11 @@ async function compile({
           // on the first run. However, if there's no server to provide the static file (in the case of a local server
           // or sandpack), then do it anyways.
           document.body.innerHTML = body;
+
+          // Preseve the head tags and add anything that comes from the template at the end of the tag
+          // This way, title and other meta tags will overwrite whatever the bundler <head> tag has,
+          // but all the preloads and scripts/styles that are needed by the bundler are kept
+          document.head.insertAdjacentHTML('beforeend', head);
```
CompuIves, line 716: Does this replace the head, or add it? It could be that we call this multiple times, so in that case we'd have more and more things in head over time. Maybe we should replace everything in head with the head from HTML. Might be worth testing if this doesn't work!

CompuIves (commented):
Loooooks good! One question.

packages/app/src/sandbox/compile.ts
alexnm: This preserves the head from the original html. I thought at first that it makes sense to preserve it because of the resources preloaded or loaded from there. but this code runs after the head is parsed, so we can also replace everything right?

CompuIves: Yes, I believe so. It would be good to test as well, but I think that's completely save to do.

PR #5649, opened 2021-04-09
Allow users to change username and display name

sbinlondon:
Requires https://github.com/codesandbox/codesandbox-server/pull/315 to be merged and deployed first.

* Removes REST API call for updating profile (bio and social links)
* Makes use of new Graphql mutation `UpdateCurrentUser` to update profile info (username, display name, bio, social links)
* Alerts user with modal when changing username, so that they know they'll need to update their external links

See it in action: https://www.loom.com/share/813da73b839141a8a16f1cfde85e73b0

I think I did all the Graphql updating okay but let me know if I missed anything since this is my first time contributing to the client repo.

### DIFF
--- a/packages/app/src/app/graphql/types.ts
+++ b/packages/app/src/app/graphql/types.ts
@@ -17,7 +17,7 @@ export type Scalars = {
    * The `DateTime` scalar type represents a date and time in the UTC
    * timezone. The DateTime appears in a JSON response as an ISO8601 formatted
    * string, including UTC timezone ("Z"). The parsed date and time string will
-   * be converted to UTC and any UTC offset other than 0 will be rejected.
+   * be converted to UTC if there is an offset.
    */
   DateTime: any;
   /**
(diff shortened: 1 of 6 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5649/commits/99c80c92bfeb7162765f0e6c2d2b3200fdc979cd is at https://pr5649.build.csb.dev/s/new.

siddharthkp:
Kudos for the video walkthrough!!

Code looks ✌️ 

Do you have thoughts if this should be in Profile or in Dashboard Setting for personal workspace? 🤔

sbinlondon:
> Do you have thoughts if this should be in Profile or in Dashboard Setting for personal workspace? 🤔

There was a cursory discussion on Slack but essentially, there wasn't a pane for profile stuff in the preferences modal, whereas if you clicked on the profile sidebar you could edit bio and social links. So I figured that it would be best to keep it where people have used it before. 

I think that if we go beyond allowing people to edit username, display name, bio, and social links, though, it'll have to move to the preferences pane because we can't fit anything else where it is now.

siddharthkp:
Makes sense to me!

siddharthkp (approved):

PR #5675, opened 2021-04-19
Don't delete files with similar name optimistically

CompuIves:
We used to return everything starting with the name of a directory in
`getModulesAndDirectoriesInDirectory`. This isn't right, because if you
have this:

- /app
- /app/index.js
- /app-components

`getModulesAndDirectoriesInDirectory(..., ..., '/app')` would also
include `/app-components`. Luckily we only do this client side, so when
someone deletes the folder `/app`, `/app-components` doesn't get
deleted. But it's confusing that the folder gets hidden. This fixes that.

I tested it by:

1. Creating 4 folders, 2 in the root and 2 in a subfolder
2. Deleting the subfolder
3. Verify that all folders in the subfolder are deleted, and other subfolder is still there.

### DIFF
--- a/packages/common/src/sandbox/modules.ts
+++ b/packages/common/src/sandbox/modules.ts
@@ -37,10 +37,7 @@ export function resolveDirectory(
   }
 
   // Split path
-  const splitPath = path
-    .replace(/^.\//, '')
-    .split('/')
-    .filter(Boolean);
+  const splitPath = path.replace(/^.\//, '').split('/').filter(Boolean);
 
   const foundDirectoryShortid = splitPath.reduce(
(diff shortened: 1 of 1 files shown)


CompuIves:
The important part: https://github.com/codesandbox/codesandbox-client/pull/5675/files#diff-840da019ca20675d7970e36406ba43a3d813104d5cd9978462f25847de7f99a9R81-R92

andrasbacsai (approved):

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5675/commits/5505be03ff59b3cdddd5cbc4b12db481bfb94764 is at https://pr5675.build.csb.dev/s/new.

PR #5680, opened 2021-04-19
Start preview only if editor is ready

DeMoorJasper:

This PR changes the start behaviour of the preview to only start when the editor has fully loaded.
This should speed up the initial load time of the editor by removing some of the load the bundler and user-code bring.

Not entirely sure if this is wanted behaviour and/or if this is the right way to implement it. Was mainly just hoping to resolve the double-compile issue.

Currently the preview loads together with the editor resulting in two compilations whenever a codesandbox is opened.

Usually most of it is cached by the second compilation so it might speedup the preview, but probably slows down the editors initial load and also caused some weird side-effects with saved vs unsaved editor state. App and error overlay flickering is one of them...

## What is the new behavior?

Preview loads when editor has finished loading.

## Checklist

- [x] Documentation
- [x] Testing 
- [x] Ready to be merged
      

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5680/commits/aee4af006c8f8f499a4337a2661764ff8fb863f9 is at https://pr5680.build.csb.dev/s/new.

CompuIves:
This one is suuper interesting. I think this could be very valuable, I want to test the flow a bit on stream to see how it feels in fast/slow connections.

aj-foster:
@DeMoorJasper hello 👋🏼 just checking in to see if this PR should remain open.

DeMoorJasper:
Don't think this'll ever get merged so probably safe to close

PR #5691, opened 2021-04-22
clone html elements in a way that allows code execution

DeMoorJasper:

Fixes https://github.com/codesandbox/sandpack/issues/44 and fixes https://github.com/codesandbox/sandpack/issues/40

Script tags aren't being executed as we currently  simply use .innerhtml

## What is the new behavior?

HTML tags are cloned in a way so they are executed properly.

## Checklist

- [x] Documentation
- [x] Testing 
- [x] Ready to be merged
      

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5691/commits/d337fe1185768c6bc03fb8839f5e772a9c68b7e0 is at https://pr5691.build.csb.dev/s/new.

packages/app/src/sandbox/compile.ts
```diff
@@ -707,20 +766,20 @@ async function compile(opts: CompileOptions) {
           process.env.LOCAL_SERVER ||
           process.env.SANDPACK
         ) {
-          // The HTML is loaded from the server as a static file, no need to set the innerHTML of the body
-          // on the first run. However, if there's no server to provide the static file (in the case of a local server
-          // or sandpack), then do it anyways.
-          document.body.innerHTML = body;
-
           // Add head tags or anything that comes from the template
```
alexnm, line 769: I think this comment is outdated, because this function doesn't override anything from the original head of the bundler page.

alexnm (approved):
looks good, tested with the branch locally in sandpack, all good, including the example from which the issue originally started

alexnm:
Just saw that 3 tests are failing

DeMoorJasper:
@alexnm about the tests I have no clue why they're failing, they also fail on my other bugfix PRs so probably unrelated to this code

alexnm:
I think the fact that you don't remove existing content from the body anymore causes this. Check the bundler content for this sandbox: https://pr5691.build.csb.dev/s/vue vs the production one: https://codesandbox.io/s/vue. Does the bundler simulate a disabled javascript environment in any way? really weird that we get the `<noscript>` content in the PR build.

Maybe we can still use your version of inserting each tag so scripts get executed, but we also clear the innerHTML for the body? @CompuIves any thoughts on this?

CompuIves:
It looks like the `<noscript></noscript>` tag shows for some reason, maybe if we dynamically add it to the dom it renders anything in it, doesn't matter if there's JS or not?

alexnm:
the `<noscript>` is in the bundler initial html file, it is not injected from the vue sandbox html in this case. when we were doing `document.innerHTML = html`, any content of the body from the initial html was being removed

DeMoorJasper:
@alexnm Ives was right it had to do with `noscript`, the DOMParser built into the browser is also a bit too strict unfortunately so changed it to use the domparser2 library

packages/app/src/sandbox/compile.ts
```diff
@@ -687,41 +688,43 @@ async function compile(opts: CompileOptions) {
         if (htmlModule && htmlModule.code) {
           html = htmlModule.code;
         }
+
         const { head, body } = getHTMLParts(html);
+        if (!lastHeadHTML && !lastBodyHTML) {
```
DeMoorJasper, line 693: This should prevent us from duplicating html content. Changed it to simply reload on any change to html.

packages/app/src/sandbox/compile.ts
```diff
@@ -687,41 +688,43 @@ async function compile(opts: CompileOptions) {
...
+              await appendHTML(body, document.body);
+            }
+          }
+        } else if (
+          (lastHeadHTML && lastHeadHTML !== head) ||
+          (lastBodyHTML && lastBodyHTML !== body)
+        ) {
+          // Always refresh if html changed
+          if (manager) {
+            manager.clearCompiledCache();
```
DeMoorJasper, line 722: Not entirely sure why this has to clear all the cache every time html changes? cc @CompuIves

DeMoorJasper:
Issue with integration tests was apparently a double body tag in the source code, so the original sandpack implementation of this was handled incorrectly.

See commit https://github.com/codesandbox/integration-sandboxes/commit/13396423eac308ef7398bb72505bbe0c9836976f for the fix

PR #5710, opened 2021-04-27
Send active team id to server for github imports

sbinlondon:
Relates https://github.com/codesandbox/codesandbox-server/issues/316 and https://github.com/codesandbox/codesandbox-server/pull/340

In order to allow users (who do not have personal pro) to import private repos into a team they are part of that does have team pro, we need to be able to tell what the active team is. This param was being sent in `forkSandbox` but not via the normal static Github URL import (`/s/github/user/repo`).

### DIFF
--- a/packages/app/src/app/overmind/namespaces/editor/actions.ts
+++ b/packages/app/src/app/overmind/namespaces/editor/actions.ts
@@ -298,7 +298,8 @@ export const sandboxChanged = withLoadApp<{
   }
 
   try {
-    const sandbox = await effects.api.getSandbox(newId);
+    const params = state.activeTeam ? { teamId: state.activeTeam } : undefined;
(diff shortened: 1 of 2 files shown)


packages/app/src/app/overmind/namespaces/editor/actions.ts
```diff
@@ -298,7 +298,8 @@ export const sandboxChanged = withLoadApp<{
   }
 
   try {
-    const sandbox = await effects.api.getSandbox(newId);
+    const params = state.activeTeam ? { teamId: state.activeTeam } : undefined;
```
alexnm, line 301: nitpick: I would use null to suggest the absence of the value, but if that calls for an extra conditional further down the line to check if params is null or not, undefined is also good

alexnm (approved):
looks good

PR #5715, opened 2021-04-28
Further Sandpack HTML HMR improvements

DeMoorJasper:

Should improve the HMR for HTML files a bit further.

Removes and re-adds all html nodes, without touching sandpack specific code. In case of a css or js file it bails out and just refreshes the page.

I can change this so scripts are also re-added similar to how the rest of html works in this PR, there's just one downside to this is that scripts can start running more than once which can cause problems...

## Checklist

- [x] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5715/commits/88e4e9f6c49ecdfd1d26adfd9b27d6280f3a842a failed.

PR #5723, opened 2021-04-29
Improve HTML HMR Logic and fix some issues

DeMoorJasper:

Should resolve issues with multiple html entry points and lean more towards how html hmr originally worked.

Original approach was just appending content... This does not work unfortunately, ideally we'd use innerHTML for everything except scripts... So we need to extract scripts out of html and append those to the html and than change the innerHTML or vice-versa

After some more research in https://codesandbox.io/s/admiring-water-hvj6e?file=/src/index.js I discovered that the ideal way to do this would be to load scripts using `document.createElement` and load everything else using `body.innerHTML` and `head.innerHTML` this should fix all issues below. We should also figure out a way to not overwrite sandpack's elements.

Some sandboxes that can be used for testing various HTML edge-cases:
- https://codesandbox.io/s/jovial-jang-qnzlg (Should render a circle svg)
- https://codesandbox.io/s/1on0w (Should be able to navigate the pages)
- https://codesandbox.io/s/github/programeasily/vue-3-basics/tree/main/ (Should be able to render it, TODO: Add this to integration tests) This has some weird template syntax that should be handled
- https://codesandbox.io/s/xenodochial-shadow-qjqx2 (Should be able to reload content without forking to a new sandbox)

## Checklist

- [x] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5723/commits/a61aa8ab33b2708dedcd6c71a9616302c1d40445 is at https://pr5723.build.csb.dev/s/new.

plondon:
Hi @DeMoorJasper, any update on this PR? Anything we can do to help?

DeMoorJasper:
@plondon I'm planning to look into this further next week

PR #5730, opened 2021-05-03
Fix character width issues in embed

CompuIves:
The font that we were loading in the embed wasn't actually loaded
because it wasn't included in the html. Now it is, so the character
width calculations of the embed should be fixed.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5730/commits/a3a62a3c8a04ea2bf703b11bc1ffc58b90530a25 is at https://pr5730.build.csb.dev/s/new.

PR #5733, opened 2021-05-04
fix(editor): refresh editor in the second launch

danilowoz:
*The big bug*

**Reproduce**

- Launch the editor in any template or sandbox;
- Go to the dashboard page using a navigation link;
- Relaunch the editor through a sandbox link or creating a new one;

**Bugs**

- The command palette no longer open;
- The Github link is the same as the previous session, causing a lot of inconsistency;

**What might be causing it?**

As long as the editor is in the browser memory, there are many stale states, and probably a few methods have not been triggering again in the second launch.

**Possible solution?**

- Reset the states on unmount component;
- Identify which methods need to be retriggered or be cleaned;

**Temp fix**

There is a proposal of a temporary workaround, which consists of checking in the second launch if the editor has loaded; if so, reload the entire page. ¯\*(ツ)*/¯

https://www.notion.so/28294a4d1f924b20b73533ab5b8c2419?v=adac55ed225d4d48a19a0c73579a02d0&p=b5fcc670c3fb46afa8a57dd05f701bcb

### DIFF
--- a/packages/app/src/app/pages/Sandbox/index.tsx
+++ b/packages/app/src/app/pages/Sandbox/index.tsx
@@ -23,6 +23,22 @@ export const Sandbox = React.memo<Props>(
     const state = useAppState();
     const actions = useActions();
 
+    /**
+     * !important Hard bug fix
+     *
+     * This address temporarily many unexpected
+     * behaviors that occurs when  navigating away
+     * from the editor and go back to it
+     * Eg: Editor -> Dashboard -> Editor
+     *
+     * @link https://www.notion.so/Unexpected-behaviors-when-navigating-away-from-the-editor-and-go-back-to-it-b5fcc670c3fb46afa8a57dd05f701bcb
+     */
+    useEffect(() => {
+      if (window.CSEditor) {
+        window.location.reload();
+      }
+    }, []);
+
     useEffect(() => {
       if (!showNewSandboxModal) {
         if (window.screen.availWidth < 800) {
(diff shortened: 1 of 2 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5733/commits/47c9b07def02179c1e99b2b86ce055b47d015b1f is at https://pr5733.build.csb.dev/s/new.

packages/app/src/app/pages/Sandbox/index.tsx
```diff
@@ -23,6 +23,22 @@ export const Sandbox = React.memo<Props>(
...
+     * from the editor and go back to it
+     * Eg: Editor -> Dashboard -> Editor
+     *
+     * @link https://www.notion.so/Unexpected-behaviors-when-navigating-away-from-the-editor-and-go-back-to-it-b5fcc670c3fb46afa8a57dd05f701bcb
+     */
+    useEffect(() => {
+      if (window.CSEditor) {
+        window.location.reload();
+      }
+    }, []);
```
danilowoz, line 40: Is this running the CodeSandbox call API twice?

danilowoz: @CompuIves Yes, it seems that it calls the API twice. In order to avoid this, I tried to put this conditional in the root of the computer, however, this `reload` was triggered in the "Fork" command, which is something that we don't want. 

Apart from this is not a so nice workaround, (very well known as "gambiarra" in Portuguese), I'd say this's the best we could do now, as far as I am concerned.

CompuIves (approved):
This looks good to me! Excited to see this go live :smile:

PR #5769, opened 2021-05-11
client auth v1

sannek:


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5769/commits/97de7f8415710d1577475ee64c4a2bdd08c8590e is at https://pr5769.build.csb.dev/s/new.

danilowoz (approved):

packages/app/src/app/pages/Client/Prompt/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+      <SubTitle>
+        <br />
+        Click the button below to continue and finish signing in.
+      </SubTitle>
+
+      <Buttons>
+        <Button
+          as="a"
+          autoWidth
+          href={deepLink}
```
CompuIves, line 69: If you want you can probably also directly put the link in here:
```suggestion
          href={`codesandbox://auth-completion/?nonce=${authToken}`}
```

CompuIves (commented):
Code looks good!! I left a nitpick (it doesn't have to be implemented), but other than that it's v clean.

I would like opinion from @DannyRuchtie as well, I think we should maybe make it a bit more in the new design of CodeSandbox with our header and button maybe different.

packages/app/src/app/pages/Client/Prompt/index.tsx
zehfernandes: I think we can align to the current brand changing the typography and adding the logo. What do you think? 

image

PR #5770, opened 2021-05-11
Add zsh & vim to Dockerfile 😇

sbinlondon:
Adding zsh and vim for when us server bods work on the client


sbinlondon:
@siddharthkp said this was ok to merge if it helps dev setup so going to do that 👀

PR #5810, opened 2021-05-20
improve tar protocol and jsdelivr-gh to support branch names

### LINKED ISSUE #5725: Support GitHub branch as npm package
# 🌈 Feature

## What is this feature?

Currently when u wanna install some package from GitHub repository, u have ability to do it so by writing
```js
// package.json
{
 "dependencies": {
    "redux-injectors": "goranurukalo/redux-injectors",
  },
}
```
By default this grabs default (master) branch and use it.
All good and nice.

It would be nice if we can implement different <branch/commit/tag> etc... (jsdelivr supports all of this [jsdelivr](https://www.jsdelivr.com/?docs=gh))

I would like to implement and provide PR in case this makes sense and that there was not issues while this was made or there are some issues that wasn't able to find

By quick search or check, file: [jsdelivr-gh.ts](https://github.com/codesandbox/codesandbox-client/blob/master/packages/sandpack-core/src/npm/dynamic/fetch-protocols/jsdelivr/jsdelivr-gh.ts) was right (or not 😅) place.

We can parse, extract values after hashtag and provide it to jsdelivr as specific field that they support.

This part of code where we can extend and provide value to "jsdelivr" to understand how and what to serve
I would like to know more about cache and stuff like that

User/Developer is writing
```js
// package.json
{
 "dependencies": {
    "redux-injectors": "goranurukalo/redux-injectors#dev",
  },
}
```
And now it can use specific branch in codesandbox

`convertGitHubURLToVersion` function could filter out (by regex?) hashtag (`#`) 
and return always just repo name
and new function can be provided to extract value after hash and added as part of first argument `url` for `fetchWithRetries` method?

## My questions
-  caching of files (how it works for this app)
-  what about `JSDelivrGHFetcher.meta` method?

## End
Thank you for reading ❤️

DeMoorJasper:

Also supports branch names in gh provider, adds tests and improves overall reliability of tar and jsdelivr-gh checks

Closes #5725

## Checklist

- [x] Documentation
- [x] Testing 
- [x] Ready to be merged
      
### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -327,7 +327,7 @@
     "jest-image-snapshot": "^1.0.1",
     "jest-junit": "^3.4.1",
     "jest-styled-components": "7.0.2",
-    "isomorphic-untar-gzip": "^1.0.3",
+    "isomorphic-untar-gzip": "^1.0.5",
     "mini-css-extract-plugin": "^0.8.0",
     "object-assign": "^4.1.1",
     "opn": "4.0.2",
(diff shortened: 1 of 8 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5810/commits/0471bbff26c5c04dcf8533c3af5c095f96ed4758 is at https://pr5810.build.csb.dev/s/new.

PR #5813, opened 2021-05-21
Resolver plugins

DeMoorJasper:
## What kind of change does this PR introduce?

This Pull Request reworks the entire resolution logic of sandpack to use a more efficient custom resolver and allow resolver plugins so we can have a ts resolver, esmodule resolver, ...

## Checklist

- [x] Documentation
- [x] Testing 
- [x] Ready to be merged
      
- [x] Added myself to contributors table

PR #5821, opened 2021-05-21
Basic ESModule url imports

DeMoorJasper:

This PR implements ESModule url imports, or at least the basis of it, this is a quicker alternative (hack) to #5813 as the other plugin will probably need a lot of time, planning and testing before we can actually release a reliably version of it.

Future work:
- Rewrite resolver from scratch
- Simplify transpilation logic to take into account remote dependencies a bit better

## Checklist

- [x] Documentation
- [x] Testing 
- [x] Ready to be merged
      

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5821/commits/9df1e9da7527845663349d65a5519bfb3f8bf339 is at https://pr5821.build.csb.dev/s/new.

packages/app/src/sandbox/eval/transpilers/babel/worker/babel-worker.js
```diff
@@ -408,6 +408,21 @@ async function compile(code, customConfig, path, isV7) {
   try {
     let result;
     try {
+      // Set modern targets for env preset, speeds up Babel significantly
```
DeMoorJasper, line 411: When debugging a resolver issue I noticed the babel preset-env was running ie9 ? which is actually less modern than the codesandbox editor and causes a lot more transpilation and therefore slower builds

DeMoorJasper: I can always update the browserlist to something less modern but ie9 seems a bit extreme

packages/app/src/sandbox/eval/transpilers/babel/index.ts
```diff
@@ -46,96 +51,96 @@ class BabelTranspiler extends WorkerTranspiler {
...
+        // If the code is commonjs and does not contain any more jsx, we generate and return the code.
+        if (!syntaxInfo.jsx && !syntaxInfo.esm) {
+          measure(`dep-collection-${path}`);
+          await Promise.all(
+            collectDependencies(ast).map(async dependency => {
               if (dependency.isGlob) {
                 loaderContext.addDependenciesInDirectory(dependency.path);
               } else {
-                loaderContext.addDependency(dependency.path);
+                await loaderContext.addDependency(dependency.path);
```
CompuIves, line 91: What's the advantage of awaiting a promise in this case for adding a dependency?

packages/sandpack-core/src/manager.ts
```diff
@@ -706,134 +712,224 @@ export default class Manager implements IEvaluator {
...
+    if (esmoduleUrl) {
+      if (this.transpiledModules[esmoduleUrl]) {
+        return this.transpiledModules[esmoduleUrl].module;
+      }
+
+      throw new Error('Cannot download ESModule dependencies synchronously');
+    }
+
+    // Node modules?
+    if (isUrl(parentPath) && !esmoduleUrl) {
```
CompuIves, line 927: Interesting, do you see this case happen as well?

CompuIves (commented):
Very nice changes, also much cleaner!!

I have left some questions. One in particular I'm wondering about is `addDependency`. If `addDependency` is awaited in the transpilation step, could it slow down transpilation? For example, a transpiler that adds 10 dependencies, in that case we need for those 10 dependencies before we can start transpiling new files. I'm not sure if that's actually a problem, but worth thinking about.

packages/sandpack-core/src/manager.ts
DeMoorJasper: Oh yes it happens all the time as esmodules can be relative and absolute urls, this is also what we do in the codesandbox esmodule cdn

packages/sandpack-core/src/manager.ts
```diff
@@ -706,134 +712,224 @@ export default class Manager implements IEvaluator {
...
+  }
+
   // ALWAYS KEEP THIS METHOD IN SYNC WITH SYNC VERSION
-  async resolveModuleAsync(
-    path: string,
-    currentPath: string,
-    defaultExtensions = ['js', 'jsx', 'json', 'mjs']
-  ): Promise<Module> {
-    const dirredPath = pathUtils.dirname(currentPath);
+  async resolveModuleAsync({
```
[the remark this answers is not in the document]
DeMoorJasper: I rewrote the entire function based on the logic of `resolveModule` as they weren't in sync no more. Maybe we could use something like https://yarnpkg.com/package/gensync at some point for this?

So the esmodule checks and download should be the only changes, at least compared to the synchronous version

packages/app/src/sandbox/eval/transpilers/babel/index.ts
DeMoorJasper: I think it was causing race conditions as we utilise the dependencies in `TranspiledModule#doTranspile` which only really became obvious due to the extra delay fetching an ESModule adds

DeMoorJasper: Just tested this a little bit and didn't see any noticeable performance impact:

https://codesandbox.io/s/muddy-lake-zpstd
Before ESM: 1.8-2.15secs
After ESM: 1.9-2.1 secs

https://codesandbox.io/s/largish-redux-example-pjc8f
Before ESM: 2.5-2.9secs
After ESM: 2.3-2.6secs

packages/sandpack-core/src/manager.ts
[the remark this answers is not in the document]
CompuIves: Oh `gensync` definitely sounds interesting, that's worth looking into.

Also happy that this function is back in sync with `resolveModule` 😄

packages/sandpack-core/src/manager.ts
CompuIves: Ahh, yes, I see. Would be good to put that scenario as a comment here for future readers (something simple like `esmodule file that requests a node module`).

packages/app/src/sandbox/eval/transpilers/babel/index.ts
CompuIves: Ohhh okay, great! Let's keep this in then :shipit:

CompuIves (approved):
Okay great! This looks and sounds good! 😄

CompuIves:
Perhaps we should wait for the new template to be in before merging this, in case existing sandboxes use http urls in imports (I'm really not sure why, but maybe it happens).

PR #5875, opened 2021-06-07
Homepage: new templates

danilowoz:
It redesigns how we introduce the templates on the homepage, so from now on, it's endorsing to use templates instead of creating a new sandbox from scratch. So basically the sandbox creating flow has been changed in order to make it easier for the user to launch the editor even quicker. 

@tromika as we're introducing a new flow, does it worth adding the sign-in wall right after the template selection?


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5875/commits/962bd360cbf32385d74f67147edeec545fae294a is at https://pr5875.build.csb.dev/s/new.

DannyRuchtie (approved):
Looks really good! 
I love the subtle animation! The responsive solution is good as-well

tromika:
@danilowoz are we deploy changes like this?

PR #5885, opened 2021-06-09
Don't overwrite setImmediate

### LINKED ISSUE #5878: setImmediate and window.setImmediate are different functions, and only window.setImmediate works correctly
# 🐛 bug report

## Preflight Checklist

      adheres to.
      to file, without success.

## Description of the problem
window.setImmediate allows args to be passed in, but the global setImmediate does not.  If indeed a global is to be provided to the environment, it should be exactly the same was window.setImmediate, or at the very least have the same signature and be able to execute arguments successfully.

I want to use setImmediate to call a function and pass arguments to it

### To Reproduce
Go to https://codesandbox.io/s/elegant-smoke-erz2n?file=/src/index.js and observe the flow execution in index.js.
This shows the difference between the two functions.

### Your Environment

| Сodesandbox      | latest             |
| Browser          | chrome latest              |
| Operating System |  win10 latest            |

DeMoorJasper:

This PR changes the logic where we overwrite setImmediate with a custom version that did not behave correctly.

Mainly wondering why we ever implemented this? (we can optionally polyfill this with a correct implementation if we want to)

Fixes #5878 

## Checklist

- [x] Documentation
- [x] Testing 
- [x] Ready to be merged
      
### DIFF
--- a/packages/sandpack-core/src/runner/eval.ts
+++ b/packages/sandpack-core/src/runner/eval.ts
@@ -1,20 +1,6 @@
-// @flow
 import buildProcess from './utils/process';
 
 const g = typeof window === 'undefined' ? self : window;
-const requestFrame = (() => {
-  const raf =
-    g.requestAnimationFrame ||
-    // @ts-ignore
-    g.mozRequestAnimationFrame ||
-    g.webkitRequestAnimationFrame ||
-    function (fn) {
-      return g.setTimeout(fn, 20);
-    };
-  return function (fn: (time: number) => void) {
-    return raf(fn);
-  };
-})();
 
 const hasGlobalDeclaration = /^const global/m;
 
@@ -39,8 +25,8 @@ export default function (
     module,
     exports,
     process,
-    setImmediate: requestFrame,
     global,
+    window: global,
     ...globals,
   };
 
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5885/commits/1ca9b5e1318cec8cca6de571e05a0f411f00877a is at https://pr5885.build.csb.dev/s/new.

CompuIves:
Hmm, checking now...

CompuIves:
I'm kind of kicking myself for this: https://github.com/codesandbox/codesandbox-client/pull/362. My theory is, and correct me if I'm wrong:

* `setImmediate` doesn't exist in browser
* We wanted to run projects that assume a `node` environment (just like Webpack, polyfill the things for `node`)
* that's how we came with `setImmediate`
* However, we need to make sure that it actually follows the `setImmediate` spec...

DeMoorJasper:
@CompuIves Yeah we should at least make it follow the spec, the reason I removed is because it's already a global variable, I think it's part of browserfs?

CompuIves:
Oh interesting, you're completely right! Then this is save to merge!

PR #5907, opened 2021-06-16
Feat/ab framework implementation + redesign sign banner

danilowoz:
This introduces the implementation of the AB framework across the apps (homepage and editor) in order to use the same user-id in both environments and it also persists the id from an anonymous user to a logged-in status.

We're going to ship an **A/A** test so that we'll able to make sure that Amplitude is getting the right values in the production, so then we'll be able to release a new one (A/B)

---

It also takes the opportunity to release a redesign for the sign-in banner at the very top in the editor, which is part of the next experiment.

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -77,7 +77,7 @@
     "@apollo/react-hooks": "^3.1.3",
     "@babel/plugin-transform-destructuring": "^7.5.0",
     "@babel/preset-env": "^7.5.5",
-    "@codesandbox/ab": "^1.0.2",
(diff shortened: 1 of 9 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5907/commits/564d3e38205acba521fafae1c7da6e4372ebf65a is at https://pr5907.build.csb.dev/s/new.

DannyRuchtie (approved):
Sweet!

tromika (approved):

PR #5920, opened 2021-06-22
feat(ab framework): update api url

danilowoz:


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5920/commits/3ce41dc86647b022ce1d97d19efbeb21cbdcfb9b is at https://pr5920.build.csb.dev/s/new.

CompuIves (approved):
Great! Merci!

packages/common/src/config/env.ts
```diff
@@ -8,7 +8,7 @@ const NODE_ENV = JSON.stringify(process.env.NODE_ENV || 'development');
 const LOCAL_SERVER = Boolean(JSON.stringify(process.env.LOCAL_SERVER));
 const STAGING_API = Boolean(JSON.stringify(process.env.STAGING_API));
 
-export const AB_TESTING_URL = 'https://ab-testing.codesandbox.workers.dev';
+export const AB_TESTING_URL = 'https://ab-testing.codesandbox.io';
```
danilowoz, line 11: It should be https://ab-testing.codesandbox.stream/ on staging and https://ab-testing.codesandbox.io/ in production.

packages/common/src/config/env.ts
```diff
@@ -8,7 +8,11 @@ const NODE_ENV = JSON.stringify(process.env.NODE_ENV || 'development');
 const LOCAL_SERVER = Boolean(JSON.stringify(process.env.LOCAL_SERVER));
 const STAGING_API = Boolean(JSON.stringify(process.env.STAGING_API));
 
-export const AB_TESTING_URL = 'https://ab-testing.codesandbox.workers.dev';
+const AB_TESTING_URL_STAGING = 'https://ab-testing.codesandbox.stream';
+const AB_TESTING_URL_PRODUCTION = 'https://ab-testing.codesandbox.io';
+export const AB_TESTING_URL = STAGING_API
```
danilowoz, line 13: @CompuIves Does it make sense? I afraid it doesn't work, as @lbogdan pointed out, we use the same build for staging and production, so the choice needs to be done at runtime.

I grabbed this logic from [here](https://github.com/codesandbox/codesandbox-client/blob/master/packages/common/src/components/Preview/index.tsx#L76), so hope it makes sense.

CompuIves: Oh yeah we could even just make it io at all times. That seems fine with me as well.

CompuIves (approved):

PR #5924, opened 2021-06-24
Fix tag fetching

CompuIves:
We used to use jsdelivr for resolving a tag to an absolute version, but
it turns out that jsdelivr caches the result. This is unintended
behavior, because when a tag updates people will see different results
based on their browser cache. This PR makes us default to unpkg instead,
which has a better caching mechanism.

### DIFF
--- a/packages/common/src/utils/dependencies.ts
+++ b/packages/common/src/utils/dependencies.ts
@@ -21,16 +21,34 @@ async function fetchWithRetries(url: string) {
 }
 
 export async function fetchPackageJSON(dep: string, version: string) {
-  try {
-    return fetchWithRetries(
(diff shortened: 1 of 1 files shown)


DeMoorJasper (approved):
Looks good to me, just wondering why we can't just always try unpkg first?

CompuIves:
I've noticed that UNPKG tends to give back a 500 error more often and takes longer to respond, this might've changed by now. Would be good to test.

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5924/commits/24b34f68637677f18a1c4ed40b7857e282b242a9 is at https://pr5924.build.csb.dev/s/new.

PR #5947, opened 2021-07-05
feat(vscode-extensions): update

danilowoz:


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5947/commits/64bf6f16c5404a96199ca7b5adc1f8d0ea1d0545 is at https://pr5947.build.csb.dev/s/new.

CompuIves (changes requested):
Because we update `codesandbox.editor.main.js`, we also need to update the `vscode` cached path (because we updated that now, not kidding 😂 ). So we need to update all instances of `vscode28` to `vscode29` and all instances of `vscode27` to `vscode28`. Then it's ready!

CompuIves (approved):
Great!! Let's merge this after a quick check on stream 😄 .

PR #5951, opened 2021-07-07
Allow to set Basic Auth using username and pw

CompuIves:
This is a more ergonomic way of adding Basic auth to your private npm. We used to ask a token for `Basic` auth, but people expect a `username` and `password` field.

Tested this with multiple combinations, verified that the generated auth token is also correct by comparing with npm and intercepting the requests with an http server to verify the auth tokens.

Screenshot 2021-07-07 at 21 23 00


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5951/commits/e53d08a586c76565b757fd3bc913be3e658d89aa is at https://pr5951.build.csb.dev/s/new.

packages/app/src/app/pages/Dashboard/Content/routes/Settings/TeamSettings/RegistryPreset/UsernamePassword.tsx
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    // First unencode the token, this is what npm does as well
+    const unencodedToken = Buffer.from(token, 'base64').toString('utf-8');
+
+    const key = `${username}:${unencodedToken}`;
+    const base64Data = Buffer.from(key);
+    const newAuthKey = base64Data.toString('base64');
+    setAuthKey(newAuthKey);
+  }, [username, token, setAuthKey]);
```
alexnm, line 44: this hook will be called every time you input another character in the username or password fields. if that's not concerning for performance reasons, that's fine, otherwise I'd switch to an `onBlur` event handler and keep the Inputs uncontrolled with `defaultValue` instead of `value`

packages/app/src/app/pages/Dashboard/Content/routes/Settings/TeamSettings/RegistryPreset/Custom.tsx
```diff
@@ -71,15 +73,23 @@ export const CustomRegistryPreset = ({
...
+      {authenticationType === AuthType.Basic ? (
+        <UsernamePassword
+          authKey={authKey}
           disabled={disabled}
-          type="password"
+          setAuthKey={setAuthKey}
         />
-      </CustomFormField>
+      ) : (
+        <CustomFormField label="Auth Token">
```
alexnm, line 83: It says only "Auth Token" here, not Token or Password

alexnm (approved):
looks good, left just two comments

packages/app/src/app/pages/Dashboard/Content/routes/Settings/TeamSettings/RegistryPreset/Custom.tsx
CompuIves: In this case that's okay, because for Bearer authentication most registries call it "Auth Token"

packages/app/src/app/pages/Dashboard/Content/routes/Settings/TeamSettings/RegistryPreset/UsernamePassword.tsx
CompuIves: Good catch!! I tested a bit more, and I think in terms of performance this is still fine, the conversion from base64 to utf-8 is fast enough, so I think we can keep this for now.

CompuIves:
Thanks for the review!!

packages/app/src/app/pages/Dashboard/Content/routes/Settings/TeamSettings/RegistryPreset/Custom.tsx
alexnm: Ah yes, I was looking at the wrong label

PR #5964, opened 2021-07-14
Simpler attempt at fixing html execution in sandpack and local env

DeMoorJasper:

This PR fixes issues when having scripts in html when locally using script tags and this also appends the content to the head tag so everything non-script works as well. See #5723 for more details...

The main difference between this and previous attempts is that this attempt tries to change as little as possible, it uses innerHTML as before with the only real difference that it extracts out the scripts from the html and uses document.createElement for these scripts instead to ensure they get fetched and executed.

Closes #5723 

## Checklist

- [x] Documentation
- [x] Testing 
- [x] Ready to be merged
      
### DIFF
--- a/packages/app/src/sandbox/compile.ts
+++ b/packages/app/src/sandbox/compile.ts
@@ -667,9 +667,19 @@ async function compile(opts: CompileOptions) {
 
       await manager.preset.preEvaluate(manager, updatedModules);
 
-      if (!manager.webpackHMR) {
-        const htmlEntries = templateDefinition.getHTMLEntries(configurations);
-        const htmlModulePath = htmlEntries.find(p => Boolean(modules[p]));
+      // HTML Hydration and HMR
+      // If the user code handles hot module reload, we don't modify html dynamically (users can refresh the page manually in this case)
+      if (!manager?.webpackHMR) {
+        let pathname = document.location.pathname;
+        if (!pathname.endsWith('.html')) {
+          pathname = '/index.html';
+        }
+        const htmlEntries = Object.keys(modules).filter(k =>
+          k.endsWith('.html')
+        );
+        const htmlModulePath = htmlEntries.find(
+          p => p === pathname || p === `/public${pathname}`
+        );
         const htmlModule = modules[htmlModulePath];
         let html =
           template === 'vue-cli'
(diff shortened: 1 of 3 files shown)


packages/app/src/sandbox/compile.ts
```diff
@@ -667,9 +668,19 @@ async function compile(opts: CompileOptions) {
...
-        const htmlEntries = templateDefinition.getHTMLEntries(configurations);
-        const htmlModulePath = htmlEntries.find(p => Boolean(modules[p]));
+      // HTML Hydration and HMR
+      // If the user code handles hot module reload, we don't modify html dynamically (users can refresh the page manually in this case)
+      if (!manager?.webpackHMR) {
+        let pathname = document.location.pathname;
+        if (!pathname.endsWith('.html')) {
+          pathname = '/index.html';
+        }
+        const htmlEntries = Object.keys(modules).filter(k =>
```
DeMoorJasper, line 678: Changed this logic as this ensures we can serve any html file in sandpack (even multi page ones)

packages/app/src/sandbox/compile.ts
```diff
@@ -701,17 +713,31 @@ async function compile(opts: CompileOptions) {
...
+              }
+              appendScripts(scripts, document.body);
+            }
+          }
 
           // Add head tags or anything that comes from the template
           // This way, title and other meta tags will overwrite whatever the bundler <head> tag has.
           // At this point, the original head was parsed and the files loaded / preloaded.
+          if (head !== lastHeadHTML) {
+            document.head.innerHTML += head;
```
DeMoorJasper, line 734: This will only get called on first load of local/sandpack environment so I think this is fine

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5964/commits/9ac90a2bf6a7fef8ab1fdd3da50d16e99d6b1b32 is at https://pr5964.build.csb.dev/s/new.

alexnm (approved):
Looks good, didn't get into the fine details with the review, but I'd test it with sandpack maybe on Friday when I'm back.

plondon:
Hey @alexnm wondering if you got a chance to check this out on sandpack? Thanks!

PR #5974, opened 2021-07-19
Fix measure warnings

DeMoorJasper:

Gets rid of the resolver measure warnings and probably also fixes a memory leak as these measures are kept in memory if not resolved which can happen quite easily as there's no cleanup call


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/5974/commits/62d830da4325a0b0dce9156a903909f249c8ed7e is at https://pr5974.build.csb.dev/s/new.

PR #6017, opened 2021-08-03
fix: Use prefetch for Babel instead of preload

DeMoorJasper:

Currently we're preloading Babel, but this throws a warning in chrome as we don't always use this script within a couple seconds, sometimes it's only used on a hot reload or not at all.

Throws a warning

## What is the new behavior?

No more warning


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6017/commits/1b321f451021a6c8255c0e8ab74fdcc041a42de3 is at https://pr6017.build.csb.dev/s/new.

PR #6021, opened 2021-08-05
fix(sandpack): manager id might undefined while consuming cache

danilowoz:
The error `undefined used as a key, but it is not a string.` has been solved in Sandpack. For some reason, the `manager` and its `id` might be `undefined` while sandpack is consuming the cache. To be honest, I'm not sure what this really means, or if this will be introducing any bug, but definitely, it wasn't working in the first place, plus it was polluting the console with warnings.

### DIFF
--- a/packages/sandpack-core/src/cache.ts
+++ b/packages/sandpack-core/src/cache.ts
@@ -187,7 +187,7 @@ export function ignoreNextCache() {
   }
 }
 
-export async function consumeCache(manager: Manager) {
+export async function consumeCache(manager?: Manager) {
   try {
     const shouldIgnoreCache =
       localStorage.getItem('ignoreCache') ||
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6021/commits/b38e9d62e353df73ce2cc0c6e2f92df7bf5db1e4 is at https://pr6021.build.csb.dev/s/new.

DeMoorJasper (changes requested):
Awesome, glad you've found where the warning is coming from.

I'm not entirely sure if this is the correct solution though, will have to look a bit more in depth in what the actual consequences are of not defining an id and whether we should pass something custom instead to ensure we don't call any codesandbox api's for react-sandpack instances.

cc @CompuIves maybe you have some ideas about this

PR #6023, opened 2021-08-06
Update job application count in header navigation on main site

twanmulder:
## What kind of change does this PR introduce?

Update job application count in header navigation on main site so it matches with the footer and current amount of roles available.

## What is the current behavior?

The old one says "2".

## What is the new behavior?

Now it's a "3".

## What steps did you take to test this? This is required before we can merge, make sure to test the flow you've updated.

1. Just updated the number from "2" to "3", nothing too complex.

## Checklist

- [ ] Documentation (N\A)
- [x] Testing
- [x] Ready to be merged
- [ ] Added myself to contributors table

PR #6012, opened 2021-08-02
Fixes #6011

ioana-chiorean:
3 Jobs listed, modified the code accordingly. 
https://github.com/codesandbox/codesandbox-client/issues/6011

## What steps did you take to test this? This is required before we can merge, make sure to test the flow you've updated.

1. Step A
2. Step B
3. Step C

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      
- [ ] Added myself to contributors table

PR #6024, opened 2021-08-06
Update /jobs and job counts

smhutch:

1. Added jobs link to mobile menu
2. Added "how we hire" section to /jobs
3. Prevent the job counts getting out of sync (https://github.com/codesandbox/codesandbox-client/pull/6012 + https://github.com/codesandbox/codesandbox-client/pull/6023). Ideally we would fetch this data from Recruitee, but I hard coded it in one place for now to avoid an additional request.

1. Job count was different in header and footer
2. Jobs UI got squashed on mobile due to fixed height

1. Open mobile menu, see link to jobs
2. On jobs page, see new content

## Checklist

- [ ] Documentation
- [x] Testing 
- [x] Ready to be merged
      

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6024/commits/592b3b8068ed39bf6cb007f322036d5873570817 is at https://pr6024.build.csb.dev/s/new.

packages/homepage/src/components/Navigation/elements.js
```diff
@@ -267,12 +254,7 @@ export const Items = styled.ul`
     }
   }
 
-  li a {
-    display: flex;
-    text-decoration: none;
-  }
-
-  span {
+  li a > span {
```
smhutch, line 257: Only target immediate children, not the JobsBadge

packages/homepage/src/pages/jobs/_elements.js
```diff
@@ -38,12 +38,11 @@ export const Job = styled.li`
   list-style: none;
   font-size: 19px;
   line-height: 33px;
-  height: 64px;
```
smhutch, line 41: Caused issues when wrapping on to two lines

packages/homepage/src/pages/jobs/_elements.js
```diff
@@ -38,12 +38,11 @@ export const Job = styled.li`
...
   color: ${props => props.theme.homepage.white};
   display: flex;
   justify-content: space-between;
-  align-items: center;
-  padding: 0 4px;
+  align-items: flex-start;
+  padding: 12px 4px;
```
smhutch, line 45: Padded added to replicate fixed height, but works on mobile too

packages/homepage/src/pages/jobs/_elements.js
```diff
@@ -38,12 +38,11 @@ export const Job = styled.li`
   list-style: none;
   font-size: 19px;
   line-height: 33px;
-  height: 64px;
   color: ${props => props.theme.homepage.white};
   display: flex;
   justify-content: space-between;
-  align-items: center;
-  padding: 0 4px;
+  align-items: flex-start;
```
smhutch, line 44: Looks better when job title wraps on to multiple lines

packages/homepage/src/pages/jobs/index.js
```diff
@@ -290,13 +336,13 @@ const Careers = () => {
                       `}
                     >
                       <Job>
-                        <span
+                        <strong
                           css={`
-                            font-weight: bold;
+                            margin-right: 1rem;
```
smhutch, line 341: Prevent job title from touching "remote"

packages/homepage/src/config/hiring.js
```diff
@@ -0,0 +1,2 @@
+// Should probably be fetched from recruitee, but it's hard coded for now to avoid an extra request.
+export const OPEN_JOBS_COUNT = 3;
```
danilowoz, line 2: You could use this implementation here to fetch the amount of open positions: https://github.com/codesandbox/codesandbox-client/blob/master/packages/homepage/src/pages/jobs/index.js#L33

danilowoz (approved):

packages/homepage/src/config/hiring.js
smhutch: Yeah this was my initial thought too, then I realised it would be firing off three requests on every page. I think we need to start using something like SWR or ReactQuery to make requests like this with caching to avoid too many network requests. WDYT?

danilowoz: Agreed, it makes sense. A React-context would also work to avoid extra requests, then it wouldn't increase the bundle with one more package in the application.

CompuIves: Makes sense to me! Another thing we could do is a cache like this, if it's easier (I'm not familiar enough with react-context):

```ts
let jobsFetchRequest: Promise<Job[]> | null = null;

function fetchJobs() {
  jobsFetchRequest = jobsFetchRequest || fetch(...);

  return jobsFetchRequest;
}

```

But that would be more hacky, though I'm not sure if we'll get more cases where we need caching 🤔 .

smhutch: [Task](https://linear.app/codesandbox/issue/CSB-980/fetch-job-count-from-recruitee) for later

PR #6027, opened 2021-08-09
fix(sandpack): Manager typings should reflect that id can be undefined in case of react-sandpack

DeMoorJasper:

Currently the manager types suggest id is always defined, which is not the case for react-sandpack. This is a copy from #6021 but instead of only checking this in consumeCache this fixes the manager types to reflect what the state is in react-sandpack as there's some other places where this could cause issues.

Closes #6021 

Manager has an id property that is typed as stirng

## What is the new behavior?

Manager has an id property that is typed as string, null or undefined

### DIFF
--- a/packages/app/src/sandbox/compile.ts
+++ b/packages/app/src/sandbox/compile.ts
@@ -777,13 +777,7 @@ async function compile(opts: CompileOptions) {
       type: 'success',
     });
 
-    saveCache(
-      sandboxId,
-      managerModuleToTranspile,
-      manager,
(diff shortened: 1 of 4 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6027/commits/3d5a5c95daf7ec3e49514db8a47001e6b6166c3d is at https://pr6027.build.csb.dev/s/new.

danilowoz (approved):

PR #6030, opened 2021-08-09
Fix issues with jspm and loader import syntax conflicts

DeMoorJasper:

Currently loading the module `https://jspm.dev/npm:shopify-buy@2.11.0!cjs` fails due to it having an explanation mark, which we also use to specify loaders in imports...

Should we consider replacing this `!` with something else like `|>` or something, to make it more explicit and less likely we conflict with an actual character someone might use?

Throws an error

## What is the new behavior?

Does not throw an error


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6030/commits/136642aacf523499ef31ca77ee7fc18e9004d7e4 is at https://pr6030.build.csb.dev/s/new.

PR #6040, opened 2021-08-13
fix(dashboard beta): feature flag wrapper over the component instead of the route

danilowoz:
Related to CSB-918

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Content/index.tsx
+++ b/packages/app/src/app/pages/Dashboard/Content/index.tsx
@@ -57,9 +57,15 @@ export const Content = withRouter(({ history }) => {
       })}
     >
       <Switch>
-        {isFeatureFlagBeta && (
-          <Route path="/dashboard/beta" component={BetaRepositoriesPage} />
-        )}
+        <Route
+          path="/dashboard/beta"
+          /**
+           * Wrap the component instead of the whole route, as the
+           * feature-flag query takes a few ms to load, and meanwhile the Route
+           * redirect the page because the route hasn't been registered yet
+           */
+          component={isFeatureFlagBeta ? BetaRepositoriesPage : () => null}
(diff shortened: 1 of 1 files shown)


packages/app/src/app/pages/Dashboard/Content/index.tsx
```diff
@@ -57,9 +57,15 @@ export const Content = withRouter(({ history }) => {
...
-          <Route path="/dashboard/beta" component={BetaRepositoriesPage} />
-        )}
+        <Route
+          path="/dashboard/beta"
+          /**
+           * Wrap the component instead of the whole route, as the
+           * feature-flag query takes a few ms to load, and meanwhile the Route
+           * redirect the page because the route hasn't been registered yet
+           */
+          component={isFeatureFlagBeta ? BetaRepositoriesPage : () => null}
```
smhutch, line 67: Would something like this work to avoid the possibility of rendering a blank page?

```suggestion
          component={isFeatureFlagBeta ? BetaRepositoriesPage : () => <Redirect to={dashboardUrls.home(activeTeam)} />}
```

smhutch (approved):
Nice find 💯

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6040/commits/d622bd3affd911e1edc6197a21443ddfad000bfb is at https://pr6040.build.csb.dev/s/new.

packages/app/src/app/pages/Dashboard/Content/index.tsx
danilowoz: This would end up with the same behavior as before because the Redirect would be rendered anyway. A final and right solution would need some kind of loading state, but I think it's over-engineered for this beta version in a scenario that we control who has access to this.

PR #6042, opened 2021-08-16
Use semver for remote transient deps

DeMoorJasper:

Fixes a bug with transient dependencies that are urls to globs/git repos/...

See https://codesandbox.io/s/summer-glade-ztrpm for an example

Bug report https://twitter.com/acdlite/status/1427001781012373505

We always try to fetch the version from unpkg for transient deps

## What is the new behavior?

We check if the transient dep is remote and fetch from remote source instead


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6042/commits/1c1ff6ba0321486f0b23528b697d5c01ee4a10f2 is at https://pr6042.build.csb.dev/s/new.

PR #6043, opened 2021-08-16
fix: Rewrite paths if esmodule url is a redirect

DeMoorJasper:
## What kind of change does this PR introduce?

Fixes a bug with esmodule sandboxes where asset paths aren't rewritten to the resolved url

## What is the current behavior?

Redirects don't influence module path

## What is the new behavior?

Redirects change the module path

PR #6044, opened 2021-08-17
fix: issue with resolved url not caching properly

DeMoorJasper:

A fix for #6043 breaking on cached builds as we don't rewrite requires (which we should probably change at some point), adding a url property to module is probably a better idea for now while we figure out how to resolve this properly.

Throws cannot resolve esmodule url synchronous

## What is the new behavior?

It doesn't change path to the resolved path and uses url for meta.url


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6044/commits/61dcfde10ca00abf0fe5281e867eabff3fd7461f failed.

PR #6054, opened 2021-08-19
feat: Rewrite file resolver

DeMoorJasper:

This PR is a ground up rewrite of the resolver, similar to `browser-resolve` but with all the modern features like aliases and exports baked in. This way we no longer have to hack on top of browser-resolve, this is also a bit faster.

To ensure we don't have to maintain 2 versions (sync & async) I've used a util built for Babel called `gensync` it made this sooo easy, we should probably use it in more places in the codebase at some point.

Limited control over resolution logic using browser-resolve and a couple hacks

## What is the new behavior?

Full control of resolve logic and a cleaner & faster resolver


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6054/commits/7821bc737c9412fc80bf562a892a533117525866 is at https://pr6054.build.csb.dev/s/new.

PR #6062, opened 2021-08-23
feat(preview): attach document event

danilowoz:
This change aims to track when the user interacts with the preview, which will improve the session metric and how the user spends their time in the code editor. So in order to make this work, I needed to find out a way to back and forth message from the iframe to the main application, **as it is not possible to attach a listener to an iframe.**

### DIFF
--- a/packages/app/src/sandbox/compile.ts
+++ b/packages/app/src/sandbox/compile.ts
@@ -794,6 +794,13 @@ async function compile(opts: CompileOptions) {
         }
       }
     }, 600);
+
+    /**
+     * For tracking purpose
+     */
+    document.addEventListener('click', () => {
+      dispatch({ type: 'document-focus' });
+    });
   } catch (e) {
     // eslint-disable-next-line no-console
     console.log('Error in sandbox:');
--- a/packages/common/src/components/Preview/index.tsx
+++ b/packages/common/src/components/Preview/index.tsx
@@ -13,7 +13,7 @@ import { getModulePath } from '../../sandbox/modules';
 import getTemplate from '../../templates';
 import { generateFileFromSandbox } from '../../templates/configuration/package-json';
 import { Module, NpmRegistry, Sandbox } from '../../types';
-import track from '../../utils/analytics';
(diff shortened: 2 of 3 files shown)


packages/app/src/sandbox/compile.ts
```diff
@@ -794,6 +794,13 @@ async function compile(opts: CompileOptions) {
         }
       }
     }, 600);
+
+    /**
+     * For tracking purpose
+     */
+    document.addEventListener('click', () => {
+      dispatch({ type: 'document-focus' });
+    });
```
danilowoz, line 803: Does it make sense to place this piece of code here?

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6062/commits/a66989713c1022efb49c52f1adb5cc1554ccec99 is at https://pr6062.build.csb.dev/s/new.

packages/app/src/sandbox/compile.ts
DeMoorJasper: Not really, it's pretty confusing and this function needs a ton of cleanup. But as far as I know it gets called on every compilation so even on hot reloads, resulting in a ton of event listeners at some point.

This is also true for the `initializeDOMMutationListener` function call at the bottom.

The correct place for this and the other function I mentioned would be `packages/app/src/sandbox/index.js`

packages/common/src/components/Preview/index.tsx
```diff
@@ -363,6 +363,10 @@ class BasePreview extends React.PureComponent<Props, State> {
             this.setState({ showScreenshot: false });
             break;
           }
+          case 'document-focus': {
+            trackWithCooldown('Browser focus', 5000);
```
tromika, line 367: @danilowoz I think it would be enough to send an event every 30 sec. Also, I'd name it as `Preview focus`.
The rest looks good to me!

DeMoorJasper (approved):

tromika (approved):
Thanks so much for the changes!

PR #6079, opened 2021-08-30
fix: homepage overflow

smhutch:
Fixes the overflow that occurs on the homepage:

Screen Shot 2021-08-29 at 16 03 12

### DIFF
--- a/packages/homepage/src/css/global.css
+++ b/packages/homepage/src/css/global.css
@@ -12,7 +12,7 @@ body {
   color: white;
   font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen,
     Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue';
-  /* overflow-x: hidden; */
+  overflow-x: hidden;
 }
 
 @media screen and (-webkit-max-device-pixel-ratio: 1.3),


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6079/commits/778ca01722e9562549aef1fb9ad286021d7b65e0 is at https://pr6079.build.csb.dev/s/new.

packages/homepage/src/css/global.css
```diff
@@ -12,7 +12,7 @@ body {
   color: white;
   font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen,
     Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue';
-  /* overflow-x: hidden; */
+  overflow-x: hidden;
```
danilowoz, line 15: Maybe uncommenting this line is not a big deal, or maybe this line was commented on for some unknown reason. I'd try to apply this in the component that has the problem and doesn't globally. As I could see, the following section is bigger than the window width in its initial state, but after scrolling, it fits into the window.
Screenshot 2021-08-31 at 09 11 29

danilowoz (approved):

PR #6121, opened 2021-09-09
fix: Data uri binary files

DeMoorJasper:

Binary files should simply return the uri, no need to base64 encode it especially not encoding the uri


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6121/commits/bd25941586b9f9aa4ded5bf3748c3817acbfed33 is at https://pr6121.build.csb.dev/s/new.

PR #6136, opened 2021-09-14
fix: conflict between core-js and bfs-process

### LINKED ISSUE #5943: core-js >=3.12 throws errors `Target is not iterable`
# 🐛 bug report

## Preflight Checklist

      adheres to.
      to file, without success.

## Description of the problem

Local usage of `core-js >= 3.12` doesn't have issues, whereas `codesandbox` throws `Target is not iterable`.

I try to keep dependencies up to date.

### To Reproduce

- go to [https://codesandbox.io/](https://codesandbox.io/)
- click on `Angular angular-cli`
- wait when the app has been rendered
* change in `dependencies` the version of `core-js` to `3.11.3`
* click refresh in the builtin browser
* wait when the app has been rendered
- change in `dependencies` the version of `core-js` to `3.12.0`
- click refresh in the builtin browser
- the error appears

if you download the project and run it locally, it works correctly.

DeMoorJasper:

This PR fixes an issue with core-js and bfs-process conflicting.

We definitely need to have a look at how to prevent all the global scope poluting we (and mainly browser-fs) does though, as these bugs can be prevented pretty easily that way.

Also probably a good idea to revisit the eval code at some point as it also has some weird behaviour with core-js, specifically this line: https://github.com/zloirock/core-js/blob/master/packages/core-js/internals/global.js#L2 

Also this one for detecting v8 engine using the nodeJS path: https://github.com/zloirock/core-js/blob/master/packages/core-js/internals/engine-v8-version.js

Closes #5943

### DIFF
--- a/packages/sandpack-core/src/runner/eval.ts
+++ b/packages/sandpack-core/src/runner/eval.ts
@@ -41,6 +41,7 @@ export default function (
   }
 
   // This is a hack to make core-js work...
+  // See https://github.com/codesandbox/codesandbox-client/pull/6136
   if (window.process.versions) {
     window.process.versions.v8 = '50.0';
   }
--- a/packages/sandpack-core/src/runner/eval.ts
+++ b/packages/sandpack-core/src/runner/eval.ts
@@ -5,6 +5,22 @@ const g = typeof window === 'undefined' ? self : window;
 
 const hasGlobalDeclaration = /^const global/m;
 
+const V8_VERSION = getV8Version();
+function getV8Version(): string {
+  const userAgent = navigator?.userAgent;
+  if (userAgent) {
+    let match = userAgent.match(/Edge\/(\d+)/);
+    // @ts-ignore
+    if (!match || match[1] >= 74) {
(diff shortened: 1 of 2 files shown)


packages/sandpack-core/src/runner/eval.ts
```diff
@@ -40,6 +40,11 @@ export default function (
     delete allGlobals.global;
   }
 
+  // This is a hack to make core-js work...
+  if (window.process.versions) {
```
DeMoorJasper, line 44: This is a pretty big hack, see the PR description for better (bigger) solutions/refactors

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6136/commits/9b6daccce63081a333940d0973d7aa22608ddc32 is at https://pr6136.build.csb.dev/s/new.

CompuIves (approved):
Makes sense, I would perhaps also put in the comment a link to this PR to make sure that we know exactly why it's happening. Very tricky, some things need `window.process`, other things break with it...

PR #6148, opened 2021-09-18
feat: New resolver and exports support

### LINKED ISSUE #6147: ESM: Support for Subpath patterns
# 🌈 Feature

Ability to handle [subpath patterns](https://nodejs.org/api/packages.html#packages_subpath_patterns) 

### What is this feature? 
A few libraries with a lot of exports would prefer to use the above pattern while creating exports and this current breaks in CSB example sandbox https://codesandbox.io/s/kind-payne-i2eb3?file=/src/App.tsx for https://github.com/pmndrs/jotai 

The feature or maybe even a bug would be to add resolution handlers for the `exports` field to have glob patterns and absolute file paths.

### How the feature should work? 

> The feature or maybe even a bug would be to add resolution handlers for the `exports` field to have glob patterns and absolute file paths.

### You have examples or an idea how it can be implemented? 

You can follow the node docs https://nodejs.org/api/packages.html#packages_subpath_patterns , as this was added to both v12 and v14 and already supported by v16 

I wouldn't mind raising a PR but I was unable to find the point of generation for packages. I'm assuming it's the `sandpack` package code but I'm not sure

DeMoorJasper:

Attempt 2 at adding exports support.

Add and fix the following integration tests before merging:

- https://codesandbox.io/s/gpgpu-curl-noise-dof-forked-yq6y8
- https://codesandbox.io/s/adoring-brook-vx55c

Closes #6147


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6148/commits/a5eb36b0b1c75ce1f1b0757f425fede7640da8e0 is at https://pr6148.build.csb.dev/s/new.

PR #6204, opened 2021-10-15
Update support email on homepage

smhutch:
Updates emails on the homepage


JamesACS:
Yep, that would be fantastic!

PR #6240, opened 2021-10-28
chore: add Ace to the team page

acekyd:

1. Step A
2. Step B
3. Step C

## Checklist

- [ ] Documentation
- [x] Testing 
- [x] Ready to be merged
      

packages/homepage/src/pages/company/index.js
```diff
@@ -232,6 +232,7 @@ export default () => (
         <TeamMember name="roman" />
         <TeamMember name="marco" />
         <TeamMember name="james" />
+        <TeamMember name="ace" />
```
gsimone, line 235: This doesn't quite work because the component will try to find the string `ace` in the names and won't find it. You can fix it in two ways:
- add your nick to the name, like `name: 'Adewale "Ace" Abati'` (which I think would be kinda cool)
- or change the `name` property to "adewale" so `<TeamMember name="adewale"`

gsimone (changes requested):

packages/homepage/src/pages/company/index.js
acekyd: Thanks for the catch @gsimone.

gsimone:
LGTM

gsimone (approved):

PR #6241, opened 2021-10-28
chore: adds maria to team page

olarclara:

- Adds myself to the team page. 🎉 
- Adds `object-fit: cover` to team member picture;

Team member profile pictures that weren't `1:1` would look "squished".

## What is the new behavior?

If the profile picture isn't `1:1` it will scale to cover the container.

1. Step A
2. Step B
3. Step C

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      

packages/homepage/src/components/TeamMember.js
```diff
@@ -142,6 +143,12 @@ const people = [
     pic: james,
     color: '#6200ff',
     team: 'Support',
+  },
+  {
+    name: 'Maria Clara',
+    pic: maria,
+    color: 'F24E62',
```
gsimone, line 150: I think you need the # here or the highlight color won't work

gsimone (changes requested):

packages/homepage/src/components/TeamMember.js
olarclara: @gsimone thanks for the catch! fixed it here: 48061a3

gsimone (approved):

PR #6254, opened 2021-11-04
fix(import): remove stackbit

danilowoz:
As Stackbit is deprecating its import utility, so we're removing it from our import section.

@CompuIves should we remove [this](https://github.com/codesandbox/codesandbox-client/blob/master/packages/common/src/templates/unibit.ts#L19) template as well? Also, I didn't find anything in `/docs`, so I'm assuming that there is no other mention about Stackbit, right? 

Screenshot 2021-11-04 at 11 04 33


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6254/commits/b0797aba5b1bef8f9a2ad123bfc0b11f847ce9d3 is at https://pr6254.build.csb.dev/s/new.

CompuIves (approved):
Code-wise this looks good to me!!

DannyRuchtie:
Screenshot 2021-11-04 at 13 09 12
@danilowoz could you set a max-width so that the content feels a bit more compact?

DannyRuchtie (changes requested):
Screenshot 2021-11-04 at 13 09 12
Could you give it a max-width so that the content is a bit more compact easier to scan

DannyRuchtie (approved):
Awsome!

PR #6255, opened 2021-11-04
Reduce retry count for jsdelivr if it fails to load

CompuIves:
When JSDelivr fails (which I think it did more often recently), sandboxes take ages to load. This reduces the try count, because knowing JSDelivr, if it fails twice, it will most probably fail 3rd, 4th etc...

### DIFF
--- a/packages/sandpack-core/src/npm/dynamic/fetch-protocols/utils.ts
+++ b/packages/sandpack-core/src/npm/dynamic/fetch-protocols/utils.ts
@@ -2,7 +2,7 @@ import delay from '@codesandbox/common/lib/utils/delay';
 
 export async function fetchWithRetries(
   url: string,
-  retries = 6,
+  retries = 2,
   requestInit?: RequestInit
 ): Promise<Response> {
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6255/commits/4c94c9cc327532a1b024189504678883a5819de6 is at https://pr6255.build.csb.dev/s/new.

MartinKolarik:
@CompuIves do you have any additional info about what kind of failures you are seeing? We'd be happy to look into it if you think there's anything that could be improved on our side.

CompuIves:
Hey @MartinKolarik, thanks for following up! I should've sent a message.

We saw some failed requests on the React docs here:
image

I think it was local, because I wasn't able to repro it. JSDelivr generally is very stable, but I think our retry count was a bit too excessive in this case.

MartinKolarik:
I see, do you know what the response status was by any chance and roughly when it happened? But I agree 6 retries is probably too many in general.

PR #6265, opened 2021-11-09
feat(dashboard beta): launch default branch

danilowoz:


packages/common/src/utils/url-generator.ts
```diff
@@ -183,11 +183,16 @@ export const optionsToParameterizedUrl = (options: Object) => {
...
-  githubUrl
-    .replace(gitHubPrefix, '/github')
+export const gitHubToSandboxBetaUrl = (githubUrl: string) => {
+  const [owner, repo] = githubUrl
+    .replace(gitHubPrefix, '')
     .replace(dotGit, '')
-    .replace(/\/tree\//, '/');
+    .replace(/\/tree\//, '/')
+    .split('/')
+    .filter(Boolean);
```
alexnm, line 192: Trying to figure out why the filter is needed at the end, can there be duplicate slashes in the url?

alexnm (approved):

packages/common/src/utils/url-generator.ts
danilowoz: The initial URL looks like this `/owner/project`, which would split into `["", "owner", "project:]`, does it make sense?

PR #6274, opened 2021-11-15
Make sure we use jsdelivr for resolving absolute versions

CompuIves:
Unpkg can sometimes take over 30s to load the version, which results in the bundler being _really_ slow. JSDelivr is more reliable for these kinds of things.

### DIFF
--- a/packages/common/src/utils/dependencies.ts
+++ b/packages/common/src/utils/dependencies.ts
@@ -1,8 +1,8 @@
-import { valid } from 'semver';
+import { valid, maxSatisfying } from 'semver';
 
-async function fetchWithRetries(url: string) {
+async function fetchWithRetries<T = any>(url: string): Promise<T> {
   let err: Error;
-  for (let i = 0; i < 5; i++) {
+  for (let i = 0; i < 2; i++) {
     try {
       // eslint-disable-next-line
       return await fetch(url).then(x => {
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6274/commits/bf3faa50af3176beb7d8d28aba7226f22a1b25ef is at https://pr6274.build.csb.dev/s/new.

packages/common/src/utils/dependencies.ts
```diff
@@ -42,10 +58,14 @@ export async function fetchPackageJSON(dep: string, version: string) {
...
+    // to see what's the maximum satisfying version. The API call is cached for only 10s.
     try {
-      return await fetchUnpkg();
+      const allVersions = await fetchAllVersions(dep);
+      return (
+        allVersions.tags[version] ||
+        maxSatisfying(allVersions.versions, version)
+      );
     } catch (e) {
       return fetchJsdelivr();
```
DeMoorJasper, line 70: Shouldn't it just fail in this case as the try is also jsdelivr?

packages/common/src/utils/dependencies.ts
```diff
@@ -42,10 +58,14 @@ export async function fetchPackageJSON(dep: string, version: string) {
     // If it is not an absolute version (e.g. a tag like `next`), we don't want to fetch
     // using JSDelivr, because JSDelivr caches the response for a long time. Because of this,
     // when a tag updates to a new version, people won't see that update for a long time.
-    // Unpkg does handle this nicely, but is less stable. So we default to JSDelivr, but
-    // for tags we use unpkg.
+    // Instead, we download all possible versions from JSDelivr, and we check those versions
+    // to see what's the maximum satisfying version. The API call is cached for only 10s.
     try {
-      return await fetchUnpkg();
+      const allVersions = await fetchAllVersions(dep);
```
DeMoorJasper, line 64: This is so much better :D

packages/common/src/utils/dependencies.ts
CompuIves: OH BOY, I merged before, apologies! I think this is actually better to fall back to unpkg yes...

CompuIves: Will add that as additional commit./

PR #6276, opened 2021-11-15
performance: Don't fetch version for absolute versions

DeMoorJasper:

If a version is absolute we don't need to do an additional resolve, this PR returns early in case of absolute versions.

### DIFF
--- a/packages/common/src/utils/dependencies.test.ts
+++ b/packages/common/src/utils/dependencies.test.ts
@@ -7,6 +7,12 @@ describe('dependencies', () => {
       expect(isAbsoluteVersion('154.78.7896')).toBe(true);
       expect(isAbsoluteVersion('1.1.0-next.1')).toBe(true);
       expect(isAbsoluteVersion('154.78.123-alpha.1-235f8723g5oi23')).toBe(true);
+      expect(isAbsoluteVersion('codesandbox/sandpack')).toBe(true);
+      expect(isAbsoluteVersion('codesandbox/sandpack#0.3.0')).toBe(true);
+      expect(isAbsoluteVersion('codesandbox/sandpack-bundler#0.3.0')).toBe(
+        true
+      );
+      expect(isAbsoluteVersion('codesandbox/sandpack-bundler')).toBe(true);
     });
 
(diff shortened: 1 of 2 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6276/commits/4f32c6c4069fbdff61e9134848d18161cfaa218b failed.

packages/common/src/utils/dependencies.test.ts
```diff
@@ -0,0 +1,19 @@
+import { isAbsoluteVersion } from './dependencies';
+
+describe('dependencies', () => {
+  describe('isAbsoluteVersion', () => {
+    it('returns true on absolute versions', () => {
+      expect(isAbsoluteVersion('1.1.0')).toBe(true);
+      expect(isAbsoluteVersion('154.78.7896')).toBe(true);
+      expect(isAbsoluteVersion('1.1.0-next.1')).toBe(true);
+      expect(isAbsoluteVersion('154.78.123-alpha.1-235f8723g5oi23')).toBe(true);
```
CompuIves, line 9: I would also add another check here for `codesandbox/sandpack`

CompuIves: GitHub versions!

CompuIves (approved):
Nice improvements! I would make one small change, then ready.

packages/common/src/utils/dependencies.ts
```diff
@@ -33,28 +33,26 @@ async function fetchAllVersions(dep: string): Promise<JsDelivrApiResult> {
...
+const resolveVersionFromUnpkg = (
+  dep: string,
+  version: string
+): Promise<string> => {
+  return fetchWithRetries(
+    `https://unpkg.com/${dep}@${encodeURIComponent(version)}/package.json`
+  ).then(x => x.version);
+};
+
 export async function getLatestVersion(
```
CompuIves, line 46: I triple checked, I think this function is nowhere else used. We could remove the `export` here.
```suggestion
async function getLatestVersion(
```

PR #6280, opened 2021-11-16
Fix exports failing when they include an extension

### LINKED ISSUE #6279: ModuleNotFoundError when importing from a subdirectory of a module
# 🐛 bug report

## Preflight Checklist

      adheres to.
      to file, without success.

## Description of the problem

This is a very recent regression. A sandbox that worked correctly just last week no longer works.

When trying to import from a subdirectory of a module, I get a `ModuleNotFoundError`.

```
ModuleNotFoundError
Could not find module in path: '@zendesk/laika/esm/laika' relative to '/src/index.ts'
```

### To Reproduce

- Add `@zendesk/laika@1.0.0` as dependency
- `import { Laika } from '@zendesk/laika/esm/laika'` or `import { Laika } from '@zendesk/laika/cjs/laika'`
- either import will result in the ModuleNotFoundError error

### Link to sandbox: [link](https://codesandbox.io/s/laika-example-3yyq0)

For comparison, see the same code at [StackBlitz](https://stackblitz.com/edit/laika-example?file=index.tsx) working correctly

### Your Environment

| Сodesandbox      |   2021-11-15   |
| Browser          |  any  |
| Operating System |   any  |

DeMoorJasper:
# Pull Request

Currently if an exports field contains an extension we keep adding the extension resulting in a path with the same extension multiple times... this PR fixes that by checking if the new addition to the string was the same as the end of the string.

It's not perfect but not sure how else to fix this without largely refactoring the resolver again...

Closes #6279

### DIFF
--- a/packages/sandpack-core/src/resolver/resolver.ts
+++ b/packages/sandpack-core/src/resolver/resolver.ts
@@ -240,7 +240,7 @@ function loadAlias(pkgJson: IFoundPackageJSON, filename: string): string {
     relativeFilepath = aliasedPath;
 
     // Simply check to ensure we don't infinitely alias files due to a misconfiguration of a package/user
-    if (count > 10) {
+    if (count > 5) {
       throw new Error('Could not resolve file due to a cyclic alias');
     }
     count++;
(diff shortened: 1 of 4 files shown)


packages/sandpack-core/src/resolver/resolver.ts
```diff
@@ -357,7 +363,7 @@ function* expandFile(
 ): Generator<any, string | null, any> {
   const pkg = yield* findPackageJSON(filepath, opts);
 
-  if (expandCount > 15) {
+  if (expandCount > 5) {
```
DeMoorJasper, line 366: Also reduced these a bit as 15 seemed a bit extreme...

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6280/commits/b375b5bc88a7a7cf67d6af0ece67e29c1646b73a is at https://pr6280.build.csb.dev/s/new.

PR #6290, opened 2021-11-17
Update React preset to no longer default to outdated version and cleanup code in general

DeMoorJasper:

Removed react-scripts v1 template as v1 is very old and we should probably default to a newer one.

Also merged v3 and v4 as the only difference is the babel config of the react babel preset


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6290/commits/ad8079996c527f59bc2565acd91ef77804042c0a is at https://pr6290.build.csb.dev/s/new.

packages/app/src/sandbox/eval/presets/create-react-app/utils/reactPreset.ts
```diff
@@ -45,15 +45,27 @@ export const reactPreset = babelConfig => {
...
-        return dependencies;
+        if (!deps['@babel/core']) {
+          deps['@babel/core'] = '^7.3.3';
+        }
+
+        if (!deps['@babel/runtime']) {
+          deps['@babel/runtime'] = '^7.3.4';
+        }
+
+        delete deps['babel-runtime'];
```
DeMoorJasper, line 65: This is to cleanup some deps we don't need, we can also start deprecating babel 6 if we want

PR #6330, opened 2021-12-01
re-enable experimental esmodule resolution

DeMoorJasper:

1. Step A
2. Step B
3. Step C

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      
### DIFF
--- a/packages/app/src/sandbox/eval/index.ts
+++ b/packages/app/src/sandbox/eval/index.ts
@@ -33,8 +33,13 @@ import customPreset from './presets/custom';
 export default async function getPreset(template: string, pkg: PackageJSON) {
   switch (template) {
     case esmReact.name:
-    case react.name:
-      return reactPreset(pkg);
+    case react.name: {
+      const preset = await reactPreset(pkg);
+      if (template === esmReact.name) {
+        preset.experimentalEsmSupport = true;
+      }
+      return preset;
+    }
     case preact.name:
       if (isPreact10(pkg.dependencies, pkg.devDependencies)) {
         return preactPreset();


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6330/commits/1e96de35c859b32e5feb227c856a73236dc98e6b is at https://pr6330.build.csb.dev/s/new.

darrinm:
This is great to see @DeMoorJasper, thanks! How can I create a new project using the esm-react preset? Can I configure sandpack to use it as well?

DeMoorJasper:
@darrinm yes, you can use the customSetup for this and set `environment` to `esm-react`, here's a sandbox with an example: https://codesandbox.io/s/sandpack-project-forked-o7tvk?file=/App.js

It's apparently missing in the typings though

darrinm:
> @darrinm yes, you can use the customSetup for this and set `environment` to `esm-react`, here's a sandbox with an example: https://codesandbox.io/s/sandpack-project-forked-o7tvk?file=/App.js

That did the trick. Thanks!

darrinm:
There's probably a better place to ask this but, why is ESM support coupled with React? Seems like it should be independent.

DeMoorJasper:
> There's probably a better place to ask this but, why is ESM support coupled with React? Seems like it should be independent.

This is because we launched this preview together with Framer's esmodule components which are React and we wanted to test it on a small number of sandboxes first. It's totally possible to make it work for every sandbox just have to remove the flag

darrinm:
> > There's probably a better place to ask this but, why is ESM support coupled with React? Seems like it should be independent.
> 
> This is because we launched this preview together with Framer's esmodule components which are React and we wanted to test it on a small number of sandboxes first. It's totally possible to make it work for every sandbox just have to remove the flag

Got it, thanks.

PR #6340, opened 2021-12-10
Disable cache on vue preset for now and cleanup the vue transpiler a bit

DeMoorJasper:

Disable cache on vue preset for now and cleanup the vue transpiler a bit

### DIFF
--- a/packages/app/src/sandbox/compile.ts
+++ b/packages/app/src/sandbox/compile.ts
@@ -620,11 +620,14 @@ async function compile(opts: CompileOptions) {
 
     if (shouldReloadManager || firstLoad) {
       // Now initialize the data the manager can only use once dependencies are loaded
-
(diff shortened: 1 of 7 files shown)


CompuIves:
Ah but sandboxes are probably broken right now?

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6340/commits/9f528c693d29e287981d4a81825935459ba5fab5 is at https://pr6340.build.csb.dev/s/new.

DeMoorJasper:
Yes it would make vue slower although from what I've tested starting from cache vs not is not that much different for vue anyway. But yes sandboxes are broken and this would temporarily fix it, while we figure out a better long-term solution that prevents issues like this and potentially improves performance as well.

PR #6356, opened 2021-12-15
Upgrade/react devtools

danilowoz:
Adds support a new version of `react-devtools-inline` only for the Sandpack scope, and keep the old version (as a legacy) working as before. 

I needed to split the initializer into two different modules because `react-devtools-inline` has as dependency a React version way ahead of the codesandbox.io, which include break changes, that's why it's loading two different versions of the same package, but for different scopes.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6356/commits/7f77b5d39398b772920071452da9bd2d069ec70a is at https://pr6356.build.csb.dev/s/new.

CompuIves (approved):
Makes sense to me! 👍 

Would it be hard to introduce the new devtools to editor as well?

DeMoorJasper (approved):

danilowoz:
@CompuIves Yes, it would. This was my first thought as well, but a bunch of dependencies start to stop working (reactkit, for example), and need to be upgraded which included break changes in components. Plus, as `react-devtools-inline` depends on `react@experimental`, it's too risky to take now in order to support a component that is already working properly for most of the React version.

PR #6487, opened 2022-02-23
fix: Resolver recursion in `getDependencyVersion`

DeMoorJasper:

Fixes an infinite recursion between `getDependencyVersion` and `resolvePath` in fetch-npm-module

Sandbox to see this in action: https://codesandbox.io/s/yq0z0j (beware this will likely crash your browser-tab)

### DIFF
--- a/packages/sandpack-core/src/npm/dynamic/fetch-npm-module.ts
+++ b/packages/sandpack-core/src/npm/dynamic/fetch-npm-module.ts
@@ -141,7 +141,8 @@ function resolvePath(
   currentTModule: TranspiledModule,
   manager: Manager,
   defaultExtensions: Array<string> = DEFAULT_EXTENSIONS,
-  meta: Meta = {}
+  meta: Meta = {},
+  ignoreDepNameVersion: string = ''
 ): Promise<string> {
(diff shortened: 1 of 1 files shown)


CompuIves (approved):
Good find. LGTM!

One case I had in mind, sometimes a dependency calls itself. E.g.

```js
// node_modules/a/b.js

require('a/a'); // <- should resolve to node_modules/a/a.js
```

I've seen this case before once, do you think that still works?

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6487/commits/a45529d93d21aec0476622b74af86763191edb50 is at https://pr6487.build.csb.dev/s/new.

DeMoorJasper:
> Good find. LGTM!
> 
> 
> 
> One case I had in mind, sometimes a dependency calls itself. E.g.
> 
> 
> 
> ```js
> 
> // node_modules/a/b.js
> 
> 
> 
> require('a/a'); // <- should resolve to node_modules/a/a.js
> 
> ```
> 
> 
> 
> I've seen this case before once, do you think that still works?

That should still work afaik, there's also an integration test for that case so tests should catch that

PR #6543, opened 2022-03-15
updated homepage banner

DannyRuchtie:
Updated the banners that linked to the youtube channel to the once that redirect to csbp


DannyRuchtie:
Don't merge before the 17e!!

PR #6653, opened 2022-04-22
Have client provide default github auth scope

sannek:
Today's login issues were caused by missing default scopes for the GH signin.

We are working on fixing this on the backend, but as an extra safeguard, we should make sure that the client passes at least the lowest default scope too.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6653/commits/b938c3217f471482ea60fda7c58e6f81cc11237d is at https://pr6653.build.csb.dev/s/new.

alexnm (approved):

PR #6498, opened 2022-02-28
upgrade typescript in editor to 4.5

CompuIves:
fixes #6497

PR #6672, opened 2022-05-03
fix: ESLint typescript Template Literal Types

### LINKED ISSUE #6630: Eslint throws when parsing TypeScript's Template Literal Types
# 🐛 bug report

## Preflight Checklist

      adheres to.
      to file, without success.

## Description of the problem

Hey!

Looks like the version of eslint codesandbox uses doesn't support [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html).

I noticed there is also an issue to disable eslint entirely for a project (#3401) which would probably solve my problem as well :)

### To Reproduce

Here is a sandbox reproducing the issue:

https://codesandbox.io/s/heuristic-kare-6siyr1?file=/src/index.ts

image

### Your Environment

| Сodesandbox      |         ?     |
| Browser          |      Chrome 100.0.4896.88 (Official Build) (x86_64)   |
| Operating System |        MacOS      |

DeMoorJasper:

Fixes ESLint typescript Template Literal Types by updating typescript

Closes #6630

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,7 @@
     "overmind-devtools": "^26.0.2",
     "prettier": "2.1.1",
     "pretty-quick": "^3.0.0",
-    "typescript": "4.0.3",
+    "typescript": "4.1.2",
     "username": "^5.1.0"
   },
   "resolutions": {
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -359,7 +359,7 @@
     "terser": "^4.1.4",
     "terser-webpack-plugin": "^1.4.1",
     "thread-loader": "^2.1.2",
-    "typescript": "4.0.3",
+    "typescript": "4.1.2",
     "url-loader": "1.0.1",
     "webpack": "^4.36.1",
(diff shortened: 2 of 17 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6672/commits/0a909cef6f60ce1af8761f485b919de963d13501 is at https://pr6672.build.csb.dev/s/new.

danilowoz:
@DeMoorJasper if you'd like to update the editor typescript version, I guess you need to do this (https://github.com/codesandbox/codesandbox-client/pull/6498), instead of the main package version. Plus, by updating the main package, I think you will need to fix a few typescheck in the codebase, due to new features & fixes from the new ts version.

DeMoorJasper:
@danilowoz Unfortunately in this case that's not true as it would already work in that case.

I've tested it and it works after this PR but not before. The eslint plugin also just uses regular requires from what I could tell so I assume everything is bundled from the local node_modules

PR #6836, opened 2022-08-22
feat(vscode): support insiders login

kettanaito:
CodeSandbox logoCodeSandbox logo&nbsp; Open in CodeSandbox <a href="https://codesandbox.io/p/github/codesandbox/codesandbox-client/feat/vscode-login-insiders?workspace=%7B%22gitSidebarPanel%22:%22PR%22,%22sidebarPanel%22:%22GIT%22%7D">Web Editor</a> | <a href="https://codesandbox.io/p/vscode?owner=codesandbox&repo=codesandbox-client&branch=feat/vscode-login-insiders">VS Code</a>

This is a new feature, a follow-up to the VS Code Insiders support in CodeSandbox Projects.

Currently, when going through the VS Code login flow from the VS Code Insiders, the callback URI is always hardcoded on the V1's side to be `vscode://`. This will open the regular VS Code app and not the insiders app that has initiated the login. 

## What is the new behavior?

- Composes the callback URI dynamically based on the `insiders` query parameter present in the `vscode/login` URL;
- Rewrites the hooks logic because you don't need to circle it through a `useEffect` to set the `deepLink` on the link. 
- Displays a warning notification if the user doesn't have VS Code Insiders installed in their system;
- Displays an error notification if there's a different error while trying to open the `vscode-insiders://` callback URI. 

1. Deployed to staging.
2. Tested manually on staging. 

## Screenshots

Screenshot 2022-08-23 at 13 17 16

Testing that clicking the button actually leads to Insiders if installed:

https://user-images.githubusercontent.com/14984911/186164247-a80d3d86-5e71-4fc9-af8c-977ad952fecd.mov

> The auth token in the video is rightfully invalid because I'm testing this from the local dev environment. 

## Checklist

- [x] Documentation (not necessary; internal change). 
- [x] Testing 
- [x] Ready to be merged
      
- [x] Clicking on the "Open in Visual Studio Code Insiders" button doesn't trigger the notification when the insiders build is not installed. On load notification is displayed correctly.


packages/app/package.json
```diff
@@ -182,6 +182,7 @@
     "posthtml-parser": "^0.4.1",
     "posthtml-render": "^1.1.0",
     "prism-react-renderer": "^1.0.2",
+    "protocol-handlers": "^0.1.2",
```
kettanaito, line 185: Using the same library we use in V2 to detect protocol handlers that don't have any registered apps in the system.

kettanaito:
The `protocol-handlers` was published with a syntax that's not supported (private class properties declaration). I've changed it to `es2020`, and the build should pass after I update the dependency.

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6836/commits/a415db9efa37a50f448f06668d491b0437e195ab is at https://pr6836.build.csb.dev/s/new.

packages/app/src/app/pages/VSCodeAuth/Prompt/index.tsx
```diff
@@ -101,12 +161,13 @@ export const Prompt: FunctionComponent = () => {
 
       <Buttons>
         <Button
-          as="a"
           autoWidth
-          href={deepLink}
           style={{ fontSize: 16, height: 40, width: '100%', marginTop: '1rem' }}
+          onClick={openInVsCode}
```
kettanaito, line 166: Converting this link to a button because if we navigate to `vscode-insiders://` from a link we cannot catch the exception that happens when the protocol handler is not registered. We use the `openUrl` from `protocol-handlers` to check for that, so we need an `onClick`.

kettanaito:
Adding support for conditional sign in urls from the extension's side in https://github.com/codesandbox/codesandbox-applications/pull/1974.

gsimone (approved):

PR #6872, opened 2022-09-08
feat: add tracking for always-on

alexnm:

Closes XTD-23

## What is the new behavior?

Adds two tracking events for always-on sandboxes, one for when the dashboard section of always-on is loaded and one for when an always-on sandbox is loaded


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6872/commits/9d26066a59ba43231fa65356df28971210d322c6 is at https://pr6872.build.csb.dev/s/new.

tristandubbeld (approved):
We're checking page views rather than button clicks? Inside `packages/app/src/app/pages/index.tsx` there's a `Route` component that seems tor track pageviews already with a `trackPageView` method. It should send pageviews to Amplitude but I'm not sure where they end up. I can check this with @tromika.

PR #6892, opened 2022-09-14
Revert "chore: enable signing in when developing in CSB (#6882)"

julien-leclercq:
This reverts commit c0b9a7fc514985af527d4d1280ed80d2e5f92714.

## What kind of change does this PR introduce?

Revert previous commit as it is breaking sse sandboxes 

## What steps did you take to test this? This is required before we can merge, make sure to test the flow you've updated.

1. Step A
2. Step B
3. Step C

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      
- [ ] Added myself to contributors table

PR #6893, opened 2022-09-14
fix: sse sandboxes ENDPOINT

CompuIves:
We never set the `ENDPOINT`, so I never thought it would break anything. But apparently it's used in places, even though we never set it. The problem is that there was not a protocol introduced as part of the `ENDPOINT`. 

Fixes #6892


julien-leclercq (approved):

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6893/commits/31a88e3132a15308aa5aae4bb1b4a61d2e4a34aa is at https://pr6893.build.csb.dev/s/new.

PR #6905, opened 2022-09-16
chore: remove old recents page and rename home

alexnm:
Closes XTD-43

Not a lot of changes, just renaming the `home` page into `recent` and cleaning up the old `recently modified page`, together with removing things from the state and unused functions.
Also removes the projects banner from the home/recent page.


packages/app/src/app/pages/Dashboard/Content/routes/Recent/index.tsx
```diff
@@ -1,77 +1,111 @@
-import React from 'react';
```
alexnm, line 1: this is actually the old `Home/index` page carried over, but renamed in place of the removed `Recent/index`, sorry about the extra diffs

packages/app/src/app/pages/Dashboard/Content/routes/Recent/index.tsx
```diff
@@ -1,77 +1,88 @@
...
-import { Helmet } from 'react-helmet';
-import { SelectionProvider } from 'app/pages/Dashboard/Components/Selection';
+import React, { useEffect } from 'react';
 import { useAppState, useActions } from 'app/overmind';
 import { sandboxesTypes } from 'app/overmind/namespaces/dashboard/types';
 import { Header } from 'app/pages/Dashboard/Components/Header';
 import { VariableGrid } from 'app/pages/Dashboard/Components/VariableGrid';
+import { SelectionProvider } from 'app/pages/Dashboard/Components/Selection';
+import { Helmet } from 'react-helmet';
```
tristandubbeld, line 7: Did it automatically move this import down in the list? Not sure if we're using import sort or anything, but it seems interesting if it moves it down.

packages/app/src/app/pages/Dashboard/Content/routes/Recent/index.tsx
```diff
@@ -1,77 +1,88 @@
...
-    dashboard: { sandboxes, recentSandboxesByTime, getFilteredSandboxes },
+    dashboard: { viewMode, sandboxes },
   } = useAppState();
   const {
     dashboard: { getPage },
   } = useActions();
 
-  React.useEffect(() => {
+  useEffect(() => {
     getPage(sandboxesTypes.RECENT);
```
tristandubbeld, line 24: Does `getPage` change when `activeTeam` changes by any chance and can we use that in the dependency array instead?

packages/app/src/app/pages/Dashboard/Content/routes/Recent/index.tsx
```diff
@@ -1,77 +1,88 @@
...
+  const templates: DashboardGridItem[] = (sandboxes.TEMPLATE_HOME || []).map(
+    (template, i) => {
+      const { sandbox, ...templateValues } = template;
 
-    return [
-      { type: 'header', title },
-      ...recentSandboxes.map(sandbox => ({
-        type: 'sandbox' as 'sandbox',
+      const dashboardTemplate: DashboardTemplate = {
+        type: 'template' as 'template',
```
tristandubbeld, line 33: `as const` would work too I think, but this is fine!

tristandubbeld (approved):
Minor thingies, other than that we're good to go!

olarclara (approved):
🏠

packages/app/src/app/pages/Dashboard/Content/routes/Recent/index.tsx
alexnm: just renamed the file, I didn't touch the imports in this one, just looks like changes because of the old `Recent/index.tsx` page

packages/app/src/app/pages/Dashboard/Content/routes/Recent/index.tsx
alexnm: Don't know exactly how the changes happen in overmind, as getPage is coming from the app state/dashboard. But I'd leave this as it is for now, it makes sense to have the fetching depend on the activeTeam.

PR #6906, opened 2022-09-16
feat(dashboard): fetch open source contributions

olarclara:
y7g3zb-3000 preview csb app_dashboard_repositories_open-source_workspace=411542e0-c5fb-4ee8-85c9-65c8cfd18e52

This new feature adds data fetching capabilities to the open source/contributions page.
The UI is not final and should be addressed by the facelifting efforts with the support of the dev team if needed.

1. Open the dashboard
2. Navigate to the page using the sidebar (currently clicking on "Open source")
3. Page behaves as expected

## Next steps

- Current implementation creates a new component to handle the UI rather than adapting the existing `Sandbox` component to avoid dealing with the complexity of the said component. The downside is that the styles might need to be adjusted in more than one place to match the designs. Would appreciate thoughts on this matter.
- Current implementation does not account for the list view mode, since I couldn't find the designs for it. I'll follow-up with the design team on this subject.
- Current implementation does not have a context menu, I also couldn't find the designs for them and will follow up with the design team.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6906/commits/3bb9736e5829790c8e59e1ba1cb040e807c9fa64 is at https://pr6906.build.csb.dev/s/new.

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -1978,3 +1979,40 @@ export const updateAlbum = async (
...
+      dashboard.contributions = null;
+
+      const contributionsData = await effects.gql.queries.getContributionBranches(
+        {}
+      );
+      if (
+        !contributionsData ||
+        !contributionsData.me ||
+        !contributionsData.me.recentBranches
+      ) {
```
tristandubbeld, line 2000: Can we use optional chaining here?

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -1978,3 +1979,40 @@ export const updateAlbum = async (
...
+        return;
+      }
+      contributionBranches = contributionsData.me.recentBranches;
+
+      if (!contributionBranches) {
+        return;
+      }
```
tristandubbeld, line 2007: We already check for the `contributionsData.me.recentBranches` in the if statement above.

I (nitpick) personally like a more "positive" approach where we check for certain values instead of doing early returns. Unless it causes multiple nested `if` statements. But right now I think it would work:

```ts
if (contributionsData?.me?.recentBranches) {
  contributionBranches = contributionsData.me.recentBranches;

  dashboard.contributions = contributionBranches.map(branch => ({
    type: 'contribution-branch',
    branch,
  }));
}
```

packages/app/src/app/pages/Dashboard/Components/ContributionBranch/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+          ':focus-visible': {
+            boxShadow: '0 0 2px 1px rgba(255, 255, 255, 0.4)',
+          },
+        })}
+        direction="vertical"
+        onClick={handleClick}
+        onKeyDown={e => e.key === 'Enter' && handleClick(e)}
+        tabIndex={0}
+        // TODO: refine semantics when/if the context menu gets implemented.
+        role="link"
```
tristandubbeld, line 55: 👍 

This can probably be an `a` element then. That way we don't have to take meta key presses and `onKeyDown` etc. in account. Or does the context menu automatically make it a button?

tristandubbeld (commented):
Love the JSDoc / comments! Makes things so much easier to understand. Also added a few suggestions here and there, lmk what you think!

packages/app/src/app/pages/Dashboard/Components/ContributionBranch/index.tsx
```diff
@@ -0,0 +1,102 @@
+import { v2BranchUrl } from '@codesandbox/common/lib/utils/url-generator';
+import { Icon, ListAction, Stack, Text } from '@codesandbox/components';
+import { useAppState } from 'app/overmind';
+import React from 'react';
+import { css } from '@styled-system/css';
+import { useHistory } from 'react-router-dom';
+import { DashboardContributionBranch } from '../../types';
+
+export const ContributionBranch: React.FC<DashboardContributionBranch> = ({
```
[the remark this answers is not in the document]
olarclara: @alexnm I think so, I'll unite them in the following PR with the repositories page implementation.

packages/app/src/app/overmind/namespaces/dashboard/state.ts
```diff
@@ -71,6 +74,8 @@ export type State = {
     older: Sandbox[];
   };
   curatedAlbums: DashboardAlbum[];
+  // TODO: fix unknown
+  contributions: DashboardContributionBranch[] | null;
```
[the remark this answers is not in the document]
olarclara: @alexnm I was a bit conflicted by this because these branches are not really sandboxes, right? I like keeping lists under the same structure but the types seem conceptually wrong, and we also have lists under the general structure such as the `curatedAlbums`, so the contributions lists are not the only thing and don't seem too out of place. Wdyt?

packages/app/src/app/pages/Dashboard/Components/ContributionBranch/index.tsx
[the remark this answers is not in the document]
alexnm: I've been working on adding the branches in the recent page: https://github.com/codesandbox/codesandbox-client/pull/6915 but I haven't implemented any UI for the cards yet. So I'll wait for this PR to be integrated and I will try to rely on the same components that you started building

packages/app/src/app/pages/Dashboard/Components/ContributionBranch/index.tsx
olarclara: @tristandubbeld because the action of clicking into these links needs to be tracked, we still need to add a click handler even if it's an anchor. So I'd keep the semantics like this so we get the benefit of not having to override default anchor styles. Wdyt?

packages/app/src/app/overmind/namespaces/dashboard/state.ts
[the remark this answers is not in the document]
alexnm: maybe the word sandboxes there is misleading, but indeed, let's see how this works in the end. for the recent page, I'm adding the branches under `RECENT_BRANCHES` inside sandboxes as it gets merged with `RECENT_SANDBOXES`

packages/app/src/app/pages/Dashboard/Components/ContributionBranch/index.tsx
tristandubbeld: I think overriding the styles of the anchor is easier / less code than the current implementation. I'm also not sure about the focusability or the registration as an interactive item right now. An `a` needs a `href` to be focusable, but I'm not sure if `role="link"` overrides this and makes it that we don't need an `href`.

olarclara: @tristandubbeld there's a tabIndex right above the `role="link"`, not sure if you saw it.

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
olarclara: @tristandubbeld Updated the action in commit: [dc65d5a](https://github.com/codesandbox/codesandbox-client/pull/6906/commits/dc65d5a820413f44ac7fb4b3a7052e0caaeac003). Personally, I prefer early returns and the other actions are following the same pattern. I don't think we have any guidelines on this yet but might be a topic for the next Huddle.js.

alexnm (approved):

packages/app/src/app/pages/Dashboard/Components/ContributionBranch/index.tsx
tristandubbeld: 👍

tristandubbeld (approved):

PR #6928, opened 2022-09-23
feat(dashboard): improve repositories loading

olarclara:

Adds progressive loading to the repositories page. I've also added the behavior of fetching only the selected repo if the user accesses the page directly.

## What is the new behavior?

Decreases initial loading time for a team with almost 100 projects and around 2000 branches from nearly 14s to less than 5s.
|Now|Previously|
|:-:|:-:|
|image|image|

### DIFF
--- a/packages/app/src/app/overmind/namespaces/dashboard/actions.ts
+++ b/packages/app/src/app/overmind/namespaces/dashboard/actions.ts
@@ -1972,15 +1972,35 @@ export const getRepositoriesByTeam = async ({ state, effects }: Context) => {
   try {
     dashboard.repositories = null;
 
-    const repositoriesData = await effects.gql.queries.getRepositoriesByTeam({
-      teamId: activeTeam,
-    });
-    const v2Repositories = repositoriesData?.me?.team?.projects;
-    if (!v2Repositories?.length) {
+    // First fetch data without syncing with GitHub
+    // to decrease waiting time.
+    const unsyncedRepositoriesData = await effects.gql.queries.getRepositoriesByTeam(
+      {
+        teamId: activeTeam,
+        syncData: false,
+      }
+    );
+    const unsyncedRepositories = unsyncedRepositoriesData?.me?.team?.projects;
+    if (!unsyncedRepositories?.length) {
       return;
     }
 
-    dashboard.repositories = v2Repositories;
+    dashboard.repositories = unsyncedRepositories;
+
+    // Then fetch data synced with GitHub to make sure
+    // what we show is up-to-date.
+    const syncedRepositoriesData = await effects.gql.queries.getRepositoriesByTeam(
+      {
+        teamId: activeTeam,
+        syncData: true,
+      }
+    );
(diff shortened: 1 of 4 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6928/commits/95bf0c0955d2ff2cea90ec987451fe29139bec1d is at https://pr6928.build.csb.dev/s/new.

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -1972,18 +1972,65 @@ export const getRepositoriesByTeam = async ({ state, effects }: Context) => {
...
+    const unsyncedRepositories = unsyncedRepositoriesData?.me?.team?.projects;
+    if (!unsyncedRepositories?.length) {
+      return;
+    }
+
+    dashboard.repositories = unsyncedRepositories;
+
+    // Then fetch data synced with GitHub to make sure
+    // what we show is up-to-date.
+    const syncedRepositoriesData = await effects.gql.queries.getRepositoriesByTeam(
```
alexnm, line 1992: I wonder if we should extract this into a separate function that runs when the page becomes idle, so it doesn't affect the first contentful paint. Not sure if this might be an issue for now, as it's just a async call that should at most trigger one re-render in the react hierarchy. Just a passing thought.

packages/app/src/app/pages/Dashboard/Content/routes/Repositories/index.tsx
```diff
@@ -15,10 +15,34 @@ export const RepositoriesPage = () => {
...
+    if (
+      path === '' &&
+      pathRef.current?.startsWith('github') &&
+      repositories.length === 1
+    ) {
+      actions.dashboard.getRepositoriesByTeam();
+    }
+
+    pathRef.current = path;
+  }, [path]);
```
alexnm, line 45: nice! this ensures the repository page with all the branches doesn't load additional data when being open from the editor for example. one small note, maybe this component should have a computed field inside like `view: 'repositories' / 'branches'`

alexnm (approved):

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
olarclara: @alexnm interesting, your idea is for the repositories page to trigger another action after getting the initial data? let's follow up in slack/linear.

packages/app/src/app/pages/Dashboard/Content/routes/Repositories/index.tsx
olarclara: @alexnm I must admit I spent more time on this than I should, at the of the day my reasoning was that the commentary and the single case were enough to explain the logic. But we can compute `currentPage` and `previousPage` if we want to handle more cases in this effect, like moving between repos (which is not possible atm).

PR #6929, opened 2022-09-23
Facelift: Extra type review

lenasmayor:
- Adjust Page Header paddings
Screen Shot 2022-09-23 at 19 17 35

- Adjust type size on Discover page
Screen Shot 2022-09-23 at 19 18 07

- Adjust type size on Archive page
Screen Shot 2022-09-23 at 19 17 12

- Add + icon to New Folder on the Sandboxes page
Screen Shot 2022-09-23 at 19 18 36

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Components/Header/index.tsx
+++ b/packages/app/src/app/pages/Dashboard/Components/Header/index.tsx
@@ -104,13 +104,7 @@ export const Header = ({
                 width: 'auto',
               })}
             >
-              <Icon
-                name="plus"
-                size={20}
-                title="New"
-                css={css({ paddingRight: 2 })}
-              />
-              Import Repo
+              + Import Repo
             </Button>
           )}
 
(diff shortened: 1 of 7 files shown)


tristandubbeld (approved):

packages/app/src/app/pages/Dashboard/Components/Header/index.tsx
```diff
@@ -83,7 +83,13 @@ export const Header = ({
               width: 'auto',
             })}
           >
-            + New Folder
+            <Icon
+              name="plus"
+              size={20}
+              title="New"
+              css={css({ paddingRight: 2 })}
```
alexnm, line 90: one note on the design system: when we move to the v2 codebase we should aim for limiting these exceptions. the design system should support the spacing in such a way that it is composable via the variants of the components, not with custom alignment

alexnm (approved):

PR #6945, opened 2022-09-29
fix: sort sandboxes on recent page by lastAccessedAt

alexnm:
Uses the new `lastAccessedAt` field to sort the sandboxes with the branches on the recent page

NOTE: Only works on staging, as the field is not yet deployed to prod


packages/app/src/app/overmind/effects/gql/dashboard/queries.ts
```diff
@@ -324,6 +324,7 @@ export const recentlyAccessedSandboxes: Query<
     me {
       recentlyAccessedSandboxes(limit: $limit, teamId: $teamId) {
         ...sandboxFragmentDashboard
+        lastAccessedAt
```
alexnm, line 327: I didn't add the field to the sandbox fragment, but not sure what to do, maybe it's better to keep the queries/types consistent? but then again, we are overfetching in a lot of situations because of the reuse of the same fragment for the sandbox

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6945/commits/32cb714a81f9bae5c319ad0853fd55667a393bc0 is at https://pr6945.build.csb.dev/s/new.

olarclara (approved):
⏲️

tristandubbeld (approved):

PR #6950, opened 2022-09-30
chore(create-modal): update quick start ids

alexnm:
image

@necoline can you please validate this? and maybe figure out what will be the nice names of all the templates for the release?

Feedback on my side:
* static v2 should be blank maybe? and have a different icon
* vue v2 should be nuxt? and again have a different icon
* remix should have a different icon


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6950/commits/aac84b10f8203bbeb0c5d2325c8eaf886c9d1e90 is at https://pr6950.build.csb.dev/s/new.

olarclara (approved):

necoline:
✅ Updated the names to be better titles
✅ static v2 should be blank maybe? 
> and have a different icon

I'm not sure about this. I think the idea with the blank is that people can start building a static site with html. What did you have in mind?
> vue v2 should be nuxt? and again have a different icon

I think having Nuxt is a great idea. I created a [template](https://codesandbox.io/s/nuxt-6xxu1m?file=/pages/index.vue) but we will need all right tags to make it a v2 template. I added it [here](https://codesandbox.io/s/3drzs?file=/src/data/official.ts). Is that all we need to do? @alexnm 

✅ remix should have a different icon

alexnm:
> ✅ Updated the names to be better titles ✅ static v2 should be blank maybe?
> 
> > and have a different icon
> 
> I'm not sure about this. I think the idea with the blank is that people can start building a static site with html. What did you have in mind?
> 
> > vue v2 should be nuxt? and again have a different icon
> 
> I think having Nuxt is a great idea. I created a [template](https://codesandbox.io/s/nuxt-6xxu1m?file=/pages/index.vue) but we will need all right tags to make it a v2 template. I added it [here](https://codesandbox.io/s/3drzs?file=/src/data/official.ts). Is that all we need to do? @alexnm
> 
> ✅ remix should have a different icon

blank -> not necessarily a static site, the users get an empty cloud sandbox (v2 editor) where they can install any package, run node scripts, so pretty much start from scratch any project. the static logo might be misleading, we use the codesandbox square for the "old" v2 dashboard, so how about that?

nuxt -> as far as I can see, that one is not marked as cloud template yet (v2: false). about the official list, i don't know how that is happening, someone from the backend team could help out maybe.

alexnm:
I updated the `id` to point to the `nuxt` template @necoline created, so when the template will be marked as v2 and official, it will show up in the quickstart.

PR #6964, opened 2022-10-04
feat(dashboard): import and redirect if user has write access

olarclara:

If the user has write access to the repository, import and open the editor. Else, continue to step 2.

### DIFF
--- a/packages/app/src/app/components/CreateNewSandbox/CreateSandbox/Import/useImportAndRedirect.tsx
+++ b/packages/app/src/app/components/CreateNewSandbox/CreateSandbox/Import/useImportAndRedirect.tsx
@@ -1,0 +1,25 @@
+import { notificationState } from '@codesandbox/common/lib/utils/notifications';
+import { v2DraftBranchUrl } from '@codesandbox/common/lib/utils/url-generator';
+import { NotificationStatus } from '@codesandbox/notifications';
+import history from 'app/utils/history';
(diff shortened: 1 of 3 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6964/commits/0631885739760aa66e0d16dd26c3a4b8803e7681 is at https://pr6964.build.csb.dev/s/new.

packages/app/src/app/components/CreateNewSandbox/CreateSandbox/Import/useImportAndRedirect.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export const useImportAndRedirect = () => {
+  async function importAndRedirect(
+    owner: string,
+    name: string,
+    csbTeamId: string
+  ) {
+    try {
+      const response = await importRepository({ owner, name, csbTeamId });
+      window.location.href = v2DraftBranchUrl(response.owner, response.repo);
```
tristandubbeld, line 14: Should we use something from the router lib we use here instead? Is that possible in the current state?

tristandubbeld: We discussed this on slack and since we're directing to the v2 application we can't use React Router.

tristandubbeld (approved):

PR #6972, opened 2022-10-05
fix: text-overflow for folders in sidebar

alexnm:
Focused on the text-overflow which is more problematic for the release, left the animation and other details for later on.

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Sidebar/index.tsx
+++ b/packages/app/src/app/pages/Dashboard/Sidebar/index.tsx
@@ -760,6 +760,7 @@ const NestableRowItem: React.FC<NestableRowItemProps> = ({
       >
         <Link
           to={folderUrl}
+          title={name}
           onClick={() => {
             const event = MAP_SIDEBAR_ITEM_EVENT_TO_PAGE_TYPE[page];
             if (event) {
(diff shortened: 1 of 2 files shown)


packages/app/src/app/pages/Dashboard/Sidebar/index.tsx
```diff
@@ -815,13 +816,14 @@ const NestableRowItem: React.FC<NestableRowItemProps> = ({
               paddingLeft: 0,
               paddingRight: 0,
               marginRight: '0',
+              width: '96%',
```
alexnm, line 819: ¯\_(ツ)_/¯

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6972/commits/9224093f28849a828d201b2609a61d01d15ef5fb is at https://pr6972.build.csb.dev/s/new.

packages/app/src/app/pages/Dashboard/Sidebar/index.tsx
tristandubbeld: It seems to work - like you said - if you add `box-sizing: border-box` to the whole chain of components, but doing that might break a lot of other existing components. I think this is fine / safe.

tristandubbeld (approved):

PR #6973, opened 2022-10-05
feat(dashboard): Implement notifications

tristandubbeld:
The only thing that we might want to update is the dismissal of the notifications and when the dots in the sidebar update. If we want to update the dots in the sidebar when a notification is dismissed, we need to keep track of the state higher up in the tree instead of just inside the notifications. Right now the notification indicator dots are gone when the routes change.

## Video

https://user-images.githubusercontent.com/7533849/194049924-10dba198-0778-4bcc-b531-f98a86c50681.mov

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Content/routes/MyContributions/index.tsx
+++ b/packages/app/src/app/pages/Dashboard/Content/routes/MyContributions/index.tsx
@@ -1,11 +1,12 @@
 import React from 'react';
 import { Helmet } from 'react-helmet';
 import { useParams } from 'react-router-dom';
-import { useAppState, useActions } from 'app/overmind';
+import { useAppState, useActions, useEffects } from 'app/overmind';
 import { Header } from 'app/pages/Dashboard/Components/Header';
 import { VariableGrid } from 'app/pages/Dashboard/Components/VariableGrid';
 import { DashboardGridItem, PageTypes } from 'app/pages/Dashboard/types';
 import { SelectionProvider } from 'app/pages/Dashboard/Components/Selection';
+import { Notification } from 'app/pages/Dashboard/Components/Notification/Notification';
 
 export const MyContributionsPage = () => {
   const params = useParams<{ path: string }>();
@@ -15,12 +16,17 @@ export const MyContributionsPage = () => {
     activeTeam,
     dashboard: { contributions },
   } = useAppState();
+  const { browser } = useEffects();
 
   React.useEffect(() => {
     actions.dashboard.getContributionBranches();
   }, [actions.dashboard]);
 
   const pageType: PageTypes = 'my-contributions';
+
+  const isNotificationDismissed = browser.storage.get(
+    'notificationDismissed'
+  )?.[pageType];
 
   const itemsToShow = (): DashboardGridItem[] => {
     if (contributions === null) {
@@ -51,6 +57,13 @@ export const MyContributionsPage = () => {
         showFilters={Boolean(param)}
         showSortOptions={Boolean(param)}
       />
+      {!isNotificationDismissed ? (
+        <Notification pageType={pageType}>
+          {itemsToShow().length === 0
+            ? 'Introducing contribution branches: the easiest way of contributing to open source.'
+            : 'Your contribution branches now live here, so you can manage your contributions easily.'}
+        </Notification>
+      ) : null}
       <VariableGrid page={pageType} items={itemsToShow()} />
     </SelectionProvider>
   );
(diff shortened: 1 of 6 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/6973/commits/a5c168dc9d3cadf63d3d6a1f73fbbaba95ae56fb is at https://pr6973.build.csb.dev/s/new.

packages/app/src/app/pages/Dashboard/Content/routes/MyContributions/index.tsx
```diff
@@ -15,13 +16,18 @@ export const MyContributionsPage = () => {
...
   } = useAppState();
+  const { browser } = useEffects();
 
   React.useEffect(() => {
     actions.dashboard.getContributionBranches();
   }, [actions.dashboard]);
 
   const pageType: PageTypes = 'my-contributions';
 
+  const isNotificationDismissed = browser.storage.get(
```
alexnm, line 27: would it be possible to move these checks inside the component and render null when the notification has been dismissed? reduces the lines of code and the times browser.storage is called in the code

packages/app/src/app/pages/Dashboard/Content/routes/MyContributions/index.tsx
```diff
@@ -51,6 +57,13 @@ export const MyContributionsPage = () => {
         showFilters={Boolean(param)}
         showSortOptions={Boolean(param)}
       />
+      {!isNotificationDismissed ? (
+        <Notification pageType={pageType}>
+          {itemsToShow().length === 0
```
alexnm, line 62: let's call `itemsToShow()` once at the top and use that result everywhere, I get performance anxiety when I see this, even though it's just a micro optimization :D

packages/app/src/app/pages/Dashboard/Content/routes/Repositories/index.tsx
```diff
@@ -89,6 +95,13 @@ export const RepositoriesPage = () => {
         showBetaBadge
         nestedPageType={pageType}
       />
+      {!isNotificationDismissed ? (
+        <Notification pageType={pageType}>
+          {itemsToShow().length === 0
+            ? 'CodeSandbox Projects is now Repositories: an improved git workflow powered by the cloud. '
+            : 'Your CodeSandbox Projects repositories now live here. Repository sandboxes are now listed under Synced sandboxes. You can find your contribution branches on My contributions inside your personal team.'}
```
alexnm, line 102: should we quote `My contributions` and `Synced sandboxes` for clarity?

packages/app/src/app/pages/Dashboard/Content/routes/Repositories/index.tsx
```diff
@@ -89,6 +95,13 @@ export const RepositoriesPage = () => {
         showBetaBadge
         nestedPageType={pageType}
       />
+      {!isNotificationDismissed ? (
+        <Notification pageType={pageType}>
+          {itemsToShow().length === 0
+            ? 'CodeSandbox Projects is now Repositories: an improved git workflow powered by the cloud. '
```
alexnm, line 101: This text sounds a bit weird to me. What if we take the chance here to explain that Repositories are just Sandboxes with a versioning/git? cc @lenasmayor @necoline

packages/app/src/app/pages/Dashboard/Content/routes/SyncedSandboxes/index.tsx
```diff
@@ -87,6 +93,12 @@ export const SyncedSandboxesPage = () => {
         showSortOptions={Boolean(param)}
         nestedPageType={pageType}
       />
+      {!isNotificationDismissed ? (
+        <Notification pageType={pageType}>
+          Repository sandboxes are now called Synced sandboxes. New imported
+          repositories will be listed under All repositories.
```
alexnm, line 99: same comment on quoting the categories

alexnm (approved):
Looks great, left some comments so maybe some small adjustments might work out, but you can also merge it now and come back later for some points. Also left a comment on a text, not sure what's the easiest way to involve Neco and Lena, maybe they can review it.

packages/app/src/app/pages/Dashboard/Content/routes/MyContributions/index.tsx
tristandubbeld: Great idea, especially since I was already doing this 😄

packages/app/src/app/pages/Dashboard/Content/routes/Repositories/index.tsx
tristandubbeld: I forgot to highlight these pieces of text.

packages/app/src/app/pages/Dashboard/Content/routes/Repositories/index.tsx
lenasmayor: Felipe suggested this so we would align with the copy for the newsletter and blog post. I guess the goal is to explain to migrating users (from v2) what is this new section. New users (from v1) will also see the empty state that will cover the features of the new repositories.

PR #7000, opened 2022-10-07
feat(sidebar): Sidebar overmind state

tristandubbeld:
### About

To make it easier to interact with the sidebar and conditionally render sidebar items using less data and requests we now have a separate sidebar state. In the future we will be able to use this for other things, e.g. showing sidebar notification indicators. In this PR we also conditionally render the Templates and Synced Sandboxes sidebar rows based on data.

### Linear

Closes XTD-198

### DIFF
--- a/packages/app/src/app/graphql/types.ts
+++ b/packages/app/src/app/graphql/types.ts
@@ -580,6 +580,8 @@ export type ProSubscription = {
   paymentProvider: Maybe<SubscriptionPaymentProvider>;
   quantity: Maybe<Scalars['Int']>;
   status: SubscriptionStatus;
+  trialEnd: Maybe<Scalars['DateTime']>;
+  trialStart: Maybe<Scalars['DateTime']>;
   type: SubscriptionType;
   unitPrice: Maybe<Scalars['Int']>;
   updateBillingUrl: Maybe<Scalars['String']>;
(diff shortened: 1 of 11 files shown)


packages/app/src/app/overmind/effects/gql/sidebar/queries.ts
```diff
@@ -0,0 +1,65 @@
...
+      }
+    }
+  }
+  ${sidebarSyncedSandboxFragment}
+  ${sidebarCollectionFragment}
+  ${sidebarTemplateFragment}
+`;
+/**
+ * 👆 fragments might actually be a bit too much for this
+ */
```
tristandubbeld, line 38: They're useful for types though :)

packages/app/src/app/graphql/types.ts
```diff
@@ -3400,10 +3400,6 @@ export type SidebarSyncedSandboxFragmentFragment = {
   __typename?: 'Sandbox';
 } & Pick<Sandbox, 'id'>;
 
-export type SidebarCollectionFragmentFragment = {
-  __typename?: 'Collection';
-} & Pick<Collection, 'id' | 'path'>;
-
```
tristandubbeld: Removed collections from the sidebar state as they are too intertwined with the dashboard.

packages/app/src/app/overmind/namespaces/sidebar/actions.ts
```diff
@@ -0,0 +1,50 @@
...
+        hasSyncedSandboxes,
+        hasTemplates,
+      };
+    } else {
+      /**
+       * Fetch data for the user
+       */
+      const {
+        me: { sandboxes, templates },
+      } = await queries.getPersonalSidebarData();
```
alexnm, line 35: nitpick: this if/else looks like it can be made around the call of `queries.getPersonalSidebarData` vs `queries.getTeamSidebarData`, but the rest of the function can be common. would keep the function shorter and less repetitive

alexnm (approved):

olarclara (approved):
🧠

packages/app/src/app/overmind/namespaces/sidebar/actions.ts
tristandubbeld: I remember doing it this way because the types were acting up, but since they still were I tried to implement it in another way. I've changed it but its not a lot shorter because I had to do some workarounds for TS. It was weird to me that my editor didn't show errors for certain things, but running `yarn typecheck` does. Something fishy.

alexnm (approved):
💯

PR #7006, opened 2022-10-10
experimental(dashboard): star repo feature

alexnm:
Not on the roadmap yet and **not supported by the backend**, but I really felt the need to have this feature while working in the dashboard. It relies on **localStorage** to persist the starred repos per team.

All the functionality is dependent on the value of `CSB_DEBUG` being set to `ENABLED`, similar with how the devtools appears for the v2 editor. v1 does not have the same feature flag system as v2, but it would be nice to add maybe?

* adds star/unstar in the context menu
* adds star/unstar for the repo page header (when showing the branches)
* shows the starred repos in the sidebar, under `All repositories`
* starred repos are reset everytime you change the workspace

Screenshot 2022-10-10 at 12 05 13
Screenshot 2022-10-10 at 12 05 22
Screenshot 2022-10-10 at 12 05 36

Why merge this?
* we can use it internally for convenience
* it gives us time to think about the UX of the feature for when we offer it for all users

### DIFF
--- a/packages/app/src/app/overmind/namespaces/dashboard/actions.ts
+++ b/packages/app/src/app/overmind/namespaces/dashboard/actions.ts
@@ -2035,3 +2035,48 @@ export const getRepositoryByDetails = async (
     );
   }
 };
+
+export const getStarredRepos = ({ state, effects }: Context) => {
+  const { dashboard, activeTeam } = state;
+
+  const persistedStarredRepos = effects.browser.storage.get(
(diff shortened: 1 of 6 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7006/commits/6caa2162c647614cde55b7ee94edd1cb2763c68c is at https://pr7006.build.csb.dev/s/new.

tristandubbeld (approved):

PR #7020, opened 2022-10-13
fix: hover interaction with more button

alexnm:
A bit of polishing on the card/list items when hovering to the more button
* bigger icon
* default cursor
* square with border radius instead of circle shape
* hover on icon reverts hover on card (using [:has](https://developer.mozilla.org/en-US/docs/Web/CSS/:has) which doesn't work in firefox unfortunately)
* applies on: sandboxes, folders, branches & repos

https://user-images.githubusercontent.com/9945366/195557649-9674efa1-f299-41a9-8c1a-8e7c4bc5ae87.mov

https://user-images.githubusercontent.com/9945366/195557700-41a3481c-9db8-4231-9c48-e14255baf838.mov

Notes
* There's a lot of duplication already with these cards, might be a good idea to think of a way to reuse the card/list item component

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Components/Branch/BranchCard.tsx
+++ b/packages/app/src/app/pages/Dashboard/Components/Branch/BranchCard.tsx
@@ -90,10 +90,7 @@ export const BranchCard: React.FC<BranchProps> = ({
             </Text>
           </Tooltip>
           <IconButton
-            css={{
-              borderRadius: '2px',
-              cursor: 'default',
-            }}
+            variant="square"
             name="more"
             size={14}
             title="Branch actions"
--- a/packages/app/src/app/pages/Dashboard/Components/Branch/BranchCard.tsx
+++ b/packages/app/src/app/pages/Dashboard/Components/Branch/BranchCard.tsx
@@ -40,6 +40,9 @@ export const BranchCard: React.FC<BranchProps> = ({
         ':hover': {
           backgroundColor: 'card.backgroundHover',
         },
+        ':has(button:hover)': {
+          backgroundColor: 'card.background',
+        },
         ':focus-visible': {
           borderColor: 'focusBorder',
         },
@@ -87,8 +90,12 @@ export const BranchCard: React.FC<BranchProps> = ({
             </Text>
           </Tooltip>
           <IconButton
+            css={{
+              borderRadius: '2px',
+              cursor: 'default',
+            }}
             name="more"
-            size={9}
+            size={14}
             title="Branch actions"
             onClick={evt => {
               evt.stopPropagation();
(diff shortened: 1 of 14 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7020/commits/7540c8dcb13ccf79a36723b33368dd37fd89de59 is at https://pr7020.build.csb.dev/s/new.

packages/app/src/app/pages/Dashboard/Components/Branch/BranchCard.tsx
```diff
@@ -40,6 +40,9 @@ export const BranchCard: React.FC<BranchProps> = ({
         ':hover': {
           backgroundColor: 'card.backgroundHover',
         },
+        ':has(button:hover)': {
+          backgroundColor: 'card.background',
+        },
```
tristandubbeld, line 45: This is some nice progressive enhancement!

packages/app/src/app/pages/Dashboard/Components/Folder/FolderCard.tsx
```diff
@@ -67,8 +70,12 @@ export const FolderCard: React.FC<FolderItemComponentProps> = ({
     >
       {!isNewFolder ? (
         <IconButton
+          css={{
+            borderRadius: '2px',
+            cursor: 'default',
+          }}
```
tristandubbeld, line 76: Is this one of the cases where if we override the `IconButton` styles it also affects the editor? We could create an abstraction of this component, like `IconButtonDash` or rename the default to `Round` or something. Or create a variant? Just emptying my thoughts here.

tristandubbeld (approved):

zehfernandes (approved):

packages/app/src/app/pages/Dashboard/Components/Folder/FolderCard.tsx
alexnm: Added a `variant` prop for the IconButton

packages/components/src/components/IconButton/index.tsx
```diff
@@ -29,7 +32,8 @@ export const IconButton: React.FC<IconButtonProps> = ({
         {
           width: '26px', // same width as (height of the button)
           paddingX: 0,
-          borderRadius: '50%',
+          borderRadius: variant === 'round' ? '50%' : '2px',
+          cursor: variant === 'round' ? 'pointer' : 'default',
```
tristandubbeld, line 36: We could add a comment here explaining why `round` gets a pointer and `square` doesn't for future reference maybe.

tristandubbeld (approved):
Good stuff!

packages/components/src/components/IconButton/index.tsx
alexnm: right!

olarclara (approved):
✨

PR #7026, opened 2022-10-14
fix: show sign in on cloud sandbox message

tristandubbeld:
The message changed and we weren't checking for that so the previously added sign in button was lost again. Talked to Sanne and we agreed that this is not the ideal way to do this, but for now in v1 it is what it is.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7026/commits/43d28671efdce8a17935174cdadea415968ff8a9 is at https://pr7026.build.csb.dev/s/new.

danilowoz (approved):
LGTM (Let's get this merged)!

PR #7027, opened 2022-10-14
feat(search): enable repository search

tristandubbeld:

We're now able to search for repositories on the dashboard 🥳

https://user-images.githubusercontent.com/7533849/195845514-2bdd20b6-b2a5-498c-93bf-94e1abfeb028.mov

Right now, the search works a bit janky. We only fetch the sandboxes and repositories when we visit an empty search page - I haven't changed anything about that logic. The search itself or fetching the sandboxes and repositories can be slow too -maybe because it fetches all the branches as well? We can also make an effort of removing the `ts-ignores` from the search feature.

### DIFF
--- a/packages/app/src/app/components/CreateSandbox/TemplateCard.tsx
+++ b/packages/app/src/app/components/CreateSandbox/TemplateCard.tsx
@@ -24,14 +24,7 @@ export const TemplateCard = ({
 
   const sandboxTitle = template.sandbox?.title || template.sandbox?.alias;
   const isV2 = template.sandbox?.isV2;
-  const teamName = template.sandbox?.collection?.team?.name; // Search is different
-
-  if (!teamName) {
-    console.log('template', template);
-  }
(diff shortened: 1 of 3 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7027/commits/971e38c8303d65e3beb15da3dad3003df6090db3 is at https://pr7027.build.csb.dev/s/new.

alexnm (approved):
Implementation looks good, but from a product perspective, I would ask if we should have a separate search based on the context in which you are: sandboxes vs repositories.

tristandubbeld:
>Implementation looks good, but from a product perspective, I would ask if we should have a separate search based on the context in which you are: sandboxes vs repositories.

Yeah that's a great idea. The search right now isn't very friendly, especially if you're specifically looking for repositories. I didn't want to go there (yet). There's a lot we can improve :)

PR #7028, opened 2022-10-15
fix(common): add default value to `container.node` in `ConfigWizard`

MichaelDeBoey:
When looking at a config file, it shows `10`, but it actually is `14` by default now.
https://codesandbox.io/s/github/remix-run/examples/tree/main/tailwindcss?file=/sandbox.config.json


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7028/commits/c4d8e320b34a2d7ace4b28f3932bfe036b1f488b is at https://pr7028.build.csb.dev/s/new.

PR #7035, opened 2022-10-18
fix(dashboard): set sandboxes to empty array when catching errors

olarclara:
Prevents the recent page from loading forever when errors are thrown. This defaults to the empty screen and relies on the user seeing the notification error since we don't have an error state for that page.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7035/commits/cc64e616a706c5ece43400c48140df80b2a2db3e is at https://pr7035.build.csb.dev/s/new.

tristandubbeld (approved):

PR #7037, opened 2022-10-19
feat(trial): Add onboarding questions

tristandubbeld:
## Todos

- [x] ~[Submit role and usage questions](https://linear.app/codesandbox/issue/XTD-190/save-answers-to-user-metadata).~

## How to test this

1. Go to the `/signin` page in an incognito window (or any other window where you're not logged in)
2. Open the browser console. paste the following code and press enter to execute:
```
localStorage.setItem('TOQ_DEBUG', 'ENABLED');
```
3. Refresh the page
4. You should now be able to see the onboarding questions

### Test the modal

1. Go to `/s` page in an incognito window (or any other window where you're not logged in)
2. Go to the second tab or any other tab with the sign in button
3. Open the browser console. paste the following code and press enter to execute:
```
localStorage.setItem('TOQ_DEBUG', 'ENABLED');
```
4. Press the sign in button
5. You should now be able to see the onboarding questions

**Note**: I wasn't able to properly test the whole flow, because of issues with redirects / auth happening on multiple different platforms and developer login opening on stream. We're only really able to tests styles at this point.

## Screenshots

image

image


olarclara:
Can you describe the steps for testing this?

tristandubbeld:
> Can you describe the steps for testing this?

Sorry, forgot to draft this! Still figuring out if we can test this properly without having to make changes in the code and "just rendering" the signin.

packages/app/src/app/pages/SignIn/Modal.tsx
```diff
@@ -121,18 +121,20 @@ export const SignInModalElement = ({
...
-  if (pendingUser) {
-    return (
-      <Wrapper oneCol>
-        <UserNameSelection />
-      </Wrapper>
-    );
+  /**
+   * 🚧 Utility to debug Trial Onboarding Questions
+   */
+  const TOQ_DEBUG = window.localStorage.getItem('TOQ_DEBUG') === 'ENABLED';
```
olarclara, line 133: It took me a sec to understand the `TOQ` 😄

packages/app/src/app/components/dashboard/InputText.tsx
```diff
@@ -0,0 +1,41 @@
...
+  }
+`;
+
+interface InputTextProps extends InputHTMLAttributes<HTMLInputElement> {
+  id: string;
+  label: string;
+  name: string;
+}
+
+export const InputText = ({
```
olarclara, line 31: Should this also handle validation to show ✅ or ❌ icons?

packages/app/src/app/pages/SignIn/components/UserNameSelection.tsx
```diff
@@ -1,88 +1,149 @@
...
+          <InputText
+            id="username"
+            name="username"
+            label="Username"
+            // TODO: Instead of onBlur we can debounce onchange
+            onBlur={async e => {
+              setLoadingUserName(true);
+              await validateUsername(e.target.value);
+              setLoadingUserName(false);
+            }}
```
olarclara, line 109: Is this working? I wasn't able to verify it.

olarclara (approved):
The UI looks good! ✨  A couple of questions but no need for a second pass, I believe.

packages/app/src/app/components/dashboard/InputText.tsx
tristandubbeld: Great question, and thanks for reminding me! In a proper design system I'd argue that it should be a different component as it has different requirements and functionality, and imo it is best to not create Swiss knife components. In this case it doesn't really matter but I'll see if it's easy to add - forgot about this.

packages/app/src/app/pages/SignIn/components/UserNameSelection.tsx
tristandubbeld: It should work exactly the same as it did before the "facelift". I will see if I can create a way to properly verify it in some way. Thanks.

packages/app/src/app/components/dashboard/InputText.tsx
tristandubbeld: I've created [a separate Linear ticket](https://linear.app/codesandbox/issue/XTD-300/onboarding-questions-validation) for this.

packages/app/src/app/pages/SignIn/ChooseProvider.tsx
```diff
@@ -0,0 +1,174 @@
...
+import history from 'app/utils/history';
+
+import { Button } from './components/Button';
+
+interface ChooseProviderProps {
+  redirectTo?: string;
+  onSignIn?: () => void;
+}
+
+export const ChooseProvider = ({
```
tristandubbeld, line 25: This component used to be `packages/app/src/app/pages/SignIn/Modal.tsx`. Most logic is copied from there, most significant changes is the `handleSignIn` function, which used to be three different functions.

packages/app/src/app/components/SignInModal.tsx
```diff
@@ -47,10 +49,35 @@ export const SignInModal = () => {
...
+                  boxSizing: 'border-box',
+                  backgroundColor: '#1D1D1D',
+                  borderRadius: '4px',
+                  overflow: 'hidden',
+                  margin: 'auto',
+                  padding: '24px',
+                  width: '100%',
+
+                  // With tokens
+                  boxShadow: '2',
```
alexnm, line 73: Do we use box shadow for modals in prism? I was thinking about what to do with things like animations, box shadows and other tokens that are a bit easier to sync between prism and the design system here, maybe it makes sense to keep the tokens with the same names (if possible) so the transition is smoother?

packages/app/src/app/components/SignInModal.tsx
```diff
@@ -47,10 +49,35 @@ export const SignInModal = () => {
...
+                })}
+              >
+                <Stack direction="vertical" gap={64}>
+                  {/* <Element paddingBottom={6}>Welcome</Element> */}
```
alexnm, line 77: I this still needed?

packages/app/src/app/components/dashboard/Button.tsx
```diff
@@ -0,0 +1,13 @@
+import styled from 'styled-components';
+import { Button as ButtonComponent } from '@codesandbox/components';
+
+export const Button = styled(ButtonComponent)`
```
alexnm, line 4: Is this actually a variant of the Button component or just some style adjustments to fit a specific scenario in the dashboard? I would either move this to the Button component if it will be a common use case (as a variant?) or I would call it `StyledButton` to make it easier when moving the screens later

packages/app/src/app/components/dashboard/InputSelect.tsx
```diff
@@ -0,0 +1,92 @@
+import React, { SelectHTMLAttributes } from 'react';
+import styled from 'styled-components';
+import { Icon, Stack } from '@codesandbox/components';
+import { Label } from './Label';
+
+const RelativeWrapper = styled.div`
```
alexnm, line 6: food for thought: in v2 we use a lot of instances of `Box` for quick styling for a div. Should we add it here as well as a copy of the `Element` component for future compatibility?. We can limit the usage of `styled.div` for a single css rule that we need in the hierarchy this way.

packages/app/src/app/pages/SignIn/Onboarding.tsx
```diff
@@ -0,0 +1,156 @@
+import React, { useState } from 'react';
+import { Element, Text, Stack } from '@codesandbox/components';
+import { useAppState, useActions } from 'app/overmind';
+import { InputText } from 'app/components/dashboard/InputText';
+import { InputSelect } from 'app/components/dashboard/InputSelect';
+import { Button } from 'app/components/dashboard/Button';
+
+const roleOptions = [
```
alexnm, line 8: nitpick: can you use UPPER_CASE here to make it clear these are constant values? not some dynamic / state provided object

packages/app/src/app/pages/SignIn/Onboarding.tsx
```diff
@@ -0,0 +1,156 @@
...
+          as="img"
+          css={{
+            width: 64,
+            height: 64,
+            borderRadius: '100%',
+          }}
+          alt={pendingUser.username}
+          src={pendingUser.avatarUrl}
+        />
+        <Text
```
alexnm, line 79: I know a lot of my comments have been about future plans, but another idea for these texts is to bring a similar API as the new typography components in prism (Heading, Paragraph, Text)

packages/app/src/app/pages/SignIn/ChooseProvider.tsx
```diff
@@ -0,0 +1,174 @@
...
+          maxWidth: '200px',
+          margin: '24px 0 0 0',
+
+          a: {
+            color: 'inherit',
+          },
+        })}
+      >
+        By continuing, you agree to CodeSandbox{' '}
+        <a href="https://codesandbox.io/legal/terms">Terms of Service</a>,{' '}
```
alexnm, line 169: target="_blank" maybe?

alexnm (approved):
Left a few comments here and there, but don't want to block the PR if work is needed on other tasks

packages/app/src/app/components/SignInModal.tsx
tristandubbeld: Some modals will have a header with a title, I had it here just in case. Waiting for the designs to be finished before removing it.

packages/app/src/app/components/dashboard/InputSelect.tsx
tristandubbeld: Yeah that's a good idea. We can use `<Element css={{}} />`.

packages/app/src/app/pages/SignIn/ChooseProvider.tsx
tristandubbeld: Copied this from somewhere but this is a good idea, I'll add it :)

packages/app/src/app/components/SignInModal.tsx
tristandubbeld: I've checked the designs, tokens and components and what their shadow values are right now. I'll discuss with Lena what we should do here.

design:
```
0px 4px 4px rgba(0, 0, 0, 0.25);
```
v1 token 2:
```
0px 4px 4px rgba(0, 0, 0, 0.12), 0px 16px 32px rgba(0, 0, 0, 0.24)
```

in v2 prism we have a popup shadow, but no dialog shadow it seems:
```
0px 2px 7px rgba(0, 0, 0, 0.15), 0px -2px 7px rgba(0, 0, 0, 0.15), 0px 5px 17px rgba(0, 0, 0, 0.3)
```

packages/app/src/app/pages/SignIn/Onboarding.tsx
tristandubbeld: This is a good idea, but it will take some time to properly create these. Should we pick this up at a later time?

PR #7040, opened 2022-10-19
feat(trial): create team and subscribe to pro

olarclara:

Adds UI for signing up for the trial on the create team flow.

## What is the new behavior?

Loom: https://www.loom.com/share/1cbc221612f444d0bf5940f0043c9618

## What steps did you take to test this? This is required before we can merge, make sure to try the flow you've updated.

⚠️ The trial flag is enabled on the back end only, so this should be tested against the stream.

1. Open the dashboard
2. In the workspace selector, click "Create a new team"
3. Go through step 1 (actual team creation)
4. Go through step 2 (placeholder for team members)
5. In step 3, click on "Start 14-day free trial"
6. In stripe, use a test card
7. After a successful subscription, get redirected to your new team settings

### DIFF
--- a/packages/app/src/app/hooks/useCreateCheckout.ts
+++ b/packages/app/src/app/hooks/useCreateCheckout.ts
@@ -34,7 +34,7 @@ export const useCreateCheckout = (): [
     } catch (err) {
       setStatus({
         status: 'error',
-        error: err?.message ?? 'Failed to create checkout link.',
+        error: err?.message ?? 'Failed to create checkout link',
       });
     }
   };
--- a/packages/app/src/app/hooks/useCreateCheckout.ts
+++ b/packages/app/src/app/hooks/useCreateCheckout.ts
@@ -1,0 +1,43 @@
+import { useState } from 'react';
+import { useEffects } from 'app/overmind';
+import { dashboard } from '@codesandbox/common/lib/utils/url-generator';
+
+type CheckoutStatus =
+  | { status: 'idle' }
+  | { status: 'loading' }
+  | { status: 'error'; error: string };
+
+export const useCreateCheckout = (): [
+  CheckoutStatus,
+  (arg: Record<'team_id' | 'recurring_interval', string>) => void
+] => {
+  const [status, setStatus] = useState<CheckoutStatus>({ status: 'idle' });
+  const { api } = useEffects();
+
+  const createCheckout = async ({
+    team_id,
+    recurring_interval,
+  }: Record<'team_id' | 'recurring_interval', string>) => {
+    try {
+      setStatus({ status: 'loading' });
+
+      const payload = await api.stripeCreateCheckout({
+        success_path: dashboard.settings(team_id) + '&payment_pending=true',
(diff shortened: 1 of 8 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7040/commits/2f016068f3258efb09c6cb1eec4287d7d5ca3a66 is at https://pr7040.build.csb.dev/s/new.

packages/app/src/app/hooks/useCreateCheckout.ts
```diff
@@ -0,0 +1,43 @@
+import { useState } from 'react';
+import { useEffects } from 'app/overmind';
+import { dashboard } from '@codesandbox/common/lib/utils/url-generator';
+
+type CheckoutStatus =
+  | { status: 'idle' }
+  | { status: 'loading' }
+  | { status: 'error'; error: string };
+
+export const useCreateCheckout = (): [
```
alexnm, line 10: I see this is a single time use async request basically. But I'm still wondering how we can integrate it with the rest of the overmind state, or if it makes sense to integrated it. More like food for thought. We could technically have a separate namespace for the signup flow, where we keep track of all the client state as the user goes through the flow. But I also see the simplicity with this hook and it is somewhat similar with what we are used to from v2.

packages/app/src/app/overmind/namespaces/pro/actions.ts
```diff
@@ -3,6 +3,14 @@ import { Context } from 'app/overmind';
 import { withLoadApp } from 'app/overmind/factories';
 import { Step, Plan, PaymentSummary, PaymentPreview } from './types';
 
+export const getPrices = async ({ state, effects }: Context) => {
+  try {
+    state.pro.prices = await effects.api.prices();
+  } catch (err) {
+    // Fail silently.
```
alexnm, line 10: is this a common scenario? should we track it in sentry?

alexnm (approved):

tristandubbeld (approved):

tristandubbeld:
@alexnm I'm merging this PR as I need it for the next step. We should definitely keep an eye on your comments though, should we create Linear tasks?

packages/app/src/app/hooks/useCreateCheckout.ts
olarclara: @alexnm yep, I tried to follow the pattern for hooks that we have on v2 since we will eventually move there.

packages/app/src/app/overmind/namespaces/pro/actions.ts
olarclara: @alexnm I don't think so but I also didn't want to trigger notifications since the price is just a footer label and is not a blocker of the experience. Should the back-end have the logs if these requests fail or is the front-end's responsibility to tell that the request has failed?

alexnm: makes sense, if it's not on the critical path, no logging is needed

PR #7049, opened 2022-10-23
feat(trial): Invite team members

tristandubbeld:
### Todo

- [x] ~Generate new GraphQL types~

No need to generate new GraphQL types, still using the `email` property and it still accepts a string.

### How to test

1. Open the dashboard
2. Create a new team
3. Input emails in the textarea
4. Press submit
5. Check if the next step is active and emails have been sent

### Video

https://user-images.githubusercontent.com/7533849/197798843-2894ca83-feee-437d-92de-a68b11f14d5b.mov

### DIFF
--- a/packages/app/src/app/pages/common/Modals/NewTeamModal/TeamMembers.tsx
+++ b/packages/app/src/app/pages/common/Modals/NewTeamModal/TeamMembers.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
-import { Button, Stack, Textarea, Text } from '@codesandbox/components';
+import { Button, Stack, Text } from '@codesandbox/components';
 import { useAppState } from 'app/overmind';
+import { StyledButton } from 'app/components/dashboard/Button';
+import { Textarea } from 'app/components/dashboard/Textarea';
 
 export const TeamMembers: React.FC<{ onComplete: () => void }> = ({
   onComplete,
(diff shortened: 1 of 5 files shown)


packages/app/src/app/pages/common/Modals/NewTeamModal/TeamMembers.tsx
```diff
@@ -1,25 +1,150 @@
...
+        : `${isFirstEmail ? '' : ', '}${invalidEmail}`
+    }`;
+  });
+}
 
 export const TeamMembers: React.FC<{ onComplete: () => void }> = ({
   onComplete,
 }) => {
   const { activeTeamInfo } = useAppState();
+  const { gql } = useEffects();
```
alexnm, line 37: this validation behavior can be extracted into a custom hook maybe. not sure if it's going to be reused later

alexnm (approved):

packages/app/src/app/pages/common/Modals/NewTeamModal/TeamMembers.tsx
```diff
@@ -1,25 +1,150 @@
...
+        onComplete();
+      } catch (error) {
+        // ❗️ TODO: Validate if this works!
+        // Copied logic from inviteToTeam function in dashboard/actions.ts
+        const message =
+          error?.response?.errors?.[0]?.message ||
+          'There was a problem sending the invitations.';
+
+        setInviteError(message);
+      }
```
tristandubbeld, line 77: Tried to validate this, but for now all "errors" return `200`. Client side validation makes sure it's properly caught, unless someone willingly turns it off. We don't supply an error for that right now (we would have to check the result of the mutation for a certain string that is susceptible to change).

I would like to keep this logic though, in case we receive a 500 or any other error. What do you think?

PR #7060, opened 2022-10-26
fix(dashboard): nullcheck for selected team after accepting invite

tristandubbeld:

Checking if we the selected team is in the state before reading properties. I have videos with current behaviour and proof but those show an invite link so I'm not uploading them here :)
 
After an invite, the dashboard breaks when joining a team since the selected team is still `undefined` on first render.

## What is the new behavior?

After an invite, the dashboard works and after the second render the selected team is defined.


alexnm (approved):
Great! Missed this edge case

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7060/commits/59f2f4f784c7b53f3dc32d9664ccc022942098e5 is at https://pr7060.build.csb.dev/s/new.

PR #7065, opened 2022-10-26
fix(dashboard): empty state

olarclara:

The `allCollections` field might be null but the type checker has not caught it, this PR handles this possibility.

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Components/TemplatesRow/TemplatesRow.tsx
+++ b/packages/app/src/app/pages/Dashboard/Components/TemplatesRow/TemplatesRow.tsx
@@ -14,9 +14,9 @@ import { TemplatesGrid } from './elements';
 export const TemplatesRow: React.FC = () => {
   const officialTemplates = useOfficialTemplates();
   const actions = useActions();
-  const { dashboard } = useAppState();
+  const { activeTeam, dashboard } = useAppState();
 
-  const collection = dashboard.allCollections.find(c => c.path === '/');
+  const collection = dashboard.allCollections?.find(c => c.path === '/');
 
   const handleOpenTemplate = (template: TemplateFragment) => {
     const { sandbox } = template;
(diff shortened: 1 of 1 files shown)


packages/app/src/app/pages/Dashboard/Components/TemplatesRow/TemplatesRow.tsx
```diff
@@ -38,6 +38,10 @@ export const TemplatesRow: React.FC = () => {
     actions.modals.newSandboxModal.close();
   };
 
+  React.useEffect(() => {
+    actions.dashboard.getAllFolders();
+  }, [actions.dashboard, activeTeam]);
```
olarclara, line 43: The sidebar already makes this call, I'm not sure if we should have this or if the change in line 19 would be enough.

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7065/commits/112e20d4b296dc00e58427567b8442694872cfa7 is at https://pr7065.build.csb.dev/s/new.

packages/app/src/app/pages/Dashboard/Components/TemplatesRow/TemplatesRow.tsx
alexnm: Would this be needed here? I see it is already called twice, once from the sidebar and once from the page, will check if that's ok in a separate PR, but I wouldn't add this.

tristandubbeld: I imagine we want to call this when the `TemplatesRow` is being used on its own - ~is the `SideBar` always rendered, e.g. on mobile?~ On mobile the `SideBar` is rendered too, so it's always being called.

The only reason we would want to call this _again_ is when we open a page with `TemplatesRow`, so potential newly created templates are fetched. It isn't the most glamorous approach though, and probably doesn't happen very often.

PR #7073, opened 2022-10-27
fix: null error and duplicate call

alexnm:
Started fixing a duplicate call to getAllFolders, ended up with 3 different things:

 - **getAllFolders** called only once on the sidebar, when the activeTeam changes
 - null/undefined error when no **subscription** exists for team in team settings
 - **button in link** replaced with a simple button for upgrading to pro

For later: wasn't able to figure out why the router push to `/pro` doesn't work, so I left it as an anchor tag instead of using react router.

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Content/routes/Sandboxes/index.tsx
+++ b/packages/app/src/app/pages/Dashboard/Content/routes/Sandboxes/index.tsx
@@ -1,12 +1,11 @@
 import React from 'react';
 import { Helmet } from 'react-helmet';
-import { useParams, useHistory } from 'react-router-dom';
+import { useParams } from 'react-router-dom';
 import { useAppState, useActions } from 'app/overmind';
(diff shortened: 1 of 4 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7073/commits/9f3d069bbd994f7c94259ffd832b14b37e9375c5 is at https://pr7073.build.csb.dev/s/new.

tristandubbeld (approved):
The `localTeam` effect looks like it's used for a specific use case, do we know what that is? It isn't really clear from the code.

alexnm:
> The `localTeam` effect that's removed looks like it's used for a specific use case, do we know what that is? It isn't really clear from the code.

Yes, it was going back to the root folder in sandboxes, when the team changes. But it's no longer needed, as we now take the users to the recent page when changing the team.

PR #7096, opened 2022-11-01
feat: useDismissible

tristandubbeld:
In the future we'll be writing more dismissible banners etc. that we want to keep track of in `localStorage`. We already created a few in the past weeks but found that keeping track and DX wasn't optimal. From now on we can use this hook to dismiss and check if items are dismissed while keeping track of all values.

Example usage
```tsx
const Component = () => {
  const { banners } = useAction();
  const [isBannerDismissed, dismissBanner] = useDismissible('BANNER');

  const onClose = () => {
    banners.banner.close();
    dismissBanner();
  };

  if (isBannerDismissed) {
    return null;
  }

  return (
    <Banner />
  );
};

```

Keep in mind that this hook does not keep track of state, most of this is handled in overmind. This hook should mostly be used for initial state of showing banners by a flag in `localStorage`.

### DIFF
--- a/packages/app/src/app/hooks/useDismissible.ts
+++ b/packages/app/src/app/hooks/useDismissible.ts
@@ -10,16 +10,17 @@ type DismissibleKeys =
 
 type Dismissibles = Partial<Record<DismissibleKeys, true>>;
 
-export const useDismissible = () => {
+export const useDismissible = (key: DismissibleKeys) => {
   const { browser } = useEffects();
+
   const dismissibles = browser.storage.get<Dismissibles>('DISMISSIBLES');
+  const isDismissed = dismissibles?.[key] === true;
 
-  const isDismissed = (key: DismissibleKeys) => {
-    return dismissibles?.[key] === true;
-  };
-
-  const dismiss = (key: DismissibleKeys) => {
-    const prevDismissibles = dismissibles || {};
+  const dismiss = () => {
+    // Getting the dismissibles again to make sure other instances aren't
+    // overwritten after the hook initialised.
+    const prevDismissibles =
+      browser.storage.get<Dismissibles>('DISMISSIBLES') || {};
 
     browser.storage.set('DISMISSIBLES', {
       ...prevDismissibles,
@@ -27,5 +28,5 @@ export const useDismissible = () => {
     });
   };
 
-  return { isDismissed, dismiss };
+  return [isDismissed, dismiss];
 };
--- a/packages/app/src/app/hooks/useDismissible.ts
+++ b/packages/app/src/app/hooks/useDismissible.ts
@@ -1,0 +1,31 @@
+import { useEffects } from 'app/overmind';
+
+/**
+ * Localstorage keys used for dismissible modals and banners.
+ */
+type DismissibleKeys =
+  // TODO - Any key in this type should refer to a dismissible item. Right
+  // now there are none yet.
+  'TODO';
+
+type Dismissibles = Partial<Record<DismissibleKeys, true>>;
+
+export const useDismissible = () => {
+  const { browser } = useEffects();
+  const dismissibles = browser.storage.get<Dismissibles>('DISMISSIBLES');
+
+  const isDismissed = (key: DismissibleKeys) => {
+    return dismissibles?.[key] === true;
+  };
+
+  const dismiss = (key: DismissibleKeys) => {
+    const prevDismissibles = dismissibles || {};
+
+    browser.storage.set('DISMISSIBLES', {
+      ...prevDismissibles,
+      [key]: true,
+    });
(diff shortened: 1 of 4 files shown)


packages/app/src/app/hooks/useDismissible.ts
```diff
@@ -0,0 +1,31 @@
...
+  const dismiss = (key: DismissibleKeys) => {
+    const prevDismissibles = dismissibles || {};
+
+    browser.storage.set('DISMISSIBLES', {
+      ...prevDismissibles,
+      [key]: true,
+    });
+  };
+
+  return { isDismissed, dismiss };
```
tristandubbeld, line 30: Iterated on the API and used multiple versions, this one felt the best. Let me know if you have questions about the implementation.

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7096/commits/340e069cfd4c2b1e422e79efa29358e276143f9c is at https://pr7096.build.csb.dev/s/new.

packages/app/src/app/hooks/useDismissible.ts
```diff
@@ -0,0 +1,31 @@
...
+type DismissibleKeys =
+  // TODO - Any key in this type should refer to a dismissible item. Right
+  // now there are none yet.
+  'TODO';
+
+type Dismissibles = Partial<Record<DismissibleKeys, true>>;
+
+export const useDismissible = () => {
+  const { browser } = useEffects();
+  const dismissibles = browser.storage.get<Dismissibles>('DISMISSIBLES');
```
alexnm, line 15: wouldn't it be safer for each instance of `useDismissible` to keep track of a single key that it could get as a param? so that two instances of the hook will not have different values inside this `dismissibles` reference?

alexnm: then there's no need for a `isDismissed` function, as the hook returns the boolean if the key passed as a param is dismissed or not

alexnm (approved):
Left a not there about a possible change, but this implementation also works

packages/app/src/app/hooks/useDismissible.ts
tristandubbeld: You mean use it like this?
```tsx
const { isDismissed: isSomethingDismissed, dismiss: dismissSomething } = useDismissible('SOMETHING');
```

tristandubbeld: > so that two instances of the hook will not have different values inside this dismissibles reference?

I think with `localStorage` this will always be a potential problem, right? When two of the same hooks are referencing the same key but for one hook it's dismissed, it doesn't re-render the other hooks and those values will still be the same. Not sure if this will ever be the case with dismissibles though?

alexnm: yes like this. not bulletproof indeed, but most likely useDismissible("key") will only be used once, whereas useDismissible() without any argument will have the reference to all the keys in parallel

tristandubbeld: @olarclara what do you think?

packages/app/src/app/overmind/effects/browser.ts
```diff
@@ -91,15 +91,15 @@ export default {
...
   },
   storage: {
-    get(key: string): unknown | null {
+    get<T = unknown>(key: string): T | null {
       try {
         const value = localStorage.getItem(key);
 
         if (value) {
-          return JSON.parse(value);
+          return JSON.parse(value) as T;
```
olarclara, line 102: Nice!

packages/app/src/app/hooks/useDismissible.ts
olarclara: @tristandubbeld I like @alexnm's suggestion.

packages/app/src/app/hooks/useDismissible.ts
olarclara, line 30: If each instance tracks per key, could the return be similar to React's own `useState`?

```suggestion
  return [isDismissed, dismiss];
```

This way we don't need to rename keys.

olarclara (approved):

packages/app/tsconfig.strictNullChecks.json
```diff
@@ -6,6 +6,9 @@
     "strictNullChecks": true,
     "skipLibCheck": true
   },
-  "include": ["./src/app/overmind/**/*.ts"],
+  "include": [
+    "./src/app/overmind/**/*.ts",
+    "./src/app/hooks/useDismissible.ts"
```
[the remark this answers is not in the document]
tristandubbeld: Added the folder and fixed 1 issue :)

PR #7098, opened 2022-11-02
Configure storybook for app

tristandubbeld:
Started configuring Storybook so we can run it for components in the app folder, but as it turns out, components are a lot more dependent on their context than expected. For now I will stop iterating on this and leave the code to run storybook in this PR.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7098/commits/3450efd7e5843e5fc5ea2b2bb54cefdd559ef8a2 is at https://pr7098.build.csb.dev/s/new.

PR #7099, opened 2022-11-02
feat(trial): view only state for dashboard cards and list items

tristandubbeld:
Dashboard cards and list items will have (slightly) different styles and a "View only" badge when they're in view only mode. Some actions in the context menu are disabled when in view only mode. 

#### How to test

`Cards` and `ListItem`s with changes now have an `isViewOnly` property. Since we don't have the data yet to properly populate this it's easiest to test by changing the destructured property to `isViewOnly = true`.

For the different context menus you can use the `isViewOnly` variable and set it to `true` instead of `false` to view the changes.

#### Note

Depending on how this mode will be implemented, the property name can change. In the frontend it will behave the same though.

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Components/Branch/BranchCard.tsx
+++ b/packages/app/src/app/pages/Dashboard/Components/Branch/BranchCard.tsx
@@ -5,6 +5,8 @@ import {
   Stack,
   Text,
   Tooltip,
+  Element,
+  Badge,
 } from '@codesandbox/components';
 import { css } from '@styled-system/css';
 import { BranchProps } from './types';
@@ -15,6 +17,7 @@ export const BranchCard: React.FC<BranchProps> = ({
   isBeingRemoved,
   selected,
   onContextMenu,
+  isViewOnly = true,
   ...props
 }) => {
   const { name: branchName, project, contribution } = branch;
@@ -57,7 +60,7 @@ export const BranchCard: React.FC<BranchProps> = ({
     >
       <Stack
         css={css({
-          backgroundColor: 'rgba(229, 229, 229, 0.04)',
+          backgroundColor: isViewOnly ? '#252525' : '#161616',
           paddingY: 11,
           position: 'relative',
         })}
@@ -66,6 +69,13 @@ export const BranchCard: React.FC<BranchProps> = ({
       >
         <Icon color="#808080" name="branch" size={32} />
       </Stack>
+      {isViewOnly ? (
+        <Element css={{ position: 'absolute', top: 8, left: 8 }}>
+          <Badge color="accent" isPadded>
+            View only
+          </Badge>
+        </Element>
+      ) : null}
       <Stack
         css={css({
           paddingY: 5,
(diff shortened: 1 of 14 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7099/commits/8fa9da035728cab6539c8141524eb118fb5aeea8 is at https://pr7099.build.csb.dev/s/new.

packages/app/src/app/pages/Dashboard/Components/Branch/BranchCard.tsx
```diff
@@ -57,7 +60,7 @@ export const BranchCard: React.FC<BranchProps> = ({
     >
       <Stack
         css={css({
```
alexnm, line 62: nitpick, can we try to get rid of the css from styled-system here?

packages/app/src/app/pages/Dashboard/Components/SyncedSandbox/SyncedSandboxListItem.tsx
```diff
@@ -19,6 +20,7 @@ export const SyncedSandboxListItem = ({
   onDoubleClick,
   onContextMenu,
   isScrolling,
+  isViewOnly,
```
alexnm, line 23: same question as above, should we even consider the synced sandboxes part of the limitation?

packages/components/src/components/Badge/index.tsx
```diff
@@ -6,28 +6,38 @@ import { Text } from '../Text';
...
-  <Stack
-    css={{
-      alignItems: 'center',
-      padding: icon ? '4px 8px' : '1px 6px',
-      userSelect: 'none',
+}) => {
+  let padding = '1px 6px';
 
-      borderRadius: '999px',
+  if (isPadded) {
```
alexnm, line 20: does it make sense to keep the logic that if it has an icon, it should be padded? so `isPadded || icon`?

alexnm (approved):
Left a few comments, but I don't think this requires a second pass

packages/components/src/components/Badge/index.tsx
tristandubbeld: The values for `isPadded`, `icon` and default are all different. Made sure that the `Badge` stays the same for previous implementations.

packages/app/src/app/pages/Dashboard/Components/SyncedSandbox/SyncedSandboxListItem.tsx
tristandubbeld: I don't know actually. When implementing this it felt weird to me that we wouldn't add the view only or restrictions to synced sandboxes, so I just did it for these as well.

packages/app/src/app/pages/Dashboard/Components/Branch/BranchCard.tsx
tristandubbeld: In this case it's using `paddingY: 11` which results in `44px`. We can update it to hard-coded `44px` and remove the `css` function, is that something we want?

PR #7100, opened 2022-11-02
feat(trial): Disabled styles for import and create branch buttons

tristandubbeld:
Added `isDisabled` state for repository buttons. This state should be enabled when the repository is in view only mode. The logic to trigger this does not yet exist.

#### How to test

`NewBranch` and `ImportRepository` now have an `isDisabled` property. Since we don't have the data yet to properly populate this it's easiest to test by changing the destructured property to `isDisabled = true`.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7100/commits/f189977c639acd1341cdfb474bca6eec3c9c0538 is at https://pr7100.build.csb.dev/s/new.

alexnm (approved):

olarclara (commented):
Should we also add `aria-disabled="true"` to these elements?

PR #7106, opened 2022-11-03
feat(trial): Reusable banner components

tristandubbeld:
Created `MessageStripe` and `Banner` components. I've also updated some components to be more flexible and to make sure we don't have to use the same `css` prop with the same values everytime.

**Message Stripe Trial**

image

**Message Stripe Warning**

image

**Message Stripe without action**

image

**Message Stripe with space between**

image

**Banner**
Note: Everett font does not work in Storybook

image

### DIFF
--- a/packages/components/src/components/Banner/Banner.tsx
+++ b/packages/components/src/components/Banner/Banner.tsx
@@ -1,0 +1,37 @@
+import React from 'react';
+
+import { Element } from '../Element';
+import { IconButton } from '../IconButton';
+
+interface BannerProps {
+  children: React.ReactNode;
+  onDismiss?: () => void;
+}
+
+export const Banner = ({ children, onDismiss }: BannerProps) => {
+  return (
+    <Element
+      css={{
+        position: 'relative',
+        padding: [4, 6, 8],
+        backgroundColor: '#252525',
+        borderRadius: '4px',
+      }}
+    >
+      {children}
+
+      {onDismiss ? (
+        <Element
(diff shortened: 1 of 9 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7106/commits/e1dc7a449273931e1b397079c564066479e4fcfc failed.

olarclara (approved):

packages/components/src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,37 @@
...
+  children: React.ReactNode;
+  onDismiss?: () => void;
+}
+
+export const Banner = ({ children, onDismiss }: BannerProps) => {
+  return (
+    <Element
+      css={{
+        position: 'relative',
+        padding: [4, 6, 8],
```
alexnm, line 16: is this a styled-component or a styled-system feature? I thought it only works with the css() function

alexnm (approved):
💯

packages/components/src/components/Banner/Banner.tsx
tristandubbeld: Styled System! It does only work with the `css()` function, yes. The `Element` component uses it to generate the component, so we can pass it with the `css` prop:
```tsx
export const Element = styled.div<IElementProps>(props =>
  css({
    // stuff
    ...(props.css || {}),
  })
);
  ```

PR #7109, opened 2022-11-04
chore(cleanup): remove console errors and warnings

tristandubbeld:
Got rid of a bunch of browser console errors.


packages/app/src/app/pages/Dashboard/Components/Sandbox/SandboxCard.tsx
```diff
@@ -235,28 +235,28 @@ export const SandboxCard = ({
       onBlur={onBlur}
       onContextMenu={onContextMenu}
       {...props}
-      css={css({
+      css={{
```
tristandubbeld, line 238: Had to get rid of the `css` function here. This was a complex one, as it merged components with the `as` prop (the `Stack` and `Link`) which both contain `styled-components` and `styled-system` styles. The wrong css prop was applied by the `Stack`, but passed to the `Link` as well, which in turn couldn't apply the css properly because it had already been used on the `Stack` (?) and styled-system works in mysterious ways.

It resulted in adding `_css2` or something attribute to an `a` tag, which resulted in a big red console error. I feel like we're better of this way, because it takes away the two styling solutions merged into each other too.

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7109/commits/eeb9e409f05c5a134470d5d63ed4a14c636aa137 is at https://pr7109.build.csb.dev/s/new.

alexnm (approved):
Great 💯

PR #7110, opened 2022-11-04
feat(pro): UI updates

olarclara:

- Extends the `Badge` component to accept more props.
- Removes the `Pro` badge.
- Adds the `Free` badge where appropriate.
- Adds the link to team settings.
- Removes unused code + organizes the folder.

1. Visit `/pro`.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7110/commits/d90c196d853159d378a16595578f491093e90332 is at https://pr7110.build.csb.dev/s/new.

packages/app/src/app/pages/Pro/components/Switcher.tsx
```diff
@@ -56,12 +61,25 @@ export const Switcher: React.FC<{
...
+          <Stack css={{ marginLeft: 24 }} direction="vertical">
+            <Stack align="center" gap={1}>
+              <WorkspaceName>
+                <span>{activeTeamInfo.name}</span>
+              </WorkspaceName>
+              {isFreeWorkspace && (
+                <Badge
+                  color="accent"
+                  css={{
+                    height: 'fit-content',
```
tristandubbeld, line 73: Can't check it right now, but `fit-content` isn't supported for Firefox according to MDN, will this cause issues on FF? I'll come back to this if I'm able to test again!

tristandubbeld: It seems to work, but removing the height style seems to look the same?

tristandubbeld (approved):

packages/app/src/app/pages/Pro/components/Switcher.tsx
alexnm: I would also avoid passing generic props/css to certain components which should be "closed" like the badge, to ensure we don't abuse this further down the line

alexnm (approved):
image
Disabled options in the select still have a weird extra padding that needs to be removed, other then that, it looks good. I would suggest some additional changes here, like not even showing the teams that are disabled, but we should merge this first and follow-up with an improvement ticket of the page for later.

PR #7122, opened 2022-11-09
feat(pro): Public sandbox limits

tristandubbeld:
#### Changes

- Fixed a few styling issues in the `MessageStripe` and made sure that the component and `Banner` are now properly exported from the package.

- Fixed an issue with `react-hot-reload` where mapping over children uses proxy components. With proxy components we aren't able to check the child type, which results in broken "augmented" children in the `MessageStripe` component.

- Added the `MessageStripe` with all the conditionals to the "All sandboxes" page.

- Created a hook `useSubscription` to contain subscription logic extracting it a way from the components and enabling reuse.

#### Testing

1. Open the code in the editor
2. Open `packages/app/src/app/pages/Dashboard/Content/routes/Sandboxes/index.tsx`
3. Change the `hasMaxSandboxes` boolean const to `true`
4. If you want to test certain scenario's use different accounts or change boolean values around

#### Screenshots

---

- Has max sandboxes ✅ 
- Eligible for trial ✅ 
- Current user is team admin ✅ 

image

---

- Has max sandboxes ✅ 
- Eligible for trial ✅ 
- Current user is team admin ❌

image

---

- Has max sandboxes ✅
- Eligible for trial ❌
- Current user is team admin ✅

image

---

- Has max sandboxes ✅
- Eligible for trial ❌
- Current user is team admin ❌

image

### DIFF
--- a/packages/app/config/webpack.common.js
+++ b/packages/app/config/webpack.common.js
@@ -359,6 +359,12 @@ module.exports = {
           name: 'static/media/[name].[hash:8].[ext]',
         },
       },
+      // https://github.com/gaearon/react-hot-loader/issues/1311
+      {
+        test: /\.js$/,
+        include: /node_modules\/react-dom/,
+        use: ['react-hot-loader/webpack'],
+      },
     ],
 
     noParse: [
--- a/packages/app/src/app/hooks/useSubscription.ts
+++ b/packages/app/src/app/hooks/useSubscription.ts
@@ -1,0 +1,63 @@
+import {
+  SubscriptionStatus,
+  SubscriptionType,
+  TeamMemberAuthorization,
+} from 'app/graphql/types';
+import { useAppState } from 'app/overmind';
+
+export const useSubscription = () => {
+  const {
+    activeTeam,
+    activeTeamInfo,
+    personalWorkspaceId,
+    activeWorkspaceAuthorization,
+  } = useAppState();
+
+  /**
+   * Personal states
+   */
+  const isPersonalSpace = activeTeam === personalWorkspaceId;
+
+  /**
+   * Team states
+   */
+  const isTeamSpace = !isPersonalSpace;
+
+  const isTeamAdmin =
+    isTeamSpace &&
+    activeWorkspaceAuthorization === TeamMemberAuthorization.Admin;
+
+  /**
+   * Subscription states
+   */
+
+  // There are different statuses for a subscription, but only ACTIVE and TRIALING
+  // should be considered an active TeamPro subscription.
+  // TODO: This might change based on how we use other statuses in the subscription (eg: PAUSED)
+  const hasActiveSubscription =
+    activeTeamInfo?.subscription?.status === SubscriptionStatus.Active ||
+    activeTeamInfo?.subscription?.status === SubscriptionStatus.Trialing;
+
+  const hasPastSubscription = Boolean(activeTeamInfo?.subscription?.status);
+
+  /**
+   * Trial states
+   */
+  const hasActiveTeamTrial =
+    isTeamSpace &&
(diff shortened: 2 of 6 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7122/commits/742672e542ea66d52c7468d2b3e8269266186fa5 is at https://pr7122.build.csb.dev/s/new.

packages/app/config/webpack.common.js
```diff
@@ -359,6 +359,12 @@ module.exports = {
           name: 'static/media/[name].[hash:8].[ext]',
         },
       },
+      // https://github.com/gaearon/react-hot-loader/issues/1311
+      {
+        test: /\.js$/,
+        include: /node_modules\/react-dom/,
+        use: ['react-hot-loader/webpack'],
+      },
```
tristandubbeld, line 367: To make sure the `MessageStripe` works with mapping over children and adding props to the `Action` child, I had to include this webpack plugin. I am by no means a webpack wizard, so please review this carefully.

packages/app/src/app/hooks/useSubscription.ts
```diff
@@ -0,0 +1,63 @@
...
+   */
+
+  // There are different statuses for a subscription, but only ACTIVE and TRIALING
+  // should be considered an active TeamPro subscription.
+  // TODO: This might change based on how we use other statuses in the subscription (eg: PAUSED)
+  const hasActiveSubscription =
+    activeTeamInfo?.subscription?.status === SubscriptionStatus.Active ||
+    activeTeamInfo?.subscription?.status === SubscriptionStatus.Trialing;
+
+  const hasPastSubscription = Boolean(activeTeamInfo?.subscription?.status);
```
alexnm, line 41: would a better name for this be: `hasExistingOrPastSubscription`?

packages/app/src/app/hooks/useSubscription.ts
```diff
@@ -0,0 +1,63 @@
...
+  const isPersonalSpace = activeTeam === personalWorkspaceId;
+
+  /**
+   * Team states
+   */
+  const isTeamSpace = !isPersonalSpace;
+
+  const isTeamAdmin =
+    isTeamSpace &&
+    activeWorkspaceAuthorization === TeamMemberAuthorization.Admin;
```
alexnm, line 28: I would keep this away from the hook, unless it is needed in the subscription fields in any way. It would make the hook more focused on subscription data only and we can even have a separate `useWorkspaceAuthorization` that can return things like: `isTeamAdmin` or `isPersonalSpace`

packages/app/src/app/pages/Dashboard/Sidebar/index.tsx
```diff
@@ -100,48 +90,24 @@ export const Sidebar: React.FC<SidebarProps> = ({
...
     : null;
 
   const showBottomMessage =
-    isTeamFree ||
-    isPersonalFree ||
-    (hasActiveTeamProTrial &&
-      trialDaysLeft !== null &&
-      trialDaysLeft <= END_OF_TRIAL_DAYS_NOTIFICATION);
+    !hasActiveSubscription ||
+    (trialDaysLeft !== null && trialDaysLeft <= END_OF_TRIAL_DAYS_NOTIFICATION);
```
alexnm, line 110: nice, this simplifies the condition! logic checks out as far as I can tell

alexnm (approved):
Logic checks out, I suggested two small changes in `useSubscription` but you can also merge this as it is and we can try making the changes in the subsequent PR

packages/app/src/app/hooks/useSubscription.ts
tristandubbeld: Yeah we can definitely do that!

PR #7144, opened 2022-11-14
feat(subscription): Public repo and sandbox max

tristandubbeld:
When working on `XTD-362` and `XTD-359` we weren't able to calculate the max public sandboxes and repositories yet. We are now, so I've added those flags to the `useSubscription` hook!


alexnm (approved):
💯

olarclara (approved):

PR #7147, opened 2022-11-14
feat: add trial information on subscription card

alexnm:
Compared to the design, I had to drop some information we unfortunately don't have for stripe users:
* We don't have the price unit to show how much users will be billed
* We don't have the next billing date

Active trial
Screenshot 2022-11-14 at 21 15 56

Cancelled trial
Screenshot 2022-11-14 at 21 15 46

I believe we need to start modelling more details around the trial and subscription. We have the concept of an `ACTIVE` subscription, but we also have a `canceledAt` date time as well as a `cancelAtPeriodEnd` flag. For now I would leave it as it is, until after the release, but we could introduce for better clarity the following booleans:
* **isPro** - instead of checking for active subscription or trialing or any other combination, just return true or false if pro features apply for this user in this workspace
* **hasActiveSubscription** - this should mean that the subscription is active and ongoing beyond the current period
* **hasActiveTrial** - this means that the user has an active trial (not canceled)

wdyt @olarclara @tristandubbeld ?

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Content/routes/Settings/TeamSettings/ManageSubscription/index.tsx
+++ b/packages/app/src/app/pages/Dashboard/Content/routes/Settings/TeamSettings/ManageSubscription/index.tsx
@@ -21,9 +21,11 @@ export const ManageSubscription = () => {
   const { activeTeamInfo: team } = useAppState();
   const {
     hasActiveSubscription,
+    hasActiveTeamTrial,
     numberOfSeats,
     isPaddle,
     isStripe,
+    subscription,
   } = useSubscription();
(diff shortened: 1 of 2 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7147/commits/c3de432bbea56169f01fc5a5bc8191f60c8d4cf9 is at https://pr7147.build.csb.dev/s/new.

olarclara (approved):
Looks good!

Regarding updating the modeling, I like it... Would be similar to the `allowedCapabilities` we have on the v2 client.

tristandubbeld (approved):
I like the distinction between `isPro` and `hasActiveSubscription` you're describing 👍

PR #7157, opened 2022-11-15
feat(subscriptions): import limitations

olarclara:

- Extracts import limits to a different file.
- Blocks the import form if the workspace has reached the limit for max public repos.
- Adds tracking to upgrade CTA when attempting to import a private repo.

- No UI limitation if the workspace has reached the limit for max public repos.

## What is the new behavior?

- Dims UI and removes form controls if importing is not allowed.

|Admin|Non-admin|
|:-:|:-:|
|qqmd2g-3000 preview csb app_dashboard_recent_workspace=608e42de-934e-4a76-80ec-8addda9c620a|qqmd2g-3000 preview csb app_dashboard_repositories_workspace=608e42de-934e-4a76-80ec-8addda9c620a|

1. Launch the dashboard
2. Select a free team with 3+ imported public repos
3. Click on the 'Create' modal
4. Open the 'Import from GitHub' tab

### DIFF
--- a/packages/app/src/app/components/CreateSandbox/Import/Import.tsx
+++ b/packages/app/src/app/components/CreateSandbox/Import/Import.tsx
@@ -222,12 +222,15 @@ export const Import: React.FC<ImportProps> = ({ onRepoSelect }) => {
         </MessageStripe>
       ) : null}
       <Element
-        as="form"
-        css={{
-          opacity: hasMaxPublicRepositories ? 0.4 : 1,
-          pointerEvents: hasMaxPublicRepositories ? 'none' : 'initial',
-        }}
-        onSubmit={handleFormSubmit}
+        {...(hasMaxPublicRepositories
+          ? {
+              as: 'div',
+              css: {
+                opacity: hasMaxPublicRepositories ? 0.4 : 1,
+                pointerEvents: hasMaxPublicRepositories ? 'none' : 'initial',
+              },
+            }
+          : { as: 'form', onSubmit: handleFormSubmit })}
       >
         <Stack gap={2}>
           <Input
--- a/packages/app/src/app/components/CreateSandbox/Import/Import.tsx
+++ b/packages/app/src/app/components/CreateSandbox/Import/Import.tsx
@@ -1,27 +1,17 @@
 import track from '@codesandbox/common/lib/utils/analytics';
 import { gitHubRepoPattern } from '@codesandbox/common/lib/utils/url-generator';
-
-import {
-  Button,
-  Element,
-  Input,
-  Link,
-  MessageStripe,
-  Stack,
-  Text,
-} from '@codesandbox/components';
-import { Link as RouterLink, useLocation } from 'react-router-dom';
+import { Button, Element, Input, Stack, Text } from '@codesandbox/components';
 import css from '@styled-system/css';
 import { GithubRepoAuthorization } from 'app/graphql/types';
 import { useSubscription } from 'app/hooks/useSubscription';
-import { useWorkspaceAuthorization } from 'app/hooks/useWorkspaceAuthorization';
 import { useActions, useAppState } from 'app/overmind';
 import React from 'react';
-import { useGetCheckoutURL } from 'app/hooks/useCreateCheckout';
+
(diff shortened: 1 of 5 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7157/commits/3e71a8351e19c0b647c68b9920425356c2562919 is at https://pr7157.build.csb.dev/s/new.

packages/app/src/app/hooks/useSubscription.ts
```diff
@@ -82,6 +82,7 @@ export const useSubscription = () => {
   const maxPublicProjects = activeTeamInfo?.limits?.maxPublicProjects;
 
   const hasMaxPublicRepositories =
+    isTeamSpace &&
```
alexnm, line 85: isn't the same restriction valid for personal spaces?

packages/app/src/app/constants.ts
```diff
@@ -1,3 +1,10 @@
 export const DIALOG_TRANSITION_DURATION = 0.25;
 export const DIALOG_WIDTH = 420;
 export const REPLY_TRANSITION_DELAY = 0.5;
+export const SUBSCRIPTION_DOCS_URLS = {
```
alexnm, line 4: we should move towards a `paths` like structure, similar with v2 and unify it it here with the `urls` from common (getting rid of those dependencies in the process)

Added in note in the RFC: https://www.notion.so/codesandbox-app/Client-codebases-merge-e63037d544814407be282d5a4c94dfd2

packages/app/src/app/components/CreateSandbox/Import/importLimits.tsx
```diff
@@ -0,0 +1,152 @@
...
+    team_id: isTeamAdmin ? activeTeam : undefined,
+    success_path: pathname,
+    cancel_path: pathname,
+  });
+
+  return (
+    <MessageStripe justify="space-between">
+      You&apos;ve reached the maximum amount of free repositories. Upgrade for
+      more.
+      {isTeamAdmin ? (
```
alexnm, line 45: this logic is reused quite a while, don't want to push for over abstraction but maybe there's room  for a `SubscribeToProMessageStripe` that only gets the message we want to show and some other props maybe and determines internally if it shows the CTA or the link to docs

alexnm: Not suggesting this for the current PR, rather for a follow-up refactor

alexnm (approved):
We can merge this and clarify the comments later

packages/app/src/app/components/CreateSandbox/Import/Import.tsx
```diff
@@ -179,23 +153,38 @@ export const Import: React.FC<ImportProps> = ({ onRepoSelect }) => {
...
+      >
         <Stack gap={2}>
           <Input
+            aria-disabled={hasMaxPublicRepositories}
             aria-describedby="form-title form-error"
             aria-invalid={Boolean(url.error)}
-            onChange={handleUrlInputChange}
+            css={{ height: '32px' }}
             placeholder="GitHub Repository URL"
+            tabIndex={hasMaxPublicRepositories ? -1 : 0}
```
tristandubbeld, line 175: I think if we use `aria-disabled` instead of `disabled` the goal is that the element is still tabbable, but removing `tabIndex` would nullify this? I think it's fine in this case to use `disabled`, or is the "disabled state" described by `aria-describedby` in this case?

packages/app/src/app/hooks/useSubscription.ts
olarclara: @tristandubbeld @alexnm on the pricing page, we mention unlimited public repositories for personal free accounts.

packages/app/src/app/components/CreateSandbox/Import/Import.tsx
olarclara: @tristandubbeld I didn't want to override the css that gets applied when the button is disabled, so I went for aria-disabled and tabIndex to have the semantics without the UI.

PR #7168, opened 2022-11-18
feat(subscriptions/sandboxes): soft limits

olarclara:

- Adds stripe banner to the sandboxes page if the workspace has 20+ items:

m6m2dw-3000 preview csb app_dashboard

- Adds stripe banner to the create sandbox modal if the workspace has 20+ items:
  - Also handles disabling the template cards. 

|Templates list|Search results list|
|:-:|:-:|
|m6m2dw-3000 preview csb app_s|m6m2dw-3000 preview csb app_s (1)|

- Removes 'Create new sandbox' from the container menu:

Screen Shot 2022-11-18 at 03 52 45

- I haven't disabled the `Fork sandbox` from the context menu because it creates sandboxes as drafts.

- Moving drafts to sandboxes
  - Right now it doesn't have any feedback such as a `not-allowed` cursor.
  - It's still possible to trigger the action through the 'Move to Folder' modal, although it will fail on the backend.
    - Should we remove the `Move to Folder` modal altogether? Might be the simplest way to prevent moving sandboxes around... The downside I see is that moving the sandboxes between folders will only be possible through drag-and-drop.
    - It's possible to move sandboxes back and forth from the templates section, should it stay as-is?

- Moving sandboxes to other teams:
Screen Shot 2022-11-18 at 03 59 51

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Components/Selection/ContextMenus/SandboxMenu.tsx
+++ b/packages/app/src/app/pages/Dashboard/Components/Selection/ContextMenus/SandboxMenu.tsx
@@ -21,7 +21,7 @@ export const SandboxMenu: React.FC<SandboxMenuProps> = ({
 }) => {
   const actions = useActions();
   const { user, activeTeam, activeWorkspaceAuthorization } = useAppState();
-  const { hasActiveSubscription, hasMaxPublicRepositories } = useSubscription();
+  const { hasActiveSubscription, hasMaxPublicSandboxes } = useSubscription();
   const {
     browser: { copyToClipboard },
   } = useEffects();
@@ -179,7 +179,7 @@ export const SandboxMenu: React.FC<SandboxMenuProps> = ({
               sandboxIds: [item.sandbox.id],
               preventSandboxLeaving:
                 item.sandbox.permissions.preventSandboxLeaving ||
-                (!hasActiveSubscription && hasMaxPublicRepositories),
+                (!hasActiveSubscription && hasMaxPublicSandboxes),
             });
           }}
         >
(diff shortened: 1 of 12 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7168/commits/fe486c095a7aa0cb860518d10622f6134c31877c is at https://pr7168.build.csb.dev/s/new.

packages/app/src/app/pages/Dashboard/Components/Selection/ContextMenus/SandboxMenu.tsx
```diff
@@ -178,7 +178,8 @@ export const SandboxMenu: React.FC<SandboxMenuProps> = ({
             actions.modals.moveSandboxModal.open({
               sandboxIds: [item.sandbox.id],
               preventSandboxLeaving:
-                item.sandbox.permissions.preventSandboxLeaving,
+                item.sandbox.permissions.preventSandboxLeaving ||
+                (!hasActiveSubscription && hasMaxPublicSandboxes),
```
alexnm, line 182: Should `hasMaxPublicSandboxes` also contain the logic of `hasActiveSubscription`? so it is always false for PRO accounts basically

alexnm (approved):

tristandubbeld (approved):

packages/app/src/app/pages/Dashboard/Components/Selection/ContextMenus/SandboxMenu.tsx
olarclara: @alexnm yes, I think it's redundant to look at `hasActiveSubscription` every time we check for usage items.

PR #7180, opened 2022-11-23
feat: strict hooks values

alexnm:
Replaces `useSubscription` with two separate tooks:
* `useWorkspaceSubscription` holds the values for free/pro flags, provider + subscription object
* `useWorkspaceLimits` holds all the flags and numbers related to the limitations

I went ahead and replaced all `!<something>` conditions with the explicit true check for `isPro`/`isFree`.

The hooks now follow a similar approach in v1/v2 where we treat the initial value as `undefined` so there's no risk of flickering elements on the screen as long as we use the true values to conditionally render.


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7180/commits/ee5b4fdf7a97d725ab3c4f20b50ea39cf3ee7535 is at https://pr7180.build.csb.dev/s/new.

.codesandbox/tasks.json
```diff
@@ -14,9 +27,6 @@
       "runAtStart": true,
       "preview": {
         "port": 3000
-      },
-      "restartOn": {
-        "files": ["yarn.lock", "packages/app/scripts/*"]
```
alexnm, line 19: wow, need to fix this, we have a bug with generating the task config, probably it omits the restartOn

packages/app/src/app/constants.ts
```diff
@@ -8,3 +8,9 @@ export const SUBSCRIPTION_DOCS_URLS = {
       '/docs/learn/introduction/workspace#managing-teams-and-subscriptions',
   },
 };
+
+// Soft limit of maximum amount of pro
+// editor a team can have. Above this,
+// we should prompt CTAs to enable
+// custom pricing.
+export const MAX_PRO_EDITORS = 20;
```
alexnm, line 16: moved from useSubscription.ts

packages/app/src/app/hooks/useWorkspaceAuthorization.ts
```diff
@@ -18,7 +18,7 @@ export const useWorkspaceAuthorization = () => {
    * Team states
    */
 
-  const isTeamSpace = !isPersonalSpace;
+  const isTeamSpace = activeTeam !== null && !isPersonalSpace;
```
alexnm, line 21: isTeamSpace should be true only after the team is loaded (unfortunately null means the team is not yet there in this case)

packages/app/src/app/pages/Dashboard/Components/SyncedSandbox/index.tsx
```diff
@@ -59,6 +59,7 @@ export const SyncedSandbox = ({
     onDoubleClick,
     onContextMenu,
     onBlur,
+    isViewOnly: false,
```
alexnm, line 62: random typecheck error

.codesandbox/tasks.json
```diff
@@ -31,16 +44,10 @@
     "storybook": {
       "name": "Run storybook",
       "command": "cd packages/components && yarn start:storybook"
-    }
-  },
-  "setupTasks": [
```
alexnm, line 36: setupTasks are moved at the beginning when the config is generated

.codesandbox/tasks.json
```diff
@@ -1,4 +1,17 @@
 {
+  // These tasks will run in order when initializing your CodeSandbox project.
```
tristandubbeld, line 2: Comments in `json`, does this not throw any errors? 👀

.codesandbox/tasks.json
tristandubbeld: Is this something that we need to address somewhere or is it limited to this repository?

tristandubbeld (approved):
Great work Alex, readable and explicit! Walked through the application a bit and didn't notice any weird popping banners or stripes.

.codesandbox/tasks.json
alexnm: created a ticket for the editor team, it's a pitcher problem

.codesandbox/tasks.json
alexnm: pitcher removes comments before deserializing the config file. these have been added a while ago to help users navigate the json file better

tristandubbeld: ~Shouldn't it be a `.jsonc` file then?~ [Doesn't look like it needs to be an explicit `.jsonc` file](https://code.visualstudio.com/docs/languages/json#_json-with-comments). So I guess we're good.

packages/app/src/app/pages/Dashboard/Components/Selection/ContextMenus/SandboxMenu.tsx
```diff
@@ -179,7 +181,7 @@ export const SandboxMenu: React.FC<SandboxMenuProps> = ({
               sandboxIds: [item.sandbox.id],
               preventSandboxLeaving:
                 item.sandbox.permissions.preventSandboxLeaving ||
-                (!hasActiveSubscription && hasMaxPublicSandboxes),
+                (isFree && hasMaxPublicSandboxes),
```
olarclara, line 184: Is `isFree` redundant here?

olarclara (approved):
Looks good!

But I think we can remove all the `isFree && hasSomeLimit` because these values get computed already taking `isFree` into account.

Issue #6762, opened 2022-06-01
Errors are reported when using vue3 template setup


jerrywu001:
# Bug report

## Packages affected

-   [👉🏽 ] sandpack-react

## Description of the problem

**Errors are reported when using vue3 template setup**

```jsx
<Sandpack template="vue3"  />
```

```html
<template>
  <main id="app">
    <h1>{{ helloWorld }}</h1>
  </main>
</template>
                             
<script setup>
const helloWorld = ref("Hello World");
</script>
```

![image](https://user-images.githubusercontent.com/12525415/171375123-fb2bfd5f-14fa-4ceb-8171-e2384b25ae52.png)

holazz:
same issue.

Yoduh:
`<script setup>` seems impossible to use with codesandbox.  I hope this gets more attention as its becoming a more and more popular way to write components with Vue 3

selucaciu:
same issue

xiangrikuil:
same issue

LowLifeArcade:
I wonder why this is still not working

zymbth:
Looking forward to the support of vue3 syntax sugar `<script setup>`. At present, I have to convert my code by `setup()`

zymbth:
您好，邮件已收到，谢谢！

danilowoz:
We just merged a fix that address:
- Error on deleting a `script` or `style` tag;
- Support for `<script setup>`;

Here's a sandbox: https://codesandbox.io/s/charming-turing-oe2n17?file=/src/App.vue

Could someone please confirm that it's working?

jerrywu001:
> We just merged a fix that address:
> 
> * Error on deleting a `script` or `style` tag;
> * Support for `<script setup>`;
> 
> Here's a sandbox: https://codesandbox.io/s/charming-turing-oe2n17?file=/src/App.vue
> 
> Could someone please confirm that it's working?

When will it be released?

danilowoz:
It should already be live on CodeSandbox and Sandpack@1.17.1

PR #7196, opened 2022-11-29
fix: vue transpiler

danilowoz:
Given a Vue template with `style`, `script`, and `template` tags, the Vue transpiler needs to consider that `style` might be deleted at some point and there is no way to transpile this piece of code.

Plus, I upgrade `@vue/compiler-sfc` in order to support `<script setup>` new template config.

Maybe fix https://github.com/codesandbox/codesandbox-client/issues/6762

### DIFF
--- a/packages/vue3-transpiler/src/templateLoader.ts
+++ b/packages/vue3-transpiler/src/templateLoader.ts
@@ -3,6 +3,7 @@ import { LoaderContext } from 'sandpack-core';
 import loaderUtils from 'sandpack-core/lib/transpiler/utils/loader-utils';
 import { compileTemplate, TemplateCompiler } from 'vue3-browser-compiler';
 import { WarningStructure } from 'sandpack-core/lib/transpiler/utils/worker-warning-handler';
+import { v5 as uuid } from 'uuid';
 
 import { VueLoaderOptions } from './index';
 import { formatError } from './formatError';
@@ -33,6 +34,7 @@ function TemplateLoader(source: string, loaderContext: LoaderContext) {
   }
 
   const compiled = compileTemplate({
+    id: uuid(),
     source,
     inMap,
     filename: loaderContext.path,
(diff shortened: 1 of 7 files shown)


tristandubbeld (commented):
🥳

DeMoorJasper (changes requested):
Looks good to me, thanks for looking into this, would be nice to have an integration test added here and the typings fixed

packages/vue3-transpiler/src/templateLoader.ts
```diff
@@ -33,6 +34,7 @@ function TemplateLoader(source: string, loaderContext: LoaderContext) {
   }
 
   const compiled = compileTemplate({
+    id: uuid(),
```
DeMoorJasper, line 37: What is this id for?

danilowoz: A new dependency from `@vue/compiler-sfc`

DeMoorJasper: Doesn't seem like we're using vue/compiler-sfc, this file has an import for https://unpkg.com/browse/vue3-browser-compiler@1.0.0/ so probably didn't change there as that module has never been updated?

Also still dunno what the id is for, if it is used for hmr we can't use uuid's that's what I was trying to get to, than it should be the module id, which is the filepath with some optional query

DeMoorJasper (approved):
Let's get this merged, hmr will likely break because of the id's but vue is broken now anyway, we can fix it in a follow-up

danilowoz:
@DeMoorJasper I just tested, and it didn't mess up the HMR. However, it completely reloads the preview when removing a style or script tag, which I guess is expected.

DeMoorJasper:
Would expect those changes to hot reload and not page reload, so that is probably the broken behaviour, but will look into it later today.

Already really nice that we finally got around to fixing this issue, thanks for working on this Danilo

PR #7201, opened 2022-11-30
fix(dashboard/repositories): silently fail get repos

olarclara:

If the action to get repos is transparent (ie: first visibly fetch unsynced data and then silently fetch synced data), hide error notifications originating from a timeout because that can happen if there's too much data from GitHub.

- Turn on throttling to a slow network to make it easier to test
- Launch the dashboard
- Select a team with repos that have hundreds of branches
- See that the first call with unsynced data succeeds
- See that the second call with synced data throws a 504

### DIFF
--- a/packages/app/src/app/overmind/namespaces/dashboard/actions.ts
+++ b/packages/app/src/app/overmind/namespaces/dashboard/actions.ts
@@ -1922,11 +1922,11 @@ export const getRepositoriesByTeam = async (
   const { activeTeam, dashboard } = state;
   const { bypassLoading = false } = options ?? {};
 
-  // If we should bypass the loading state, we don't need to make two
-  // queries (synced and unsynced) since the loading time won't be
-  // perceived by the user.
-  try {
-    if (!bypassLoading) {
+  // If we should show the loading state, we need to make two
+  // queries (synced and unsynced) since the waiting time to
+  // get up-to-date data from GitHub is too long.
+  if (!bypassLoading) {
+    try {
       dashboard.repositories = null;
 
       // First fetch data without syncing with GitHub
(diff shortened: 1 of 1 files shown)


packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -1960,6 +1966,11 @@ export const getRepositoriesByTeam = async (
 
     dashboard.repositories = syncedRepositories.sort(sortByNameAscending);
   } catch (error) {
+    if (!bypassLoading && error?.response?.status === 504) {
```
olarclara, line 1969: Wondering if `bypassLoading` being false is enough to completely hide errors from the second request, it's probably weird to see an error but have data on the screen. Thoughts?

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7201/commits/8b6b59fc104b72cfd9ef8c5af7c2b7ac2963bed2 is at https://pr7201.build.csb.dev/s/new.

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
tristandubbeld: If the user is looking for a certain repository that has yet to be synced but something fails, we probably still want to show an error. We might want to word the error differently though, something like: "There was a problem syncing your repositories with GitHub, data might not be up to date."

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -1960,8 +1966,13 @@ export const getRepositoriesByTeam = async (
...
+      // Fail silently since this is a secondary request.
+      return;
+    }
+
     effects.notificationToast.error(
-      'There was a problem getting your repositories'
+      'There was a problem syncing your repositories with GitHub, data show might not be up to date'
```
tristandubbeld, line 1975:
```suggestion
      'There was a problem syncing your repositories with GitHub, data shown might not be up to date'
```

tristandubbeld (approved):
Does this request always fail on workspaces with repos with a ton of branches? I didn't have to throttle for it to fail. The message showed up though!

olarclara:
@tristandubbeld it times out when there are thousands of branches, I had to remove `test-sandbox` from the imported projects in my personal workspace to get it to load properly.

PR #7204, opened 2022-12-01
fix(vue): Use the correct id for vue compiler to ensure correct hmr logic

DeMoorJasper:

Vue transpiler should use the hash of filename/query as the id, not a uuid, I assume it's because of hmr and replacing the correct element.

Haven't been able to break the current logic so not sure where this would show, but best to stay in sync with the vue plugins the maintainer of vue wrote.

Some vite code for reference: https://github.com/vitejs/vite/blob/main/packages/plugin-vue/src/template.ts#L109

Update: Tested a bunch of things, doesn't seem to be making a difference, but at least it should be faster and perhaps they don't use the id's for script yet? (also wouldn't be very semver compatible if they changed this now)

Uses uuid for hmr

## What is the new behavior?

Uses filename hash

### DIFF
--- a/packages/vue3-transpiler/src/index.ts
+++ b/packages/vue3-transpiler/src/index.ts
@@ -18,7 +18,6 @@ import { selectBlock } from './select';
 import { genHotReloadCode } from './hotReload';
 import { genCSSModulesCode } from './cssModules';
 import { formatError } from './formatError';
-import { v4 as uuid } from 'uuid';
 
 import vueTemplateLoader from './transpilers/template-loader';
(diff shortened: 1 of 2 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7204/commits/ad802f454ccda91ff9e4f495b1f955c602be03cf is at https://pr7204.build.csb.dev/s/new.

danilowoz (approved):
Let's get this merged! 🚀

PR #7209, opened 2022-12-05
fix: vercel integration

alexnm:
The vercel callback passes multiple params with the newer integration, so this is fixed by taking the code value only from the query string

### DIFF
--- a/packages/app/src/app/pages/VercelAuth/index.js
+++ b/packages/app/src/app/pages/VercelAuth/index.js
@@ -15,7 +15,9 @@ const VercelSignIn = () => {
   useEffect(() => {
     if (document.location.search.match(/\?code=(.*)/)) {
       // eslint-disable-next-line
-      const [_, code] = document.location.search.match(/\?code=(.*)/);
+      const queryParams = new URLSearchParams(document.location.search);
+      const code = queryParams.get('code');
+
       if (window.opener) {
         if (window.opener.location.origin === window.location.origin) {
           window.opener.postMessage(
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7209/commits/16299ca7bb0660df021ff1e8ed43f212737f49ce is at https://pr7209.build.csb.dev/s/new.

packages/app/src/app/pages/VercelAuth/index.js
```diff
@@ -14,8 +14,9 @@ const VercelSignIn = () => {
 
   useEffect(() => {
     if (document.location.search.match(/\?code=(.*)/)) {
```
Utwo, line 16: I don't think this will work if the query params start with something else other than `?code=`

Eg. `?auth=true&code=123`

Utwo: Maybe `includes("code=")` would be a better option.

alexnm: moved code computation above the if

packages/app/src/app/pages/VercelAuth/index.js
```diff
@@ -13,9 +13,10 @@ const VercelSignIn = () => {
   const [error, setError] = useState(null);
 
   useEffect(() => {
-    if (document.location.search.match(/\?code=(.*)/)) {
-      // eslint-disable-next-line
-      const [_, code] = document.location.search.match(/\?code=(.*)/);
+    const queryParams = new URLSearchParams(document.location.search);
+    const code = queryParams.get('code');
+
+    if (code) {
```
Utwo, line 19: We can also use early returns here instead of imbricating so many if's.

PR #7217, opened 2022-12-07
feat(dashboard/components): ExternalCard: article and video variant

olarclara:
Closes XTD-481.
Closes XTD-482.

- Creates `VideoCard`;
- Creates `ArticleCard`;

They are placed in the components section to make visual testing easier with Storybook since this component will be integrated on an upcoming task.

|Article|Video|
|:-:|:-:|
|Screen Shot 2022-12-07 at 18 23 01|Screen Shot 2022-12-07 at 18 23 19|

### DIFF
--- a/packages/components/src/components/ExternalCard/VideoCard.tsx
+++ b/packages/components/src/components/ExternalCard/VideoCard.tsx
@@ -10,6 +10,7 @@ import {
 
 type Video = {
   duration: string;
+  durationLabel: string;
   title: string;
   thumbnail: string;
   url: string;
@@ -19,6 +20,7 @@ type VideoCardProps = { onClick?: () => void } & Video;
 
 export const VideoCard: React.FC<VideoCardProps> = ({
   duration,
+  durationLabel,
   title,
   thumbnail,
   url,
@@ -34,7 +36,7 @@ export const VideoCard: React.FC<VideoCardProps> = ({
       {...props}
     >
       <StyledContent>
-        <StyledDetails>
+        <StyledDetails gap={3}>
           <Stack
             align="center"
             css={{
@@ -49,8 +51,10 @@ export const VideoCard: React.FC<VideoCardProps> = ({
           >
             <Icon name="start" />
           </Stack>
-          <StyledTitle>{title}</StyledTitle>
-          <StyledTitle clamp={false}>{duration}</StyledTitle>
+          <StyledTitle clamp>{title}</StyledTitle>
+          <StyledTitle aria-label={durationLabel} clamp={false}>
+            {duration}
+          </StyledTitle>
         </StyledDetails>
       </StyledContent>
     </StyledWrapper>
--- a/packages/components/src/components/ExternalCard/VideoCard.tsx
+++ b/packages/components/src/components/ExternalCard/VideoCard.tsx
@@ -1,0 +1,58 @@
+import React from 'react';
+import { Icon } from '../Icon';
+import { Stack } from '../Stack';
+import {
+  StyledContent,
+  StyledDetails,
+  StyledTitle,
+  StyledWrapper,
+} from './elements';
+
+type Video = {
+  duration: string;
+  title: string;
+  thumbnail: string;
+  url: string;
+};
+
+type VideoCardProps = { onClick?: () => void } & Video;
+
+export const VideoCard: React.FC<VideoCardProps> = ({
+  duration,
+  title,
+  thumbnail,
+  url,
+  ...props
+}) => {
+  return (
+    <StyledWrapper
+      as="a"
+      direction="vertical"
+      href={url}
+      target="_blank"
+      thumbnail={thumbnail}
+      {...props}
+    >
+      <StyledContent>
+        <StyledDetails>
+          <Stack
+            align="center"
+            css={{
+              flexShrink: 0,
+              height: '32px',
+              width: '32px',
+              borderRadius: '50%',
+              background: 'rgba(255, 255, 255, 0.2)',
+              backdropFilter: 'blur(2.68809px)',
+            }}
+            justify="center"
+          >
+            <Icon name="start" />
+          </Stack>
+          <StyledTitle>{title}</StyledTitle>
+          <StyledTitle clamp={false}>{duration}</StyledTitle>
+        </StyledDetails>
+      </StyledContent>
+    </StyledWrapper>
+  );
+};
--- a/packages/components/src/components/ExternalCard/VideoCard.tsx
+++ b/packages/components/src/components/ExternalCard/VideoCard.tsx
@@ -40,16 +40,15 @@ export const VideoCard: React.FC<VideoCardProps> = ({
           <Stack
             align="center"
             css={{
+              padding: '8px',
               flexShrink: 0,
-              height: '32px',
-              width: '32px',
               borderRadius: '50%',
               background: 'rgba(255, 255, 255, 0.2)',
               backdropFilter: 'blur(2.68809px)',
             }}
             justify="center"
           >
-            <Icon name="start" />
+            <Icon css={{}} name="start" />
           </Stack>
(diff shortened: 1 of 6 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7217/commits/2ba9d00f5eb3f6c643df689cc5c6e6097ecdd596 is at https://pr7217.build.csb.dev/s/new.

packages/components/src/components/ExternalCard/elements.ts
```diff
@@ -0,0 +1,83 @@
+import styled from 'styled-components';
+import { Stack } from '../Stack';
+import { Text } from '../Text';
+
+export const StyledWrapper = styled(Stack)<{ thumbnail: string }>`
+  position: relative;
+  overflow: hidden;
+  height: 156px;
+  width: 276px;
+  background: #999999;
```
tristandubbeld, line 10: Before the image loads the background is quite light, what do you think about changing it to `#161616` to be in line with the other cards? When the image fails to load this would also be a nice fallback to make sure the text is still readable.

packages/components/src/components/ExternalCard/elements.ts
```diff
@@ -0,0 +1,83 @@
...
+    background-size: cover;
+    background-position: center;
+    transform: scale(1);
+    transition: transform ease-in-out;
+    transition-duration: 100ms;
+  }
+
+  &:hover::after {
+    transform: scale(1.01);
+  }
```
tristandubbeld, line 31: Missing `focus` or `focus-visible` styles and `hover` "overlay"-color change.

packages/components/src/components/ExternalCard/elements.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export const StyledWrapper = styled(Stack)<{ thumbnail: string }>`
+  position: relative;
+  overflow: hidden;
+  height: 156px;
+  width: 276px;
+  background: #999999;
+  border-radius: 4px;
+
+  &:after {
```
tristandubbeld, line 13: Super nitpick: `::after` 😛

packages/components/src/components/ExternalCard/elements.ts
```diff
@@ -0,0 +1,83 @@
...
+    background-image: url(${props => props.thumbnail});
+    background-size: cover;
+    background-position: center;
+    transform: scale(1);
+    transition: transform ease-in-out;
+    transition-duration: 100ms;
+  }
+
+  &:hover::after {
+    transform: scale(1.01);
```
tristandubbeld, line 30: Love this addition!

packages/components/src/components/ExternalCard/elements.ts
```diff
@@ -0,0 +1,83 @@
...
+      rgba(25, 25, 25, 0) 100%
+    ),
+    linear-gradient(0deg, rgba(25, 25, 25, 0.6) 0%, rgba(25, 25, 25, 0) 100%),
+    linear-gradient(0deg, rgba(25, 25, 25, 0.8) 0%, rgba(25, 25, 25, 0) 100%);
+`;
+
+export const StyledDetails = styled(Stack)`
+  margin: auto 24px 24px;
+`;
+StyledDetails.defaultProps = {
```
tristandubbeld, line 57: It seems that [`defaultProps` will be deprecated one day](https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#deprecate-defaultprops-on-function-components) (or already is in v18, dunno) so I think we should move these to the component implementation.

packages/components/src/components/ExternalCard/elements.ts
```diff
@@ -0,0 +1,83 @@
...
+    `  
+  width: 100%;
+  overflow: hidden;
+  display: -webkit-box;
+  -webkit-line-clamp: 3;
+  -webkit-box-orient: vertical;
+  `}
+`;
+StyledTitle.defaultProps = {
+  as: 'p',
```
tristandubbeld, line 81: Because it's inside an `a` element, we should use `span`. I think this is the default for the `Text` component so we can remove this line. Also, the `margin: 0;` can be removed this way.

packages/components/src/components/ExternalCard/VideoCard.tsx
```diff
@@ -0,0 +1,58 @@
...
+      {...props}
+    >
+      <StyledContent>
+        <StyledDetails>
+          <Stack
+            align="center"
+            css={{
+              flexShrink: 0,
+              height: '32px',
+              width: '32px',
```
tristandubbeld, line 43: In this case we can use `padding` instead if we want.

packages/components/src/components/ExternalCard/elements.ts
olarclara: @tristandubbeld is there a specific reason for using the `span`? There's no semantic impeditive to it, quoting the spec:

> The [a](https://www.w3.org/TR/2011/WD-html5-author-20110809/the-a-element.html#the-a-element) element may be wrapped around entire paragraphs, lists, tables, and so forth, even entire sections, so long as there is no interactive content within (e.g. buttons or other links).

Reference: https://www.w3.org/TR/2011/WD-html5-author-20110809/the-a-element.html.

packages/components/src/components/ExternalCard/VideoCard.tsx
olarclara: @tristandubbeld would it be better to use the padding instead of defining the height?

packages/components/src/components/ExternalCard/elements.ts
olarclara: @tristandubbeld oh, I wasn't aware. Removed the default props here: https://github.com/codesandbox/codesandbox-client/pull/7217/commits/24bed18cc0845f5ca5698afbf07a6ef69cecb37e.

packages/components/src/components/ExternalCard/elements.ts
olarclara: @tristandubbeld good catch, updated here: https://github.com/codesandbox/codesandbox-client/pull/7217/files#diff-c3fb0bcbf16750175cf4ed35ed5e3b66e6b0f026e36a21933f896816fba5e04bR11. I also added a `ThumblessVariant` showcasing the UI when the image is loading/fails to load.

packages/components/src/components/ExternalCard/elements.ts
olarclara: Screen Shot 2022-12-08 at 19 17 44

packages/components/src/components/ExternalCard/elements.ts
tristandubbeld: Hmm yeah you're right. It's interactive elements that aren't allowed as children. My reasoning was that since it's just a few words `p` wouldn't make much sense. But it's not wrong indeed, thanks for pointing this out!

packages/components/src/components/ExternalCard/VideoCard.tsx
tristandubbeld: Sorry, forgot to explain why I thought this is the case. Whenever the user has a bigger personal font size preference set in the browser (and when the page properly uses `rem` values) the text might scale and be bigger than the set width and height. In this case, since we're talking about an `Icon` it doesn't matter, but when it's text it does. This is why I usually - in all cases, just to be sure and for consistency - prefer padding over a set width and height.

tristandubbeld (approved):
Good to go! Since the `height` and `width` aren't a big deal in this codebase I approved, feel free to change it to padding if you'd like :)

packages/components/src/components/ExternalCard/VideoCard.tsx
olarclara: @tristandubbeld oh, that makes sense. Thank you for explaining. 🙇🏽‍♀️

PR #6283, opened 2021-11-16
CodeSandbox CI docs: update node versions and add node 16

lbogdan:

PR #6107, opened 2021-09-07
docs: update SSE node versions and add node 16

lbogdan:

PR #7220, opened 2022-12-08
chore(node): Update to version 10.24.1

tristandubbeld:
According to #6283 and #6107 the proper version should be `10.24.1` but this was never updated in the `.nvmrc` and `package.json` files of the root and `components` package.

Ideally we upgrade this even more and deprecate node 10, but I think we need some devops work for that. In the meantime I'm keeping an eye on the status of adding node 18 to see how that should work.


tristandubbeld:
Ah, here we go - it's not as easy as it seems!

image

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7220/commits/daf2429a2e5e62249984c4566e28974bf9214888 failed.

tristandubbeld:
Closed in favour of #7466!

PR #7223, opened 2022-12-09
feat(sandboxes): update empty state

olarclara:
There's a bit of code duplication at the moment. I want to pick up another task to see what can be abstracted into a template and what's better left as is.

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Content/routes/Sandboxes/EmptyState.tsx
+++ b/packages/app/src/app/pages/Dashboard/Content/routes/Sandboxes/EmptyState.tsx
@@ -34,6 +34,13 @@ const TEMPLATE_IDS = [
   '9qputt', // React + Vite (browser)
 ];
 
+const DESCRIPTIONS = {
+  TEAM:
+    'Sandboxes are a great way to prototype your ideas with zero startup costs and with everything you need: a code editor, previews, dev servers, unit tests, Storybook and many other devtools.<br /><br />Sandboxes in this folder are visible to all team members.',
+  PERSONAL:
+    'Sandboxes are a great way to prototype your ideas with zero startup costs and with everything you need: a code editor, previews, dev servers, unit tests, Storybook and many other devtools.',
+};
+
 export const EmptyState: React.FC = () => {
   const { isPersonalSpace } = useWorkspaceAuthorization();
   const { hasMaxPublicSandboxes } = useWorkspaceLimits();
@@ -84,11 +91,10 @@ export const EmptyState: React.FC = () => {
       gap={10}
     >
       {isPersonalSpace ? (
-        <StyledEmptyDescription as="p">
-          Sandboxes are a great way to prototype your ideas with zero startup
-          costs and with everything you need: a code editor, previews, dev
-          servers, unit tests, Storybook and many other devtools.
-        </StyledEmptyDescription>
+        <StyledEmptyDescription
+          as="p"
+          dangerouslySetInnerHTML={{ __html: DESCRIPTIONS.PERSONAL }}
+        />
       ) : (
         <Stack
           css={{
@@ -100,14 +106,10 @@ export const EmptyState: React.FC = () => {
           direction="vertical"
           gap={6}
         >
-          <StyledEmptyDescription as="p">
-            Sandboxes are a great way to prototype your ideas with zero startup
-            costs and with everything you need: a code editor, previews, dev
-            servers, unit tests, Storybook and many other devtools.
-          </StyledEmptyDescription>
-          <StyledEmptyDescription as="p">
-            Sandboxes in this section are visible to all team members.
-          </StyledEmptyDescription>
+          <StyledEmptyDescription
+            as="p"
+            dangerouslySetInnerHTML={{ __html: DESCRIPTIONS.TEAM }}
+          />
         </Stack>
       )}
       <Stack direction="vertical" gap={6}>
--- a/packages/app/src/app/pages/Dashboard/Content/routes/Sandboxes/EmptyState.tsx
+++ b/packages/app/src/app/pages/Dashboard/Content/routes/Sandboxes/EmptyState.tsx
@@ -48,6 +48,14 @@ export const EmptyState: React.FC = () => {
   const actions = useActions();
   const { dashboard } = useAppState();
 
+  const filteredTemplates = React.useMemo(() => {
+    return officialTemplates.state === 'ready'
+      ? officialTemplates.templates.filter(t =>
+          TEMPLATE_IDS.includes(t.sandbox.id)
+        )
+      : undefined;
+  }, [officialTemplates]);
+
   const handleOpenTemplate = (template: TemplateFragment) => {
     const { sandbox } = template;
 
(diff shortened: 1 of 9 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7223/commits/cb8e3d0b4785ecfb2afeeeca961a6e17012d86d9 is at https://pr7223.build.csb.dev/s/new.

packages/app/src/app/pages/Dashboard/Content/routes/Sandboxes/MaxSandboxesBanner.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { useWorkspaceAuthorization } from 'app/hooks/useWorkspaceAuthorization';
+import { useWorkspaceSubscription } from 'app/hooks/useWorkspaceSubscription';
+import { useAppState } from 'app/overmind';
+import track from '@codesandbox/common/lib/utils/analytics';
+import { dashboard as dashboardUrls } from '@codesandbox/common/lib/utils/url-generator';
+import { Element, MessageStripe } from '@codesandbox/components';
+import { Link } from 'react-router-dom';
+import React from 'react';
+
+export const MaxSandboxesBanner: React.FC = () => {
```
alexnm, line 11: nitpick: would call this `RestrictionsBanner` to be compatible with any future message stripe we need to show here

packages/app/src/app/pages/Dashboard/Content/routes/Sandboxes/EmptyState.tsx
```diff
@@ -0,0 +1,177 @@
...
+            width: '100%',
+            '@media screen and (min-width: 768px)': {
+              width: '560px',
+            },
+          }}
+          direction="vertical"
+          gap={6}
+        >
+          <StyledEmptyDescription as="p">
+            Sandboxes are a great way to prototype your ideas with zero startup
```
alexnm, line 104: Can you put this text in a variable to reuse? we might accidentally update only one of them later on

packages/app/src/app/pages/Dashboard/Content/routes/Sandboxes/EmptyState.tsx
```diff
@@ -0,0 +1,177 @@
...
+            costs and with everything you need: a code editor, previews, dev
+            servers, unit tests, Storybook and many other devtools.
+          </StyledEmptyDescription>
+          <StyledEmptyDescription as="p">
+            Sandboxes in this section are visible to all team members.
```
alexnm, line 109: copy suggestion: use `folder` instead of `section` as you see this view in an empty folder as well.

packages/app/src/app/pages/Dashboard/Content/routes/Sandboxes/EmptyState.tsx
```diff
@@ -0,0 +1,177 @@
...
+                    disabled={hasMaxPublicSandboxes}
+                    template={template}
+                    onOpenTemplate={handleOpenTemplate}
+                    onSelectTemplate={handleSelectTemplate}
+                  />
+                </Stack>
+              ))}
+
+          {((officialTemplates.state === 'ready' &&
+            officialTemplates.templates.length === 0) ||
```
alexnm, line 159: shouldn't we show the option to create from template (which opens the create modal) for all scenarios? as the last card here.

packages/app/src/app/pages/Dashboard/Content/routes/Sandboxes/MaxSandboxesBanner.tsx
olarclara: @alexnm done here: [086356c](https://github.com/codesandbox/codesandbox-client/pull/7223/commits/086356c8eb43e921aabb350399fda1b122deba08)

packages/app/src/app/pages/Dashboard/Content/routes/Sandboxes/EmptyState.tsx
olarclara: @alexnm that's not in the designs but I can ask Lena.

packages/app/src/app/pages/Dashboard/Content/routes/Sandboxes/EmptyState.tsx
```diff
@@ -0,0 +1,185 @@
...
+            : null}
+          {filteredTemplates && filteredTemplates.length > 0
+            ? filteredTemplates.map(template => (
+                <Stack as="li" key={template.id}>
+                  <TemplateCard
+                    key={template.id}
+                    disabled={hasMaxPublicSandboxes}
+                    template={template}
+                    onOpenTemplate={handleOpenTemplate}
+                    onSelectTemplate={handleSelectTemplate}
```
tristandubbeld, line 161: I was really confused about what `handleSelectTemplate` would do, because it doesn't seem selectable? What I think happens here is that user / team template logic leaks into the official template buttons. These cards don't need all the functionality and, imo should be (relatively) simple links.

What I'd suggest here is that we create a different component for these (official) template links. The template card feels like a bit like Swiss knife to me this way. The modal close in the `handleSelectTemplate` function also doesn't feel like it belongs.

packages/app/src/app/pages/Dashboard/Content/routes/Sandboxes/EmptyState.tsx
tristandubbeld: Another way to do it might be reusable components instead of using `dangerouslySetInnerHTML`?

packages/app/src/app/pages/Dashboard/Content/routes/Sandboxes/EmptyState.tsx
olarclara: @tristandubbeld I guess I'd rather have the template card behave in a predictable way across the app than look the same but perform different actions in different places.

tristandubbeld: But it doesn't behave the same, right? That's what confused me because it's not selectable. I'm confused now what `TemplateCard` should be used for, only for templates to create a new sandbox from or also for template sandboxes that are editable? If it's only the former I don't think it needs an `onSelectTemplate` function because it doesn't seem to do anything.

olarclara: @tristandubbeld I think the naming is off, but selecting a template actually forks it, and opening a template (using cmd + click/cmd + open) just opens it in a new tab.

tristandubbeld: Interesting, do we actually want that to be the behaviour?

packages/app/src/app/pages/Dashboard/Content/routes/Sandboxes/EmptyState.tsx
alexnm: Yes, just a thought, you have the option to create a sandbox always from the context menu

packages/app/src/app/pages/Dashboard/Content/routes/Sandboxes/EmptyState.tsx
olarclara: @tristandubbeld it's been like this forever I think, lol. I've added tracking to see if people ever just open a sandbox.

tristandubbeld: Nice! Let's see 😄

PR #7230, opened 2022-12-13
fix(dependencies): show right version in version dropdown

CompuIves:
If you've configured a version for a dependency in `package.json` that doesn't match the version list, we randomly select a version from the version selector. Like here:

image

That's a bit confusing. With this change, we add the dependency that's in package.json to the list in case it's not in the list, so you'll get a view like this:

image

### DIFF
--- a/packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/Explorer/Dependencies/Dependency/index.tsx
+++ b/packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/Explorer/Dependencies/Dependency/index.tsx
@@ -189,7 +189,7 @@ export const Dependency = ({
           {versions.length === 0 ? null : (
             <Select
               css={{ width: '80px' }}
-              value={versionFromDropdown}
+              value={versionFromDropdown || dependencies[dependency]}
               onChange={e => onRefresh(dependency, e.target.value)}
             >
               {versionFromDropdown === undefined && (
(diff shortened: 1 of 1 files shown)


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7230/commits/d5e2180677fea992d93b82950a1c1825b06dad53 is at https://pr7230.build.csb.dev/s/new.

packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/Explorer/Dependencies/Dependency/index.tsx
```diff
@@ -185,9 +189,14 @@ export const Dependency = ({
           {versions.length === 0 ? null : (
             <Select
               css={{ width: '80px' }}
-              value={versions.find(v => v === dependencies[dependency])}
+              value={versionFromDropdown}
```
olarclara, line 192: Since the version from dropdown might be undefined, should the value be version from dropdown or the defined version that's added to the options?

CompuIves: Oh good point, that's more explicit. I'll update this!

CompuIves: Updated!

olarclara (approved):

PR #7234, opened 2022-12-14
feat(dashboard/drafts): update empty page

olarclara:

- Extracts variable grid constants to a separate file, this was needed to avoid an error where the templates row would try to get the values before their initialization.
- Updates the `TemplatesRow` to use the styled empty page elements.
- Updates the drafts page to match the new UI. 

codesandbox stream_dashboard_drafts_workspace=608e42de-934e-4a76-80ec-8addda9c620a

## What is the new behavior?

s2hrv7-3000 preview csb app_dashboard_drafts_workspace=608e42de-934e-4a76-80ec-8addda9c620a (1)


packages/app/src/app/pages/Dashboard/Components/TemplatesRow/TemplatesRow.tsx
```diff
@@ -43,110 +77,61 @@ export const TemplatesRow: React.FC = () => {
...
-              .map(template => (
+    <Stack direction="vertical" gap={6}>
+      {title !== undefined && title !== '' ? (
+        <EmptyPage.StyledGridTitle as="h2">
+          Start from a template
+        </EmptyPage.StyledGridTitle>
+      ) : null}
+      <EmptyPage.StyledGrid as="ul">
+        {officialTemplates.state === 'loading'
+          ? new Array(templateIds.length).fill(undefined).map((_, i) => (
```
tristandubbeld, line 91: Now that we're using `templateIds.length` we can just map over `templateIds`, right? Do we still need this loading state and will we still need to fetch `officialTemplates` or are we just using hardcoded templates?

packages/app/src/app/pages/Dashboard/Content/routes/Drafts/EmptyState.tsx
```diff
@@ -0,0 +1,38 @@
...
+const TEMPLATE_IDS = [
+  'k8dsq1', // NodeJS (cloud)
+  'vanilla', // Vanilla JS (browser)
+  'uo1h0', // NextJS (cloud)
+  '9qputt', // React + Vite (browser)
+];
+
+const DESCRIPTIONS = {
+  TEAM:
+    'By default, every sandbox you create will show up on this folder.<br />Sandboxes in My Drafts are not visible to your team members unless moved to the "All sandboxes" section.',
```
alexnm, line 15: Reading this makes me wonder if `All sandboxes` is the right term, since it clearly doesn't keep ALL sandboxes :) cc @lenasmayor

packages/app/src/app/pages/Dashboard/Content/routes/Drafts/EmptyState.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export const EmptyState: React.FC = () => {
+  const { isPersonalSpace } = useWorkspaceAuthorization();
+
+  return (
+    <EmptyPage.StyledWrapper>
+      <EmptyPage.StyledDescription
+        as="p"
+        dangerouslySetInnerHTML={{
+          __html: isPersonalSpace ? DESCRIPTIONS.PERSONAL : DESCRIPTIONS.TEAM,
```
alexnm, line 28: I guess same comment as in the other PR, would this be solved with two paragraph tags instead of one?

packages/app/src/app/pages/Dashboard/Components/TemplatesRow/TemplatesRow.tsx
olarclara: @tristandubbeld are you suggesting we hardcode everything related to the templates rather than using the selected ids to find the full info after fetching the official templates?

tristandubbeld: Hmm yeah didn't think about it that way, wasn't aware most information was fetched - doing it like this makes most sense! We do still need the loading state then, but we can map over the id's instead of filling an array with `undefined`.

olarclara: @tristandubbeld makes sense! I'll update it :)

olarclara: [2b48d29](https://github.com/codesandbox/codesandbox-client/pull/7234/commits/2b48d29b09b7ab80b6b35436002d3027e21bfb55)

olarclara:
Merging this to unblock the recent page.

PR #7243, opened 2022-12-15
feat(dashboard/recent): hide upgrade banner on first visit

olarclara:

Creates a hook to check if the dashboard has been visited before.
If the dashboard has not been visited, hide the upgrade banner on the "Recent" page.

The upgrade banner is shown for first-time visitors.

## What is the new behavior?

The upgrade banner is only visible after the 2nd visit to the dashboard.

- Launch the app
- Go to the dashboard (`/dashboard/recent`)
- The upgrade banner should not be visible
- Refresh the page
- The upgrade banner should be visible


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7243/commits/8fb40934a19a0bf52c0c24be863a39d83bf5ec22 is at https://pr7243.build.csb.dev/s/new.

packages/app/src/app/hooks/useIsFirstVisit.ts
```diff
@@ -0,0 +1,21 @@
...
+  const { browser } = useEffects();
+
+  const isFirstVisit = browser.storage.get<boolean>(KEY) ?? true;
+
+  useEffect(() => {
+    const setHasVisited = () => {
+      browser.storage.set(KEY, false);
+    };
+
+    window.addEventListener('beforeunload', setHasVisited);
```
tristandubbeld, line 16: I'm not sure if the `beforeunload` listener is the best option here. I've been reading about it a bit and both [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) and [Chrome](https://developer.chrome.com/blog/page-lifecycle-api/) don't recommend it for cases like these. The code now only triggers when the hook is rendered on a page, so when moving from `/recent` to `/discover` and then closing the website does not trigger `setHasVisited`.

I'm not sure what we define as a "first visit", is it actually closing CodeSandbox after the first time, or is it also moving to a different page than recent? This does make implementation harder. The closest thing tracking the first visit - no matter which page on the dashboard has been visited - would be to actually check the second visit.

tristandubbeld (commented):
Not sure if this will work how we want it, but an option would be the following:

On the first visit, as close to the root component as possible to make sure it's only rendered once, we can set a storage item with `DASHBOARD_FIRST_VISIT=1` (or something like that) when the item doesn't exist yet. Navigating pages doesn't change this because it shouldn't re-render.

On the second visit (when the dashboard is rendered for the 2nd time, and the `DASHBOARD_FIRST_VISIT` item exists) we can check for `DASHBOARD_FIRST_VISIT=1` and set it to `DASHBOARD_FIRST_VISIT=2`. We can then set some kind of `hasVisited` flag. I also noticed that this does already exist, but is set immediately after visiting for the first time?

Not sure if my comments are clear and it might be too much work for a "small" feature like this, but I'm wondering if it's worth it hiding the banner on the first visit.

olarclara:
Closing in favor of #7308.

PR #7245, opened 2022-12-15
feat(recent): "invite team members" action

olarclara:

Opens the "create team" modal in the "invite members" step from the empty recent page.

### Default flow
1. Sidebar workspace select -> "Create a new team"
2. Go through the team creation flow as expected

### Invite members to the current team flow
1. Empty recent page -> "Invite team members"
2. Invite new members
3. Modal gets dismissed after inviting users
4. There's no "skip" button


lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7245/commits/b11f3866a347e278b9fc129ce298fbf5f5dd9c8a is at https://pr7245.build.csb.dev/s/new.

packages/app/src/app/pages/Dashboard/Components/NewTeamModal/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+const NEXT_STEP: Record<TeamStep, TeamStep | null> = {
+  info: 'members',
+  members: 'subscription',
+  subscription: null,
+};
+
+type NewTeamProps = {
+  step?: TeamStep;
+  hasNextStep?: boolean;
```
tristandubbeld, line 30: PP: since the functionality isn't necessarily related to hiding the next step but rather being invite only `hasNextStep` can be renamed to something like `isInviteOnly`. We'd have to switch around the bools, but I think it covers the functionality that way.

alexnm (approved):

packages/app/src/app/pages/Dashboard/Components/NewTeamModal/index.tsx
olarclara: @tristandubbeld I wanted to keep this generic because the inner state gets the next step based on a map of states, by introducing this specific flag (`isInviteOnly`), then we would have to introduce specific checks and increase maintainability.

tristandubbeld: Hmm yeah you're right, this makes sense now that we're passing `step: 'members'` when opening the modal. What threw me off was the `typeof hasNextStep === 'undefined' || hasNextStep === true` to determine the next step in combination with `hideSkip` prop for `TeamMembers`. Seemed like a lot was happening.

PR #7247, opened 2022-12-19
feat: improve repository fetching performance

alexnm:
Started with some performance improvements but ended up with a number of changes, which I thought are for the best with the improved query speed.

There are now two separate page: Repositories and RepositoryBranches, to avoid conditional rendering inside the same page/component. Also, there are now two types in GraphQL, Project and ProjectWithBranches, the first one fetching only the branchCount field, not the entire list of branches. Each of the two pages relies on a different caching mechanism in overmind, one holds the project list by team id (similar with what Maria first implemented) the other holds each project with branches indexed by team/owner/name.

Modified the loading sequence for the repositories page:
- when no repositories exist for the current team, fetch the server cached version of the repos (fast) and then trigger a second fetch with syncData (slow)
- when repositories have been previously fetched for this team, show them first and fetch the synced data in the background (slow)

Also modified the flows for removing branches/repositories
- redirect to main page when removing a repository from within the repository details page
- clean all the references of projects/branches from overmind when removing them (after the success of the operation)
- refetch only recent page data to get new entries in the recent page


packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -1928,91 +1931,69 @@ export const getContributionBranches = async ({ state, effects }: Context) => {
 
 type RepositoriesActionOptions = {
   teamId: string;
-  bypassLoading?: boolean;
+  syncData?: boolean;
```
alexnm, line 1934: decided to pass directly the `syncData` flag as a param and call this function twice if needed, instead of duplicating the logic inside it

packages/app/src/app/overmind/namespaces/dashboard/utils.ts
```diff
@@ -22,7 +22,19 @@ export function sortByNameAscending(
   a: ProjectFragment,
   b: ProjectFragment
 ): number {
-  const repoA = a.repository.owner + '/' + a.repository.name;
-  const repoB = b.repository.owner + '/' + b.repository.name;
-  return repoA < repoB ? -1 : 1;
+  return a.repository.name.toLowerCase() < b.repository.name.toLowerCase()
```
alexnm, line 25: repositories are sorted just by repo name now, as it is more easy to spot one in a longer list with the new design

lbogdan:
Build for latest commit https://github.com/codesandbox/codesandbox-client/pull/7247/commits/784ef7b8cf6087a1eadf62a2eef882c72305700f is at https://pr7247.build.csb.dev/s/new.

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
tristandubbeld: Like the name change!

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -1928,91 +1931,69 @@ export const getContributionBranches = async ({ state, effects }: Context) => {
...
-      [teamId]: syncedRepositories.sort(sortByNameAscending),
+    dashboard.repositoriesByTeamId = {
+      ...dashboard.repositoriesByTeamId,
+      [teamId]: repositories.sort(sortByNameAscending),
     };
+
+    // If syncData was false, the initial call was made (faster), so we can
+    // safely trigger the second call to sync with GitHub (slower)
+    if (!syncData) {
+      actions.dashboard.getRepositoriesByTeam({ teamId, syncData: true });
```
tristandubbeld, line 1962: Hmm I get this code, but it feels strange that when `syncData` is `false`, we actually do sync data (but later). Sounds more like "lazy revalidation" instead of syncing? Can we clear up this concept by changing variable names in a way?

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -2097,34 +2081,35 @@ export const removeBranchFromRepository = async (
...
-        repository.branches = repository.branches.filter(b => b.id !== id);
-      }
-
-      actions.dashboard.getRepositoriesByTeam({
-        teamId: activeTeam,
-        bypassLoading: true,
-      });
+    // Clean the branch references from all possible locations in the state
+    // 1. Repository with branches
+    // 2. Contributiob branches
```
tristandubbeld, line 2086:
```suggestion
    // 2. Contribution branches
```

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -2097,34 +2081,35 @@ export const removeBranchFromRepository = async (
...
-        dashboard.contributions?.filter(b => b.id !== id) ?? [];
+    dashboard.contributions =
+      dashboard.contributions?.filter(b => b.id !== id) ?? [];
 
-      actions.dashboard.getContributionBranches();
-    }
+    dashboard.sandboxes.RECENT_BRANCHES =
+      dashboard.sandboxes.RECENT_BRANCHES?.filter(b => b.id !== id) ?? [];
 
     if (page === 'recent') {
```
tristandubbeld, line 2111: This means that the branch is visible on the recent page?

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -2158,35 +2143,42 @@ export const removeRepositoryFromTeam = async (
...
 
-      // Then sync in the background.
-      actions.dashboard.getRepositoriesByTeam({
-        teamId: activeTeam,
-        bypassLoading: true,
-      });
+    const key = getProjectUniqueKey({ teamId, owner, name });
+    const repositoryWithBranches = dashboard.repositoriesWithBranches[key];
+    if (repositoryWithBranches) {
+      delete dashboard.repositoriesWithBranches[key];
```
tristandubbeld, line 2164: Does / should this `delete` re-trigger a state update equal to creating a new object?

packages/app/src/app/overmind/namespaces/dashboard/utils.ts
```diff
@@ -22,7 +22,19 @@ export function sortByNameAscending(
...
+    ? -1
+    : 1;
+}
+
+export function getProjectUniqueKey({
```
tristandubbeld, line 30: What is the reason we're calling this `project` instead of `repository`, to make a distinction between GitHub and CSB repositories? I think BEOPS does this too, right? Should we define this somewhere e.g. in the glossary?

packages/app/src/app/pages/Dashboard/Content/routes/Repositories/index.tsx
```diff
@@ -1,123 +1,66 @@
...
   } = useWorkspaceLimits();
 
   const pageType: PageTypes = 'repositories';
-  let selectedRepo:
-    | { owner: string; name: string; private: boolean }
-    | undefined;
 
   const getItemsToShow = (): DashboardGridItem[] => {
-    if (teamRepos === null) {
+    if (teamRepos === undefined) {
```
tristandubbeld, line 44: 👍

packages/app/src/app/pages/Dashboard/Content/routes/RepositoryBranches/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+    // to directly access a repository, we should fetch said
+    // repository only.
+    if (!repositoryProject && activeTeam) {
+      actions.dashboard.getRepositoryByWithBranches({
+        activeTeam,
+        owner,
+        name,
+      });
+    }
+  }, [path, activeTeam]);
```
tristandubbeld, line 38: Do we need to specify `path` here? Since it's a page, it should re-render every time the page is changed, calling this effect anyway? Also, when changing a team we direct the user to the `/recent` page right? If all these are the case, we can move this out of the effect completely I reckon!

tristandubbeld (commented):
This one has gotten big, but it looks great!! Few questions and comments here and there, but overall super happy with how the code looks, nice one.

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
alexnm: I wondered if the two successive calls should be orchestrated from the UI instead of here, but unfortunately the UI action call is fire and continue, so I can't really do sequential work there. Maybe I could rename the param to `fetchCachedDataFirst`?

packages/app/src/app/overmind/namespaces/dashboard/utils.ts
alexnm: Yes, and to avoid `repository.repository` when accessing the data. I can add this to the glossary

packages/app/src/app/pages/Dashboard/Content/routes/RepositoryBranches/index.tsx
alexnm: Right, this might be leftover from the original page I didn't consider

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
alexnm: As far as I tested it does work, now that you ask, I don't think this triggers a re-render specifically, but something else does? I will re-test it

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
alexnm: don't know if you're talking about the conditional, but it means the user is on the recent page right now, so a fetch is needed

tristandubbeld: Sorry yes, I should remember that in the actual comment it doesn't highlight the selected line. And yeah I see now, it threw me off because `page` was deconstructed from `branch`.

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
tristandubbeld: Yeah something like that sounds good! I was thinking about `revalidateCache` or something like that, however, looking at the code again it does make a lot more sense to call it `fetchCachedDataFirst`. The function is a bit confusing at times because it calls itself. Don't have to change it, I might take another look at this function later to see what exactly is confusing me.

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -1928,91 +1931,69 @@ export const getContributionBranches = async ({ state, effects }: Context) => {
...
+    }
   }
 };
 
-// If the repository page is accessed directly, we can use this
-// to avoid fetching all repos.
-export const getRepositoryByDetails = async (
+// Fetch the repository with all its branches
+// Cached by `team/owner/name` key
+export const getRepositoryByWithBranches = async (
```
olarclara, line 1984:
```suggestion
export const getRepositoryWithBranches = async (
```

packages/app/src/app/pages/Dashboard/Content/routes/Repositories/index.tsx
```diff
@@ -1,123 +1,66 @@
...
-      teamRepos.length === 1
-    ) {
       actions.dashboard.getRepositoriesByTeam({ teamId: activeTeam });
+    } else {
+      actions.dashboard.getRepositoriesByTeam({
+        teamId: activeTeam,
+        syncData: true,
+      });
     }
+  }, [activeTeam]);
```
olarclara, line 34: Could this be simplified into:

```
React.useEffect(() => {
      actions.dashboard.getRepositoriesByTeam({
        teamId: activeTeam,
        syncData: !teamRepos,
      });
}, [activeTeam])
```

packages/app/src/app/pages/Dashboard/Content/routes/Repositories/index.tsx
```diff
@@ -126,26 +69,16 @@ export const RepositoriesPage = () => {
...
 
-      {isReadOnlyRepo && (
-        <Element paddingX={4} paddingY={2}>
-          {selectedRepo?.private && <PrivateRepoFreeTeam />}
-        </Element>
-      )}
-
-      {!selectedRepo &&
-      (hasMaxPublicRepositories || hasMaxPrivateRepositories) ? (
+      {hasMaxPublicRepositories || hasMaxPrivateRepositories ? (
```
olarclara, line 81: Maybe `hasMaxPublicRepositories || hasMaxPrivateRepositories` should just be `hasMaxRepos` since the check is done twice in this file.

packages/app/src/app/overmind/namespaces/dashboard/state.ts
```diff
@@ -68,13 +69,18 @@ export type State = {
...
    * overrides the current team data.
    * @see {@link https://linear.app/codesandbox/issue/XTD-375}
    */
-  repositories: Record<string, Repository[] | null> | null;
+  repositoriesByTeamId: Record<string, Repository[] | undefined>;
+
+  /**
+   * Repository with branches cached based on `team/owner/repo`
+   */
+  repositoriesWithBranches: Record<string, RepositoryWithBranches | null>;
```
olarclara, line 83: Why change the type from `undefined` to `null`?

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -2158,35 +2143,42 @@ export const removeRepositoryFromTeam = async (
...
+
+        return branchRepo.owner === owner && branchRepo.name === name;
+      }) ?? [];
+
+    // Refetch start page data in the background to fill the remaining slots
+    if (page === 'recent') {
+      actions.dashboard.getStartPageSandboxes();
+    }
+
+    // Redirect to main dashboard page when removin the repo from within the page
```
olarclara, line 2179:
```suggestion
    // Redirect to main dashboard page when removing the repo from within the page
```

packages/app/src/app/overmind/namespaces/dashboard/state.ts
olarclara: I guess it's because of the filter when deleting a repo but that could be changed to a `null` if it returns `undefined`.

alexnm: Looks like a mistake, changed  to `undefined`

packages/app/src/app/pages/Dashboard/Content/routes/Repositories/index.tsx
alexnm: I saw this but I wouldn't introduce the change in this PR, as it's already grown a bit out of scope

tristandubbeld (approved):
🚀 good speedy stuff!

PR #7255, opened 2022-12-20
feat(tracking): updates events + utm tracking

olarclara:

- Adds events to match this table on Notion: https://www.notion.so/codesandbox-app/Dashboard-Onboarding-888d5e7f16684175abc1c67995e5f07d#adc9614b198b41f8bc5e56b342a99bd5
- Adds utm source parameter to URLs.

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Content/routes/Recent/DocumentationRow.tsx
+++ b/packages/app/src/app/pages/Dashboard/Content/routes/Recent/DocumentationRow.tsx
@@ -1,3 +1,4 @@
+import track from '@codesandbox/common/lib/utils/analytics';
 import { ArticleCard, VideoCard, Stack } from '@codesandbox/components';
 import { EmptyPage } from 'app/pages/Dashboard/Components/EmptyPage';
 import React from 'react';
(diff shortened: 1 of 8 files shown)


packages/app/src/app/pages/Dashboard/Content/routes/Recent/DocumentationRow.tsx
```diff
@@ -7,6 +8,13 @@ type VideoProps = React.ComponentProps<typeof VideoCard>;
 
 type DocsItem = { label: string } & (VideoProps | ArticleProps);
 
+export const appendOnboardingTracking = (url: string): string => {
```
tristandubbeld, line 11: This is almost like the "add stripe search params" function I created. Whenever both are merge we could look into creating a general util for it maybe :)

tristandubbeld (approved):

packages/app/src/app/pages/Dashboard/Content/routes/Recent/DocumentationRow.tsx
olarclara: @tristandubbeld do you think we should have a generic `appendParams`? It can take a single object, a list of objects, etc?

PR #7262, opened 2022-12-21
feat(invite): conditional authorization

olarclara:
Instead of removing the call to action, I updated the internals of the invitation to make the role of invitees based on the current user's role.

- If the user is an admin, the role of invitees is "write".
- If the user's role is other than admin, the role of invitees is "read".


alexnm (approved):

alexnm:
We should mention this in the modal perhaps? On the other hand, not sure how common it will be the an editor to see this card.

tristandubbeld (approved):

olarclara:
@alexnm since the goal is to keep this first row even in the non-empty state, I say users will see (at least the CTA) all the time. I'll create a linear ticket for this.

olarclara:
Created a follow-up task to add copy and cc'd @necoline and @lenasmayor for a decision. https://linear.app/codesandbox/issue/XTD-551/team-invitations-add-note-explaining-the-roles-of-the-invitees

PR #7265, opened 2022-12-22
sandboxes/empty: replace templates w/ single card

olarclara:
Replaced the empty state on the sandboxes with a single card "New Sandbox". Due to its simplicity, I removed the external component and placed the UI inline.

znclpp-3000 preview csb app_dashboard_sandboxes__workspace=6db97960-5b19-448d-8185-ef76bf1d1b6f


alexnm (approved):

alexnm:
I like this, simple and efficient

PR #7266, opened 2022-12-22
pages(deleted): update loading state

olarclara:
- Imports the loading component from codesandbox-applications.
- Updates the logic to manage the internal state of the deleted page.
- Renders the loading instead of the variable grid according to the current state.

https://user-images.githubusercontent.com/24959348/209046784-3332ee21-82aa-409b-8ea3-7b320042179b.mov

https://user-images.githubusercontent.com/24959348/209046787-68328327-f97d-49fd-b699-4ecbf7098273.mov


alexnm (approved):
💯

PR #7273, opened 2022-12-23
feat: new branch card styles

alexnm:
Branch cards redesign
* main branch card (without lastCommit info which is not yet fetched from BE)
* update skeleton view for branch loading
* update contribution branches
* create branch card
* import repository card

PR takes away the common styles from the `Card` component and creates a `StyledCard` to be reused in all dashboard pages.

The last two are created using the new `ActionCard` which is simple icon + text component.


packages/components/src/components/InteractiveOverlay/InteractiveOverlay.tsx
```diff
@@ -7,11 +7,13 @@ const StyledItem = styled.a`
   border: none;
   padding: 0;
   background: transparent;
+  display: flex;
```
alexnm, line 10: lazy solution, gets rid of the extra line height when the anchor tag is default (inline)

packages/app/src/app/pages/Dashboard/Components/Branch/BranchCard.tsx
```diff
@@ -18,117 +17,86 @@ export const BranchCard: React.FC<BranchProps> = ({
...
+
+              <Stack gap={2}>
+                {contribution ? (
+                  <Icon color="#EDFFA5" name="contribution" size={16} />
+                ) : (
+                  <Icon color="#999999" name="branch" size={16} />
+                )}
+                <InteractiveOverlay.Item
+                  as="a"
+                  aria-label={ariaLabel}
```
tristandubbeld, line 59: I wonder if we should use a visually hidden label and add `aria-hidden` around the `branchName` instead of using `aria-label`. The main reason to do this would be [a11y improvement for translation](https://gomakethings.com/revisting-aria-label-versus-a-visually-hidden-class/). This would mean that we also need to [wrap the branch and repo names in elements with `translate="no"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/translate) though 😄.

packages/app/src/app/pages/Dashboard/Components/VariableGrid/index.tsx
```diff
@@ -150,20 +150,33 @@ const ComponentForTypes: IComponentForTypes = {
...
+      icon="plus"
       disabled={item.disabled}
-    />
+    >
+      Create branch
+    </ActionCard>
+  ),
+  'import-repository': ({ item }) => (
+    <ActionCard
+      as="button"
```
tristandubbeld, line 164: We probably need to tackle this when using `asChild`.

packages/components/src/components/InteractiveOverlay/InteractiveOverlay.tsx
```diff
@@ -47,12 +50,12 @@ type ItemElementProps =
       href?: never;
     });
 
-type ItemProps = {
+export type InteractiveOverlayItemProps = {
```
tristandubbeld, line 53: How do you feel about instead of exporting this, using
```
React.ComponentProps<typeof InteractiveOverlay.Item>
````

tristandubbeld (approved):

packages/app/src/app/pages/Dashboard/Components/Branch/BranchCard.tsx
alexnm: I don't know if it's worth doing this right now, but it is a great idea for adapting this to the new codebase. Maybe we can add this to our guidelines? https://www.notion.so/codesandbox-app/JS-and-TS-Coding-Guidelines-6d9af67651c740a8ab97746ece59a484 we can create a separate section on accessibility

alexnm: That guidelines doc is long looking for some love, hope to get some time for it in the first part of the year

packages/app/src/app/pages/Dashboard/Components/VariableGrid/index.tsx
alexnm: will leave it as is until we fix this in the main branch of the feature

PR #7282, opened 2022-12-29
feat(InteractiveOverlay): update Item and add Button and Link children

tristandubbeld:
Check out the components in Storybook. Also changed implementations of the `InteractiveOverlay` in the different cards, so make sure to check those out too.

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Components/shared/StyledCard.tsx
+++ b/packages/app/src/app/pages/Dashboard/Components/shared/StyledCard.tsx
@@ -12,9 +12,6 @@ export const StyledCard = styled(Card)<StyledCardProps>(({ dimmed }) => ({
   '&:hover': {
     backgroundColor: '#252525',
   },
-  '&:has(button:hover)': {
-    backgroundColor: '#1D1D1D',
-  },
   color: '#999999',
   height: '100%',
   display: 'flex',
--- a/packages/components/src/components/InteractiveOverlay/InteractiveOverlay.tsx
+++ b/packages/components/src/components/InteractiveOverlay/InteractiveOverlay.tsx
@@ -1,21 +1,7 @@
 import React, { AnchorHTMLAttributes, ButtonHTMLAttributes } from 'react';
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
 
-const StyledItem = styled.a`
-  // Reset button styles
-  // The actual styling of the element should happen in the children
-  border: none;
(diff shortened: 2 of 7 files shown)


packages/app/src/app/pages/Dashboard/Components/shared/StyledCard.tsx
```diff
@@ -12,9 +12,6 @@ export const StyledCard = styled(Card)<StyledCardProps>(({ dimmed }) => ({
   '&:hover': {
     backgroundColor: '#252525',
   },
-  '&:has(button:hover)': {
-    backgroundColor: '#1D1D1D',
-  },
```
tristandubbeld, line 17: This was messing with the `Card` hover when it only contained one button. If we want to "reset" the card background when a (icon) button is hovered, we need to rethink this.

packages/components/src/components/InteractiveOverlay/InteractiveOverlay.tsx
```diff
@@ -40,32 +26,76 @@ const StyledItem = styled.a`
...
+      if (React.isValidElement(child)) {
+        // TODO: fix ts-ignore "Property 'className' does not exist on type 'unknown'."
+        // @ts-ignore
+        return React.cloneElement(child, { className: styledClassName });
+      }
+
+      return child;
+    });
+
+    if (augmentedChildren.length) {
```
alexnm, line 88: nitpick: can you add > 0 for the check here? for sanity :D

packages/components/src/components/InteractiveOverlay/InteractiveOverlay.tsx
```diff
@@ -88,7 +118,9 @@ const Wrapper = styled(
     className: styledClassName,
   }: WrapperProps) => {
     if (React.Children.count(children) > 1) {
-      throw new Error('The Wrapper component can only contain one child.');
+      throw new Error(
+        'The StyledOverlay component can only contain one child.'
```
alexnm, line 122: What is StyledOverlay from the perspective of the user? (design system consumer). Should we reference the InteractiveOverlay component instead?

alexnm (approved):
left a few notes, but looks good

packages/components/src/components/InteractiveOverlay/InteractiveOverlay.tsx
tristandubbeld: Sharp, will update this!

PR #7284, opened 2022-12-29
feat: synced sandbox card

alexnm:
Rather incomplete, because the synced sandboxes are a bit weird, I would love to check with Necoline and/or Ives on this, but for now this is covering:
* renaming a bunch of old references to `Repo` in favor of `SyncedSandbox`
* adjusting/cleanining up the synced sandboxes cards (turning them into links from buttons)
* using the new card primitives


tristandubbeld (approved):
Code looks good, interaction looks good too! Glad the double click is gone and the selection is no more (because it didn't have a use before, I double checked). The renaming makes things a lot more clear!

PR #7286, opened 2022-12-29
chore: use Text truncate for existing cards

alexnm:
Adding truncate on existing card implementations. Still need to rely on some overflows placed around certain elements, which I don't particularly dig, but I also don't see another solution in a fluid card sizing

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Components/SyncedSandbox/SyncedSandboxCard.tsx
+++ b/packages/app/src/app/pages/Dashboard/Components/SyncedSandbox/SyncedSandboxCard.tsx
@@ -19,15 +19,7 @@ export const SyncedSandboxCard = ({ name, path, url, ...props }) => {
           css={{ height: '100%' }}
         >
           <Stack direction="vertical" gap={1}>
-            <Text
-              css={{
-                color: '#808080',
-                whiteSpace: 'nowrap',
-                overflow: 'hidden',
-                textOverflow: 'ellipsis',
-              }}
-              size={12}
-            >
+            <Text color="#999" size={12} truncate>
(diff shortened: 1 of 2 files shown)


packages/app/src/app/pages/Dashboard/Components/SyncedSandbox/SyncedSandboxCard.tsx
```diff
@@ -36,19 +28,12 @@ export const SyncedSandboxCard = ({ name, path, url, ...props }) => {
                 <Link
                   to={url}
                   css={{
+                    display: 'flex',
+                    overflow: 'hidden',
                     textDecoration: 'none',
                   }}
                 >
```
tristandubbeld, line 35: We might want a wrapper around the `react-router` `Link` component to ensure we know exactly what we're rendering. It feels weird to have `css` added straight to the `Link`. I know I started, with the `textDecoration` haha, but we might want to tackle this in the near future.

tristandubbeld (approved):

packages/app/src/app/pages/Dashboard/Components/SyncedSandbox/SyncedSandboxCard.tsx
alexnm: Right, but in this case `textDecoration` feels like a global reset (maybe it should be considered in the future) while the others are related to the use of the Link as an element wrapping something else.

alexnm: let's consider the Link in prism first, combined with the new typography components

PR #7296, opened 2023-01-09
fix(sandpack-core): correctly handle overwrite in emitModule - react-hmr issue

DeMoorJasper:
…dule

emitModule was incorrectly overwriting the module when overwrite was false, probably just a typo that we never really fixed

when overwrite is false, the module is being overwritten

## What is the new behavior?

module is only overwritten when overwrite is true (currently never)

This also improves performance quite a bit, as we no longer update the refresh module for every single module

Run the react-hmr test, failed before, works now

Before:

https://user-images.githubusercontent.com/2175521/211288145-14ba7245-de65-4d40-a601-37017c2cb8d4.mov

After: 

https://user-images.githubusercontent.com/2175521/211288229-ce53c33b-e5a7-41a9-b262-f52ae3394046.mov

### DIFF
--- a/packages/sandpack-core/src/transpiled-module/transpiled-module.ts
+++ b/packages/sandpack-core/src/transpiled-module/transpiled-module.ts
@@ -451,7 +451,7 @@ export class TranspiledModule {
         };
 
         let transpiledModule: TranspiledModule | undefined;
-        if (!overwrite) {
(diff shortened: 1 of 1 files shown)


CompuIves (approved):
Makes sense, indeed looks like a typo!

PR #7297, opened 2023-01-09
feat(sidebar): Update bottom messages related to payment method

tristandubbeld:
Adds message and CTA to add payment details when trialing and no payment details are attached to the subscription. If the link doesn't redirect to the payment method page this will be handled in another PR. A Linear task already exists to check and implement this if necessary.

### Testing

This is only testable when actively changing certain values in the editor, or when you have a trialing team that has no payment method attached and an active trial with 2 days left.

When the development environment is connected to stream you should be able to start a trial and only update the `DAYS_LEFT_TO_SHOW_PAYMENT_MESSAGE` constant to make sure it shows up.

### Screenshot
Screenshot 2023-01-09 at 14 39 04


alexnm (approved):

tristandubbeld:
Drafted due to an error when adding `paymentMethodAttached` property to the `teamFragmentDashboard`.

tristandubbeld:
Bug was fixed which makes `paymentMethodAttached` work for CSB employees again 😬!

PR #7308, opened 2023-01-12
feat(dashboard/recent): hide upgrade banner on first visit

olarclara:
Closes XTD-517.

Instead of using cookies, I decided to use the local storage following @tristandubbeld's suggestion here: https://github.com/codesandbox/codesandbox-client/pull/7243#discussion_r1050630876.

https://www.loom.com/share/e4b6610b135b4b7ba569568b6791fec2

### DIFF
--- a/packages/app/src/app/pages/Dashboard/index.tsx
+++ b/packages/app/src/app/pages/Dashboard/index.tsx
@@ -35,7 +35,7 @@ export const Dashboard: FunctionComponent = () => {
   const { browser, notificationToast } = useEffects();
   const actions = useActions();
   const { subscription } = useWorkspaceSubscription();
-  const [, trackVisit] = useDashboardVisit();
+  const { trackVisit } = useDashboardVisit();
 
   // only used for mobile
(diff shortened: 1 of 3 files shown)


packages/app/src/app/pages/Dashboard/index.tsx
```diff
@@ -34,6 +35,7 @@ export const Dashboard: FunctionComponent = () => {
   const { browser, notificationToast } = useEffects();
   const actions = useActions();
   const { subscription } = useWorkspaceSubscription();
+  const [, trackVisit] = useDashboardVisit();
```
tristandubbeld, line 38: Is it better to return an object from the `useDashboardVisit` hook instead so we don't need the `[, ...]`? Not sure if that has a name 😄.

tristandubbeld (approved):
Love the simplicity, nice job!

packages/app/src/app/pages/Dashboard/index.tsx
olarclara: Definitely! Will change

PR #7313, opened 2023-01-13
feat(sandboxcard): Update custom thumbnail prominence

tristandubbeld:
We're working on making the sandboxes with custom thumbnails more prominent in the dashboard. We do this by hiding some content by default and showing more details on hover to make sure the thumbnail is not obstructed. When hovering we show a scrim to make sure the content is readable. We also determine if the color of the thumbnail is light or dark on average to make the sandbox title contrast. Ideally we should do this on the backend, but we can't right now.

Note: To improve performance I've scaled the image to be way smaller when the calculation is done. This way we don't have to loop over 10 to a 100 thousand pixels in JS. Also, when an image is equally dark and light we can't guarantee the readability of the text. When doing this server side, or when calculating this when uploading the thumbnail we can go more crazy.

#### Style todos:

- [x] Update selected state (hidden behind scrim right now)
- [ ] ~Give a more distinct hover state to the card with an automatically generated thumbnail (better contrast)~
- [x] Double check cards without thumbnails (only showing up after initial creation)
- [x] Dynamic foreground color based on thumbnail
- [x] Transitions
- [x] Show scrim and details always on restricted sandboxes

#### Tech todos:

- [x] Combine `useSandboxThumbnail` with `useImageBrightness` hooks (better perf)
- [x] Only trigger `useImageBrightness` calculations for custom thumbnlails (better perf)
- [ ] Add `try` `catch` around canvas methods (less error prone)

### Testing

_The easy way_

1. Open this branch in the editor and navigate to `/dashboard` in the preview.
2. When you're testing on stream there should be a folder inside "All sandboxes" called "cards". On production there should be a folder called card iterations.
3. Inside the folder you'll find examples with dark and light thumbnails.

_The slightly harder way_

1. Open this branch in the editor and navigate to `/dashboard` in the preview.
2. Create a new sandbox or open an existing sandbox without thumbnail.
3. Add a thumbnail to the sandbox. This can be any image, go wild.
4. Check out the sandbox in the dashboard and see if the text is readable.

### Video

https://user-images.githubusercontent.com/7533849/212685064-a9065f76-1e15-41a6-88a3-a8fb481a2661.mov

### Screenshot

Screenshot 2023-01-16 at 14 12 00


lenasmayor:
@tristandubbeld  the idea is to not show any automatic thumbnails, only the custom uploaded ones. So for any card without an uploaded thumbnail, there will be no background image and the card should follow the original gray layout ([figma here](https://www.figma.com/file/kOIe27lGhBqgKha48CZqmt/Dashboard-Onboarding?node-id=367%3A10468&t=KbkL0VeF9vQcHjuG-1)).
image

Also, could we add a transition to the hover state? Right now it feels a bit harsh.

lenasmayor (changes requested):
@tristandubbeld the idea is to not show any automatic thumbnails, only the custom uploaded ones. So for any card without an uploaded thumbnail, there will be no background image and the card should follow the original gray layout ([figma here](https://www.figma.com/file/kOIe27lGhBqgKha48CZqmt/Dashboard-Onboarding?node-id=367%3A10468&t=KbkL0VeF9vQcHjuG-1)).

Screen Shot 2023-01-16 at 12 22 45


Also, could we add a transition to the hover state? Right now the color/opacity change feels a bit harsh.

olarclara (commented):
Great job calculating the brightness, looking forward to seeing this in production!

---

There's a regression in sandboxes with long titles:

|Stream|Preview|
|:-:|:-:|
|codesandbox stream_dashboard_recent_workspace=411542e0-c5fb-4ee8-85c9-65c8cfd18e52 (1)|dq5q2g-3000 preview csb app_dashboard_recent_workspace=411542e0-c5fb-4ee8-85c9-65c8cfd18e52|

I tried debugging but couldn't figure out why this is happening. Looks like the parent `CardContent` needs `overflow: hidden` too.

---

When polishing comes, finding a way to transition between the different types of cards will provide a great feel to the navigation. I would love to help iterate on this. :)


https://user-images.githubusercontent.com/24959348/212764992-6578b149-ec95-454a-aa43-9781035a4aab.mov

---

The `useBrightness` hook is great! I believe we can improve it by splitting the actual calculations into a utility that we can even write tests for later on.

tristandubbeld:
Updated the restricted sandboxes to always show a scrim. On the screenshot you can see the regular restricted sandbox on the left, and the **hovered** restricted sandbox in the center.

image

lenasmayor (approved):

PR #7326, opened 2023-01-18
feat(cloud sandboxes): Banner for upgrading to V2

DavidHollins6:

This PR adds a new banner at the top of SSE sandboxes giving the opportunity for users to upgrade their sandbox into a V2 cloud sandbox. If they do not have "write_code" permissions, it will instead fork it into a V2 cloud sandbox.

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      
### DIFF
--- a/packages/app/src/app/components/StripeMessages/UpgradeSSEToV2.tsx
+++ b/packages/app/src/app/components/StripeMessages/UpgradeSSEToV2.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { useEffects, useAppState } from 'app/overmind';
 import { MessageStripe } from '@codesandbox/components';
 import { sandboxUrl } from '@codesandbox/common/lib/utils/url-generator';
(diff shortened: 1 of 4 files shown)


packages/app/src/app/components/StripeMessages/UpgradeSSEToV2.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const state = useAppState();
+  const effects = useEffects();
+  const updateSandbox = useEffects().api.updateSandbox;
+
+  return (
+    <MessageStripe variant="primary">
+      <span>
+        This sandbox runs much faster in our new Editor. Do you want to convert
+        it to a Cloud Sandbox?
+      </span>
```
DavidHollins6, line 17: When a user has less than "write_code" permissions, we fork the sandbox instead of upgrading. Do we want to change the copy for that scenario?

cc @zehfernandes @filipelima18

packages/app/src/app/components/StripeMessages/UpgradeSSEToV2.tsx
```diff
@@ -0,0 +1,61 @@
...
+      <MessageStripe.Action
+        onClick={async () => {
```
tristandubbeld, line 19: Do we want some loading state? Not sure how long it takes for the sandbox to be transformed into a v2 sandbox and the user is redirected?

packages/app/src/app/components/StripeMessages/UpgradeSSEToV2.tsx
CompuIves: I would say that it's a good idea. I would lowercase "Editor" and change "convert" to "fork" if it's going to be a forked sandbox. I can also write some docs that we can link to from here, which we can add in another PR.

packages/app/src/app/components/StripeMessages/UpgradeSSEToV2.tsx
zehfernandes: If we need it you can use the loading button:
 
image

packages/components/src/components/MessageStripe/MessageStripe.tsx
```diff
@@ -34,11 +38,13 @@ export const MessageAction = ({
 const backgroundVariants: Record<Variant, string> = {
   trial: '#644ED7',
   warning: '#F7CC66',
+  primary: 'button.background',
```
tristandubbeld, line 41: Lovely `'button.background'` 😄. You can leave this for now, but in the future I think we want to convert the values in this codebase to hex to make it slightly easier to turn it into proper design system components. Just a thought that popped up when reviewing.

yarn.lock
```diff
@@ -32131,6 +32131,11 @@ tempfile@^2.0.0:
     temp-dir "^1.0.0"
     uuid "^3.0.1"
 
+template-icons@^1.0.0:
```
tristandubbeld, line 32134: I believe that if you merge `master` into this branch, this change is gone from the PR.

tristandubbeld (commented):
Do you have an active SSE sandbox somewhere I can test with? I remember a conversation on Slack but can't find it at the moment :(

yarn.lock
DavidHollins6: > Do you have an active SSE sandbox somewhere I can test with? I remember a conversation on Slack but can't find it at the moment :(

Here you go @tristandubbeld - https://veeexz-3000.preview.csb.app/s/clever-boyd-n41wst

tristandubbeld (approved):
Looks good! Only thing I'm wondering is if we should show the banner for anonymous users too, what do you think:

image

PR #7336, opened 2023-01-20
fix(icons): Rename remix to remix-starter to fix icon

tristandubbeld:
The name of the Remix template was named `remix-starter` so I changed the name to fix the icon. It currently shows the React icon though, so we should update that.

### Testing

1. Open the create new modal
2. Click through all the tabs to see if it doesn't crash
3. In the official templates, there should be a Remix template with a React icon

### DIFF
--- a/packages/common/src/templates/index.ts
+++ b/packages/common/src/templates/index.ts
@@ -100,7 +100,7 @@ export type TemplateType =
   | 'docusaurus'
   | 'babel-repl'
   | 'esm-react'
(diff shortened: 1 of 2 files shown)


alexnm (approved):

tristandubbeld:
@IvesCodeSandbox do you need to re-build and publish this for `codesandbox-applications`?

CompuIves:
Good point. Indeed that needs to be done. I'll take a look

PR #7346, opened 2023-01-23
fix: template owner

alexnm:
Changed the team name query for templates from `collection -> team -> name` to `team -> name` as sometimes templates are created in draft and moved to templates by the team, so they don't belong to any collection. Not sure why the query was originally like that.

This changes the `GitHub` fallback label under some of the templates to `CodeSandbox` or the correct owner (user or team)

Because `/official` and `/essential` are still two REST endpoints we use to show some of the templates in the create modal, I decided to hardcode `CodeSandbox` there for now, to have it show nice in prod and later we can update the endpoints.

Before

image
image

After
image
image

### DIFF
--- a/packages/app/src/app/graphql/types.ts
+++ b/packages/app/src/app/graphql/types.ts
@@ -611,6 +611,8 @@ export type Project = {
   repo: Scalars['String'];
   /** Git repository for the project, as it appears on the original git provider */
   repository: Repository;
+  /** Compute resources for the project, based on settings and team subscription */
+  resources: Resources;
   /**
    * Team to which this project is assigned
    *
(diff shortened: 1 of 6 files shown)


packages/app/src/app/graphql/types.ts
```diff
@@ -715,6 +717,21 @@ export type Repository = GitHubRepository;
 /** GitHub webhook event about a repository. */
 export type RepositoryEvent = InstallationEvent;
 
+/**
+ * Available computing resources for the project
+ *
+ * These resources may be custom for a project, or they may be determined by the team's subscription.
+ */
+export type Resources = {
```
alexnm, line 725: nice! we're going to use this for the VM specs in the other codebase :))

tristandubbeld (approved):
Looks good!

CompuIves (approved):
This makes sense, looks great!

PR #7350, opened 2023-01-23
fix(upgrade): Pass correct property to fork api call

DavidHollins6:

I was passing the incorrect value into the fork api endpoint. It was not properly converting the sandbox into V2.

It would redirect to the v2 fork of the sandbox, but if you re-opened it from the dashboard it would be v1.

## What is the new behavior?

The newly forked sandbox will always be v2

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      
### DIFF
--- a/packages/app/src/app/components/StripeMessages/UpgradeSSEToV2.tsx
+++ b/packages/app/src/app/components/StripeMessages/UpgradeSSEToV2.tsx
@@ -42,7 +42,7 @@ export const UpgradeSSEToV2Stripe = () => {
             window.location.href = sandboxV2Url;
           } else {
             const forkedSandbox = await effects.api.forkSandbox(sandboxId, {
-              isV2: true,
+              v2: true,
               teamId: state.activeTeam,
             });
 


CompuIves (approved):
Noooice!

packages/app/src/app/components/StripeMessages/UpgradeSSEToV2.tsx
```diff
@@ -42,7 +42,7 @@ export const UpgradeSSEToV2Stripe = () => {
             window.location.href = sandboxV2Url;
           } else {
             const forkedSandbox = await effects.api.forkSandbox(sandboxId, {
-              isV2: true,
+              v2: true,
```
tristandubbeld, line 45: Interesting that this wasn't caught!

tristandubbeld (approved):

packages/app/src/app/components/StripeMessages/UpgradeSSEToV2.tsx
DavidHollins6: The redirect was hiding it 😄

DavidHollins6: oh you mean by typescript?

tristandubbeld: Yes `typescript`, not so strict for this codebase.

PR #7351, opened 2023-01-23
fix(sidebar): Show payment method text and CTA when applicable

tristandubbeld:
Changed the conditional to show the "update payment method" message to always (5 days before end) show up properly rather than just two days before the end of the trial.

Closes XTD-607

image


necoline (approved):

olarclara (approved):
Nice!

PR #7354, opened 2023-01-24
fix(dashboard): create/edit team name

olarclara:
Closes XTD-609.

- Allows whitespace in team names when creating a new team.
- Fixes whitespace behavior in team names when editing an existing team.

### Create team:

- [ ] It's possible to create a team with spaces in the name (eg: `my team`)
- [ ] It's not possible to create a team with only spaces in the name (eg: `   `)
- [ ] It's not possible to create a team if the user is already part of a team with the same name

### Edit team:

- [ ] It's possible to edit a team with spaces in the name (eg: `my team`)
- [ ] It's not possible to edit a team with only spaces in the name (eg: `   `)
- [ ] It's not possible to edit a team if the user is already part of a team with the same name

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Content/routes/Settings/TeamSettings/WorkspaceSettings.tsx
+++ b/packages/app/src/app/pages/Dashboard/Content/routes/Settings/TeamSettings/WorkspaceSettings.tsx
@@ -112,10 +112,30 @@ export const WorkspaceSettings = () => {
     });
   };
 
+  const handleTeamNameChange = event => {
+    const { value } = event.target;
+
+    // Get the input and remove any whitespace from both ends.
+    const trimmedName = value?.trim() ?? '';
+
+    // Validate if the name input is filled with whitespaces.
+    if (trimmedName) {
+      event.target.setCustomValidity('');
+    } else {
+      event.target.setCustomValidity('Team name is required.');
+    }
+  };
+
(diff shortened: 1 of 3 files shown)


packages/app/src/app/pages/Dashboard/Content/routes/Settings/TeamSettings/WorkspaceSettings.tsx
```diff
@@ -112,10 +116,34 @@ export const WorkspaceSettings = () => {
...
+
+    const name = event.target.name.value?.trim();
+    const description = event.target.description.value?.trim();
+
+    if (!name) {
+      return;
+    }
+
     setLoading(true);
     try {
```
tristandubbeld, line 148: Should we also add a `catch`?

tristandubbeld (approved):
Super interesting that we didn't actually trim and check for existing names before. Code looks good!

packages/app/src/app/pages/Dashboard/Content/routes/Settings/TeamSettings/WorkspaceSettings.tsx
olarclara: @tristandubbeld I don't think so,`setTeamInfo` catches errors internally and dispatches a notification when they occur. I added this info as a comment to the code.

PR #7360, opened 2023-01-25
Draft/nifty newton

necoline:

Changes text for the home banner to highlight the lack of credit card requirements for trials

https://linear.app/codesandbox/issue/XTD-615/recent-banner

## What is the new behavior?
[x]Change headline text to Try Team Pro for free when trial is enabled 
   [x]show 'Upgrade to Pro' when trial is not enabled
[x]Change subtitle text to '14 days free trial. No credit card required'.
   [x]show 'Enjoy the full CodeSandbox experience' when trial is not enabled.
[x]Review text and order of features list

image

I viewed this on the preview

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      
### DIFF
--- a/packages/app/src/app/pages/Dashboard/Components/UpgradeBanner/UpgradeBanner.tsx
+++ b/packages/app/src/app/pages/Dashboard/Components/UpgradeBanner/UpgradeBanner.tsx
@@ -93,7 +93,7 @@ export const UpgradeBanner: React.FC<UpgradeBannerProps> = ({ teamId }) => {
     >
       <Element css={{ overflow: 'hidden' }}>
         <StyledTitle color="#EDFFA5" weight="500" block>
-          Upgrade to Pro
+          {isEligibleForTrial ? 'Try Team Pro for free' : 'Upgrade to Pro'}
         </StyledTitle>
         <Stack
           css={{
@@ -103,7 +103,9 @@ export const UpgradeBanner: React.FC<UpgradeBannerProps> = ({ teamId }) => {
         >
           <Stack direction="vertical" justify="space-between">
             <StyledTitle block>
-              Enjoy the full CodeSandbox experience.
+              {isEligibleForTrial
+                ? '14 days free trial. No credit card required.'
+                : 'Enjoy the full CodeSandbox experience.'}
             </StyledTitle>
             <Stack
               css={{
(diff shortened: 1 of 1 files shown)


tristandubbeld (approved):
Nice! There are a two more instances where the features (server and amount of editors) are defined:
- `/packages/app/src/app/constants.ts` (two times)
- `packages/app/src/app/pages/common/Modals/SubscriptionCancellation/SubscriptionCancellation.tsx`

The first looks like this right now:

image

Do we want to change these too?

necoline:
Thanks for pointing that out
There is [this ticket](https://linear.app/codesandbox/issue/XTD-616/start-trial-step-on-create-team-flow) that mentions the other locations. It makes sense to address both in one PR

necoline:
@tristandubbeld I made the changes you listed in the comment 👆 
I also added the changes from this ticket
 - [x] Move team name to modal header 
         🚨 I don't think this is right solution since the team name calculates after the render so you can see the previously active team before showing the new team name
 - [x] Add subtitle with '14 days free trial. No credit card required'
        🚨 This is my first component that I've put together from designs, so I'm sure there are some issues with margins and size that you will want to evaluate
 - [x] Review list order 
 - [x] Add comparative tags (in yellow)
 - [x] Remove modal footer text

packages/app/src/app/pages/Dashboard/Components/NewTeamModal/TeamSubscription.tsx
```diff
@@ -1,33 +1,32 @@
...
+  Feature,
+  TEAM_FREE_FEATURES,
+  TEAM_PRO_FEATURES_WITH_PILLS,
+} from 'app/constants';
+
+const StyledTitle = styled(Text)`
+  font-size: 24px;
+  line-height: 32px;
+  letter-spacing: -0.019em;
+  margin: 0;
```
tristandubbeld, line 22: I think this `margin` can be removed, but I'm not 100% sure.

tristandubbeld (approved):
> I don't think this is right solution since the team name calculates after the render so you can see the previously active team before showing the new team name

This happens in the other places where the team name is shown too, so it doesn't seem to be a new problem. If we want to fix this I suggest we create a new task for it!

> This is my first component that I've put together from designs

Looks good :)

PR #7363, opened 2023-01-26
fix(legacy-pro): default to dollar when currency is unavailable

tristandubbeld:
Fixes an error where we render `null`:

image


olarclara (approved):
💵

PR #7369, opened 2023-01-30
feat(dashboard): make content scroll horizontally

olarclara:
|Before|
|:-:|
|codesandbox stream_dashboard_recent_workspace=cab9e635-ec49-4c82-806c-c521babd3a6c|

|After|
|:-:|
|localhost_3000_dashboard_recent_workspace=cab9e635-ec49-4c82-806c-c521babd3a6c|

Adds "carousel" to the dashboard recent page.

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Content/routes/Recent/DocumentationRow.tsx
+++ b/packages/app/src/app/pages/Dashboard/Content/routes/Recent/DocumentationRow.tsx
@@ -1,6 +1,7 @@
 import track from '@codesandbox/common/lib/utils/analytics';
-import { ArticleCard, VideoCard, Stack } from '@codesandbox/components';
+import { ArticleCard, VideoCard } from '@codesandbox/components';
 import { useWorkspaceAuthorization } from 'app/hooks/useWorkspaceAuthorization';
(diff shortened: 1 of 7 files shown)


packages/app/src/app/pages/Dashboard/Content/routes/Recent/DocumentationRow.tsx
```diff
@@ -80,41 +81,35 @@ export const DocumentationRow: React.FC = () => {
     });
   };
 
+  const buildItems = () => {
```
tristandubbeld, line 84: Could be changed to `const items = DOCS.map` variable instead of function if we want.

tristandubbeld (approved):
Works great! One thing I found is that the last item doesn't have padding on the right when the carousel is scrolled all the way to the end. Would it be easy to add it to make it look slightly better?

packages/app/src/app/pages/Dashboard/Content/routes/Recent/DocumentationRow.tsx
olarclara: Great suggestion! Addressed it here: [d273581](https://github.com/codesandbox/codesandbox-client/pull/7369/commits/d27358144657bc873a762469e032fbeaf4caf497)

PR #7372, opened 2023-01-30
feat: make private package available for sandpack

danilowoz:
**Data journey:** 
- Sandpack sends a codesandbox team id to the bundler;
- The bundler looks at this team-id and tries to fetch custom dependencies;

**If it fails:**
- It sends back an error: `NPM_REGISTRY_UNAUTHENTICATED_REQUEST`;
- Sandpack launches an error screen with a sign-in button;

**If it succeeds:**
- The bundler uses a sandpack token (ready-only, with access to this team and private package capabilities) to fetch the custom registry;
- If at some point it fails, it removes the sandpack token from the cookie

**Sign-in flow:**
- Sandpack sends a request to the bundler to sign-in;
- The bundler launches a popup to `auth/sandpack/[team-id]`
- This pop-up goes to csb.io and tries to fetch a custom sandpack token based on the given team-id;
- Then, sends this token back to the parent window and add it to the cookie

https://www.notion.so/codesandbox-app/Sandpack-Pro-Seamless-private-package-60deb673b08e4ffdb6c1f909e9eca8a8?pvs=4#056304e0199b41ac988698ba4219d90c


packages/app/src/app/pages/SandpackSecret/index.tsx
```diff
@@ -0,0 +1,114 @@
+import React, { useEffect } from 'react';
```
danilowoz, line 1: Pop-up to get a read-only sandpack-token from a team id and send it back to the parent window.

packages/app/src/sandbox/compile.ts
```diff
@@ -335,8 +341,57 @@ async function initializeManager(
     }
   );
 
+  /**
+   * If a team-id is provided, we can mount a registryUrl and proxy
+   * dependencies request through CSB proxy
+   */
+  if (teamId) {
```
danilowoz, line 348: If a team-id is provided, we assume it's a sandpack instance

packages/app/src/sandbox/compile.ts
```diff
@@ -335,8 +341,57 @@ async function initializeManager(
...
+      registry_auth_key: string;
+      registry_type: string;
+      registry_url: string;
+    };
+
+    customNpmRegistries.push({
+      enabledScopes: registry.enabled_scopes,
+      limitToScopes: registry.limit_to_scopes,
+      proxyEnabled: registry.proxy_enabled,
+      registryUrl:
```
danilowoz, line 381: If there is no custom registry URL, it defaults to CSB proxy

packages/app/src/sandbox/index.ts
```diff
@@ -68,6 +72,14 @@ requirePolyfills().then(() => {
             data: {},
           });
         }
+      } else if (data.type === 'sign-in') {
+        await requestSandpackSecretFromApp(data.teamId);
+
+        window.location.reload();
```
danilowoz, line 78: Once it gets a sandpack-token, it reloads the entire iframe, which will ensure the security headers policy will be applied by the worker

packages/sandpack-core/src/npm/dynamic/fetch-protocols/npm-registry.ts
```diff
@@ -124,8 +128,25 @@ export class NpmRegistryFetcher implements FetchProtocol {
...
+     * Private packages conditionals:
+     * 1. Explicit token: if `authToken` is provide, add it to the header
+     * 2. Proxy disabled: then it's a custom registry, so do not anything
+     * 3. Proxy is enabled and team-id is provide: it's a private package provided by CSB
+     */
+
     if (this.authToken) {
-      headers.append('Authorization', `Bearer ${this.authToken}`);
+      // Custom registry url
+      headers.append('Authorization', `${this.authType} ${this.authToken}`);
```
danilowoz, line 141: Added support for custom auth type

package.json
```diff
@@ -44,7 +44,7 @@
     "build:embed": "lerna run build:embed --scope app --stream && gulp",
     "build:prod": "lerna run build --scope app --stream && lerna run copy-assets --scope app --stream",
     "commit": "concurrently \"yarn typecheck\" \"yarn lint\" && git commit -m",
-    "postinstall": "yarn lerna run install-dependencies --stream --concurrency 1 && opencollective postinstall && patch-package",
```
danilowoz, line 47: Removed `patch-package`, because we used to use on `homepage`

DeMoorJasper (approved):

PR #7374, opened 2023-01-31
fix(dashboard): user settings form

olarclara:
Closes XTD-645.

Because the `submit` is outside of the `form` element, the action was never triggered. This adds a relationship between the form and the submit button.


CompuIves (approved):
Ahh, smart find!!

PR #7375, opened 2023-01-31
fix(dashboard): Gracefully change to personal workspace

tristandubbeld:
Closes XTD-632
Closes XTD-630
Closes XTD-631

#### Before

https://user-images.githubusercontent.com/7533849/215761133-8d62ea0e-cabd-4030-85fa-97221828bb3f.mov

#### After

1. First we show that the application gracefully changes to the personal workspace when no id is present in the url or `localStorage`. No message is shown.

2. We then show behaviour when the suggested id is wrong. We change to the personal workspace and make sure we're not trying to fetch any data with the wrong id. We also notify the user that the "saved" id is wrong and we've activated the personal workspace.

3. We than show the same behaviour, but with an existing team id the user does not have access to. This behaves the same way as described above.

4. And finally we use the url with a wrong team id. This behaves the same way as described above.

https://user-images.githubusercontent.com/7533849/215761249-705b7656-4130-4b77-9969-7c1e80f6959f.mov

### DIFF
--- a/packages/app/src/app/overmind/internalActions.ts
+++ b/packages/app/src/app/overmind/internalActions.ts
@@ -12,6 +12,7 @@ import { NotificationStatus } from '@codesandbox/notifications';
 import { hasPermission } from '@codesandbox/common/lib/utils/permission';
 import values from 'lodash-es/values';
 
+import { TeamFragmentDashboardFragment } from 'app/graphql/types';
 import { ApiError } from './effects/api/apiFactory';
 import { defaultOpenedModule, mainModule } from './utils/main-module';
 import { parseConfigurations } from './utils/parse-configurations';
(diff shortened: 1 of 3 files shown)


CompuIves:
This flow shows soo much better

packages/app/src/app/overmind/internalActions.ts
```diff
@@ -590,61 +591,114 @@ export const setViewModeForDashboard = ({ effects, state }: Context) => {
...
-  actions.internal.setActiveTeamFromUrl() ||
-  actions.internal.setActiveTeamFromLocalStorage() ||
-  actions.internal.setActiveTeamFromPersonalWorkspaceId();
+const INVALID_ID_TITLE = 'Workspace not recognized.';
+const INVALID_ID_MESSAGE =
+  "The workspace in the URL or stored in your browser is unknown. We've automatically switched to your personal workspace.";
 
-export const setActiveTeamFromLocalStorage = ({
+// Rename to initializeTeam (because we also use personalWorkspaceId);
+export const setActiveTeamFromUrlOrStore = async ({
```
CompuIves, line 599: Do we call this often? Since it does an API call.

CompuIves (approved):
This is sooooo much better!!

packages/app/src/app/overmind/internalActions.ts
olarclara, line 599:
```suggestion
export const setActiveWorkspaceFromUrlOrStore = async ({
```

Works for personal and team.

packages/app/src/app/overmind/internalActions.ts
```diff
@@ -590,61 +591,114 @@ export const setViewModeForDashboard = ({ effects, state }: Context) => {
...
 
-  return null;
+  const isSuggestedTeamValid = userTeams?.some(
+    team => team.id === suggestedTeamId
+  );
+
+  return {
+    id: suggestedTeamId,
+    isValid: isSuggestedTeamValid,
+  };
```
olarclara, line 660: What do you think about changing the return?

```suggestion
  return isSuggestedTeamValid ? suggestedTeamId : null;
```

Because the id isn't useful unless it's valid.

olarclara (approved):
Nice!

packages/app/src/app/overmind/internalActions.ts
tristandubbeld: Nice that this stands out. I have been thinking about this change for a bit when I made it. I thought it made sense to always return the `id` no matter what because the function was called `getTeamIdFromUrlOrStore`. But I think you're right, we don't use it anywhere so I think it's safe to apply your suggestion.

tristandubbeld:
Thanks for the suggestions @olarclara! I've applied one, but I will create a task for the other and some potential improvements and todos left in the code.

PR #7376, opened 2023-01-31
fix(dashboard): Only close modal when modal is open

tristandubbeld:
Closes XTD-634

Close modal was called for a modal which caused all modals to close. This extra check makes sure it's only called when necessary and the modal is actually open.


olarclara (approved):
I would imagine that `close` would only perform the action if the modal is the current. 🤔

tristandubbeld:
> I would imagine that close would only perform the action if the modal is the current. 🤔

I thought the same thing! But it was the only relevant thing I was able to find to the modal closing, and it turns out that it sets all modals to `null`:
```
state.modals.current = null;
```

PR #7378, opened 2023-01-31
fix(dashboard): Import link with right URL and close modal after navigation

tristandubbeld:
Closes XTD-634

And this time for real. I've reverted the buggy behaviour that was supposed to close modals on navigation, as it wasn't necessary. I found that the logic to generate the URL and close the create new modal was behaving weird. The "upgrade to pro" URL shows up when a free team tries to import a private repository.

image

The link can either point to `/pro`, `/docs/*` or Stripe. When it's pointing to `/pro` it stays on the "dashboard" pages and uses a React Router link to navigate. This is done client side, so modals won't close. Adding an `onClick` where we close the modal for this case fixes this.

- The link is pointing to `/pro` when the user is **admin and the stripe URL hasn't been generated yet**
- The link is pointing to `/docs` when the user is **not admin and not on a personal space**
- The link is pointing to Stripe when the user is **admin and the stripe URL is generated**

Generating the Stripe URL happens async and takes time.

### Proof that the modal closes

https://user-images.githubusercontent.com/7533849/215874494-801c92ee-6b6c-499e-b8f2-5b3ad0ea48f5.mov

### Proof that the create team modal shows up

https://user-images.githubusercontent.com/7533849/215875593-94172615-c924-4930-af1c-06fd5f23ac80.mov

### DIFF
--- a/packages/app/src/app/components/CreateSandbox/Import/importLimits.tsx
+++ b/packages/app/src/app/components/CreateSandbox/Import/importLimits.tsx
@@ -100,38 +100,42 @@ export const PrivateRepoFreeTeam: React.FC = () => {
     cancel_path: pathname,
   });
 
-  const checkoutURL = React.useMemo(() => {
-    if (isTeamAdmin || isPersonalSpace) {
-      if (checkout.state === 'READY') {
-        return checkout.url;
-      }
-      return '/pro';
-    }
+  let checkoutURL: string;
 
-    return isEligibleForTrial
+  if (isTeamAdmin || isPersonalSpace) {
+    if (checkout.state === 'READY') {
+      checkoutURL = checkout.url;
+    } else {
+      checkoutURL = '/pro';
+    }
+  } else {
+    // Not team admin or personal workspace points to docs
+    checkoutURL = isEligibleForTrial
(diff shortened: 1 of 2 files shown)


CompuIves (approved):
Makes sense! If this has been tested with opening other modals, let's merge it :shipit:

CompuIves:
Will test the flows once more once it is deployed!

packages/app/src/app/components/CreateSandbox/Import/importLimits.tsx
```diff
@@ -100,38 +100,42 @@ export const PrivateRepoFreeTeam: React.FC = () => {
...
+    }
+  } else {
+    // Not team admin or personal workspace points to docs
+    checkoutURL = isEligibleForTrial
       ? '/docs/learn/plan-billing/trials'
       : '/docs/learn/introduction/workspace#managing-teams-and-subscriptions';
-  }, [checkout, isEligibleForTrial, isTeamAdmin, isPersonalSpace]);
+  }
+
+  const isDashboardLink = checkoutURL.startsWith('/pro');
```
olarclara, line 118: Clever.

olarclara (commented):
Nice!

PR #7380, opened 2023-02-01
Ignore nested package.jsons when the root of a package has exports

Andarist:
An example of a broken codesandbox can be found here: https://codesandbox.io/s/vigilant-pike-wh60h2?file=/src/App.js

this particular codesandbox imports from:
1. `@emotion/react`
2. `@emotion/react/jsx-runtime`

1\. is resolved `https://wh60h2.csb.app/node_modules/@emotion/react/dist/emotion-react.cjs.js`
2\. is resolved to `https://wh60h2.csb.app/node_modules/@emotion/react/jsx-runtime/dist/emotion-react-jsx-runtime.browser.esm.js`

the problem here is that those are files from different "sets", the main entry point doesn't resolve to a file from a "browser" set, but somehow that additional `/jsx-runtime` entry is resolved to a browser-specific file

this is an issue because there is a React context there, that they are supposed to share but accidentally they are using its different copies

I've noticed because I was trying to configure an Emotion's cache, with a custom key and some options, but this is the generated class name:
```html
class="css-vyc7dn-App"
```

This is a bug in the CSB resolver, if we just download this sandbox as zip and run it locally with webpack then the generated class name is correct.

TLDR; when `pkg/package.json#exports` is available then it should be the **only** source of the information on how to resolve its module specifiers etc. Perhaps some tool-specific aliases can also be used - like if they come from the tool's configuration. But all nested `package.json` files and all "traditional" in-package.json main fields/aliases should be ignored (including `package.json#main`, `package.json#module`, `package.json#browser`, and more). `exports` are meant to represent the full manifest of importable paths

cc @dbismut (whose [`use-gesture`](https://github.com/pmndrs/use-gesture) is also affected)
cc @danilowoz


danilowoz:
Thanks, @Andarist!

@DeMoorJasper, could you please take a look at this?

packages/sandpack-core/src/resolver/resolver.test.ts
```diff
@@ -503,6 +503,30 @@ describe('resolve', () => {
...
+        extensions: ['.ts', '.tsx', '.js', '.jsx'],
+        isFile,
+        readFile,
+      });
+      expect(resolved).toBe(
+        '/node_modules/@emotion/react/jsx-runtime/dist/emotion-react-jsx-runtime.cjs.js'
+      );
+    });
+
+    it('should always use exports from the root of the package and not from the closest package.json', () => {
```
DeMoorJasper, line 519: Is there a specification you can point to for this test? I always thought having a package.json means that is the root of said package. So it should always use the closest package.json no? And how can we reliably detect the root package.json, in case of pnpm or something this logic no longer means it will be under `node_modules/pkg/package.json` but instead somewhere `.store/pkg@version/node_modules/pkg/package.json`

packages/sandpack-core/src/resolver/resolver.ts
```diff
@@ -184,12 +200,17 @@ function* resolveNodeModule(
 
       try {
         const pkgFilePath = pathUtils.join(rootDir, pkgSpecifierParts.filepath);
-        const pkgJson = yield* loadPackageJSON(pkgFilePath, opts, rootDir);
+        const rootPkgJson = yield* loadPackageJSON(rootDir, opts);
```
DeMoorJasper, line 203: Shouldn't this fail if rootPkgJson does not exist? Given that it isn't a node_module in that case

packages/sandpack-core/src/resolver/resolver.ts
```diff
@@ -184,12 +200,17 @@ function* resolveNodeModule(
...
+        const pkgJson =
+          rootPkgJson && rootPkgJson.content.hasExports
+            ? rootPkgJson
+            : yield* loadNearestPackageJSON(pkgFilePath, opts, rootDir);
```
DeMoorJasper, line 207: Isn't this duplicate of rootPkgJson, when would this not be rootPkgJson?

packages/sandpack-core/src/resolver/utils/pkg-json.ts
```diff
@@ -48,8 +49,11 @@ export function processPackageJSON(
     }
   }
 
+  let hasExports = false;
+
   // load exports if it's not the root pkg.json
   if (content.exports && pkgRoot !== '/') {
+    hasExports = true;
```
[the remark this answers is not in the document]
DeMoorJasper: This looks fine, no need to complicate it, the less changes the quicker we can get this merged :D

DeMoorJasper (commented):
Left a couple questions, but in general this looks really good. Amazing to see a big contribution like this.

Didn't know how to handle the nowadays very mixed package.json's in terms of specs, but ignoring non-exports in case there are exports makes total sense to me

Thanks @danilowoz for the ping

packages/sandpack-core/src/resolver/utils/pkg-json.ts
```diff
@@ -48,8 +49,11 @@ export function processPackageJSON(
     }
   }
 
+  let hasExports = false;
+
   // load exports if it's not the root pkg.json
   if (content.exports && pkgRoot !== '/') {
```
DeMoorJasper, line 55: Doesn't seem like it ignores non-exports in this case though? It still processes all other things in the package.json. Would be great to hear some more about this, I can help out with getting this PR over the finish line.

Andarist:
> Left a couple questions, but in general this looks really good. Amazing to see a big contribution like this.

Don't you make me blush

> Didn't know how to handle the nowadays very mixed package.json's in terms of specs, but ignoring non-exports in case there are exports makes total sense to me

It doesn't only make sense - it's a requirement. TypeScript even ignores `package.json#types` when there are `package.json#exports` (you have to use the `types` condition in the `package.json#exports`) - to read `exports` they require an opt-in from the project though (the new `moduleResolution: node16/nodenext` option)

packages/sandpack-core/src/resolver/resolver.test.ts
Andarist: For other reviewers, you can have package layouts like this:
```ts
pkg-a/package.json
pkg-a/extra/package.json
```

When requesting `pkg-a/extra` we have to:
- first check if there is `pkg-a/package.json#exports`
- if it's there then it becomes the only source of information about available module specifiers within the package etc
- it that's not there then `pkg-a/extra/package.json` should be used as the possible source of information, main fields like `pkg-a/extra/package.json#main`, `pkg-a/extra/package.json#module`, `pkg-a/extra/package.json#browser` might have to be interpreted and tried

At the same time, `pkg-a/extra/package.json#exports` should **always** be ignored. `exports` is reserved for the "root" `package.json`

packages/sandpack-core/src/resolver/resolver.ts
Andarist: This is called in the `resolveNodeModule` so I assumed that this already makes an assumption about being called for something in `node_modules`. Theoretically, it is possible to put manually something into `node_modules` without putting a `package.json` into it though. Should I handle this case? What should happen? Should I simply throw? Is there any logic in place that will catch this error and handle it gracefully?

packages/sandpack-core/src/resolver/resolver.ts
Andarist: I mention this in [the other comment](https://github.com/codesandbox/codesandbox-client/pull/7380/files#r1095856464) - "nested" `package.json` files are possible and actually participate in the "classic" resolution algorithm. So it's still required to read from them - but only if the root one doesn't have `exports`

packages/sandpack-core/src/resolver/utils/pkg-json.ts
Andarist: Yeah, this makes sense and you are right. What I propose:
- move this block to the top of the function since `exports` always have the top priority + return early with the `exports`-based aliases if there are any. We also need to be cautious about `package.json#exports: false` - it is also a valid value. We probably should just return an empty `aliases` object in such a case or something.
- if there is no `exports` field then we should continue reading from the other fields

How does it sound?

packages/sandpack-core/src/resolver/resolver.ts
DeMoorJasper: Makes sense, thanks for explaining

packages/sandpack-core/src/resolver/utils/pkg-json.ts
Andarist: Pushed out the change. No extra tests added though - it's hard to craft on the spot one that would fail with the previous shape of the code

DeMoorJasper: Awesome thanks, could you alsk update the jest snapshot using yarn test -u so we can get this merged?

Andarist: Done

Andarist:
One additional q - in what contexts is this resolver used? Is it just on the web? Is it ever supposed to resolve things on the server side?

DeMoorJasper:
@Andarist it is used for resolving on the web in our sandpack-bundler (so any sandbox starting with https://codesandbox.io/s/ in the url).

However we're working on some new things that use the standalone version of this resolver (https://github.com/codesandbox/sandpack-resolver - will port over this bugfix to that repo probably next week, the standalone resolver is also a bit faster due to some compiler magic) that are more like a server environment than the current sandpack-bundler. We also use this in the new sandpack-bundler which is used by the react docs for their examples. So this resolver gets used a lot, quite a couple million users each month. 

Currently there are no plans to use this in our actual vm sandboxes, but never know what the future holds.

Andarist:
I've asked because this resolver here converts `package.json` to an alias map but it doesn't take into account any "conditions". So the knowledge about conditions is being lost along the way (I think?). So my fear is that in its current state, it's not capable of resolving things completely correctly for different environments.


I'm thinking about cases like this (and more):
```ts
exports: {
  browser: './browser-file.js',
  node: './node-file.js'
}
```

DeMoorJasper:
> I've asked because this resolver here converts `package.json` to an alias map but it doesn't take into account any "conditions". So the knowledge about conditions is being lost along the way (I think?). So my fear is that in its current state, it's not capable of resolving things completely correctly for different environments.
> 
> 
> 
> 
> 
> I'm thinking about cases like this (and more):
> 
> ```ts
> 
> exports: {
> 
>   browser: './browser-file.js',
> 
>   node: './node-file.js'
> 
> }
> 
> ```

The logic that converts the exports into aliases keeps this in mind, the standalone resolver I mentioned also allows us to configure these condition keys so we can use the resolver in node, web, ...

https://github.com/codesandbox/codesandbox-client/blob/master/packages/sandpack-core/src/resolver/utils/exports.ts#L4

Andarist:
Hm, the code that you linked to only lists "supported" conditions. The comment there mentions priorities - but that doesn't seem to be correct to me. You can't configure priorities for `exports` - the first key in the object "wins". So the priority is decided by the order of keys in the `exports` object. It's not decided by the requester - nor by the user, nor by the tool. It's up to the package author.

DeMoorJasper:
> Hm, the code that you linked to only lists "supported" conditions. The comment there mentions priorities - but that doesn't seem to be correct to me. You can't configure priorities for `exports` - the first key in the object "wins". So the priority is decided by the order of keys in the `exports` object. It's not decided by the requester - nor by the user, nor by the tool. It's up to the package author.

Oh had no clue that's how it worked. We can change it to act more as a supported list of conditions and respect the order.

I however expect things to break after making that change, but still worth a try.

Andarist:
Yeah, I get you - this is always the risk. It would be really great if you could find other projects that have exhaustive test cases for this stuff. It's really hard to get it right - but I think it's worth getting it right even if you break some stuff along the way. From my PoV, it's super important that every tool resolves my modules in the very same way :p

DeMoorJasper:
@Andarist we've had a report of some broken sandboxes, I've been investigating it and it appears somehow this change makes a module resolve to the node.js version instead of the browser version. (sandbox in question https://codesandbox.io/s/basic-demo-rrppl0y8l4)

In this case a module requires `fflate` (https://unpkg.com/browse/fflate@0.6.10/package.json) that resolves to `fflate/lib/index.cjs` which requires `lib/node-worker.cjs` which in the pkg.json gets aliased using the browser field to `./lib/worker.cjs` but we no longer look at the browser field, so it ends up with the node version... which requires worker_threads

However I noticed that respecting the order of exports would fix this given that `import` comes before `require` for fflate, but looking at the rest of the pkg.json I think this is a bug in fflate and they should probably make their require point to `lib/browser.cjs` to be consistent...

Not sure what to do here, we can double-down on the actual spec and respect the order as you suggested and/or still respect the browser field? A second opinion from you is very welcome here, will make the appropriate changes tomorrow or later tonight

Andarist:
> Not sure what to do here, we can double-down on the actual spec and respect the order as you suggested and/or still respect the browser field? A second opinion from you is very welcome here, will make the appropriate changes tomorrow or later tonight

If you ask my opinion, I would respect the "spec" here. Deviating from it introduces a weird situation when a package works in bundler A but doesn't work in bundler B and that doesn't help anyone (maybe with the exception of users of bundler A :P). 

I think it's important to not spread "incorrect" patterns by allowing those in resolution algorithms because it creates a butterfly effect. People often look at how other packages are configured and might copy-paste incorrect things and if we allow those to be loaded successfully then they might not double-check other bundlers etc. 

I think that package authors should absorb this complexity and they should correct their mistakes. The alternative is that users will suffer from misconfigured packages and there are more users than package authors.

One of the motivating factors here was that the same codesandbox was executing differently when running in codesandbox and when downloaded and executed with webpack. I've tried to replicate the same problem with the codesandbox that you linked to and I couldn't.

So I went investigating this and I think that webpack might actually allow alias fields even with exports and maybe my original issue had some extra precondition there. I tried to analyze the webpack's algorithm but their code is quite convoluted and it's hard to do that on the spot. I managed to find some relevant tests [here](https://github.com/webpack/enhanced-resolve/blob/90aa4017e8f1fd86991527cb350b74419ea56e91/test/exportsField.js#L2005-L2051)

I **think** this is what happens:
1. the request is first resolved using `exports`
2. and only then the alias field is "tried"

This would match what we can observe in those tests, the first one is using `conditionNames: ["webpack"]` whereas the second one is using `conditionNames: ["node"]`. So even though they request the same path and that path exists in the `package.json#browser`... they don't end up resolving the same file. That's because the `webpack` condition redirects the request to the `/lib2/` directory ([here](https://github.com/webpack/enhanced-resolve/blob/90aa4017e8f1fd86991527cb350b74419ea56e91/test/fixtures/exports-field/node_modules/exports-field/package.json#L12)). So the request path gets changed and that final path is not in the `package.json#browser`.

When it comes to the second test case - the request path also gets changed (from `/dist/` to `/lib/`) but the new path is present in the `package.json#browser` and thus gets "redirected" once again.

I think it might be quite hard to replicate this behavior with your "flat" object of `aliases`.

DeMoorJasper:
> > Not sure what to do here, we can double-down on the actual spec and respect the order as you suggested and/or still respect the browser field? A second opinion from you is very welcome here, will make the appropriate changes tomorrow or later tonight
> 
> If you ask my opinion, I would respect the "spec" here. Deviating from it introduces a weird situation when a package works in bundler A but doesn't work in bundler B and that doesn't help anyone (maybe with the exception of users of bundler A :P).
> 
> I think it's important to not spread "incorrect" patterns by allowing those in resolution algorithms because it creates a butterfly effect. People often look at how other packages are configured and might copy-paste incorrect things and if we allow those to be loaded successfully then they might not double-check other bundlers etc.
> 
> I think that package authors should absorb this complexity and they should correct their mistakes. The alternative is that users will suffer from misconfigured packages and there are more users than package authors.
> 
> One of the motivating factors here was that the same codesandbox was executing differently when running in codesandbox and when downloaded and executed with webpack. I've tried to replicate the same problem with the codesandbox that you linked to and I couldn't.
> 
> So I went investigating this and I think that webpack might actually allow alias fields even with exports and maybe my original issue had some extra precondition there. I tried to analyze the webpack's algorithm but their code is quite convoluted and it's hard to do that on the spot. I managed to find some relevant tests [here](https://github.com/webpack/enhanced-resolve/blob/90aa4017e8f1fd86991527cb350b74419ea56e91/test/exportsField.js#L2005-L2051)
> 
> I **think** this is what happens:
> 
> 1. the request is first resolved using `exports`
> 2. and only then the alias field is "tried"
> 
> This would match what we can observe in those tests, the first one is using `conditionNames: ["webpack"]` whereas the second one is using `conditionNames: ["node"]`. So even though they request the same path and that path exists in the `package.json#browser`... they don't end up resolving the same file. That's because the `webpack` condition redirects the request to the `/lib2/` directory ([here](https://github.com/webpack/enhanced-resolve/blob/90aa4017e8f1fd86991527cb350b74419ea56e91/test/fixtures/exports-field/node_modules/exports-field/package.json#L12)). So the request path gets changed and that final path is not in the `package.json#browser`.
> 
> When it comes to the second test case - the request path also gets changed (from `/dist/` to `/lib/`) but the new path is present in the `package.json#browser` and thus gets "redirected" once again.
> 
> I think it might be quite hard to replicate this behavior with your "flat" object of `aliases`.

Thanks a lot for your view on this, it helped a lot in deciding on which route to take.

You're right we should follow the spec more closely (still not sure where to find this spec though), I tried to do that in #7404 but that appears to break a lot of things will investigate that further and get that in at some point.

For now I also created #7405 which does still respect the aliasing and shouldn't have any impact on file resolution besides the root pkg.json (like your PR did originally) this way we hopefully can gradually move to a more spec compliant resolver

Andarist:
> You're right we should follow the spec more closely (still not sure where to find this spec though)

Yeaah, this is a little bit of a problem 🤣 The `exports` stuff is nicely documented in node docs (and in webpack docs) but it's not that clear how this interacts with some of the less standard fields, like `package.json#browser`. I'm hesitant to say that every tool actually uses `package.json#browser` when there are `package.json#exports` - but apparently webpack does (just remember that `exports` are still resolved first and only then `browser` is consulted using the resolution's result).

I have seen a lot of quirks related to those algorithms over the years - and I would consider myself an "expert" on that stuff but... it's so hard that I still miss some stuff (like this whole webpack thing discussed here). 

Happy to discuss those things in more detail if you want - I've suffered enough learning those things, so there is no reason why anybody would have to figure this out on their own from scratch.

Andarist:
I also verified with Rollup (and its `@rollup/plugin-node-resolve`, `nodeResolve({ browser: true })`) that the `package.json#browser` field also kicks in there after resolving `package.json#exports`. In Rollup you need to

PR #7382, opened 2023-02-01
fix(dashboard/settings): user invite hangs

olarclara:

Replaces the alert modal with a banner if adding another editor will increase the subscription charges.

In the first render, the modal does not appear and leaves the user hanging. This is hard to reproduce because it goes away after a refresh.

## What is the new behavior?

I couldn't find a way to fix it (but thinking now, maybe forcing a re-render inside the component might help), so I removed the confirmation step and added a banner informing about the charges that will occur when adding a new editor.

cyt2tt-3000 preview csb app_dashboard_settings_workspace=aa86c532-6129-4c06-a8cc-979bfcb72424

I also fixed the form input to prevent performing any action if it was empty.

1. Go to a pro team where you are an admin
2. See the banner
3. Inviting new members performs the action w/o any steps in-between
4. Copying the invite link performs the action w/o any steps in-between

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      

tristandubbeld (approved):
The behaviour of inviting others is really curious, it seems to work very differently in multiple scenarios. Sometimes the modal doesn't show up, sometimes it does. Sometimes it invites a user without confirming the dialog too. I think we need to spend some time to look into the deeper layers when we can. This PR looks good to me though!

packages/app/src/app/pages/Dashboard/Content/routes/Settings/TeamSettings/AdditionalEditorsCharges.tsx
```diff
@@ -0,0 +1,59 @@
...
+        ((subscription.unitPrice || 0) / 100).toFixed(2) +
+        '/' +
+        subscription.billingInterval?.toLowerCase()
+      );
+    }
+
+    if (!prices) return null;
+
+    const workspaceType =
+      (activeTeam === personalWorkspaceId ? 'pro' : 'teamPro') ?? 'pro';
```
tristandubbeld, line 33: Isn't it always `teamPro`? Since personal workspaces can't invite new users.

tristandubbeld (changes requested):
Actually! I just tried to reproduce this a few times on production and it seems that removing the buggy code that closed the "create new" modal was also closing this one. Can you confirm this?

olarclara:
@tristandubbeld thank you for catching this! I'll close and keep an eye out if the issue happens again.

PR #7385, opened 2023-02-01
feat(dashboard/import): github authorization

olarclara:

1. Step A
2. Step B
3. Step C

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      
### DIFF
--- a/packages/app/src/app/components/CreateSandbox/Import/Import.tsx
+++ b/packages/app/src/app/components/CreateSandbox/Import/Import.tsx
@@ -63,7 +63,7 @@ export const Import: React.FC<ImportProps> = ({ onRepoSelect }) => {
   const { hasLogIn, activeTeam } = useAppState();
   const importAndRedirect = useImportAndRedirect();
 
-  const ghAuth = useGitHubAuthorization();
+  const { allowsPublicRepos } = useGitHubAuthorization();
   const { isFree } = useWorkspaceSubscription();
   const { hasMaxPublicRepositories } = useWorkspaceLimits();
 
@@ -153,7 +153,7 @@ export const Import: React.FC<ImportProps> = ({ onRepoSelect }) => {
   const limitImportBasedOnSubscription =
     privateRepoFreeAccountError === url.raw;
 
-  const disableImport = hasMaxPublicRepositories || !ghAuth.allowsPublicRepos;
+  const disableImport = hasMaxPublicRepositories || !allowsPublicRepos;
 
   return (
     <Stack direction="vertical" gap={4}>
@@ -170,7 +170,7 @@ export const Import: React.FC<ImportProps> = ({ onRepoSelect }) => {
         Enter the GitHub repository URL to import
       </Text>
       {hasMaxPublicRepositories ? <MaxPublicRepos /> : null}
-      {!ghAuth.allowsPublicRepos ? <UnauthorizedGitHub /> : null}
+      {!allowsPublicRepos ? <UnauthorizedGitHub /> : null}
       <Element
         {...(disableImport
           ? {
--- a/packages/app/src/app/components/CreateSandbox/Import/Import.tsx
+++ b/packages/app/src/app/components/CreateSandbox/Import/Import.tsx
@@ -12,7 +12,11 @@ import { GithubRepoToImport } from './types';
 import { useGithubRepo } from './useGithubRepo';
 import { useImportAndRedirect } from './useImportAndRedirect';
 import { getOwnerAndRepoFromInput } from './utils';
-import { MaxPublicRepos, PrivateRepoFreeTeam } from './importLimits';
+import {
+  MaxPublicRepos,
+  PrivateRepoFreeTeam,
+  UnauthorizedGitHub,
+} from './importLimits';
 
 const UnauthenticatedImport: React.FC = () => {
   const actions = useActions();
@@ -59,8 +63,7 @@ export const Import: React.FC<ImportProps> = ({ onRepoSelect }) => {
   const { hasLogIn, activeTeam } = useAppState();
   const importAndRedirect = useImportAndRedirect();
 
-  useGitHubAuthorization();
-
+  const ghAuth = useGitHubAuthorization();
   const { isFree } = useWorkspaceSubscription();
   const { hasMaxPublicRepositories } = useWorkspaceLimits();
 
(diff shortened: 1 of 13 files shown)


packages/app/src/app/hooks/useGitHubAuthorization.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  const scopes = data.scopes;
+
+  return {
+    allowsPrivateRepos: scopes.includes('repo'),
+    allowsPublicRepos:
+      scopes.includes('repo') || scopes.includes('public_repo'),
+  };
+};
```
olarclara, line 30: @tristandubbeld do you think it makes sense to have these fields? Or should we aim for a simpler return type?

tristandubbeld: I really like the explicitness @olarclara. I think it's way better than typing it as a string (or number as we do with privacy right now 😬). Do we want to include something like `hasGithubAuthorization` too to maybe check if the user has any authorization at all?

olarclara: @tristandubbeld I spent some time debating about this, at first I was going to return `null` if there was no authorization but I preferred to follow the pattern of `useWorkspaceSubscription` and return sensible defaults. Wdyt?

tristandubbeld: I think that's great!

packages/app/src/app/components/CreateSandbox/Import/Import.tsx
```diff
@@ -58,6 +63,7 @@ export const Import: React.FC<ImportProps> = ({ onRepoSelect }) => {
   const { hasLogIn, activeTeam } = useAppState();
   const importAndRedirect = useImportAndRedirect();
 
+  const ghAuth = useGitHubAuthorization();
```
tristandubbeld, line 66: We can destructure `allowsPublicRepos` here

packages/app/src/app/components/CreateSandbox/Import/Import.tsx
```diff
@@ -162,14 +170,11 @@ export const Import: React.FC<ImportProps> = ({ onRepoSelect }) => {
         Enter the GitHub repository URL to import
       </Text>
       {hasMaxPublicRepositories ? <MaxPublicRepos /> : null}
+      {!ghAuth.allowsPublicRepos ? <UnauthorizedGitHub /> : null}
       <Element
-        {...(hasMaxPublicRepositories
+        {...(disableImport
           ? {
               as: 'div',
-              css: {
```
tristandubbeld, line 169: Now that the `css` is removed from this element and we disable the `input`, do we still need to use `as: 'div'`? I think a form is fine. What do you think?

packages/app/src/app/components/CreateSandbox/Import/importLimits.tsx
```diff
@@ -154,3 +154,23 @@ export const PrivateRepoFreeTeam: React.FC = () => {
...
+  const { isLoadingGithub } = useAppState();
+
+  return (
+    <MessageStripe justify="space-between" variant="warning">
+      Adjust your GitHub App permissions to have full access to your
+      repositories.
+      <MessageStripe.Action
+        loading={isLoadingGithub}
+        // Temporarily ask for private repo access because
+        // public access only returns an API error.
```
tristandubbeld, line 169: We're working on this right 👀?

packages/app/src/app/components/CreateSandbox/Import/importLimits.tsx
```diff
@@ -154,3 +154,23 @@ export const PrivateRepoFreeTeam: React.FC = () => {
...
+        onClick={() => signInGithubClicked('private_repos')}
+      >
+        Review GH permissions
```
tristandubbeld, line 172:
```suggestion
        Review GitHub permissions
```

packages/app/src/app/pages/Dashboard/Content/routes/Repositories/index.tsx
```diff
@@ -85,6 +92,30 @@ export const RepositoriesPage = () => {
         </Element>
       ) : null}
 
+      {!allowsPrivateRepos ? (
+        // && !isGHBannerDismissed
```
tristandubbeld, line 96: We can maybe call this `permissionsBanner` instead of GH. Just a nitpick, not a required change if you disagree! Permissions is a longer word though.

packages/app/src/app/pages/Dashboard/Content/routes/Repositories/index.tsx
```diff
@@ -85,6 +92,30 @@ export const RepositoriesPage = () => {
         </Element>
       ) : null}
 
+      {!allowsPrivateRepos ? (
```
tristandubbeld, line 95: If we're going to use it like this most often (meaning as inverted boolean) we might want to change it to something that explicitly says unauthorised. Just something to keep in mind.

tristandubbeld (commented):
Going to test this properly one more time, but the code looks good so far! Really like the hook 👍

tristandubbeld (approved):
Checked it and works great, except for the stream thingies we know and talked about. Nice job :)

packages/app/src/app/components/CreateSandbox/Import/Import.tsx
olarclara: https://github.com/codesandbox/codesandbox-client/pull/7385/commits/e6b22d0e4e3f5e72d5af2aef96e4e1a06a67d31d

packages/app/src/app/components/CreateSandbox/Import/importLimits.tsx
olarclara: 8f635d7a3796d4fe9a8edfff723ae7e5eb2d9695

packages/app/src/app/components/CreateSandbox/Import/Import.tsx
olarclara: [9347827](https://github.com/codesandbox/codesandbox-client/pull/7385/commits/9347827d661b3c1b5f8798362aa2cee6c9a68c04)

olarclara:
Closing in favor of #7407.

PR #7388, opened 2023-02-02
feat: update provider auth

olarclara:

This PR is needed to unblock further work in: https://github.com/codesandbox/codesandbox-client/pull/7385

- Adjusts the function signature to improve consistency.
- Handles opening `.stream` to test the GitHub flow in development mode.

### DIFF
--- a/packages/app/src/app/overmind/actions.ts
+++ b/packages/app/src/app/overmind/actions.ts
@@ -14,6 +14,7 @@ import { TEAM_ID_LOCAL_STORAGE } from './utils/team';
 import { Context } from '.';
 import { DEFAULT_DASHBOARD_SANDBOXES } from './namespaces/dashboard/state';
 import { FinalizeSignUpOptions } from './effects/api/types';
+import { AuthOptions } from './utils/auth';
 
 export const internal = internalActions;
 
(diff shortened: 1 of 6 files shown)


packages/app/src/app/overmind/utils/auth.ts
```diff
@@ -0,0 +1,8 @@
+export type AuthOptions =
+  | {
+      provider: 'github';
+      useExtraScopes?: true; // Omit the flag if there's no need for extra scopes.
```
tristandubbeld, line 4: What are "extra scopes" referring to, should we explain this somewhere?

packages/app/src/app/overmind/actions.ts
```diff
@@ -304,25 +305,9 @@ export const toggleSignInModal = ({ state }: Context) => {
...
-    });
-    state.signInModalOpen = false;
-    state.cancelOnLogin = null;
-    return;
-  }
-  await actions.internal.signIn({
-    useExtraScopes,
-    provider,
-  });
+  await actions.internal.signIn(options);
```
tristandubbeld, line 310: Amazing improvement!

packages/app/src/app/overmind/internalActions.ts
```diff
@@ -163,33 +161,44 @@ export const authorize = async ({ state, effects }: Context) => {
...
 
   if (provider === 'github') {
-    if (useExtraScopes) {
-      authPath.searchParams.set('scope', 'user:email,repo,workflow');
-    } else {
-      authPath.searchParams.set('scope', 'user:email');
-    }
+    authPath.searchParams.set(
+      'scope',
+      'useExtraScopes' in options ? 'user:email,repo,workflow' : 'user:email'
```
tristandubbeld, line 192: I believe that when `useExtraScopes` is explicitly set to `undefined` this also results in `true` when using `in`:

```js
const a = { b: undefined };
console.log('b' in a); // true
```
I think you changed it so `useExtraScopes` is always omitted (`signInButtonClicked({ provider });`) when we don't need it, but there might be cases where it's not omitted and might be undefined. Should we change this to `options.useExtraScopes` instead to be sure? Wdyt?

packages/app/src/app/overmind/utils/auth.ts
olarclara: Extra scopes is referring to whether we ask for the user's email only or for the workflow and repos too. I'll document this type when I update it to be more flexible since we might want to ask for public repos only.

packages/app/src/app/overmind/internalActions.ts
olarclara: @tristandubbeld nice catch! I fixed it here: [2ab2c57](https://github.com/codesandbox/codesandbox-client/pull/7388/commits/2ab2c57a6a7c80dd726cbf8bd9066c0ee99fe893)

PR #7390, opened 2023-02-02
fix(pro): Make personal team user team admin by default

tristandubbeld:
When creating the checkout URL on the pro page we check if the user is a team admin, but this was only true for team spaces. Changed it to make sure we have an `isAdmin` property unrelated to personal or teams, but just to the user.


packages/app/src/app/hooks/useWorkspaceAuthorization.ts
```diff
@@ -38,6 +45,7 @@ export const useWorkspaceAuthorization = () => {
     isTeamAdmin,
```
olarclara, line 45: Maybe change `isTeamAdmin` to `isTeamSpace && isAdmin`.

olarclara (approved):

PR #7215, opened 2022-12-07
feat: update project requests

alexnm:
Changes:
* Pass `workspace_id` to the delete branch endpoint
* Pass `team` param when querying project branches
* Implement new `importProject` mutation in favor of the old import REST endpoint (closes XTD-413)
* Implement new `deleteProject` mutation for assigned projects (old style projects are still unlinked with a rest endpoint)
* Refactor the URL builders to support workspaceId
* Change branch URLs from the dashboard to include the workspaceId in the query string
* Does not create a branch anymore after import/fork, as seamless branch fixes the problem of the editing capabilities (closes XTE-382)

[Testing scenarios in notion](https://www.notion.so/codesandbox-app/Client-s-updates-ef9ec14daea347d19fb68f858bd04854)

PR #7395, opened 2023-02-03
fix: redirect to default branch opposed to a new draft one

alexnm:
I already solved this in #7215 but this might go to prod faster. Doesn't make sense to create draft branches by default now that we have the seamless branching in production

Need to test the onboarding flow on the editor side


tristandubbeld (approved):
I was wondering where else we were using `v2DraftBranchUrl`, but you updated it to the default branch in all the right spots where it makes sense 👍

olarclara (approved):
Nice!

PR #7405, opened 2023-02-06
fix(sandpack-bundler): Don't use nested pkg.json when exports are present

DeMoorJasper:

We should not use the nested pkg.json's as a reference when the root package.json has an exports field present.

This is based on PR https://github.com/codesandbox/codesandbox-client/pull/7380 this is probably not as spec compliant but at least it doesn't break anything.

We need to investigate this further with a more advanced test suite soonish


packages/sandpack-core/src/resolver/utils/pkg-json.ts
```diff
@@ -68,5 +50,23 @@ export function processPackageJSON(
     }
   }
 
+  // load exports if it's not the root pkg.json
+  if (content.exports && pkgRoot !== '/') {
```
Andarist, line 54: Note that likely you only need to still process alias fields - but main fields don't have to be processed.

DeMoorJasper: You're right good catch, updated it accordingly

CompuIves (approved):
Looks good!

drcmda:
@CompuIves @Andarist @DeMoorJasper i think this one broke it once again. all our boxes are down, see: https://codesandbox.io/s/basic-demo-rrppl0y8l4

DeMoorJasper:
> @CompuIves @Andarist @DeMoorJasper i think this one broke it once again. all our boxes are down, see: https://codesandbox.io/s/basic-demo-rrppl0y8l4 i don't know what the pr is trying to do, but if it would require we change something we wouldn't be able to survive it, we have too many sandboxes, poimandres with all its libraries is practically leaning on csb.

Sorry about this, will revert this PR for now and add some test cases to ensure this does not happen next time

The aim of this PR is to follow the exports spec better, supporting the new esm packages

Andarist:
Do we know what broke this time? Which package was it and what could be the reason there?

DeMoorJasper:
@Andarist yes, I've added some tests to this PR https://github.com/codesandbox/codesandbox-client/pull/7412 with the resolutions that broke

PR #7409, opened 2023-02-07
fix(pro): Undefined user crash and update CTA when eligible for trial

tristandubbeld:
Found a bug that when visiting `/pro` directly made the application crash because `user` was still `undefined`. Also changed the CTA text when the user is eligible for trial to start a trial rather than "proceed to checkout".


filipelima18 (approved):
Tested & looks good to me!

CompuIves (approved):

olarclara (approved):

PR #7412, opened 2023-02-07
fix: root exports

DeMoorJasper:

Only use root pkg.json for exports third try at this

TODO:
- [x] Add test for its-fine
- [ ] Add test for fflate resolving `lib/node-worker.cjs` to `lib/worker.cjs` in browser (https://unpkg.com/browse/fflate@0.6.10/package.json)


packages/sandpack-core/src/resolver/fixture/node_modules/its-fine/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "its-fine",
```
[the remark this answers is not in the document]
Andarist: i know this is a real package but this makes it even funnier here 😅

packages/sandpack-core/src/resolver/fixture/node_modules/its-fine/package.json
```diff
@@ -0,0 +1,57 @@
...
+  "type": "module",
+  "types": "./out/index.d.ts",
+  "main": "./out/index.cjs",
+  "module": "./out/index.js",
+  "react-native": "./out/index.js",
+  "exports": {
+    "types": "./out/index.d.ts",
+    "require": "./out/index.cjs",
+    "import": "./out/index.js"
+  },
```
Andarist, line 32: 1. what is requested for this package?
2. what does it resolve to right now?
3. what does it resolve to with this patch?

packages/sandpack-core/src/resolver/fixture/node_modules/its-fine/package.json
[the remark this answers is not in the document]
DeMoorJasper: Haha, best package name ever, at least it makes it a bit fun when a bug report comes in :D 

its-fine

packages/sandpack-core/src/resolver/fixture/node_modules/its-fine/package.json
DeMoorJasper: Currently it resolves to nothing, gonna debug it now

DeMoorJasper: Seems like the resolver did not handle the exports correctly if it's a root exports, omitting the `.` was not possible.

Thanks for pointing to the webpack docs btw, really nice reference for this

Andarist: Oh, it makes sense that it didn't work for `exports: Record<Condition, FilePath>`. Nice catch! I assume that a regular shorthand, like this `exports: './file.js'`, works OK, right?

DeMoorJasper:
Tested this on various sandboxes, still waiting for some more data on most viewed sandboxes, will try those as well and than merge this when that all looks good

packages/sandpack-core/src/resolver/resolver.test.ts
```diff
@@ -503,6 +503,61 @@ describe('resolve', () => {
...
+      const resolved = resolveSync('fflate', {
+        filename: '/node_modules/its-fine/out/index.cjs',
+        extensions: ['.ts', '.tsx', '.js', '.jsx'],
+        isFile,
+        readFile,
+      });
+      expect(resolved).toBe('/node_modules/fflate/lib/index.cjs');
+    });
+
+    // exports would return invalid value, so aliasing has to happen afterwards to fix this
```
Andarist, line 551: This comment doesn't sound completely right but maybe I'm misreading this. This test case tests if an internal request (within a package) is able to be resolved through the alias field (`package.json#browser`). In such a situation - the `exports` shouldn't even be examined (when the target specifier is a relative path).

packages/sandpack-core/src/resolver/resolver.test.ts
```diff
@@ -503,6 +503,61 @@ describe('resolve', () => {
...
+      const resolved = resolveSync('its-fine', {
+        filename: '/foo.js',
+        extensions: ['.ts', '.tsx', '.js', '.jsx'],
+        isFile,
+        readFile,
+      });
+      expect(resolved).toBe('/node_modules/its-fine/out/index.cjs');
+    });
+
+    it('resolve fflate correctly', () => {
```
Andarist, line 541: is this testing if a package can use a "self request"? perhaps you didn't want to use an internal `filename` here?

DeMoorJasper: Not sure I follow what you mean with internal filename, but this test has `/node_modules/its-fine/out/index.cjs` as filename intentionally as that's the real use-case that failed here

packages/sandpack-core/src/resolver/resolver.test.ts
DeMoorJasper: I'll rephrase it a bit, because it is as you mention it should just be able to respect browser field, has nothing to do with exports.

Currently it also still behaves incorrectly, because now it will read exports even when doing relative requests

DeMoorJasper: But I'll make this a two step approach in a follow-up.

If it's a node_module request => use/read exports and main fields
Afterwards and relative paths => read/use aliases

Andarist: > Afterwards and relative paths => read/use aliases

just note here that `import 'pkg'` from within `pkg` should still resolve `exports`. But that might just work out of the box - it most likely only depends on what `node_modules` paths you are looking in for "external requests"

packages/sandpack-core/src/resolver/fixture/node_modules/its-fine/package.json
DeMoorJasper: Yes the regular shorthand worked fine and was also tested, but didn't know the root `Record<Condition, FilePath>` was a thing, so didn't implement it

packages/sandpack-core/src/resolver/resolver.test.ts
Andarist: > Not sure I follow what you mean with internal filename, but this test has /node_modules/its-fine/out/index.cjs as filename intentionally as that's the real use-case that failed here

This is kinda what I meant by the internal request - although I would call this one specifically a "self request".

1. external request: from `./src/index.js` request `fflate`
2. internal request: from `./node_modules/fflate/dist/index.js` request `./node_modules/fflate/dist/other.js`
3. self request: `./node_modules/fflate/dist/index.js` request `flatte`

Those are not official names though :P

packages/sandpack-core/src/resolver/resolver.test.ts
DeMoorJasper: Yes, I think we need to add some flag to options like `relativeRequest: boolean` that is defined based on the first character of the original import request being anything but a letter or `@` symbol, pointing to it being an alias/relative import

But that requires quite a bit of refactoring as this also means splitting our pkg.json parsing in two and adding a bunch more tests, will do that in a follow-up, don't wanna make big steps here.

packages/sandpack-core/src/resolver/resolver.test.ts
Andarist: "resolve fflate correctly" sounds like it just tries to request `flatte` from the consuming project - but it looks to me like maybe this is testing the whole "self request" thing so maybe the test title could be more specific about that

packages/sandpack-core/src/resolver/resolver.test.ts
Andarist: It also reminds me of something... you might not handle `package.json#imports` at all. Those are for "internal" requests, the structure of this field is basically the same as `package.json#exports` but all specifiers have to start with `#`.

packages/sandpack-core/src/resolver/resolver.test.ts
DeMoorJasper: It's from its-fine to fflate so it is not requesting itself but a different module, but yes maybe the naming could be better, naming is hard

Andarist: ah, i totally missed that it's requesting from one package to another - makes sense. Then I guess that the self-request thing could be added as an extra test case.

packages/sandpack-core/src/resolver/resolver.test.ts
DeMoorJasper: The standalone resolver does support this and hoping to migrate over to using that one later, but it's really dangerous to introduce any big change in this repo

https://github.com/codesandbox/sandpack-resolver/blob/main/src/utils/imports.ts

PR #7415, opened 2023-02-08
fix(pricing): Reflect true pricing

tristandubbeld:
Change pricing to reflect the actual pricing in Stripe.

image


CompuIves:
Nice followup! Missed this, thanks  a lot!

CompuIves (approved):

PR #7425, opened 2023-02-12
feat(editor): Add Docker navigation item to convert to cloud sandbox

tristandubbeld:
This PR adds a Docker navigation item to the sidebar of the v1 editor. The intent is to convert sandboxes for users that are interested into cloud sandboxes to use docker.

### Linear

Closes OPP-17

### Screenshot

image


packages/app/src/app/overmind/utils/items.ts
```diff
@@ -107,6 +112,7 @@ export function getDisabledItems(store: any): INavigationItem[] {
   return [];
 }
 
+//
```
olarclara, line 115: Is there a comment that should be added here?

olarclara (approved):
🐳

packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/Docker/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+        </Text>
+        <Stack align="center" gap={4}>
+          <Button
+            onClick={perform}
+            loading={loading}
+            variant="primary"
+            autoWidth
+          >
+            Yes, {canConvert ? 'convert' : 'fork and convert'}
+          </Button>
```
olarclara, line 37: I noticed that there's no error handling if converting fails. Took a look and seems that the hook does not catch errors so I created a task in Linear: https://linear.app/codesandbox/issue/XTD-698/handle-errors-when-upgrading-from-v1-to-v2. The priority is medium but I believe we can increase it next week.

tristandubbeld: Really good one!

filipelima18 (approved):
I made a small copy update and everything else looks good to me.
@tristandubbeld - can you let me know the name of the Amplitude event that will be used to track when users click "Yes, convert"?

tristandubbeld:
@filipelima18 that would be: `Editor - Docker menu Convert to Cloud Sandbox`!

PR #7248, opened 2022-12-19
feat(XTD-348): Stripe success param

tristandubbeld:
Shows notification when a subscription through Stripe has been successful. Based on if the subscription webhook from Stripe came through fast enough we well show the user more information and an action.

### Testing

1. Open the dashboard with a team without trial
2. Start a trial
3. When redirected by Stripe, it should show a notification on the dashboard

Based on if your internet speed and device are faster than the webhook, you should see different notifications. When the webhook didn't come yet in you should see a refresh action in the notifications. To mock the refresh action notification, you can also open the dashboard for a team without trial and add `&stripe=success` to the url. To mock the notification without action, you can add `&stripe=success` to the url for a team that already has a subscription.

### Screenshots

**When webhook didn't come through yet**
image

**When webhook came through and subscription is available**
image

PR #7430, opened 2023-02-15
feat(dashboard): support opening new team modal from the url

olarclara:

Makes the new team modal appear when opening `dashboard?create_team=true`.

## What is the new behavior?

https://www.loom.com/share/eac98b5064794efdb8f99f22fd7008a1

1. Go to `/dashboard?create_team=true`.
2. The URL will change to `/dashboard?workspace={id}&create_team=true`
3. The create team modal will open automatically.

## Checklist

- [ ] Documentation
- [x] Testing 
- [ ] Ready to be merged
      
### DIFF
--- a/packages/common/src/utils/url-generator/dashboard.test.ts
+++ b/packages/common/src/utils/url-generator/dashboard.test.ts
@@ -1,0 +1,41 @@
+import * as dashboard from './dashboard';
+
+const teamId = 'foo-bar-123';
+
+describe('dashboard url generator', () => {
+  test('sandboxes', () => {
+    expect(dashboard.sandboxes('/path', teamId)).toBe(
+      '/dashboard/sandboxes/path?workspace=foo-bar-123'
+    );
+  });
+
+  test('recent', () => {
+    expect(dashboard.recent(teamId)).toBe(
+      '/dashboard/recent?workspace=foo-bar-123'
+    );
+  });
+
+  test('recent with extra params', () => {
+    expect(dashboard.recent(teamId, { create_team: 'true' })).toBe(
+      '/dashboard/recent?workspace=foo-bar-123&create_team=true'
+    );
+  });
+
+  test('search', () => {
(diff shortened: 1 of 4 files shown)


packages/common/src/utils/url-generator/dashboard.test.ts
```diff
@@ -0,0 +1,41 @@
+import * as dashboard from './dashboard';
+
+const teamId = 'foo-bar-123';
+
+describe('dashboard url generator', () => {
```
tristandubbeld, line 5: Nice!!

packages/common/src/utils/url-generator/dashboard.ts
```diff
@@ -51,8 +51,29 @@ export const syncedSandboxes = (teamId?: string | null) =>
 export const templates = (teamId?: string | null) =>
   appendTeamIdQueryParam(`${DASHBOARD_URL_PREFIX}/templates`, teamId);
 
-export const recent = (teamId?: string | null) =>
-  appendTeamIdQueryParam(`${DASHBOARD_URL_PREFIX}/recent`, teamId);
+export const recent = (
```
tristandubbeld, line 54: Not sure if it makes things simpler but it might, have you considered using `new URL` and appending search params like that? I don't think we need checks for `?` or `&` in that case. [We do this for the stripe success param](https://github.com/codesandbox/codesandbox-client/pull/7248/files#diff-c18c2f44efab0a0400faf6f5240fbc37d3d69e50ecea267767e8d15e59c2070bR21-R31). This isn't a blocker but I was curious!

tristandubbeld (approved):

packages/common/src/utils/url-generator/dashboard.ts
olarclara: I don't like how it's implemented either but because we only have the path, the URL constructor fails because it expects a full URL. :/

PR #7433, opened 2023-02-16
fix: pass search params to redirect after sign-in

olarclara:
The redirect to the dashboard on sign-in removed the search params from the URL thus not prompting the new team modal to open, this PR fixes it.

https://www.loom.com/share/502cf0b0bf3c4233add5e38a0b4596b6


tristandubbeld (approved):
Nice!

PR #7438, opened 2023-02-17
fix(import): extract owner and repo from input

olarclara:
Updates the `getOwnerAndRepoFromInput` to handle more cases. I tried getting rid of the most complex regular expressions because they can make maintainability harder, instead, it checks some cases separately.

- [ ] Import a repo with an URL ending in a trailing slash, eg (`https://github.com/codesandbox/codesandbox-client/`) 
- [ ] Import a repo with a dot (`.`) in the repo name, eg (`https://github.com/tailwindlabs/tailwindcss.com`)
- [ ] Import a repo in the format `owner/repo`


tristandubbeld (approved):
Love the readability!

PR #7443, opened 2023-02-21
feat(live): add restricted go live feature flag and view message

tristandubbeld:
Closes XTD-711

Adds the recently added `restrictions` object with `liveSessionsRestricted` flag to the types, and use it to view a message when going live is restricted.

### Testing

Open the team "free-team" on stream. This team should have the feature flag enabled, and go live restricted. Go to any sandbox and press go live, either as editor or admin. A message should pop up.

### Video
As an admin.

https://user-images.githubusercontent.com/7533849/220380810-985670dc-05bc-440a-bbd7-fb5750a6508e.mov

### Screenshot
As non-admin.

image

### DIFF
--- a/packages/app/src/app/pages/common/Modals/LiveSessionRestricted/index.tsx
+++ b/packages/app/src/app/pages/common/Modals/LiveSessionRestricted/index.tsx
@@ -1,0 +1,104 @@
+import React from 'react';
+import { Button, Stack, Text } from '@codesandbox/components';
+import { sandboxUrl } from '@codesandbox/common/lib/utils/url-generator';
+import track from '@codesandbox/common/lib/utils/analytics';
+
+import { useWorkspaceAuthorization } from 'app/hooks/useWorkspaceAuthorization';
+import { useCreateCheckout } from 'app/hooks';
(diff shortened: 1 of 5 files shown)


packages/app/src/app/pages/common/Modals/LiveSessionRestricted/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+              onClick={() => {
+                track('Live Session - learn more clicked', {
+                  codesandbox: 'V1',
+                  event_source: 'UI',
+                });
+              }}
+              autoWidth
+            >
+              Learn more
+            </Button>
```
olarclara, line 103: Is this the same as the button in line 52 (except for the variant)?

olarclara (approved):
🔴

packages/app/src/app/pages/common/Modals/LiveSessionRestricted/index.tsx
tristandubbeld: I think you mean line `34` and yeah it's the same except for the variant!

PR #7446, opened 2023-02-22
fix(sentry): Ensure continue action is not called when forking is not allowed

tristandubbeld:
Should fix Sentry issue 3587048851.

### The culprit

The error had to do with an unauthorized API call (returned a `401`). The API was called when an anonymous user tried to save a file which resulted in CSB trying to stop it from being forked. This was done successfully, but the "save" action wasn't properly canceled after that. This PR changes that behaviour and throws an error rather than returning nothing to end up in a `catch` block - which then properly stops the save action from happening.

I've also made sure we don't trigger an "unknown" editor error while also showing the fork alert. The fork alert with sign in button should be enough.

### Before

https://user-images.githubusercontent.com/7533849/220662233-961e6e2c-7755-48d1-ad0b-3f460ebd8920.mov

### After

https://user-images.githubusercontent.com/7533849/220662297-c239a1fc-94a2-4648-9156-7b0d9360b538.mov

### DIFF
--- a/packages/app/src/app/overmind/factories.ts
+++ b/packages/app/src/app/overmind/factories.ts
@@ -128,14 +128,10 @@ export const withOwnedSandbox = <I>(
           sandboxId: sandbox.id,
         });
       } catch (e) {
-        const isAnonServerForkMessage = e.message.includes(
-          'server container as an anonymous user'
-        );
-
         // If we know the error is caused by an anonymous user trying to save a
         // file and fork a server sandbox, we don't want to show the vscode error
         // with the cancelAction and we don't want to continue either.
-        if (isAnonServerForkMessage) {
+        if (e.message === 'ERR_ANON_SSE_FORK') {
           return () => {};
         }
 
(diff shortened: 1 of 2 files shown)


packages/app/src/app/overmind/factories.ts
```diff
@@ -127,6 +128,17 @@ export const withOwnedSandbox = <I>(
           sandboxId: sandbox.id,
         });
       } catch (e) {
+        const isAnonServerForkMessage = e.message.includes(
```
DeMoorJasper, line 131: nitpick: Can we use error codes for this?

DeMoorJasper (approved):
Looks really good, amazing to see the v1 editor getting some polish :D

packages/app/src/app/overmind/factories.ts
tristandubbeld: Yeah I was thinking the same thing. I remember this was done somewhere else so I went with it, but since we're not using the message at all we can definitely use error codes. Do we have a way to generate these somehow or can I just come up with one?

PR #7451, opened 2023-02-24
feat(pricing): Use dynamic pricing fetched from stripe on pro page

tristandubbeld:
Made the pricing on the Pro page dynamic so it updates automatically when we update it.

### Personal screenshot

image

### Team screenshot

image

### DIFF
--- a/packages/app/src/app/overmind/effects/api/apiFactory.ts
+++ b/packages/app/src/app/overmind/effects/api/apiFactory.ts
@@ -4,10 +4,18 @@ import { camelizeKeys, decamelizeKeys } from 'humps';
 
 const API_ROOT = '/api';
 
-// If the path starts with `/beta`, do not append
-// `/v1` to the api root url.
-const getBaseApi = (path: string) =>
-  path.startsWith('/beta') ? API_ROOT : `${API_ROOT}/v1`;
+/**
+ * If the path starts with `/beta`, do not append `/v1` to the api root
+ * url. Alternatively we have the useRoot param for when we just want to
+ * use the root path.
+ */
+const getBaseApi = (path: string, useRoot: boolean = false) => {
+  if (useRoot) {
+    return API_ROOT;
+  }
+
+  return path.startsWith('/beta') ? API_ROOT : `${API_ROOT}/v1`;
+};
 
 export type ApiError = AxiosError<
   { errors: string[] } | { error: string } | any
@@ -22,7 +30,17 @@ type Options = {
 };
 
 export type Api = {
-  get<T>(path: string, params?: Params, options?: Options): Promise<T>;
+  get<T>(
+    path: string,
+    params?: Params,
+    options?: Options,
+    /**
+     * Our API has three different base paths: /beta, /v1 and a path
+     * without one of those. We can use the useRoot parameter to use the
+     * latter.
+     */
+    useRoot?: boolean
+  ): Promise<T>;
   post<T>(path: string, body: any, options?: Options): Promise<T>;
   patch<T>(path: string, body: any, options?: Options): Promise<T>;
   put<T>(path: string, body: any, options?: Options): Promise<T>;
@@ -45,9 +63,9 @@ export default (config: ApiConfig) => {
       : {}),
   });
   const api: Api = {
-    get(path, params, options) {
+    get(path, params, options, useRoot) {
       return axios
-        .get(getBaseApi(path) + path, {
+        .get(getBaseApi(path, useRoot) + path, {
           params,
           headers: createHeaders(config.provideJwtToken),
         })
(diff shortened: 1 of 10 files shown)


packages/app/src/app/overmind/effects/api/apiFactory.ts
```diff
@@ -4,10 +4,18 @@ import { camelizeKeys, decamelizeKeys } from 'humps';
...
-// If the path starts with `/beta`, do not append
-// `/v1` to the api root url.
-const getBaseApi = (path: string) =>
-  path.startsWith('/beta') ? API_ROOT : `${API_ROOT}/v1`;
+/**
+ * If the path starts with `/beta`, do not append `/v1` to the api root
+ * url. Alternatively we have the useRoot param for when we just want to
+ * use the root path.
+ */
+const getBaseApi = (path: string, useRoot: boolean = false) => {
```
tristandubbeld, line 12: Adding `useRoot` was the simplest change to make sure we were able to reuse most of the `api` and `apiFactory` logic, without having to write custom fetch methods or update the factory in a big way.

packages/app/src/app/overmind/namespaces/pro/actions.ts
```diff
@@ -3,17 +3,31 @@ import { withLoadApp } from 'app/overmind/factories';
...
+    state.pro.prices = await effects.api.getPrices();
+  } catch (e) {
+    // e
+  }
+};
+
 export const pageMounted = withLoadApp(async ({ effects, state, actions }) => {
-  state.pro.prices = await effects.api.prices();
+  // We have to call the api effects directly rather than using an action because
+  // for some reason actions don't work.
```
tristandubbeld, line 26: I think I can move the functions `getLegacyPrices` and `getPrices` above, but I'm not 100% sure if overmind does anything else with them. I don't think so.

packages/app/src/app/pages/Pro/Upgrade.tsx
```diff
@@ -242,10 +258,9 @@ export const ProUpgrade = () => {
               }
             >
               <Stack gap={1} direction="vertical" css={{ flexGrow: 1 }}>
-                <Text aria-hidden size={32} weight="400">
+                <Text size={32} weight="400">
```
tristandubbeld, line 261: I checked with the screenreader again and we don't need to aria hide `$number` and have a written, visually hidden alternative. This is only the case when we use cents and want to read out the price in a particular way. This is also true for the other instances where I removed `aria-hidden` and `VisuallyHidden`.

packages/app/src/app/pages/common/Modals/MemberPaymentConfirmation/index.tsx
```diff
@@ -12,7 +12,7 @@ export const MemberPaymentConfirmation: React.FC<{ title: string }> = ({
   const {
     activeTeamInfo,
     personalWorkspaceId,
-    pro: { prices },
+    pro: { legacyPrices },
```
olarclara, line 15: Can't we use the `prices` value here?

tristandubbeld: Yes we should, overlooked that one!

tristandubbeld: I've updated it in [2010fa8](https://github.com/codesandbox/codesandbox-client/pull/7451/commits/2010fa88cd5c18cfe343e4a4ba38bc27074b5452)! I've also updated the pro page to use the existing `formatCurrency` function instead of a new function.

packages/app/src/app/pages/Pro/Upgrade.tsx
```diff
@@ -260,13 +260,22 @@ export const ProUpgrade = () => {
                 isHighlighted
               >
                 <Stack gap={1} direction="vertical">
-                  <Text aria-hidden size={32} weight="500">
-                    $12
+                  <Text size={32} weight="500">
+                    {formatCurrency({
+                      currency: 'USD',
+                      amount: pro?.prices?.individual.year.usd / 12,
```
olarclara, line 266: `formatCurrency` expects `amount` to be a number, but if `pro?.prices?.individual.year.usd` evaluates to `undefined`, it'll receive `NaN` and possibly error.

```suggestion
                      amount: (pro?.prices?.individual.year.usd || 0) / 12,
```

olarclara (approved):
💰

packages/app/src/app/pages/Pro/Upgrade.tsx
tristandubbeld: Ah yes I should fix this. We don't want to show $0 though I think, so I'm making undefined an option and return null instead. It shouldn't ever be falsy unless it's still loading the prices from somewhere.

PR #7453, opened 2023-02-24
fix(vercel): Signing into vercel sometimes throws an error but still succeeds

tristandubbeld:
Turns out that the error is actually thrown because we weren't able to fetch the user from Vercel, but we are still able to deploy to Vercel via the ingegration. Updated the messages to reflect this.

### DIFF
--- a/packages/app/src/app/overmind/actions.ts
+++ b/packages/app/src/app/overmind/actions.ts
@@ -346,6 +346,11 @@ export const signInVercelClicked = async ({
   actions,
 }: Context) => {
   state.isLoadingVercel = true;
+
+  /**
+   * We're opening a browser popup here with the /auth/zeit page but then do a server
+   * side redirect to the Vercel sign in page. This only works on production, not locally.
+   */
   const popup = browser.openPopup('/auth/zeit', 'sign in');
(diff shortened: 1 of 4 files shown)


packages/app/src/app/overmind/actions.ts
```diff
@@ -354,10 +359,22 @@ export const signInVercelClicked = async ({
...
+    try {
       await actions.deployment.internal.getVercelUserDetails();
+
+      // Not sure if we ever reach the catch clause below because the error has already
+      // been caught in getVercelUserDetails.
     } catch (error) {
       actions.internal.handleError({
-        message: 'Could not authorize with Vercel',
+        message:
+          'We were not able to fetch your Vercel user details. You should still be able to deploy to Vercel, please try again if needed.',
```
olarclara, line 377: Could we add this message to a constant?

olarclara (approved):

packages/app/src/app/overmind/actions.ts
tristandubbeld: Yeah that's not a bad idea. I spent some time looking but I can't really find a good file to add it to right now, and I don't want to create a new file for just this string. Do we want a task to extract more consts like these?

PR #7454, opened 2023-02-24
feat(live): Replace privacy settings with live sessions pro perk in banner

tristandubbeld:
image

Closes XTD-721


olarclara (approved):
🔴

PR #7461, opened 2023-02-27
feat(import): show in which teams the repository has been imported

olarclara:

Adds a notice that the repository has already been imported into another team(s).

## What is the new behavior?

https://user-images.githubusercontent.com/24959348/222799776-77ed00ab-ac07-434e-88ec-b4ae92004dd6.mov

1. Open the create new modal
2. Go to the import tab
3. Type/paste a repo that exists in one of your workspaces (eg. test-sandbox)
4. A notice should appear that this repo has been imported into (at least one) workspace already

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      

tristandubbeld:
The message is slightly delayed as we need to fetch the repository first. Do we want to disable importing before the message shows up?

olarclara:
@tristandubbeld that's an excellent question! I reached out to @christianalfoni in [Slack](https://codesandboxapp.slack.com/archives/C040U0806P2/p1677511658307119) with a similar thing in mind:

> My worry about querying as the user types is that the time it takes to perform the request if that repo has already been imported is longer than the time to press enter or click "import" and that will make the notice useless.

From the ticket:

>  It should not prevent the user from importing, just a friendly "Btw…" 

So we could have a loading state to let users know that we are fetching info about the repo they are trying to import, I think we need @zehfernandes input here.

tristandubbeld (approved):
🙌

PR #7462, opened 2023-02-27
chore(vercel): Rename zeit to vercel

tristandubbeld:
Renaming all occurrences of `zeit` to `vercel`. However, `zeit` does still come from one of our API endpoints so in order to map the value to `vercel` I've added a small helper function. This ensures that we need less changes when we've renamed `zeit` to `vercel` coming from the API.

I wasn't able to rename the `zeit` route in the frontend to `vercel` because the server returns "invalid `redirect_uri`".


tristandubbeld:
Drafted this PR for a bit to wait a little longer until all the changes on the server side have been deployed.

packages/app/src/app/pages/index.tsx
```diff
@@ -220,7 +220,7 @@ const RoutesComponent: React.FC = () => {
             <Route path="/cli/login" component={CLI} />
             <Route path="/client/login" component={MobileAuth} />
             <Route path="/vscode/login" component={VSCodeAuth} />
-            <Route path="/auth/zeit" component={VercelSignIn} />
+            <Route path="/auth/vercel" component={VercelSignIn} />
```
olarclara, line 223: Should we add a redirect from the old route to `/vercel` too?

packages/app/src/app/overmind/effects/api/index.ts
```diff
@@ -538,8 +538,9 @@ export default {
       avatar,
     });
   },
-  createVercelIntegration(code: string): Promise<CurrentUser> {
-    return api.post(`/users/current_user/integrations/zeit`, {
+  // change here
```
olarclara, line 541: Can you add a comment on what needs to be changed?

packages/app/src/app/pages/index.tsx
tristandubbeld: We're doing server side redirects for these routes, so it will be automatically redirected. But yeah better safe than sorry. I can add one!

packages/app/src/app/overmind/effects/api/index.ts
tristandubbeld: Oops, old comment. Thanks for pointing this out!

PR #7465, opened 2023-02-28
feat(import): update suggested private repository action when on a free tier

tristandubbeld:
Closes XTD-731

https://user-images.githubusercontent.com/7533849/221875497-2f2471b8-4fa1-4db1-91cd-1834af7f638e.mov

### DIFF
--- a/packages/app/src/app/components/CreateSandbox/Import/SuggestedRepositories.tsx
+++ b/packages/app/src/app/components/CreateSandbox/Import/SuggestedRepositories.tsx
@@ -33,7 +33,7 @@ export const SuggestedRepositories = () => {
   const { modals } = useActions();
   const { restrictsPrivateRepos } = useGitHuPermissions();
   const { isTeamSpace } = useWorkspaceAuthorization();
-  const { isFree } = useWorkspaceSubscription();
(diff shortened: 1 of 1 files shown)


packages/app/src/app/components/CreateSandbox/Import/SuggestedRepositories.tsx
```diff
@@ -151,7 +162,40 @@ export const SuggestedRepositories = () => {
...
-                    <StyledIndicator aria-hidden>Import</StyledIndicator>
+                    {isFree && repo.private ? (
+                      <StyledIndicator>
+                        <InteractiveOverlay.Item>
+                          <Link
+                            as={RouterLink}
+                            to="/pro"
+                            onClick={() => {
+                              modals.newSandboxModal.close();
+                            }}
```
olarclara, line 173: Should we track this action and/or add a source to `/pro`?

olarclara (approved):

packages/app/src/app/components/CreateSandbox/Import/SuggestedRepositories.tsx
tristandubbeld: Yep, I think we should. I'll add it :)

PR #7466, opened 2023-02-28
chore: remove homepage/update node to 16

DeMoorJasper:
Remove old homepage gatsby project and update to node 16

Seems to be failing on some polyfills in overmind, if anyone's feeling adventurous feel free to have a look


DeMoorJasper:
@tristandubbeld seems like I fixed the build, was just some more unused polyfills. Appears like core-js was a dependency of homepage and not of app. Seemed better to just remove those polyfills as we no longer even target IE11 anyway. (doubt csb even still worked in IE11)

Ran yarn start and that worked really well, would be nice if you could have a look as well, to make sure everything still works as expected.

Next up: moving webpack to a newer version so we can finally use some modern syntax

DeMoorJasper:
CircleCI now also works 🎉 

Just needs to double check testing locally and than get the integration test dockerfile to node 16

tristandubbeld:
This typecheck is currently failing, but has proven to be flaky in the past and should be resolved in a second run:
```
lerna ERR! npm run typecheck exited 2 in 'app'
lerna ERR! npm run typecheck stdout:

> app@0.1.1 typecheck
> tsc --noEmit -p tsconfig.check.json && tsc --noEmit -p tsconfig.strictNullChecks.json

src/app/utils/executor-manager.ts(9,24): error TS2307: Cannot find module '@codesandbox/executors/dist/executor' or its corresponding type declarations.
```

@DeMoorJasper do you have any ideas on how we can improve this executor package to make sure it'll be built (that's my guess with what's wrong)?

tristandubbeld:
> Proper manual check if everything still works

So far so good, dashboard seems to work, signing in seems to work and React and Vanilla sandboxes work too. Not seeing any weird issues but I'll keep clicking around for a bit longer.

DeMoorJasper:
> @DeMoorJasper do you have any ideas on how we can improve this executor package to make sure it'll be built (that's my guess with what's wrong)?

Not sure, from the logs it seems like it built successfully. A good starting point is probably migrating that package to native tsc, or at least updating tsdx to the latest version and perhaps also updating lerna

tristandubbeld (approved):
I've tested multiple flows again, inside a CodeSandbox preview and locally. I was also able to run everything in CodeSandbox and I don't think I've ever had a branch restart that smooth. 

Let's ship it 😎!

PR #7480, opened 2023-03-02
fix(import): Authorizing GitHub organizations via Account picker

tristandubbeld:
Fixes XTD-755

Instead of using the "grant access to private repos" function we redirect to the GitHub OAuth application page. This ensures users will always be able to grant access to organizations, no matter if access is granted to private repos for other orgs already.


olarclara (approved):
✅

PR #7483, opened 2023-03-03
feat(import): Suggest imports in create team modal

tristandubbeld:
Closes XTD-623

### Testing

- [ ] Create a new team as a user that has already granted access to GitHub via the app. You should be able to see suggested repositories based on the authorization (just public or both public and private repos).
- [ ] Create a new team as a user that hasn't granted access to GitHub yet. You should be able to grant access via the UI in the create team modal.

### Videos

In the following videos we skip to the suggested imports step immediately.

No authorization:

https://user-images.githubusercontent.com/7533849/223712703-9f59a5a9-d03a-4fcd-b66e-a14db9dd5e95.mov

Public authorization:

https://user-images.githubusercontent.com/7533849/223712741-c2fb9a8b-a63b-44fc-88f8-9152b71fb2f0.mov

Full authorization:

https://user-images.githubusercontent.com/7533849/223712730-78039655-56fc-406b-89e5-e733f15e24aa.mov

#### Full flow

https://user-images.githubusercontent.com/7533849/223713455-ac64f4c1-fb51-44cb-91af-9002567ee2d2.mov


packages/app/src/app/overmind/namespaces/dashboard/actions.ts
```diff
@@ -2284,6 +2284,30 @@ export const importGitHubRepository = async (
   }
 };
 
+export const importGitHubRepositoryWithoutRedirect = async (
+  { state, effects }: Context,
+  { owner, name }: { owner: string; name: string }
```
olarclara, line 2289: Could you change the parameters of `importGitHubRepository` in line 2255? This way, we don't duplicate the function, and defaulting it to `true` does not require any changes to other usages of that action.

```suggestion
  { owner, name, redirect = true }: { owner: string; name: string; redirect?: boolean }
```

packages/app/src/app/pages/Dashboard/Components/NewTeamModal/index.tsx
```diff
@@ -51,7 +53,6 @@ const NewTeam: React.FC<NewTeamProps> = ({ step, hasNextStep, onClose }) => {
 
   return (
     <>
-      {' '}
       <Element padding={6}>
```
olarclara, line 56: Any chances we can set a height in this modal to prevent it from jumping around if there are no suggestions?

olarclara (approved):
I have suggestions for improvement but the functionality works well! Thank you for the detailed screen recordings, they are helpful. :)

packages/app/src/app/overmind/namespaces/dashboard/actions.ts
tristandubbeld: Great suggestion! I think I split it first because I wanted to return something else, but I can do it by using the redirect bool too.

tristandubbeld:
@olarclara @lenasmayor for test purposes I'm only showing one step which makes the modal close after importing, but usually it will go to the next step. Here's what it looks like right now:

https://user-images.githubusercontent.com/7533849/224040841-8e7e0196-9df4-4c53-9661-e87b161e0baa.mov

olarclara (approved):
💯

PR #7484, opened 2023-03-03
chore: Remove ie11 from babel config

DeMoorJasper:

Remove ie11 from babel config

### DIFF
--- a/packages/app/config/webpack.prod.js
+++ b/packages/app/config/webpack.prod.js
@@ -40,33 +40,7 @@ module.exports = merge(commonConfig, {
     minimizer: [
       new TerserJSPlugin({
         terserOptions: {
-          parse: {
-            // We want terser to parse ecma 8 code. However, we don't want it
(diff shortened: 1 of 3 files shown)


packages/app/config/webpack.prod.js
```diff
@@ -40,33 +40,7 @@ module.exports = merge(commonConfig, {
     minimizer: [
       new TerserJSPlugin({
         terserOptions: {
-          parse: {
-            // We want terser to parse ecma 8 code. However, we don't want it
```
tristandubbeld, line 44: Have the issues these comments describe been resolved in the meantime?

DeMoorJasper: As far as I understand this issue is that terser might replace old syntax with new syntax as it's faster/smaller, but breaks ie11 support, which this PR is removing anyway

tristandubbeld (approved):
Looks like it's working for me too!

PR #7504, opened 2023-03-14
feat(import): Suggest repositories to import on empty repositories page

tristandubbeld:
Adds suggested repositories to the empty repositories page. Also moves a bunch of files to shared folders.

### Preview

https://user-images.githubusercontent.com/7533849/224970841-bae45af4-e159-4693-b98a-2a7e079d4e47.mov

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Content/routes/Repositories/EmptyRepositories.tsx
+++ b/packages/app/src/app/pages/Dashboard/Content/routes/Repositories/EmptyRepositories.tsx
@@ -85,7 +85,7 @@ export const EmptyRepositories: React.FC = () => {
         />
       </EmptyPage.StyledGrid>
       <RestrictedImportDisclaimer />
-      <Element css={{ minHeight: 8 }} />
+      <Element css={{ minHeight: 32 }} />
       <Suggested />
     </EmptyPage.StyledWrapper>
   );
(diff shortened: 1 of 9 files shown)


lenasmayor:
Can we change the row's copy from `Import from *org dropdown*` to `Start by importing from *org dropdown*`?

packages/app/src/app/pages/Dashboard/Content/routes/Repositories/EmptyRepositories.tsx
```diff
@@ -57,6 +85,233 @@ export const EmptyRepositories: React.FC = () => {
...
+          />
+        )}
+      </Stack>
+
+      {githubRepos.state === 'loading' ? (
+        <EmptyPage.StyledGrid>
+          <SolidSkeleton viewMode="grid" />
+          <SolidSkeleton viewMode="grid" />
+          <SolidSkeleton viewMode="grid" />
+        </EmptyPage.StyledGrid>
```
olarclara, line 199: What do you think about making the number of skeletons the same as the number of cards we intend to show?

olarclara (commented):
Great work! Left only a couple of minor comments.

I know that there are some ideas of reducing the number of suggestions shown and adding an A/B test with a carousel, do you want to implement it in this PR?

PR #7509, opened 2023-03-16
feat: add "billing" role

olarclara:
Closes XTD-770.
Closes XTD-758.

In the server, an authorization looks like this:

```ts
authorization: ADMIN | WRITE | READ;
teamManager: true | false;
```

In the client, we transform this into:

```
billingManager: authorization === ADMIN || teamManager;
role: EDITOR | VIEWER;
teamAdmin: authorization === ADMIN
```

And we keep the original authorization as a reference to perform the necessary mutations.

## Test scenarios

As a team admin, you can:

- Change the role of a user who is not an admin:

wydfby-3000 csb app_dashboard_settings_workspace=0790fe87-05e9-49b0-8345-64d73d292a0f (3)

**Notes**:
- If the workspace has reached the max amount of editors and the user is a viewer, it won't be possible to change their role.
- Changing from viewer to editor will trigger a confirmation modal if the workspace is paid and there are no available seats.

---

- Grant billing rights to a non-admin:

wydfby-3000 csb app_dashboard_settings_workspace=0790fe87-05e9-49b0-8345-64d73d292a0f (4)

**Note** Their role will stay the same, only `teamManager` will be set to `true`.

---

- Grant admin rights to a non-admin:

wydfby-3000 csb app_dashboard_settings_workspace=0790fe87-05e9-49b0-8345-64d73d292a0f (5)

**Notes**
- This will change the user authorization to `authorization: ADMIN, teamManager: true`.
- This will trigger a confirmation modal if the workspace is paid and there are no available seats.
- This will be disabled If the workspace has reached the max amount of editors and the user is a viewer.

---

- Revoke billing rights of a non-admin:

wydfby-3000 csb app_dashboard_settings_workspace=0790fe87-05e9-49b0-8345-64d73d292a0f (7)

**Note** Their role will stay the same, only `teamManager` will be set to `false`.

---

- Revoke admin rights (from yourself or others) if there's more than one admin:

wydfby-3000 csb app_dashboard_settings_workspace=0790fe87-05e9-49b0-8345-64d73d292a0f (2)

This will change from `authorization: ADMIN, teamManager: true` to `billingManager: false, teamAdmin: false, role: EDITOR`. Then, it'll be possible to change this user's role.

---

As a non-admin, you can:

- Leave a team:

Screen Shot 2023-03-21 at 19 28 37

---

- Revoke pending invitations:

Screen Shot 2023-03-21 at 19 46 22

**Note**: team viewers can't invite or revoke invitations.


lenasmayor:
Hey Maria, I wasn't able to verify all the scenarios but I noticed a few things that we need to work on:

- Admins should not show a billing badge next to their username. If a billing user is changed into an admin, they display only the admin badge. 

- Admin users do not have the 'Add billing rights' option enabled for them. (I wasn't able to verify if this behavior is implemented or not).

- Disabled options should still be shown in the more options menu but not be interactive.

- We should order the team members by admins >> billing >> other team members. All by alphabetical order inside the section. Your username should follow the same rules (be placed in whatever positions makes sense according to theses rules)

olarclara:
@lenasmayor updated the PR and added the test cases, can you take another look when you have a chance?

packages/app/src/app/pages/Dashboard/Content/routes/Settings/TeamSettings/WorkspaceSettings.tsx
```diff
@@ -430,7 +226,7 @@ export const WorkspaceSettings = () => {
               <Button
                 autoWidth
                 variant="link"
-                disabled={loading}
+                disabled={teamInfoLoading}
```
tristandubbeld, line 229: Do we really need to disable this button here? I've tried updating the team but it never seems to resolve and is stuck in the loading state. It's either in a loop [related to this `useEffect`](https://github.com/codesandbox/codesandbox-client/pull/7509/files#diff-2e799785ade7c78119b46da1285e889ba2c1afe1dacff2493ad214762cda2435R94-R96), or something else is going on. I think it's simpler if we keep the loading state inside the `TeamInfo` component rather than lifting it up to `WorkspaceSettings`.

tristandubbeld (changes requested):
Found a bug related to editing the team info that needs to be fixed. The code for the billing and admin role changes look great and those scenarios work well!

packages/app/src/app/pages/Dashboard/Content/routes/Settings/TeamSettings/WorkspaceSettings.tsx
olarclara: @tristandubbeld makes sense to not disabled, I tried to keep the behaviour when refactoring but since the only blocker to deleting a team is having an active subscription, we can remove the check to make things easier.

packages/app/src/app/pages/Dashboard/Content/routes/Settings/components/TeamInfo.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+    setLoading(true);
+    // no try/catch because setTeamInfo dispatches
+    // a notification toast on error.
+    await actions.dashboard.setTeamInfo({
+      name,
+      description,
+      file,
+    });
+    setEditing(false);
```
tristandubbeld, line 88: Found that after submitting and clicking the edit button again everything was still loading and I wasn't able to change the name again. Setting loading state to false should resolve this.

```suggestion
    setLoading(false);
    setEditing(false);
```

tristandubbeld (approved):
Nice changes ✨

Found one little bug that's easy to fix, all the other stuff looks great :)

packages/app/src/app/pages/Dashboard/Content/routes/Settings/components/TeamInfo.tsx
olarclara: @tristandubbeld thank you for catching this! 🙇🏽‍♀️

PR #7514, opened 2023-03-17
fix(suggested repos): do not query if there's no team id

olarclara:

Closes XTD-808.

https://user-images.githubusercontent.com/24959348/225953712-d3fa9907-be44-478a-9138-5733b966b7c0.mov

## What is the new behavior?

https://user-images.githubusercontent.com/24959348/225955208-59b8a561-1465-47ca-b2fe-0f9149d8895c.mov

1. Visit https://nqqsyr-3000.csb.app/s
2. Select "Import from GitHub"
3. Go through the "Sign in"
4. "There was a problem getting your repositories" error doesn't happen


tristandubbeld (approved):
Oh interesting find! Thanks for fixing this!

PR #7503, opened 2023-03-13
feat: leva setup

olarclara:
Closes XTD-794.

## What kind of change does this PR introduce?

Since we upgraded to Node 16, we shouldn't have problems installing Leva. The tool will help us mock test scenarios more easily.

https://user-images.githubusercontent.com/24959348/224812871-6de7b3cf-e66d-40cb-8719-16ada66b9d8e.mov

## What steps did you take to test this? This is required before we can merge, make sure to test the flow you've updated.

1. Launch the preview
2. Open the dashboard
3. Set `CSB_DEBUG` to `true` 
4. Reload the page
5. Controls should appear at the left bottom of the screen

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      
- [ ] Added myself to contributors table

PR #7515, opened 2023-03-17
fix: leva setup

olarclara:
It adds back: https://github.com/codesandbox/codesandbox-client/pull/7503

And fixes the panel showing in production because `useControls` instance will render the panel even if we return early in the `Debug`, we need to explicitly set its value to `hidden` based on local storage or environment.

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -154,6 +154,7 @@
     "jest-snapshot": "^22.1.2",
     "jszip": "^3.1.3",
     "keyboardjs": "^2.5.1",
+    "leva": "^0.9.34",
     "localforage": "^1.5.5",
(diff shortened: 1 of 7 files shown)


tristandubbeld (changes requested):
I think we need to move the `Debug` component to the root route. It's showing up on `/s` for me due to `useControls` being used there (via `useWorkspaceSubscription` in the create new modal). This causes Leva to render for some reason, but without the different checks because the `Debug` component isn't rendered.

olarclara:
@tristandubbeld good catch, fixed in [6c5728b](https://github.com/codesandbox/codesandbox-client/pull/7515/commits/6c5728b622c100b1a67b6e6536afc74ac9a7e01e).

PR #7519, opened 2023-03-20
style(dashboard): remove fake button to upload team image at the create team flow

marcovincit:
This image makes the user to think that there is functionality to upload an image. While the feature is not implemented, it makes no sense to keep this "fake button".

Old:
image

New:
image


olarclara:
Now that the content is shorter, is there a big jump in the height when moving to the next step?

marcovincit:
> Now that the content is shorter, is there a big jump in the height when moving to the next step?

Yes, this is the current behavior, with or without the icon.

lenasmayor (approved):

olarclara:
@marcovincit you could set a fixed height for the wrapper element.

PR #7525, opened 2023-03-22
feat: adds suport to "create_sandbox" search param

olarclara:
Closes xtd-821.

`dashboard?create_sandbox=true` will send users to the recent page in the dashboard and open the new sandbox modal.


olarclara:
@tristandubbeld that's a great question! I thought about keeping the previous parameters for backward compatibility and introducing a new param `modal` which could be `github_import`, `create_sandbox`, `new_team`, and so on. On one hand, it'll centralize the logic making it easier to reason, on the other hand, I think we lose a bit of the explicitness/readability. Wdyt?

PR #7527, opened 2023-03-23
fix(trial): Create customer portal instead of checkout url and combine components

tristandubbeld:
Combined the mid trial reminder with the cancel subscription modal so they share one style and one component to view the lost benefits.

Note: I've also fixed an **important** bug where we would try and load the checkout url for a team that's currently on trial - which doesn't work. Instead we're now creating a customer portal.

### Mid trial reminder updates
image

### Cancel trial or subscription updates
image

### DIFF
--- a/packages/app/src/app/pages/common/Modals/MidTrialModal/index.tsx
+++ b/packages/app/src/app/pages/common/Modals/MidTrialModal/index.tsx
@@ -90,7 +90,14 @@ export const MidTrialModal = () => {
                 })}`}{' '}
             left on your trial.
           </Text>
-          <Text as="p" size={19} color="#C2C2C2" align="center">
+          <Text
+            as="p"
+            size={19}
+            color="#C2C2C2"
+            align="center"
+            lineHeight="28px"
+            css={{ margin: 0 }}
+          >
             {!experimentValue ? <SkeletonText /> : null}
 
             {experimentValue === ExperimentValues.A
@@ -228,7 +235,7 @@ const LoseAfterTrial = () => {
               <Text size={16} color="#FFFFFF" lineHeight="24px">
                 {feature.free}
               </Text>
-              {pillValue ? <Badge variant="error">{pillValue}</Badge> : null}
+              {pillValue ? <Badge variant="warning">{pillValue}</Badge> : null}
             </Stack>
           </div>
         );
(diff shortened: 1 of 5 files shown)


packages/components/src/components/Badge/index.tsx
```diff
@@ -4,7 +4,7 @@ import { Stack } from '../Stack';
 import { Text } from '../Text';
 
 export interface BadgeProps {
-  variant?: 'trial' | 'neutral' | 'warning' | 'highlight' | 'error';
+  variant?: 'trial' | 'neutral' | 'warning' | 'highlight';
```
olarclara, line 7: Should we keep this? The `Badge` in prism has an `error` variant.

packages/app/src/app/pages/common/Modals/MidTrialModal/index.tsx
```diff
@@ -58,48 +56,70 @@ export const MidTrialModal = () => {
...
       codesandbox: 'V1',
       event_source: 'UI',
     });
 
-    dismissMidTrialReminder();
+    // Custom implementation based on useGetCheckoutURL hook because the modal
+    // needs to be dismissed before redirecting.
+    try {
+      setLoadingPortal(true);
+      const payload = await api.stripeCustomerPortal(activeTeam, pathname);
```
olarclara, line 69: Can you replace this with the `useCreateCustomerPortal` hook?

olarclara (changes requested):
Looks good! I like the reusable `LoseFeatures`, requested changes because I believe we can use an existing hook to create the customer portal.

packages/components/src/components/Badge/index.tsx
tristandubbeld: Oh I didn't realise! I created this variant in the previous PR where I also added the mid-trial modal. Next week Alex and I will take a quick look into using prism components in v1, and it would be great if we can replace the badge with that one :)

packages/app/src/app/pages/common/Modals/MidTrialModal/index.tsx
tristandubbeld: I shall take a look, thanks!

olarclara: @tristandubbeld we could pass an optional "onComplete" callback to the hook but if you think it's out of scope, we can merge this as is.

PR #7533, opened 2023-03-24
fix(vercel): Update vercel user endpoint

tristandubbeld:
Started out by changing the Vercel user endpoint to the latest version, but found that there is a lot of stuff we're not using. This PR doesn't get rid of all of it, but I was able to remove a bunch of unused code and update a few more endpoints in the time left I wanted to spend on this.

### Testing

This can be tested on stream by deploying (v1) sandboxes to Vercel via the UI. Let me know if you need help getting a Sandbox template to fork and deploy.


olarclara (approved):
Looks good to me!

Is it too much out of scope to make the preview link in this modal open in new tabs?

k6txce-3000 csb app_s_reverent-nova-xupzy5

PR #7543, opened 2023-03-28
feat(pricing): Determine user location and show appropriate price

tristandubbeld:
In this PR we're fixing the price calculation. There were some bugs and we also needed to calculate and format the pricing based on the users locale and currency.

### Determining location

We're determining the user location based on their timezone. This isn't fool proof but we expect Stripe to take care of that. The timezone will tell us what the country of origin is.

#### Testing

1. On an Apple device, turn off automatic time and location in the date & time settings. Change your "closest city" to an Indian one (e.g. Kolkata). This will update your timezone.
2.  Visit the pro page and select a personal workspace without subscription.
3. You should now see the price in Rupees.

#### Rupees
We base the formatting locale on the currency. When the timezone is set to the Indian one, the currency will be Rupees and the locale will be set to `hi-IN`.

#### Euros
When the currency is in `EUR`, this is the case with Paddle subscriptions, we will default to the `nl-NL` locale. This formats the price with the euro sign in front of the price. In e.g. the `de-DE` locale the euro sign is placed behind the price, and we don't want that in this case.

#### Dollars
When the currency is in `USD`, which is the case for most other subscriptions we default the locale to `en-US`. This will format the price accordingly.

### Currency not available in stripe

When a certain currency is not available in Stripe (like right now, we haven't activated INR on production yet) we fall back to USD prices and currency.

### Loading states

When the workspace has an active subscription, we're loading the `subscription`. This might take some time, and we show a skeleton rather than nothing. We previously rendered `NaN`, this has now been resolved.

When the workspace does not have a subscription, we're loading the `prices`. This might take some time, and we show a skeleton rather than nothing. We previously rendered `NaN`, this has now been resolved.

### Active subscriptions

For active subscriptions we didn't format the price in the right way. It always showed dollars, and the `unitPrice` in cents wasn't divided to full dollars (or other full currencies). This has been resolved.

#### Personal free

| Loading  | Ready |
| ------------- | ------------- |
| personal-free-loading | personal-free |

#### Personal Patron subscription

| Loading  | Ready |
| ------------- | ------------- |
| patron-personal-active-loading | patron-personal-active |

#### Personal upsell card

The upsell to team subscription card (on the right).

| Loading  | Ready |
| ------------- | ------------- |
| personal-upsell-loading | personal-upsell-active |

#### Team free

| Loading  | Ready |
| ------------- | ------------- |
| team-free-loading | team-free |

#### Team Stripe subscription

| Loading  | Ready |
| ------------- | ------------- |
| stripe-team-active-loading | stripe-team-active |

#### Team Paddle subscription

| Loading  | Ready |
| ------------- | ------------- |
| paddle-team-active-loading | paddle-team-active |

#### Personal free with Rupees

| Ready |
| ------------- |
| personal-free-rupee |

### DIFF
--- a/packages/app/src/app/pages/Pro/Upgrade.tsx
+++ b/packages/app/src/app/pages/Pro/Upgrade.tsx
@@ -42,9 +42,9 @@ const useCurrencyFromTimeZone = () => {
 
   if (timeZone === 'Asia/Kolkata' || timeZone === 'Asia/Calcutta') {
     return 'INR';
-  } else {
-    return 'USD';
   }
+
+  return 'USD';
 };
 
 export const ProUpgrade = () => {
--- a/packages/app/src/app/pages/Pro/Upgrade.tsx
+++ b/packages/app/src/app/pages/Pro/Upgrade.tsx
@@ -28,6 +28,7 @@ import { formatCurrency } from 'app/utils/currency';
 import { useGetCheckoutURL } from 'app/hooks';
 import { useWorkspaceAuthorization } from 'app/hooks/useWorkspaceAuthorization';
 import { useWorkspaceSubscription } from 'app/hooks/useWorkspaceSubscription';
+import { useCurrencyFromTimeZone } from 'app/hooks/useCurrencyFromTimeZone';
 import { Switcher } from './components/Switcher';
 import { SubscriptionPaymentProvider } from '../../graphql/types';
 import { SubscriptionCard } from './components/SubscriptionCard';
@@ -36,16 +37,6 @@ import type { CTA } from './components/SubscriptionCard';
 import { StyledPricingDetailsText } from './components/elements';
 import { TeamSubscriptionOptions } from '../Dashboard/Components/TeamSubscriptionOptions/TeamSubscriptionOptions';
 import { NewTeamModal } from '../Dashboard/Components/NewTeamModal';
-
-const useCurrencyFromTimeZone = () => {
-  const timeZone = Intl.DateTimeFormat().resolvedOptions().timeZone;
-
-  if (timeZone === 'Asia/Kolkata' || timeZone === 'Asia/Calcutta') {
-    return 'INR';
-  }
-
-  return 'USD';
-};
 
 export const ProUpgrade = () => {
   const {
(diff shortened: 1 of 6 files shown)


packages/app/src/app/pages/Pro/Upgrade.tsx
```diff
@@ -162,6 +173,24 @@ export const ProUpgrade = () => {
...
+    type: 'individual' | 'team',
+    period: 'year' | 'month'
+  ) => {
+    const priceInCurrency =
+      pro?.prices?.[type]?.[period]?.[currency.toLowerCase()] ||
+      pro?.prices?.[type]?.[period]?.usd; // fallback to usd
+
+    // Divide by 12 if the period is year to get monthly price for yearly
+    // subscriptions
+    const price = period === 'year' ? priceInCurrency / 12 : priceInCurrency;
```
olarclara, line 186: I think this can resolve to `NaN` or `undefined` if we don't have the data that we expect (or it's loading and the connection is slow) and because `formatCurrency` only checks if the `amount` is `null`, it results in the following in a slow connection:

Screen Shot 2023-03-28 at 11 36 05

[This](https://www.speedtest.net/global-index/india) shows that India ranks 81st in internet speed, so we could add a loading state, not render anything or show a fallback to avoid having this issue.

tristandubbeld: This was actually a bug when loading the active `subscription`! I've fixed this in [f28e0ce](https://github.com/codesandbox/codesandbox-client/pull/7543/commits/f28e0cebc0dec00aa014abcd1999b63b03751a49), among other things.

packages/app/src/app/pages/Pro/Upgrade.tsx
```diff
@@ -162,6 +165,35 @@ export const ProUpgrade = () => {
       team.subscription?.paymentProvider === SubscriptionPaymentProvider.Paddle
   );
 
+  const getPricePerMonth = (
```
tristandubbeld, line 168: The reason I'm keeping this function as a copy of the one in `UpsellTeamProCard.tsx` is that this might change in a relatively short time. We might want to show actual subscription prices to admins with an active subscription which will need some extra checks and changes in this function.

olarclara (approved):
LGTM!

PR #7544, opened 2023-03-28
fix(settings): editor seats

olarclara:
Closes XTD-854.

- Checks if it should restrict new editors before showing the upgrade alert.
- Fix the number of seats for teams that never had a subscription.


tristandubbeld (approved):

packages/app/src/app/hooks/useWorkspaceSubscription.ts
```diff
@@ -50,6 +50,7 @@ export const useWorkspaceSubscription = (): WorkspaceSubscriptionReturn => {
     return {
       ...NO_SUBSCRIPTION,
       isEligibleForTrial: isTeamSpace, // Currently, only teams are eligible for trial.
+      numberOfSeats: activeTeamInfo.limits.maxEditors,
```
tristandubbeld, line 53: We can remove the `numberOfSeats` in `NO_SUBSCRIPTION` if we want or we need to update the object and the types to reflect the `maxEditors` type.

PR #7545, opened 2023-03-29
Allow clipboard permissions in iframe preview

kevin940726:

I would say it's both a bug fix and a feature? Not sure.

Currently, copying and pasting from the clipboard doesn't work in the preview iframe in the editor. This is because the preview frame is in another domain and some browsers (chrome and firefox for instance) block cross-origin access to the clipboard. Based on this [article](https://sites.google.com/a/chromium.org/dev/Home/chromium-security/deprecating-permissions-in-cross-origin-iframes), the solution is to add the necessary permission to the `allow` attribute for the iframe. However, clipboard permissions are missing in the allow list.

## What is the new behavior?

Add `clipboard-read` and `clipboard-write` to the preview frame to allow accessing the clipboard.

1. Visit https://codesandbox.io/s/eager-bird-087o05?file=/index.html
2. Click on the button in the iframe preview. Notice there's an error outputted on top but not the copied content.
3. Use devtools to add `clipboard-read` to the allow list of the iframe, and reload the iframe (it will not work without reloading).
4. Try again, and the copied content should now show on the screen.

## Checklist

- [ ] Documentation N/A
- [x] Testing 
- [x] Ready to be merged
      
### DIFF
--- a/packages/common/src/components/Preview/index.tsx
+++ b/packages/common/src/components/Preview/index.tsx
@@ -634,7 +634,7 @@ class BasePreview extends React.PureComponent<Props, State> {
               <>
                 <StyledFrame
                   key="PREVIEW"
-                  allow="accelerometer; ambient-light-sensor; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
+                  allow="accelerometer; ambient-light-sensor; camera; clipboard-read; clipboard-write; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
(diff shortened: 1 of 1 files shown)


packages/common/src/components/Preview/index.tsx
```diff
@@ -634,7 +634,7 @@ class BasePreview extends React.PureComponent<Props, State> {
               <>
                 <StyledFrame
                   key="PREVIEW"
-                  allow="accelerometer; ambient-light-sensor; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
+                  allow="accelerometer; ambient-light-sensor; camera; clipboard-read; clipboard-write; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
```
danilowoz, line 637: dwada
dwada

PR #7620, opened 2023-05-10
fix(compile): create one TestRunner between sandbox compiles

SSHari:

Bug fix

This is a proposed upstream fix for a [related Sandpack issue](https://github.com/codesandbox/sandpack/issues/927).

If you call the `compile` logic multiple times in a short amount of time, then the create `TestRunner` logic won't work as expected. Since the `testRunner` is shared between compiles, this is what ends up happening:

1. Call compile (**Compile A**)
2. Call compile again immediately (**Compile B**)
3. **Compile A**: `setTimeout` is called (wait `600ms`)
4. **Compile B**: `setTimeout` is called (wait `600ms`)
5. **Compile A**: `setTimeout` runs
6. **Compile A**: Because `testRunner` is `undefined`, await `jest-lite` module import
7. **Compile B**: `setTimeout` runs (step 6 is suspended because of the `await`)
8. **Compile B**: Because step 6 was suspended because of the `await`, `testRunner` is still `undefined`, so await `jest-lite` module import
9. **Compile A**: Set `testRunner` to `new TestRunner(manager)`
10. **Compile B**: Set `testRunner` to `new TestRunner(manager)`

The last two steps are the key to this issue. Because we call `new TestRunner(manager)` _twice_, we end up with _two_ [Code Sandbox message listeners](https://github.com/codesandbox/codesandbox-client/blob/master/packages/app/src/sandbox/eval/tests/jest-lite.ts#L112). When the `TestRunner` responds to a message (e.g. `run-all-tests`), it handles the message twice which means the tests are run twice. Since there's a shared global state used for tests and reporting results, the two test runs end up conflicting with each other and we get the error that's seen in that linked Sandpack issue.

Verifying this is easier to do in Sandpack because the issue currently exists there. In Sandpack, we call `compile` multiple times when a client is initialized which is reproducing this problem. While I think we may be able to make some improvements on the Sandpack side so that we don't compile more than once initially, I don't think we should rely on that to prevent this bug. I think the better fix for this issue is to address it in this package instead.

I couldn't figure out how to reproduce this error directly in this package. I _was_ able to simulate this behavior/issue by adding the following to the [app package's sandbox index file](https://github.com/codesandbox/codesandbox-client/blob/master/packages/app/src/sandbox/index.ts#L58):

```ts
setTimeout(() => compile(data), 1);
```

When I added the above line and opened a Jest Code Sandbox locally, I got the same error after running the test:

Screenshot 2023-05-10 at 1 06 03 AM

## What is the new behavior?

Calling `compile` multiple times doesn't throw this error anymore because we only create the `TestRunner` once. The key part of the change is the line:

```ts
testRunner = testRunner || new TestRunner(manager);
```

This line happens _after_ the `await`, which means that when it runs for the second compile (i.e. **Compile B**), `testRunner` _will_ be set.

**codesandbox-client**

I simulated the error in this package using the method above. After I made the fix, I no longer saw the error in this package.

**sandpack**

1. I made the change locally
2. I started the app package's Sandpack script (i.e. `start:sandpack-sandbox`)
3. I started Sandpack locally and pointed the `bundlerURL` to my local sandbox (running on `localhost:3000`)
4. I no longer saw the error on the Sandpack side

## Checklist

- [ ] Documentation N/A
- [x] Testing 
- [x] Ready to be merged
      
### DIFF
--- a/packages/app/src/sandbox/compile.ts
+++ b/packages/app/src/sandbox/compile.ts
@@ -890,11 +890,9 @@ async function compile(opts: CompileOptions) {
 
     setTimeout(async () => {
       try {
-        testRunner =
-          testRunner ||
-          (await import('./eval/tests/jest-lite')
-            .then(s => s.default)
-            .then(TestRunner => new TestRunner(manager)));
+        const jestLiteModule = await import('./eval/tests/jest-lite');
+        const TestRunner = jestLiteModule.default;
+        testRunner = testRunner || new TestRunner(manager);
 
         sendTestCount(modules);
       } catch (e) {
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1691,6 +1691,15 @@
         "bug",
         "code"
       ]
+    },
+    {
+      "login": "SSHari",
+      "name": "Sai Hari",
+      "avatar_url": "https://avatars.githubusercontent.com/u/11434879?v=4",
+      "profile": "https://thesshguy.com/",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "contributorsPerLine": 7,
--- a/README.md
+++ b/README.md
@@ -274,6 +274,7 @@ Thanks goes to these wonderful people
     <td align="center"><a href="http://baobangdong.cn"><img src="https://avatars.githubusercontent.com/u/36991862?v=4" width="100px;" alt="B2D1"/><br /><sub><b>B2D1</b></sub></a><br /><a href="https://github.com/codesandbox/codesandbox-client/commits?author=B2D1" title="Documentation">📖</a></td>
     <td align="center"><a href="https://github.com/kettanaito"><img src="https://github.com/kettanaito.png" width="100px;" alt="Artem"/><br /><sub><b>Artem</b></sub></a><br /><a href="https://github.com/codesandbox/codesandbox-client/commits?author=kettanaito" title="Code">💻</a></td>
     <td align="center"><a href="http://hugos.computer"><img src="https://avatars.githubusercontent.com/u/1035169?v=4" width="100px;" alt="Hugo Wiledal"/><br /><sub><b>Hugo Wiledal</b></sub></a><br /><a href="https://github.com/codesandbox/codesandbox-client/issues?q=author%3Awiledal" title="Bug reports">🐛</a> <a href="https://github.com/codesandbox/codesandbox-client/commits?author=wiledal" title="Code">💻</a></td>
+    <td align="center"><a href="https://thesshguy.com/"><img src="https://avatars.githubusercontent.com/u/11434879?v=4" width="100px;" alt="Sai Hari"/><br /><sub><b>Sai Hari</b></sub></a><br /><a href="https://github.com/codesandbox/codesandbox-client/commits?author=SSHari" title="Code">💻</a></td>
   </tr>
 </table>
 


packages/app/src/sandbox/compile.ts
```diff
@@ -890,11 +890,9 @@ async function compile(opts: CompileOptions) {
 
     setTimeout(async () => {
       try {
-        testRunner =
-          testRunner ||
-          (await import('./eval/tests/jest-lite')
-            .then(s => s.default)
-            .then(TestRunner => new TestRunner(manager)));
+        const jestLiteModule = await import('./eval/tests/jest-lite');
```
danilowoz, line 893: Maybe I'm misreading this. But isn't it doing exactly the same thing but in a different order? 

Plus, maybe we could make import conditional in order to prevent loading/fetching it once again:

```js
if(!testRunner) {
	const jestLiteModule = await import('./eval/tests/jest-lite');
	const TestRunner = jestLiteModule.default;
	testRunner = new TestRunner(manager);
}
```

What do you think?

SSHari: Hey @danilowoz, so it's subtle, but they're not exactly doing the same thing. The difference comes down to _when_ the `testRunner` variable conditional is checked.

In the current code and the code that you suggested, we're checking whether `testRunner` exists _before_ we call `await import('./eval/tests/jest-lite');`. What's happening is that when the first `compile` call reaches the `await import('./eval/tests/jest-lite');` line, the process suspends its execution. Before the import is returned and `testRunner` can be set, the second `compile` call reaches the _same_ `await import('./eval/tests/jest-lite');` line.

Because the first call was suspended and `testRunner` hasn't been set yet, the second call will _also_ fail this check and start the import process again. This results in `new TestRunner(manager)` being called twice, leading to the bug we're seeing.

The difference with the code in the PR is that we check if `testRunner` has been set _after_ the `await import('./eval/tests/jest-lite');` line (i.e. `testRunner = testRunner || new TestRunner(manager);`). Here's a rough explanation of what I think is happening in this scenario:

1. First `compile` call reaches the `await import('./eval/tests/jest-lite');` line
2. First `compile` call suspends its execution because of the `await`
3. Second `compile` call reaches the `await import('./eval/tests/jest-lite');` line
4. Second `compile` call suspends its execution because of the `await`
5. First `compile` call resumes execution and  sets `testRunner` because it's `undefined` and then it finishes executing
6. Second `compile` call resumes execution and **doesn't** set `testRunner` because we do the check at this point _after_ Step 5 finishes

In all 3 scenarios, we're still calling the `await import('./eval/tests/jest-lite');` twice, but by moving the check to see if `testRunner` exists to _after_ the `await`, we're able to skip the second call to `new TestRunner(manager);`.

I put a minimal example together that outlines each of the 3 scenarios: https://codesandbox.io/s/mystifying-dream-lcfdtz?file=/src/index.js. If you scroll to the bottom of the sandbox, there are 3 versions of the main function that represent the 3 different scenarios we're talking about:
1. The current code
2. The code you're suggesting
3. The code in the PR

If you open the console and try uncommenting each of those main functions, you should be able to see how the code flow is logged. The key part is that only the last example calls the `createTestRunner` function once.

One other thing to note is that even though we're calling `await import("./createTestRunner")` twice in the example CodeSandbox, the `console.log` at the top level of the imported file is only executed once. I can't say for sure, but it looks like even though we `import` the same module more than once, it doesn't actually load it multiple times (maybe it's served from a cache).

danilowoz: Gotcha!! Great example, btw.

Let's get this merged

danilowoz (approved):

PR #7654, opened 2023-05-24
Debounce Algolia Search

sannek:
Adds a debounce to the algolia search, so we don't fire off too many requests while someone is still typing.

Let me know if there's a better way to do this? But I noticed we use this `debounce` function in lots of places, so it seemed like the right thing to do?

1. Step A
2. Step B
3. Step C

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      

alexnm (approved):
Looks good, let's see if we get any feedback and need o tweak the delay

PR #7673, opened 2023-06-01
Open preferences modal from notification actions

sannek:
Stacks into #7672 

I've never felt so accomplished by a modal 🧙 

Adds a new action 'Manage notification preferences' to the notifications menu, which opens the preferences modal at the 'notifications' tab. (It seemed too complicated to find this, so opening it from the notifications seemed sensible)

image

## What steps did you take to test this? This is required before we can merge, make sure to test the flow you've updated.

1. Step A
2. Step B
3. Step C

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      
- [ ] Added myself to contributors table

PR #7675, opened 2023-06-06
Sanne/pc 702 open preferences by url

sannek:
Replaces #7673 

- I moved the `openUserPreferences` function to `actions > preferences` so we can use the same function wherever we want to open this modal
(Being able to open this modal via URL is one of the number 1 requests from support, since otherwise it's so tricky to find)

Changed to a query parameter `?preferences=true` that will work anywhere within `/dashboard`

1. Step A
2. Step B
3. Step C

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      
### DIFF
--- a/packages/app/src/app/components/Notifications/Filters.tsx
+++ b/packages/app/src/app/components/Notifications/Filters.tsx
@@ -6,10 +6,13 @@ import css from '@styled-system/css';
 export const Filters = () => {
   const { userNotifications } = useAppState();
   const {
-    filterNotifications,
-    markAllNotificationsAsRead,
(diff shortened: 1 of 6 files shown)


christianalfoni (commented):
So this looks great! 👏 😄 

But I would argue that `preferences` should not be a url param, but a query? So it does not really matter what Dashboard URL you open, if you have `?preferences=true` in the url, it will open that modal on top? Or is it specifically constrained to the `recents`?

christianalfoni (approved):
Awesome @sannek !!

PR #7678, opened 2023-06-13
fix: debounce error in dependencies

alexnm:
Using the debounce inline was not passing the event argument properly. I also moved it into a useCallback to avoid calling debounce again on each re-render.

Closes PC-981


sannek (approved):
Thanks for fixing!

PR #7699, opened 2023-06-29
feat(CLI): add dev and local token

christianalfoni:
This PR changes the `post` approach to rather:

- Enable automatic exchange token approval on `codesandbox.stream` as you have to have a `codesandbox.io` account to use `.stream`
- Instead of `post` we now do `local`, which writes the exchange token to local storage and rather posts a confirmation when it is written, accessible by the iFrame by domain security, bypassing passing the token itself to the preview


christianalfoni:
The `console-feed` package will be published again: https://github.com/codesandbox/codesandbox-client/pull/7700

PR #7700, opened 2023-06-30
feat(CLI): add dev and preview token support

christianalfoni:
This PR changes the `post` approach to rather:

- Enable automatic exchange token approval on `codesandbox.stream` as you have to have a `codesandbox.io` account to use `.stream`
- Instead of `post` we now do `preview`, which writes the exchange token to local storage, accessible by the iFrame by domain security, bypassing passing the token itself to the preview

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -115,7 +115,7 @@
     "comlink-loader": "^2.0.0",
     "compare-versions": "^3.1.0",
     "console": "^0.7.2",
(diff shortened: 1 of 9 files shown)


christianalfoni:
Just waiting for this: https://github.com/samdenty/console-feed/issues/124#issuecomment-1613891672

CompuIves (approved):

PR #7845, opened 2023-07-27
feat: sso support

alexnm:
Finalize the SSO flow from the client
* New option in sign in component to sign in with SSO
* Sign in with SSO should work as oauth signin, but it has a custom URL that the backend manages
* When successful, the user goes through the signin or signup flow and gets redirect to the dashboard

image

Fixes/changes:
* Create team modal not shown when the newly registered user already is part of more then 1 workspace
* Redirect to dashboard after registration finished


packages/app/src/app/overmind/effects/api/apiFactory.ts
```diff
@@ -14,6 +14,11 @@ const getBaseApi = (path: string, useRoot: boolean = false) => {
     return API_ROOT;
   }
 
+  // Special case for /auth/workos requests which are not on /api/v1
```
christianalfoni, line 17: Why is it not on /api/v1? On premise?

alexnm: the /auth/workos routes are not in /api/v1, some of them are json endpoints, others are html

christianalfoni (approved):
Great stuff 👍

PR #7866, opened 2023-08-02
refactor(pages): update onboarding roles

filipelima18:

This PR updates the role options on the first post-signup question so that it matches the roles we have defined internally.

I couldn't figure out a way to simulate a new user onboarding. Also, I'm not sure if this requires any back-end work so that the new roles get registered to our database. Ideally, that should be tested during the review.

Closes MKT-354


alexnm (approved):

packages/app/src/app/pages/SignIn/Onboarding.tsx
```diff
@@ -8,11 +8,13 @@ import { StyledButton } from 'app/components/dashboard/Button';
...
 
 const ROLE_OPTIONS = [
-  { value: 'frontend', label: 'Frontend developer' },
-  { value: 'backend', label: 'Backend developer' },
-  { value: 'fullstack', label: 'Fullstack developer' },
+  { value: 'frontend', label: 'Front-End Developer' },
+  { value: 'backend', label: 'Back-End Developer' },
+  { value: 'fullstack', label: 'Full-Stack Developer' },
+  { value: 'designer', label: 'Lead Developer' },
+  { value: 'designer', label: 'Engineering Manager' },
```
alexnm, line 15: I think there's a wrong value here for Lead Developer and Engineering Manager, they shouldn't have value `designer` right?

alexnm (dismissed):

packages/app/src/app/pages/SignIn/Onboarding.tsx
filipelima18: True. Not sure how I missed that. Just committed an update. Can you have a look @alexnm ?

christianalfoni (approved):
This looks good to me! Though we should validate with BE if there are any expectations on values here 👍

filipelima18 (commented):
Updated.

PR #7870, opened 2023-08-02
feat: update workspace selector

alexnm:
Still partially done, since the `legacy-free` vs `deactivated` badges need the legacy subscription check

image

* Personal workspaces are shown with the name of the user
* Legacy team free is shown with the Free badge
* Legacy personal pro is shown with a Pro badge
* Deactivated teams (cancelled or expired subscriptions) are shown with the Deactivated badge
* Also removes the `isPatron` flag and components as these do not exist anymore


packages/app/src/app/components/WorkspaceSelect/WorkspaceSelect.tsx
```diff
@@ -150,12 +168,16 @@ export const WorkspaceSelect: React.FC<WorkspaceSelectProps> = React.memo(
...
                       </Text>
 
-                      {isTeamFree && <Badge variant="trial">Free</Badge>}
+                      {isTeamFree && !isDeactivated && (
+                        <Badge variant="trial">Free</Badge>
+                      )}
+                      {isPersonalProLegacy && <Badge variant="pro">Pro</Badge>}
+                      {isDeactivated && (
+                        <Badge variant="neutral">Deactivated</Badge>
+                      )}
```
danilowoz, line 180: Wondering if we can make something like this to make it more readable
```suggestion
											{isDeactivated ? 
  											<Badge variant="neutral">Deactivated</Badge>
                       : isTeamFree ? 
                        <Badge variant="trial">Free</Badge>
                       : isPersonalProLegacy && <Badge variant="pro">Pro</Badge>}
```

packages/app/src/app/hooks/useWorkspaceSubscription.ts
```diff
@@ -76,10 +88,6 @@ export const useWorkspaceSubscription = (): WorkspaceSubscriptionReturn => {
   const numberOfSeats =
     (isFree ? activeTeamInfo.limits.maxEditors : subscription.quantity) || 1;
 
-  const isPatron =
```
danilowoz, line 79: 👋

danilowoz (approved):

packages/app/src/app/components/WorkspaceSelect/WorkspaceSelect.tsx
alexnm: I messed it up in the last step, because we're in this intermediate state where deactivated and free can overlap. but after we get a new field on the team, each flag should be a single condition, so I'd keep it linear

```
{isA && A}
{isB && B}
{isC && C}
```

hope that makes sense

PR #7880, opened 2023-08-03
feat(api): add new endpoint for retrieving features

alexnm:
Integrates the new `/meta/features` endpoint that tells the client (for now) which login options are available. SSO and OAuth login options are now conditioned by these flags.

The new api routes that are no longer on /v1 will need some better support with the current implementation of apiFactory, but I'll change that in a subsequent PR to avoid dragging this for too long or blowing the scope here.

Closes PC-1113

### DIFF
--- a/packages/app/src/app/overmind/effects/api/apiFactory.ts
+++ b/packages/app/src/app/overmind/effects/api/apiFactory.ts
@@ -10,13 +10,13 @@ const API_ROOT = '/api';
  * use the root path.
  */
 const getBaseApi = (path: string, useRoot: boolean = false) => {
-  if (useRoot) {
-    return API_ROOT;
-  }
-
   // Special case for /auth/workos requests which are not on /api/v1
   if (path.startsWith('/auth/workos')) {
     return '';
+  }
+
+  if (useRoot) {
+    return API_ROOT;
   }
 
   return path.startsWith('/beta') ? API_ROOT : `${API_ROOT}/v1`;
--- a/packages/app/src/app/overmind/state.ts
+++ b/packages/app/src/app/overmind/state.ts
@@ -10,6 +10,7 @@ import {
 } from 'app/graphql/types';
 import { derived } from 'overmind';
 import { hasLogIn } from '@codesandbox/common/lib/utils/user';
+import { MetaFeatures } from './effects/api/types';
 
 export type PendingUserType = {
   avatarUrl: string | null;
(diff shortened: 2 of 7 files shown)


packages/app/src/app/overmind/effects/api/apiFactory.ts
```diff
@@ -10,15 +10,15 @@ const API_ROOT = '/api';
...
-  }
-
   // Special case for /auth/workos requests which are not on /api/v1
   if (path.startsWith('/auth/workos')) {
     return '';
   }
 
+  if (useRoot) {
+    return API_ROOT;
+  }
```
alexnm, line 20: moved down since `/auth/workos` should have precedence

packages/app/src/app/pages/SignIn/ChooseProvider.tsx
```diff
@@ -281,26 +288,29 @@ export const ChooseProvider: React.FC<ChooseProviderProps> = ({
...
-                  width: '100%',
-                  height: '48px',
-                }}
-                loading={loadingAuth.github}
-                onClick={() => handleSignIn('github')}
-              >
-                <GitHubIcon />
-                <Text
-                  as="span"
+              {features.loginWithGithub && (
```
alexnm, line 291: just wrapped the button in a conditional

packages/app/src/app/pages/SignIn/ChooseProvider.tsx
```diff
@@ -309,46 +319,50 @@ export const ChooseProvider: React.FC<ChooseProviderProps> = ({
...
-                  secondary
-                >
-                  <GoogleIcon />
-                  <Text
-                    as="span"
-                    css={{ lineHeight: 1 }}
-                    size={3}
-                    weight="medium"
+                {features.loginWithGoogle && (
+                  <Button
```
alexnm, line 323: just wrapped the button in a conditional

packages/app/src/app/overmind/state.ts
```diff
@@ -144,4 +152,10 @@ export const state: State = {
     github: false,
   },
   isProcessingPayment: false,
+  features: {
+    // Fallback values for when the features endpoint is not available
+    loginWithApple: true,
```
danilowoz, line 157: Shouldn't it init as false? I guess such features should be unavailable until the API confirms its state, otherwise the user might see a layout shift if the API returns `false`

danilowoz (approved):
Left some comments

packages/app/src/app/overmind/state.ts
alexnm: I was thinking about this, but I was more concerned with breaking something in prod. This call is super fast, so there shouldn't be any shift as far as I tested. But wanted to make sure that in case it fails (for whatever reason) the default options are available as a fallback

PR #7925, opened 2023-08-10
feat: update pro page

alexnm:
NOTE: This is still subject to change, but I'm wrapping up the work for this week.

NOTE 2: Pricing is not yet updated

/pro page now has 3 possibilities

### 1. Legacy pro - for paddle users
technically, this page should not be accessible, unless the user types the URL manually. all paddle links from the dashboard are directly taking the user to the customer portal

image

### 2. Upgrade pro - for flows of upgrading legacy free teams or inactive teams to pro, or for managing pro subscriptions for a specific workspace. these URLs need a ?workspace= query param

Pro (new and old)
image

Free Team (legacy)
image

Inactive Team (expired, cancelled)
image

Personal Pro (legacy)
image

this one should be accessible when the user is not presented with the pro benefits, so we use it as an intermediate step before redirect to stripe. eg: from the team settings -> permissions or registries banners on the legacy free teams.

### 3. Create pro - the default version of the /pro page which is the starting point for creating a pro workspace

image

this is the standard view the users sees when going to /pro the page is considering the personal free workspace as the starting point, so no information about any existing subscription is presented.


necoline:
Some content notes for the cards (this should apply to upsell banners in the dashboard as well)
- instead of 20 editors, it should be `unlimited editors`
- let's combine the lines on the sandboxes and repos to be `private sandboxes and repositories`

PR #7926, opened 2023-08-10
chore: update amplitude and conditional tracking

alexnm:
A couple of steps taken:
* Amplitude integrated with their official JS package, not through GTM
* GTM + GA tracking removed
* Amp and sentry keys are hardcoded until we merge the other PR with the dynamic env


christianalfoni (commented):
Haha, I know @CompuIves made this function originally 😂 

`isDoNotTrackEnabled`

I am so fascinated how your brain works, why does it not aim for `isTrackEnabled`? 😂 

Anyways, this makes sense to me, though it is not obvious with the code change that it will fix all tracking. Is this used by the tracking effect in Overmind on V1 etc.? 🤔

CompuIves:
It's because it's based on an existing name [image: 😄]

https://en.m.wikipedia.org/wiki/Do_Not_Track

packages/common/src/utils/analytics/amplitude.ts
```diff
@@ -1,129 +1,105 @@
...
-    ) {
-      amplitude.getInstance().setUserId(null);
-      amplitude.getInstance().regenerateDeviceId();
-    }
-  } else {
+  if (!amplitude) {
     debug('[Amplitude] NOT resetting user id because Amplitude is not loaded');
   }
+
+  amplitude.reset();
```
CompuIves, line 57: Are we sure that we reset the user id fully here? Looking at the previous code (with the explicit null setting), I have a feeling that there's a bigger story here and maybe it requires the clear null setting.

packages/common/src/utils/analytics/google.ts
```diff
@@ -1,13 +0,0 @@
-import { global } from './utils';
-
-export const track = (_, data) => {
-  if (typeof global.ga !== 'undefined') {
-    global.ga('send', data);
```
CompuIves, line 5: We'll lose Google Analytics from this, but that should be fine, right?

packages/app/src/embed/index.html
```diff
@@ -52,19 +52,6 @@
     rel="icon"
   />
     <title>Embed - CodeSandbox</title>
-    <!-- Google Tag Manager -->
```
CompuIves, line 55: Do we need to change anything in `embed/index.js` to make this work properly?

CompuIves (commented):
Some small questions!

packages/common/src/utils/analytics/google.ts
alexnm: Yes, my understanding is that we want to get rid of GA anyway, for privacy concerns

packages/common/src/utils/analytics/amplitude.ts
alexnm: I was looking at the docs for the browser SDK and the reset does both user reset and device id: https://www.docs.developers.amplitude.com/data/sdks/browser-2/#reset-when-user-logs-out

CompuIves (approved):

PR #7928, opened 2023-08-10
use updated astring generator

barelyhuman:

Enhancement / Fix

JSXFragment, JSXSpreadAttributes will not be rendered properly

## What is the new behavior?

Handles the above with added code in the original library

N/A

## Checklist

- [ ] Documentation - N/A
- [ ] Testing
- [ ] Ready to be merged

Waiting on the instruction, to do the above if the team still uses the babel transpiler and is not already handling the above mentioned Tokens elsewhere

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -94,6 +94,7 @@
     "apollo-link-batch-http": "^1.2.12",
     "apollo-link-context": "^1.0.18",
     "astring": "^1.7.4",
+    "@barelyhuman/astring-jsx": "^1.0.3",
     "axios": "^0.19.0",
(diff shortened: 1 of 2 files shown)


CompuIves:
Hey! Thanks for the contribution, do you have an example where we're currently not parsing/rendering this correctly?

barelyhuman:
Not sure, I didn't go through the entire codebase but the reason for update was the missing `JSXEmptyExpression` and `JSXFragment` writers missing in the original code which was derived from v0.1.x of `astring-jsx` which hasn't been updated with the new additions in `acorn-jsx`

PR #8062, opened 2023-09-08
feat: move workspace name in post-pro flow

alexnm:
Closes PC-1222

Removes the step to name the workspace / create team before going to stripe.

On the pro page, when you start trial / upgrade to pro you go directly to stripe
When stripe redirects to the dashboard, you end up with `new_workspace=true` in the query string.
The modal window now has 3 steps:
image
image
image

A few style adjustments to match on all 3 screens


necoline (approved):
Love the styling changes!

danilowoz (approved):

PR #8082, opened 2023-09-20
feat: add all checkout options on the pro page

alexnm:
Deploying it on stream for testing the full flow (+stripe)

The main change is on the /pro page where the dropdown appears instead of the single CTA button. All other changes are just params passing and renames / cleanup


danilowoz (approved):
LGTM

PR #8089, opened 2023-09-27
fix(pro): Ensure /pro link does not upgrade personal accounts

alexnm:
Last two places where it was possible to go to the pro page with the personal workspaceId in the URL


tromika (approved):
yayy

PR #8097, opened 2023-09-28
fix: handling legacy  persisted tab

alexnm:


CompuIves (approved):
Makes sense! Could you also link the issues here that are fixed with "Fixes ...", so they automatically get closed when this is merged?

PR #8123, opened 2023-10-12
fix: dev server crash

alexnm:


packages/app/config/webpack.common.js
```diff
@@ -257,6 +257,12 @@ module.exports = {
         test: /\.mjs$/,
         include: /node_modules/,
         type: 'javascript/auto',
+        use: [
```
CompuIves, line 260: This one we can remove i think

CompuIves (approved):

PR #8124, opened 2023-10-12
feat: small changes for pro workflows

alexnm:
Closes PC-1336

* Pro workspace badge shown at the top but not in the dropdown
image

* Redirect after login to the first pro workspace (in the future this will be the primary workspace instead of the personal one)
* Remove upsell banner when the user is already part of a pro workspace

### DIFF
--- a/packages/app/src/app/components/WorkspaceSelect/WorkspaceSelect.tsx
+++ b/packages/app/src/app/components/WorkspaceSelect/WorkspaceSelect.tsx
@@ -30,8 +30,8 @@ export const WorkspaceSelect: React.FC<WorkspaceSelectProps> = React.memo(
     const history = useHistory();
     const {
       isLegacyFreeTeam,
-      isLegacyPersonalPro,
       isInactiveTeam,
+      isPro,
     } = useWorkspaceSubscription();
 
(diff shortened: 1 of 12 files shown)


packages/app/src/app/components/WorkspaceSelect/WorkspaceSelect.tsx
```diff
@@ -30,8 +30,8 @@ export const WorkspaceSelect: React.FC<WorkspaceSelectProps> = React.memo(
     const history = useHistory();
     const {
       isLegacyFreeTeam,
-      isLegacyPersonalPro,
```
danilowoz, line 33: Can't we remove the `isLegacyPersonalPro` check from the hook?

danilowoz (approved):
Left a question, but it looks good to me

packages/app/src/app/components/WorkspaceSelect/WorkspaceSelect.tsx
alexnm: I want to leave it as a reference and it is still used to conditionally render some components in the pro page for example. it will be removed with the work starting on Monday

Issue #2058, opened 2019-06-13
Hot reloading fails: TypeError: Cannot read property 'Ctor' of undefined


tony19:
# 🐛 bug report

## Description of the problem
A previously working Codesandbox is now crashing with:
<details>
<summary><code>Cannot read property 'Ctor' of undefined</code></summary>
<pre>sandbox.115b0c69.js:1 TypeError: Cannot read property 'Ctor' of undefined
    at VueComponent.eval (eval at de (sandbox.115b0c69.js:1), <anonymous>:88:19)
    at invokeWithErrorHandling (VM3644 vue.common.dev.js:1859)
    at callHook (VM3644 vue.common.dev.js:4210)
    at VueComponent.Vue._init (VM3644 vue.common.dev.js:4988)
    at new VueComponent (VM3644 vue.common.dev.js:5137)
    at createComponentInstanceForVnode (VM3644 vue.common.dev.js:3286)
    at init (VM3644 vue.common.dev.js:3117)
    at createComponent (VM3644 vue.common.dev.js:5963)
    at createElm (VM3644 vue.common.dev.js:5910)
    at VueComponent.patch [as __patch__] (VM3644 vue.common.dev.js:6460)
    at VueComponent.Vue._update (VM3644 vue.common.dev.js:3936)
    at VueComponent.updateComponent (VM3644 vue.common.dev.js:4057)
    at Watcher.get (VM3644 vue.common.dev.js:4468)
    at new Watcher (VM3644 vue.common.dev.js:4457)
    at mountComponent (VM3644 vue.common.dev.js:4064)
    at VueComponent.Vue.$mount (VM3644 vue.common.dev.js:9034)
    at VueComponent.Vue.$mount (VM3644 vue.common.dev.js:11919)
    at mount (vue-test-utils.js:6187)
    at shallowMount (vue-test-utils.js:6218)
    at Object.eval (MyComponent.spec.js? [sm]:11)
    at vendors~sandbox.b6b2495b.chunk.js:310
    at new Promise (<anonymous>)
    at t.callAsyncFn (vendors~sandbox.b6b2495b.chunk.js:310)
    at sandbox.115b0c69.js:1
    at g (common-sandbox.52624a55.chunk.js:15)
    at Generator._invoke (common-sandbox.52624a55.chunk.js:15)
    at Generator.e.<computed> [as next] (common-sandbox.52624a55.chunk.js:15)
    at o (common-sandbox.52624a55.chunk.js:15)
[...]

PR #8127, opened 2023-10-14
fix(sandpack-core): loaderContext.emitModule() may replace tModule directly

arcturus011:
Bug fix

vue2 project HRM fails

https://github.com/codesandbox/codesandbox-client/issues/2058

## What is the new behavior?
vue2 project HRM works

1. create a vue2 project
2. modiy a App.vue file
3. see if HRM work correctly

## Checklist

- [ x] Documentation
- [ x] Testing 
- [ ] Ready to be merged
      
- [ x] Added myself to contributors table
      

## Cause Analysis
when a vue componet is created, [vue-hot-reload-api](https://github.com/vuejs/vue-hot-reload-api/blob/master/src/index.js#L3) will call `createRecord()` and keep its records in module scope. when HRM is triggered, `vue-hot-reload-api` will call `rerender` which reference saved records. so if we replace its tModule, we lost the records, thus raise an error.


CompuIves:
This is incredible! What a great find, I'm super impressed! I have a couple more meetings today, but I expect to properly review and merge this tomorrow.

Thanks for doing this fix, super impressive!

arcturus011:
> This is incredible! What a great find, I'm super impressed! I have a couple more meetings today, but I expect to properly review and merge this tomorrow.
> 
> Thanks for doing this fix, super impressive!

Thanks, i am looking forward to your code review!

PR #8147, opened 2023-10-24
feat: remove personal workspace dependency

alexnm:
Quite a big PR, but there's a lot of code removed in the diffs.

Main objectives:
* client implementation no longer depends on the existence of a personalWorkspaceId
* primaryWorkspaceId field used as a backup in preparation for the next releases
* free workspaces can be created again with a simple flow which does not involve going to stripe
* restrictions are lifted on workspaces (max public sandboxes and max public repos)
* inactive and free team badges are removed, in favor of showing a pro badge for the pro workspaces


packages/app/src/app/overmind/internalActions.ts
```diff
@@ -581,62 +582,68 @@ export const setViewModeForDashboard = ({ effects, state }: Context) => {
...
-      // Also set state while we're at it
-      state.dashboard.teams = teams.me.workspaces;
-      state.personalWorkspaceId = teams.me.personalWorkspaceId;
-      state.userCanStartTrial = teams.me.eligibleForTrial;
+    if (firstProWorkspace) {
+      actions.setActiveTeam({ id: firstProWorkspace.id });
+    } else if (firstWorkspace) {
+      actions.setActiveTeam({ id: firstWorkspace.id });
+    } else {
+      // TODO: Redirect to workspace setup
```
danilowoz, line 635: Is this an uncomplete step?

packages/app/src/app/pages/Dashboard/index.tsx
```diff
@@ -107,10 +107,7 @@ export const Dashboard: FunctionComponent = () => {
 
     const searchParams = new URLSearchParams(location.search);
 
-    if (JSON.parse(searchParams.get('new_workspace'))) {
```
danilowoz, line 110: I didn't get how this change is related to this scope, is this intentional?

packages/app/src/app/pages/Pro/index.tsx
```diff
@@ -46,12 +43,7 @@ export const ProPage: React.FC = () => {
...
-  // When  you are on pro?workspaceId=... the flow is set to manage existing workspace
-  // either by upgrading to pro or managing the subscription
-  if (urlWorkspaceId) {
-    return <ProUpgrade />;
-  }
+  return <ProUpgrade />;
 
-  // When you are on the /pro page, the flow is set to create a new pro subscription
-  return <ProCreate />;
+  // TODO: Cleanup ProCreate which is not used anymore
```
danilowoz, line 48: Is this a valid TODO?

packages/app/src/app/overmind/internalActions.ts
alexnm: Yes, this will be implemented in https://linear.app/codesandbox/issue/PC-1341/handle-empty-state-for-primary-workspaces, will drop the link there as well

packages/app/src/app/pages/Dashboard/index.tsx
alexnm: Yes. Before this change, the flow for creating a workspace was: stripe / name workspace / invite collaborators / import repos.

But now the workspace is created as a free one again, so: name / invite / import.

This param was used to trigger the workspace rename/invite/import flow after subscribing to pro, which we don't do now anymore.

packages/app/src/app/pages/Pro/index.tsx
alexnm: Yes, I think it's safe to remove that component, but I would wait until we have the first test completed on stream.

CompuIves (approved):
Huge PR! It's hard to review all code changes, but I'd say this code can be approved from the thorough testing that we've done.

PR #8166, opened 2023-11-03
feat: exprimental option to open sandboxes in v2

alexnm:


packages/common/src/utils/url-generator.ts
```diff
@@ -110,15 +110,29 @@ export const editorUrl = () => `/s/`;
...
 
-export const sandboxUrl = (sandboxDetails: SandboxUrlSourceData) => {
-  const baseUrl = sandboxDetails.isV2
-    ? `${v2EditorUrl()}sandbox/`
-    : editorUrl();
+export const sandboxUrl = (
+  sandboxDetails: SandboxUrlSourceData,
+  hasBetaEditorExperiment?: boolean
+) => {
+  const baseUrl =
```
danilowoz, line 117: Shoun't we discard SSE sandboxes here?

necoline: I was thinking that if it's an SSE sandbox, we should just convert it to cloud, but then you wouldn't be able to revert. So I think skipping the beta for SSE sandboxes makes sense

danilowoz (approved):

PR #8185, opened 2023-11-17
Feat: Hide .eslintrc.cjs file

christianalfoni:
To enable smoother transition to Devboxes we are adding the ESLint configuration, but hiding it for Sandboxes.

### DIFF
--- a/packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/Explorer/Files/DirectoryEntry/DirectoryChildren/index.tsx
+++ b/packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/Explorer/Files/DirectoryEntry/DirectoryChildren/index.tsx
@@ -1,4 +1,7 @@
-import { HIDDEN_DIRECTORIES } from '@codesandbox/common/lib/templates/constants/files';
+import {
+  HIDDEN_DIRECTORIES,
+  HIDDEN_FILES,
+} from '@codesandbox/common/lib/templates/constants/files';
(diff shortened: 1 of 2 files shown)


packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/Explorer/Files/DirectoryEntry/DirectoryChildren/index.tsx
```diff
@@ -79,7 +82,11 @@ export const DirectoryChildren: React.FC<IDirectoryChildrenProps> = ({
           />
         ))}
       {sortBy(
-        modules.filter(x => x.directoryShortid === parentShortid),
+        modules
+          .filter(x => x.directoryShortid === parentShortid)
+          .filter(
+            x => !(x.directoryShortid == null && HIDDEN_FILES.includes(x.title))
```
alexnm, line 88: does this work for nested files? or it is explicitly implemented to ignore only root level files?

alexnm (approved):
Left a question, but good to go

packages/app/src/app/pages/Sandbox/Editor/Workspace/screens/Explorer/Files/DirectoryEntry/DirectoryChildren/index.tsx
christianalfoni: Yes, only for root files 👍 

This stuff is history soon, so it is just to handle the `.eslintrc.cjs` file which is always on the root 👍

PR #8195, opened 2023-11-21
style: Update AI tool access labels and permissions

danilowoz:
Closes PC-1489


packages/app/src/app/constants.ts
```diff
@@ -30,10 +30,6 @@ export const TEAM_PRO_FEATURES: Feature[] = [
     key: 'boxes',
     label: 'Unlimited private devboxes and sandboxes',
   },
-  {
```
necoline, line 33: Can we move this to the free card instead of just removing it?

necoline (changes requested):

necoline (approved):

PR #8198, opened 2023-11-22
fix: ensure beta is switched off first when the disable form is submitted

alexnm:
There were reports of the switch not being reset after the form submitted. Could not reproduce but the only cause might be an exception thrown by the rows form submit, so I flipped the order to ensure the flag is always set to false.


christianalfoni:
Hm, I do not think this is the issue, cause the `effects` is not blocking. Did you try reproducing by only toggling, not fill in any Feedback and navigate away... then go back?

PR #8199, opened 2023-11-22
feat: Update gap and width in Sidebar components

alexnm:
Update gap from 2 to 1 in NestableRowItem. Update width from 40px to 36px with padding in RowItem.

Before
Screenshot 2023-11-22 at 11 51 52

After
Screenshot 2023-11-22 at 11 51 21


christianalfoni (approved):
Looking good to me 😄

PR #8208, opened 2023-11-29
Feat/tar normalize

Ku3mi41:

Feature

`normalizeTar` does nothing for non-standard packages with a folder name other than "package" in the root, such as "parse-json" in case of `@types/parse-json` package. As a result, the calling code cannot retrieve any files from the archive using paths relative to the "package/" folder. The error on the screenshot was caused by a call with the argument "/package.json", but for `@types/parse-json` we still got the path "parse-json/package.json".

[Exist Sandpack issue](https://github.com/codesandbox/sandpack/issues/641)

image

## What is the new behavior?

Instead of replacing the first word "package" in the path, any first word will be replaced. Thus, we have the same behavior for archives containing a `package` in the root, and raised by one level paths for archives with any other folder in the root.

1. Fixed sandpack dist file with new regex.
2. Now my sandpack can load any package from my private npm registry.

## Checklist

- [ ] Documentation  "N/A"
- [x] Testing
- [x] Ready to be merged

p.s.: If we assume that someday we will receive an archive with files directly in the root, then this method will still not work, just like the previous one.


CompuIves (approved):
Good change, and makes sense! Thank you!

PR #8213, opened 2023-12-01
feat: ubb waitlist banner

alexnm:
A few small UI differences compared to the figma design:
- `Pay only for what you use` in Everett, not in Inter
- no border radius to match the subscription cards

image

### DIFF
--- a/packages/app/src/app/constants.ts
+++ b/packages/app/src/app/constants.ts
@@ -20,15 +20,15 @@ export interface Feature {
 export const FREE_FEATURES: Feature[] = [
   {
     key: 'editor',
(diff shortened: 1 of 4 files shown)


necoline (approved):

ed-codesandbox:
I also had mixed thoughts about the border radius. It doesn't conform to the style of pricing boxes, but at the same time signals that this is something that's outside the "normal" version of the page. If you think square corners work better, let's go with those.

Everett also works - if we use it, it should be medium weight though. Is this medium weight? It seems narrower but that might just be because it's dark on light / rendered in the browser.

PR #8233, opened 2023-12-20
feat: ubb flows

alexnm:


packages/app/src/app/components/WorkspaceSetup/WorkspaceSetup.tsx
```diff
@@ -0,0 +1,59 @@
...
+}) => {
+  const startFromIndex = startFrom ? steps.indexOf(startFrom) : 0;
+  const [currentStepIndex, setCurrentStepIndex] = React.useState<number>(
+    startFromIndex
+  );
+
+  const currentStep = steps[currentStepIndex];
+  const Component = STEP_COMPONENTS[currentStep];
+
+  if (currentStepIndex >= steps.length) {
```
danilowoz, line 31: Great way to avoid an useEffect hook!

packages/app/src/app/overmind/effects/api/types.ts
```diff
@@ -48,3 +48,14 @@ export type MetaFeatures = {
...
+export type VMTier = {
+  name: string;
+  shortid: VMType;
+  cpu: number;
+  memory: number;
+  storage: number;
+  creditBasis: number;
+};
+
+export type VMType = 'vm-1' | 'vm-2' | 'vm-3' | 'vm-4' | 'vm-5' | 'vm-6';
```
danilowoz, line 61: This is duplicated on `packages/app/src/app/constants.ts`

packages/app/src/app/overmind/internalActions.ts
```diff
@@ -615,6 +615,7 @@ export const initializeActiveWorkspace = async ({
       state.dashboard.teams = teams.me.workspaces;
       state.primaryWorkspaceId = teams.me.primaryWorkspaceId;
       state.userCanStartTrial = teams.me.eligibleForTrial;
+      state.userFeatureFlags = teams.me.featureFlags;
```
danilowoz, line 618: I thought we should get it from workspace, is this expected?

danilowoz (dismissed):

packages/app/src/app/overmind/internalActions.ts
alexnm: this is a separate flag at the user level, if the users are allowed to create new ubb workspaces

packages/app/src/app/overmind/effects/api/types.ts
alexnm: Right, good catch. One is for deserializing the rest endpoint result, the other is for hardcoded data. This one from overmind can actually be just a string, but I went the extra length to define its possible values like this. I wouldn't change it now, but later we will get more data from BE and I will refactor the usage

danilowoz (dismissed):

danilowoz (approved):

PR #8241, opened 2023-12-22
feat: improve templates management

alexnm:
### Added some consistency for the badges we use in the dashboard
* Default badge is `Devbox/Sandbox`
* Icon is always reflecting the type of box
* If turned into a template, the label `Template` replaces the `Devbox/Sandbox`
* If restricted (will be deprecated soon) the label turns into `Restricted` with a warning color instead of the bright purple one
* Same logic applies for the list view now as well

Screenshot 2023-12-22 at 14 14 31
image

### Fix context menu for templates
image

### Notification when the first template is created. Also the templates collection appears now in the dashboard immediately after creating the first template.
Screenshot 2023-12-22 at 14 15 01

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Components/Sandbox/SandboxBadge.tsx
+++ b/packages/app/src/app/pages/Dashboard/Components/Sandbox/SandboxBadge.tsx
@@ -22,7 +22,7 @@ export const SandboxBadge: React.FC<SandboxBadgeProps> = ({
   }
 
   if (isTemplate) {
-    boxTypeLabel = 'Template';
+    boxTypeLabel += ' template';
   }
 
   return (
--- a/packages/app/src/app/pages/Dashboard/Components/Sandbox/SandboxBadge.tsx
+++ b/packages/app/src/app/pages/Dashboard/Components/Sandbox/SandboxBadge.tsx
@@ -1,0 +1,36 @@
+import { Icon, Stack, Text } from '@codesandbox/components';
+import React from 'react';
+
+export interface SandboxBadgeProps {
+  isDevbox: boolean;
+  isTemplate: boolean;
+  restricted: boolean;
+}
+
+export const SandboxBadge: React.FC<SandboxBadgeProps> = ({
+  isDevbox,
+  isTemplate,
+  restricted,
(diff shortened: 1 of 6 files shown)


packages/app/src/app/pages/Dashboard/Components/Sandbox/SandboxBadge.tsx
```diff
@@ -0,0 +1,36 @@
...
+  let boxTypeColor = isDevbox ? '#FFFFFF' : '#A6A6A6';
+  let boxTypeLabel = isDevbox ? 'Devbox' : 'Sandbox';
+
+  if (restricted) {
+    boxTypeColor = '#F7CC66';
+    boxTypeLabel = 'Restricted';
+  }
+
+  if (isTemplate) {
+    boxTypeLabel = 'Template';
```
danilowoz, line 25: I think we could compound the template label with the project type, as I did on the editor:
Screenshot 2024-01-03 at 09 43 04

Do you think there is room for a longer label?

packages/app/src/app/pages/Dashboard/Components/Selection/ContextMenus/SandboxMenu.tsx
```diff
@@ -307,22 +307,24 @@ export const SandboxMenu: React.FC<SandboxMenuProps> = ({
...
             onSelect={() => {
               actions.dashboard.makeTemplates({
                 sandboxIds: [sandbox.id],
+                isOnRecentPage: location.pathname.includes('recent'),
               });
             }}
             disabled={restricted}
           >
-            Make {boxType} a template
+            Convert into a template
```
danilowoz, line 327: I like it! It is clearer for me now

danilowoz (dismissed):
Left a comment, but it's not a blocker

packages/app/src/app/pages/Dashboard/Components/Sandbox/SandboxBadge.tsx
alexnm: that's a good idea, I like it!

danilowoz (approved):

PR #8248, opened 2024-01-05
feat: replace workspace settings with workspace portal

alexnm:
All URLs pointing to the dashboard settings page are absolute paths to the workspace portal
Dashboard settings remains for now until we get this tested

Closes PC-1578

### DIFF
--- a/packages/app/src/app/pages/Dashboard/Content/routes/Settings/TeamSettings/TeamSettings.tsx
+++ b/packages/app/src/app/pages/Dashboard/Content/routes/Settings/TeamSettings/TeamSettings.tsx
@@ -14,7 +14,6 @@ import {
   Switch as RouterSwitch,
   useLocation,
 } from 'react-router-dom';
-import * as dashboardUrls from '@codesandbox/common/lib/utils/url-generator/dashboard';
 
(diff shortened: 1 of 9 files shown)


packages/app/src/app/pages/Dashboard/Content/routes/Settings/TeamSettings/TeamSettings.tsx
```diff
@@ -54,15 +53,15 @@ export const TeamSettings = () => {
               <RouterSwitch location={location}>
                 <Route
                   component={RegistrySettings}
-                  path={dashboardUrls.registrySettings()}
+                  path="/dashboard/settings/npm-registry"
```
danilowoz, line 56: It looks good to me, but why don't we keep these URLs dynamic? It's seems easier to maintain if we have a single place to change it

alexnm: These will go away soon, the page should not render normally unless you access it by its old URL. and the dynamic functions are now returning portal URLs which don't work as route paths

sannek (approved):

PR #8254, opened 2024-01-09
fix: update feedback form URL to v2 previews

lbogdan:
We're currently using a [sandbox](https://codesandbox.io/p/devbox/webhook-forwarder-s2973) for the feedback form submit API, which was recently moved from v1 SSE to v2. This PR updates the URL accordingly.


sannek (approved):
💯  Thanks

alexnm (approved):

PR #8268, opened 2024-01-12
feat: frozen experience

alexnm:
image

Updated screen
* All banners below have been centered
* Wording updated based on the feedback from @ed-codesandbox 
* Buttons that create branches, devboxes or import repos have been disabled while frozen

image

========== INITIAL COMMIT 

Running out of credits
Screenshot 2024-01-12 at 11 57 59

Out of credits
Screenshot 2024-01-12 at 11 56 32

Close to spending limit
Screenshot 2024-01-12 at 12 00 28

At spending limit
Screenshot 2024-01-12 at 12 24 07


ed-codesandbox (dismissed):
This looks great! Also, I laughed out loud when I saw the first pic. :)

A few things:

- If we went with centered content on our other similar banners, we should be consistent with that.
- Since both sentences are on the same line here, the copy of the "close to spending limit" banner is a bit repetitive. Let's change it to `<strong>Your workspace is close to hitting its spending limit.</strong> To keep using CodeSandbox, increase your spending limit.`

packages/app/src/app/hooks/useWorkspaceLimits.ts
```diff
@@ -49,7 +49,7 @@ export const useWorkspaceLimits = (): WorkspaceLimitsReturn => {
   const isCloseToOutOfCredits =
     isFree === true && !frozen && usage.credits > 350; // Random pick for now
   const isAtSpendingLimit = isPro === true && frozen; // TODO
```
danilowoz, line 51: Are we okay to merge without implementing these two conditionals above?

danilowoz (approved):

danilowoz:
BTW, Lint and Typescheck job have failed

packages/app/src/app/hooks/useWorkspaceLimits.ts
alexnm: the spending limit is actually ok like this, I will remove the TODO. the other is set to false until we have the exact credit usage from the BE and we can show the right banner

packages/app/src/app/hooks/useWorkspaceLimits.ts
```diff
@@ -1,95 +1,85 @@
...
 
-  const hasOver20Sandboxes = isFree === true && publicSandboxesQuantity > 20;
-  const hasOver200Sandboxes = isFree === true && publicSandboxesQuantity > 200;
+  const onDemandCreditsLimit = limits.onDemandCreditLimit ?? 0;
+  const isCloseToSpendingLimit =
+    !friendOfCsb &&
+    ubbBeta &&
+    isPro === true &&
+    !frozen &&
+    usage.credits / (limits.includedCredits + onDemandCreditsLimit) >
```
danilowoz, line 46:
```suggestion
    usage.credits / (limits.includedCredits + onDemandCreditsLimit) >=
```

PR #8274, opened 2024-01-15
feat: sign up for ubb self-serve flow

alexnm:
`/signup-beta` works the same as `/upgrade`, it just runs the new mutation AJ implemented, so the workspaces are flipped to UBB during the flow

### DIFF
--- a/packages/app/src/app/components/WorkspaceSetup/steps/SelectWorkspace.tsx
+++ b/packages/app/src/app/components/WorkspaceSetup/steps/SelectWorkspace.tsx
@@ -1,10 +1,10 @@
 import React, { useState } from 'react';
 import { Button, Stack } from '@codesandbox/components';
 import { useActions, useAppState } from 'app/overmind';
-import { SubscriptionStatus } from 'app/graphql/types';
+import { SubscriptionStatus, TeamMemberAuthorization } from 'app/graphql/types';
 import { useURLSearchParams } from 'app/hooks/useURLSearchParams';
 import { InputSelect } from 'app/components/dashboard/InputSelect';
-import { useHistory } from 'react-router-dom';
+import { useHistory, useLocation } from 'react-router-dom';
 import { StepHeader } from '../StepHeader';
 import { AnimatedStep } from '../elements';
 import { StepProps } from '../types';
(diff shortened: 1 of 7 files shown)


packages/app/src/app/components/WorkspaceSetup/steps/SelectWorkspace.tsx
```diff
@@ -16,18 +16,35 @@ export const SelectWorkspace: React.FC<StepProps> = ({
...
 
-  const workspacesEligibleForUpgrade = dashboard.teams.filter(
-    t =>
-      !(
-        t.subscription?.status === SubscriptionStatus.Active ||
-        t.subscription?.status === SubscriptionStatus.Trialing
-      ) && t.featureFlags.ubbBeta
-  );
+  const workspacesEligibleForUpgrade = dashboard.teams
+    .filter(
```
alexnm, line 27: rewrote this filter for clarity. filters out workspaces that are not free and those for which you are not an admin, then it looks at the ubbFlag based on the `upgrade` vs `signup-beta` flow

packages/app/src/app/pages/WorkspaceFlows/SignUpUBB.tsx
```diff
@@ -0,0 +1,67 @@
+import React, { useEffect, useState } from 'react';
```
alexnm, line 1: this is mostly a duplicate of `Upgrade.tsx`

packages/app/src/app/pages/index.tsx
```diff
@@ -221,6 +221,7 @@ const RoutesComponent: React.FC = () => {
             <Route path="/standalone/:componentId" component={StandalonePage} />
             <Route path="/create-workspace" component={CreateWorkspace} />
             <Route path="/upgrade" component={UpgradeWorkspace} />
+            <Route path="/signup-beta" component={SignUpUBB} />
```
danilowoz, line 224: Is there room for changing this URL? I think sign up might be misleading to a sort of authentication process instead of a plan. What about `/workspace-new`, `/workspace-beta`?

danilowoz (approved):
Left a comment, but not a blocker

packages/app/src/app/pages/index.tsx
alexnm: This is a temporary URL only and will be shared directly to people that want to subscribe to the new ubb models. I think this is the wording used in the banners that advertised ubb before the release. cc @necoline, this is an easy change, so whichever URL format works best we can deploy that to prod today

PR #8279, opened 2024-01-18
chore: cleanup workspace settings

alexnm:
image


danilowoz (approved):
image

OMG! This is big! 

I tested the dashboard it looks good to me. What do you think about dropping the Account section inside settings too? I believe the WS portal can do this too. 

Plus, I'd leave to merge this on Monday morning, just in case.

PR #8280, opened 2024-01-20
feat: new checkout options

alexnm:
Simplified plans page
image

Addons page
image

Spending limit with summary in place
image


packages/app/src/app/components/WorkspaceSetup/steps/Plans.tsx
```diff
@@ -386,61 +343,64 @@ const PlanAndPricing: React.FC<{ plan: PricingPlan }> = ({ plan }) => {
...
+      <Text css={{ marginLeft: '28px' }} size={3} color={secondaryColor}>
+        {heading}
       </Text>
-    ) : null}
+    )}
+    <Stack as="ul" css={{ padding: 0, margin: 0 }} gap={4} direction="vertical">
+      {features.map(feature => (
+        <Stack as="li" key={feature.toString()} gap={2} align="flex-start">
+          <Icon color={secondaryColor} name={itemIcon} size={20} />
+          {includeTooltips ? (
```
danilowoz, line 374: To simplify it, we could only use `TextWithTooltips` as it defaults to `Text` in case there's no tooltip, then you could drop `includeTooltips` prop too.

packages/app/src/app/pages/WorkspaceFlows/SignUpUBB.tsx
```diff
@@ -46,11 +43,6 @@ export const SignUpUBB = () => {
     );
   }
 
-  if ((workspaceId && ubbBeta === true) || isAdmin === false) {
-    // Page was accessed with a non-ubb workspace id
-    return <Redirect to={dashboardUrls.recent(workspaceId)} />;
```
danilowoz, line 51: Is this an intentional change?

danilowoz (approved):
Left two comments, but they are not blockers

packages/app/src/app/pages/WorkspaceFlows/SignUpUBB.tsx
alexnm: Yes, but the admin part is still needed, need to revert

danilowoz (approved):

PR #8281, opened 2024-01-21
fix: limit create branch buttons to a single row

alexnm:
Based on user feedback, for edge cases where a lot of repos are in use and work in primarily done on the main branch, I limited the number of cards that show to create a new branch to a single row in the grid

See the video (with mocked data)

https://github.com/codesandbox/codesandbox-client/assets/9945366/e28ab726-0860-4f24-a2e9-bd6f1f820646


christianalfoni (approved):
Great 👍 As long as it still shows most recent I think it makes a lot of sense 😄

PR #8289, opened 2024-01-24
feat: allow pro workspaces to use /upgrade to convert to ubb

alexnm:
image

To test:
* Ensure you have an old pro workspace (seat based, not flagged as ubb)
* Go to `/upgrade`, select the workspace
* Choose the pro plan, choose some addons, go over spending limit
* Last step is the accept changes now which should redirect to the portal instead of going to stripe

### DIFF
--- a/packages/app/src/app/components/WorkspaceSetup/steps/ChangePlan.tsx
+++ b/packages/app/src/app/components/WorkspaceSetup/steps/ChangePlan.tsx
@@ -1,0 +1,111 @@
+import React from 'react';
+import { Stack, Button, Text, Icon } from '@codesandbox/components';
+import * as dashboardUrls from '@codesandbox/common/lib/utils/url-generator/dashboard';
+
+import { useURLSearchParams } from 'app/hooks/useURLSearchParams';
+import { useActions, useAppState } from 'app/overmind';
+import { StepProps } from '../types';
+import { StepHeader } from '../StepHeader';
+import { AnimatedStep } from '../elements';
+
+export const ChangePlan: React.FC<StepProps> = ({
+  onPrevStep,
+  onDismiss,
+  currentStep,
+  numberOfSteps,
+}) => {
+  const actions = useActions();
+  const { checkout } = useAppState();
+  const { getQueryParam } = useURLSearchParams();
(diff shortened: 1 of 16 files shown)


ed-codesandbox (approved):
I don't have a valid WS to check with at the moment - but the screenshot looks great 👌

packages/app/src/app/components/WorkspaceSetup/steps/Create.tsx
```diff
@@ -70,6 +70,7 @@ export const Create: React.FC<StepProps> = ({
 
         setQueryParam('workspace', team.id);
         setFreshWorkspaceId(team.id);
+        await actions.dashboard.getTeams();
```
danilowoz, line 73: Does this fix the issue with the sidebar, which couldn't find the current workspace?

packages/app/src/app/components/WorkspaceSetup/steps/ChangePlan.tsx
```diff
@@ -0,0 +1,111 @@
...
+            <Text>{checkout.totalCredits} VM credits</Text>
+            <Text>{checkout.totalSandboxes} Sandboxes</Text>
+          </Stack>
+          <Text size={6} color="#e5e5e5">
+            ${checkout.totalPrice} / month
+          </Text>
+        </Stack>
+
+        <Text>
+          Additional VM credits are available on-demand for $0.018/credit.
```
danilowoz, line 65: Isn't this value available in a constant?

danilowoz (approved):

packages/app/src/app/components/WorkspaceSetup/steps/ChangePlan.tsx
alexnm: Yes, could be a good idea as it's used in a few places

packages/app/src/app/components/WorkspaceSetup/steps/Create.tsx
alexnm: Yes, I think so. Weirdly on stream that problem was not reproducing, I'm curios what happens when this goes to prod

PR #8293, opened 2024-01-26
fix: better signaling for drafts in create modal

alexnm:
image

Closes PC-1669


CompuIves:
@alexnm I was wondering if we could add an option here to choose where the dev/sandbox is put when creating it? Like a dropdown of all folders. I can look into implementing that if you like it.

alexnm:
> @alexnm I was wondering if we could add an option here to choose where the dev/sandbox is put when creating it? Like a dropdown of all folders. I can look into implementing that if you like it.

Yes! we talked about that, but I feel like it's a bit more work. Feel free to pick it up or I can have a look immediately after the release next week

danilowoz (approved):

PR #8301, opened 2024-01-29
feat: add a "Go to Dashboard" button to /s

CompuIves:
Noticed that people can get lost when they get redirected to the "Create Sandbox" modal, with this change we do give them a way to go to the dashboard:

image

### DIFF
--- a/packages/app/src/app/components/Create/GenericCreate.tsx
+++ b/packages/app/src/app/components/Create/GenericCreate.tsx
@@ -1,9 +1,16 @@
 import React, { useEffect } from 'react';
 
-import { Stack, Text, IconButton, Element } from '@codesandbox/components';
+import {
+  Stack,
+  Text,
+  IconButton,
+  Element,
+  Button,
+} from '@codesandbox/components';
 import track from '@codesandbox/common/lib/utils/analytics';
 import { useActions } from 'app/overmind';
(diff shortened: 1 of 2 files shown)


alexnm:
Isn't `/s` mostly used by anonymous users? we should check if the user is signed in and only then show the link maybe?

alexnm (changes requested):
Isn't `/s` mostly used by anonymous users? we should only show the link when the user is signed in maybe

CompuIves:
Yeah that's interesting, so in this case, the user wanted to go create an account from the homepage immediately. So they pressed "Get started" from the homepage and got here, but they wanted to be in the dashboard to sign up and create a workspace. So I think this is especially useful for anonymous users.

CompuIves:
Yeah hmm, I think we should still give people the opportunity to explore the editor/product before putting in a sign in wall. That's also why I was wondering if we could allow people to open the editor of devbox templates without signing in.

alexnm:
ok, sounds good

maybe in the future we can differentiate some of the CTAs on the website, like have one for `trying out` the product (sounds very weird when typing it :D) and another CTA for sign up / sign in.

alexnm (approved):

CompuIves:
Yeah, I agree. Good idea!

PR #8302, opened 2024-01-29
feat: add support for selecting folder when creating box

CompuIves:
image
image
image

### DIFF
--- a/packages/app/src/app/components/Create/CreateBox/CreateBoxForm.tsx
+++ b/packages/app/src/app/components/Create/CreateBox/CreateBoxForm.tsx
@@ -14,11 +14,14 @@ import { useWorkspaceSubscription } from 'app/hooks/useWorkspaceSubscription';
 import { useGlobalPersistedState } from 'app/hooks/usePersistedState';
 import { sandboxes } from '@codesandbox/common/lib/utils/url-generator/dashboard';
 import { docsUrl } from '@codesandbox/common/lib/utils/url-generator';
+import { PATHED_SANDBOXES_FOLDER_QUERY } from 'app/pages/Dashboard/queries';
+import { Query } from 'react-apollo';
 import { CreateParams } from '../utils/types';
 
 interface CreateBoxFormProps {
   type: 'sandbox' | 'devbox';
-  isDraft?: boolean;
+  collectionId: string | null;
+  setCollectionId: (collectionId: string | undefined) => void;
   onCancel: () => void;
   onSubmit: (params: CreateParams) => void;
 }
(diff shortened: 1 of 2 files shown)


packages/app/src/app/components/Create/CreateBox/CreateBoxForm.tsx
```diff
@@ -14,11 +14,14 @@ import { useWorkspaceSubscription } from 'app/hooks/useWorkspaceSubscription';
...
 import { sandboxes } from '@codesandbox/common/lib/utils/url-generator/dashboard';
 import { docsUrl } from '@codesandbox/common/lib/utils/url-generator';
+import { PATHED_SANDBOXES_FOLDER_QUERY } from 'app/pages/Dashboard/queries';
+import { Query } from 'react-apollo';
 import { CreateParams } from '../utils/types';
 
 interface CreateBoxFormProps {
   type: 'sandbox' | 'devbox';
-  isDraft?: boolean;
+  collectionId: string | null;
```
alexnm, line 23:
```suggestion
  collectionId: string | undefined;
```

packages/app/src/app/components/Create/CreateBox/CreateBoxForm.tsx
```diff
@@ -38,6 +42,7 @@ export const CreateBoxForm: React.FC<CreateBoxFormProps> = ({
   const effects = useEffects();
   const nameInputRef = useRef<HTMLInputElement>(null);
   const { isPro } = useWorkspaceSubscription();
+  const isDraft = collectionId == null;
```
alexnm, line 45:
```suggestion
  const isDraft = collectionId === undefined;
```

I'd keep consistency and stronger checks if this was defined as string | undefined in the top component

packages/app/src/app/components/Create/CreateBox/CreateBoxForm.tsx
```diff
@@ -124,6 +129,37 @@ export const CreateBoxForm: React.FC<CreateBoxFormProps> = ({
           />
         </Stack>
 
+        <Stack direction="vertical" gap={1}>
+          <Text size={3} as="label">
+            Folder
+          </Text>
+
+          <Query
```
alexnm, line 137: didn't know we have this :D

alexnm (approved):
I think this is useful for further clarifying the difference between drafts and folder sandboxes. I wonder if we should move the VM specs to the left column, as it's more like the starting specs based on the template, rather than a choice

packages/app/src/app/components/Create/CreateBox/CreateBoxForm.tsx
CompuIves: I found it in another component I remembered (the sandbox move modal) where we show all collections 😄

danilowoz (commented):
👍 LGTM

Issue #8299, opened 2024-01-28
Define API `query` parameter regression


oliviertassinari:
# 🐛 bug report

## Preflight Checklist

- [x] I have read the [Contributing Guidelines][contributing] for this project.
- [x] I agree to follow the [Code of Conduct][code_of_conduct] that this project
      adheres to.
- [x] I have searched the issue tracker for an issue that matches the one I want
      to file, without success.

## Description of the problem

The `query` param described in https://codesandbox.io/docs/learn/sandboxes/cli-api#supported-parameters doesn't seem to work.

## How has this issue affected you? What are you trying to accomplish?

I noticed because in https://mui.com/material-ui/react-switch/#color opening a codesandbox does no longer set the file in focus. The DX could be better

Actual:

![Screenshot 2024-01-28 at 17 50 35](https://github.com/codesandbox/codesandbox-client/assets/3165635/299bcc59-8d45-438a-af7b-5f0830a0b959)

Expected:

![Screenshot 2024-01-28 at 17 50 40](https://github.com/codesandbox/codesandbox-client/assets/3165635/605710a2-6217-4270-a22c-695e79fc83b4)

### To Reproduce

https://codesandbox.io/p/sandbox/define-api-get-request-forked-v5zl4m?file=%2Findex.js%3A16%2C5

taken from https://codesandbox.io/docs/learn/sandboxes/cli-api#get-request. In practice, we use the POST API but I assume the bug is shared, as it seems to happen during an internal codesandbox redirection.

### Your Environment

| Software         | Name/Version |
| ---------------- | ------------ |
| Сodesandbox      |              |
| Browser          |              |
| Operating System |              |

[code_of_conduct]: https://github.com/codesandbox/codesandbox-client/blob/master/CODE_OF_CONDUCT.md
[contributing]: https://github.com/codesandbox/codesandbox-client/blob/master/CONTRIBUTING.md

CompuIves:
This is a regression! Fix is underway.

CompuIves:
Hmm, I can't reproduce it yet. Does it work for you in an incognito?

CompuIves:
I'm expecting that this could be a race condition, where the initial file is set too late (only after the editor is initialized), but I cannot repro it yet.

oliviertassinari:
A recording in incognito with the provided reproduction sandbox:

https://github.com/codesandbox/codesandbox-client/assets/3165635/941242e7-417a-42d3-9dfe-6ee49a1df153

[...]

PR #8309, opened 2024-01-30
fix: pass on query params when redirecting to v2

CompuIves:
Fixes https://github.com/codesandbox/codesandbox-client/issues/8299

When redirecting to v2, we were not passing on the query params. Because of this, e.g. the `file` parameter would get lost.

Tested this on the dev server inside CodeSandbox, now it's passed along.


christianalfoni (dismissed):
👍

danilowoz (commented):
👍 LGTM

PR #8310, opened 2024-01-31
feat: allow anonymous user to open devbox templates from /s

alexnm:
We don't really have a nice variety of devboxes on stream but I removed the sign in button from the templates list and I changed the action for `selectTemplate` to open for anonymous users and fork for signed in users.

Clicking on the template opens in the same window
Cmd+click still works to open in a new tab


CompuIves (approved):
Nice change!

PR #8323, opened 2024-02-04
feat: allow only pro workspaces when moving out of current workspace

alexnm:
This is more of a business/product decision, but since we know some teams use this feature, should we allow moving content into other free workspaces? Could be a way to bypass restrictions, especially since we don't yet enforce a fixed number of free workspaces users can create. 

Before
image

After
image


CompuIves (approved):
The code looks good to me! The only thing I wonder is how we can communicate this. Should we show some text explaining it, or should we maybe add this to an FAQ in the pricing page or the docs?

necoline (approved):

PR #8335, opened 2024-02-07
feat: use pricing endpoint and refactor plans and addons data

alexnm:
Everything should run as before, but now the pricing endpoint is called when the plans are loaded. I did keep the previous data types as the client stores some additional numbers like save percentage and hardcoded strings for the UI

Closes PC-1638


packages/app/src/app/overmind/namespaces/checkout/constants.ts
```diff
@@ -1,50 +1,123 @@
...
+    'Unlimited API',
+    'VMs up to 64 vCPUs + 128 GB RAM',
+    'On-premise options',
+    'Private managed cloud',
+    'Dedicated support',
+    'SSO',
+  ],
+};
+
+export const ADDON_CREDITS_500: CreditAddon = {
```
danilowoz, line 75: Can't we get this from endpoint too?

danilowoz (approved):

packages/app/src/app/overmind/namespaces/checkout/constants.ts
alexnm: good catch! I made another commit with extracting the addon values, I forgot about them at the end

danilowoz: So maybe we can remove them from `constants.ts` to avoid future confusion

alexnm: ah no, they are still needed there as the constants define the initial value and we store some additional things on the client, like the percentage deduction or the price before the deduction

alexnm: actually, my bad, we override credits and price, which is all we care about and type is kinda redundant there (I think). I'd still keep the values here because it doesn't hurt to have something on page load

danilowoz (approved):

PR #8342, opened 2024-02-12
fix: allow moving into friend of csb workspaces

alexnm:
Forgot to account for friend of csb in the filter there


sannek (approved):
🙌 Thanks!

danilowoz (approved):

PR #8355, opened 2024-02-20
chore: add tracking for requesting repo access

CompuIves:
Minor PR that will give us more insight into the drop-off rate of users when asked to sign in for repo access


alexnm (commented):
👍 looks good

PR #8357, opened 2024-02-21
fix: redirect to workspace after invitation accepted

alexnm:
Noticed a bug yesterday on the stream, as accepted invitations were not redirecting you to the actual workspace.
Also added a query param for future welcome banners


christianalfoni (approved):
Great stuff 👍

PR #8364, opened 2024-02-27
feat: add postgres as featured template

CompuIves:
Make the Postgres template more features to get a better sense if people like to see more full-stack templates.


danilowoz (commented):
👍 LGTM

PR #8368, opened 2024-02-28
fix(embed): show team name instead of author name

CompuIves:
This was a request from someone on support, it will show the team name rather than the author name.


alexnm (commented):
👍

PR #8377, opened 2024-03-07
feat: add cookie consent

alexnm:
Same as: https://github.com/codesandbox/codesandbox-site/pull/404

### DIFF
--- a/packages/app/src/app/index.js
+++ b/packages/app/src/app/index.js
@@ -3,16 +3,19 @@ import '@codesandbox/common/lib/global.css';
 import 'normalize.css';
 
 import './split-pane.css';
+import './utils/cookieConsent/cookieConsentTheme.css';
+import 'vanilla-cookieconsent/dist/cookieconsent.css';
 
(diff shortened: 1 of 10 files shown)


packages/app/src/app/index.js
```diff
@@ -94,10 +97,7 @@ if (process.env.NODE_ENV === 'production') {
   ];
 
   try {
-    initializeAnalytics({
-      sentryDSN: window._env_?.SENTRY_DSN,
-      amplitudeApiKey: window._env_?.AMPLITUDE_API_KEY,
-    });
+    initializeSentry(window._env_?.SENTRY_DSN);
```
alexnm, line 100: decoupled sentry and amplitude here, as only the latter is conditioned by the cookie consent

danilowoz: I didn't get how you enable Sentry only after the cookie consent, because I could only see amplitude being enabled after that

danilowoz (approved):

PR #8387, opened 2024-03-13
Revert "fix: issue with npm fetcher in sandpack, looks stuck in a loop (#8361)

### LINKED ISSUE #8384: The "Could not find module in path" error on an attempt to load a package that contains esm and cjs modules
# 🐛 bug report

## Preflight Checklist

      adheres to.
      to file, without success.

## Description of the problem

The CodeSandBox cannot load load the 'devextreme' package which is imported inside the 'devextreme-react' package

We used CodeSandBox to generate examples from our [demos](https://js.devexpress.com/React/Demos/WidgetsGallery/Demo/DataGrid/SimpleArray/MaterialBlueLight/). See "Copy To CodeSandBox" 

So, I'm unable to use CodeSandBox check issues or show demos to our clients.

### To Reproduce

Please open the example I shared below below.

### Your Environment

| Browser          |      Chrome   version 122      |
| Operating System |    Windows 11          |

## Additional information

The issue doesn't specific to the "devextreme-react" package. The issue is the same for the "devextreme-angular" and "devextreme-vue" packages. It started on March 8 according to our experience. The example I shared below was created earlier and working fine for some time.

CSB_Issue

CompuIves:
Fixes #8384

Tested the repro case without this change, and it worked again. I think we might be returning too early somewhere here.

### DIFF
--- a/packages/sandpack-core/src/manager.ts
+++ b/packages/sandpack-core/src/manager.ts
@@ -21,8 +21,7 @@ import {
   SerializedTranspiledModule,
 } from './transpiled-module/transpiled-module';
 import { Preset } from './preset';
-import {
-  fetchModule,
+import fetchModule, {
   setCombinedMetas,
   combinedMetas,
 } from './npm/dynamic/fetch-npm-module';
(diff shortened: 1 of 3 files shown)


alexnm (commented):
👍

PR #8390, opened 2024-03-13
feat: add param for prefill sandbox template

alexnm:
In order to test this, try and sandbox/devbox id in the following queries:
/dashboard/recent?create_devbox=devbox-id
/dashboard/recent?create_sandbox=sandbox-id


danilowoz (approved):
Tested and it looks good to me

PR #8393, opened 2024-03-14
fix: improve sort order for org repos

alexnm:
Uses the new sort option for organizationRepos, so there's no more jumping around when the entire list is loaded

Also made some changes to the tracking events, to ensure consistency


packages/app/src/app/overmind/effects/gql/dashboard/queries.ts
```diff
@@ -344,7 +344,12 @@ export const getPartialOrganizationRepos: Query<
   GetPartialGitHubOrganizationReposQueryVariables
 > = gql`
   query GetPartialGitHubOrganizationRepos($organization: String!) {
-    githubOrganizationRepos(organization: $organization, perPage: 10, page: 1) {
+    githubOrganizationRepos(
+      organization: $organization
+      sort: PUSHED
```
danilowoz, line 349: Cool!

danilowoz (approved):

PR #8403, opened 2024-03-22
feat: create sandbox from standalone modal

CompuIves:
Couple of changes:

- Folder & privacy are now coupled to each-other
- We can create sandboxes from the standalone modal
- We can pass in a minimum privacy

### DIFF
--- a/packages/app/src/app/components/Create/CreateBox.tsx
+++ b/packages/app/src/app/components/Create/CreateBox.tsx
@@ -18,8 +18,6 @@ import { pluralize } from 'app/utils/pluralize';
 import { ModalContentProps } from 'app/pages/common/Modals';
 import { useGlobalPersistedState } from 'app/hooks/usePersistedState';
 import { useWorkspaceLimits } from 'app/hooks/useWorkspaceLimits';
-import { SignInForTemplates } from 'app/pages/common/Modals/SignInForTemplates';
-import { SignIn } from 'app/pages/SignIn/SignIn';
 import {
   Container,
   Tab,
--- a/packages/app/src/app/components/Create/CreateBox.tsx
+++ b/packages/app/src/app/components/Create/CreateBox.tsx
@@ -42,6 +42,7 @@ import { TemplateInfo } from './CreateBox/TemplateInfo';
 import { useFeaturedTemplates } from './hooks/useFeaturedTemplates';
 import { useAllTemplates } from './hooks/useAllTemplates';
 import { mapSandboxGQLResponseToSandboxToFork } from './utils/api';
+import { WorkspaceSelect } from '../WorkspaceSelect';
 
 type CreateBoxProps = ModalContentProps & {
   collectionId?: string;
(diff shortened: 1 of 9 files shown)


packages/app/src/app/components/Create/CreateBox.tsx
```diff
@@ -63,6 +83,11 @@ export const CreateBox: React.FC<CreateBoxProps> = ({
     initialCollectionId
   );
 
+  const parsedUrl = new URL(window.location.href);
+  const sandboxIdToFork =
+    sandboxIdToForkProp ?? parsedUrl.searchParams.get('sandbox');
```
alexnm, line 88: is `sandbox=<id>` passed only when the components mounts on the `<Standalone>` page? I would suggest we parse the URL in that page and pass `sandboxIdToFork` as a prop to this component, so that it behaves similarly in the two scenarios:
- dashboard URL with `create_sandbox=<id>' query param -> opens the component with the id as a prop
- standalone page with `sandbox=<id>` -> opens the component with the id as a prop

Does that make sense?

packages/app/src/app/components/Create/CreateBox.tsx
```diff
@@ -162,22 +187,41 @@ export const CreateBox: React.FC<CreateBoxProps> = ({
...
+        body: {
+          title: name,
+          collectionId,
+          privacy: permission,
+          v2: createAs === 'devbox',
+        },
+      })
+      .then(forkedSandbox => {
+        if (forkedSandbox && isStandalone) {
+          window.parent.postMessage(
```
alexnm, line 208: Nice!

packages/app/src/app/overmind/namespaces/editor/actions.ts
```diff
@@ -796,11 +798,15 @@ export const forkExternalSandbox = async (
     } else {
       state.editor.sandboxes[forkedSandbox.id] = forkedSandbox;
 
-      effects.router.updateSandboxUrl(forkedSandbox, {
-        openInNewWindow,
-        hasBetaEditorExperiment,
-      });
+      if (!preventNavigation) {
```
alexnm, line 801: nitpick:  I would reverse this into a param called `redirectAfterFork` that is passed only when it's true and is false by default

alexnm (changes requested):
Just a few thoughts on the changes

packages/app/src/app/components/Create/CreateBox.tsx
CompuIves: Good one, will change it!

packages/app/src/app/overmind/namespaces/editor/actions.ts
CompuIves: I'll take a look. I was initially doing `preventNavigation` because we wouldn't have to change all other call-sites to this, but maybe this function is called from a single place.

alexnm (dismissed):

alexnm (commented):
👍

PR #8404, opened 2024-03-25
feat: Add workspaceId prop to StepHeader component

danilowoz:
- Fix an issue on the Select Workspace step
- Add a workspace caption on every step 

Screenshot 2024-03-25 at 09 45 32

Closes PC-1778

### DIFF
--- a/packages/app/src/app/components/WorkspaceSetup/Layout.tsx
+++ b/packages/app/src/app/components/WorkspaceSetup/Layout.tsx
@@ -31,6 +31,7 @@ export const WorkspaceFlowLayout: React.FC<{
         >
           <Stack
             css={{
+              width: '100%',
               height: '100%',
               padding: '64px',
               flex: 1,
--- a/packages/app/src/app/components/WorkspaceSetup/StepHeader.tsx
+++ b/packages/app/src/app/components/WorkspaceSetup/StepHeader.tsx
@@ -1,10 +1,12 @@
 import React from 'react';
 import { Icon, Stack, Text, Button } from '@codesandbox/components';
+import { useAppState } from 'app/overmind';
 
-export interface StepHeaderProps {
+interface StepHeaderProps {
   numberOfSteps: number;
   currentStep: number;
   title: string;
+  workspaceId?: string;
   onPrevStep: () => void;
   onDismiss: () => void;
 }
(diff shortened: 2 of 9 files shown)


packages/app/src/app/components/WorkspaceSetup/steps/SelectWorkspace.tsx
```diff
@@ -84,6 +84,7 @@ export const SelectWorkspace: React.FC<StepProps> = ({
             name="workspace"
             label="Select the workspace you want to upgrade"
             onChange={ev => setWorkspaceId(ev.target.value)}
+            value={workspaceId}
```
danilowoz, line 87: Here's the original problem: as no initial workspace ID was passed, the select showed the first item as selected, although it could be any option.

packages/app/src/app/components/WorkspaceSetup/Layout.tsx
```diff
@@ -31,6 +31,7 @@ export const WorkspaceFlowLayout: React.FC<{
         >
           <Stack
             css={{
+              width: '100%',
```
danilowoz, line 34: Fix some responsive issues

packages/app/src/app/components/WorkspaceSetup/StepHeader.tsx
```diff
@@ -15,48 +17,56 @@ export const StepHeader = ({
...
-        margin={0}
-        as="h1"
-        color="#fff"
-        weight="medium"
-        fontFamily="everett"
-        size={32}
-      >
-        {title}
-      </Text>
+        {workspaceName && <Text>Upgrade {workspaceName}&apos;s workspace</Text>}
```
alexnm, line 68: The same component is reused when you create a new workspace, so `Upgrade` will not work here. I think it's ok to use just the workspace name

packages/app/src/app/components/WorkspaceSetup/steps/SelectWorkspace.tsx
alexnm: nice catch!

alexnm (approved):

packages/app/src/app/components/WorkspaceSetup/StepHeader.tsx
danilowoz: This caption only shows up after the create and select step, but the implementation didn't make it clear

PR #8412, opened 2024-03-28
fix(esm-convert): read + write cases

CompuIves:
When converting ESM, and there's an expression like `--counter`, where `counter` is exported, we would wrongly convert it to `--exports.counter = counter`, which is invalid syntax. In this case, it's sufficient to do `--exports.counter`, and with this change we do that.


alexnm (commented):
👍

PR #8413, opened 2024-03-28
feat: use official templates from sandbox-templates

alexnm:
Core feature has been implemented:
* devbox templates are fetched from github (json file)
* sandbox templates are fetched from the old /official endpoint

Still to decide:
* what categories we show for devboxes, based on the tags
* should we order templates based on some priority or is it ok to show them in alphabetic order or based on number of forks?
* rely on `featured` tag instead of the FEATURED_IDS


CompuIves (approved):
Really clean code, nice additional code reuse!

PR #8422, opened 2024-04-09
fix: converting write ref with esm to cjs

CompuIves:
Important fix. Our quick esm->cjs converter doesn't handle write + read references well. When there's code like:

```js
  character = position > 0 ? charat(characters, --position) : 0
```

It converts `--position` to `--exports.position`. That's wrong, because now the local variable does not get changed. We can solve this by resolving the ast node higher, and change it to:

```js
  exports.character = character = position > 0 ? charat(characters, exports.position = exports.position = --position) : 0;
```

But right now we cannot do that. Because of that, we throw an error and rely on the slower converter (babel) to convert esm to cjs. That will solve an infinite loop problem that people using emotion are seeing, which crashes the whole browser..


DeMoorJasper (commented):
👍

PR #8428, opened 2024-04-16
feat: self-serve yearly

alexnm:
Not a lot of changes visually, but this allows for the `annual` plan to be selected by users and passed to stripe. 
The checkout calculations are aligned with the annual switch as well.


danilowoz (commented):
👍 LGTM

PR #8432, opened 2024-04-24
feat: auto-join workspace after sign-up

danilowoz:
Screenshot 2024-04-26 at 08 43 15

Closes PC-1806


christianalfoni (commented):
Thinking that we should disable the "next" button while it is figuring out eligible workspace to join? There can be latency and it feels a bit weird that it starts enabled and then suddenly becomes disabled. Is there a loading indicator while it is figuring out eligible workspace?

christianalfoni (approved):

PR #8447, opened 2024-05-13
fix(context-menu): use parent privacy on fork

danilowoz:


packages/app/src/app/overmind/namespaces/editor/actions.ts
```diff
@@ -766,11 +766,11 @@ export const forkExternalSandbox = async (
     customVMTier?: number;
     redirectAfterFork?: boolean;
     body?: {
-      collectionId: string;
+      collectionId?: string;
       alias?: string;
       v2?: boolean;
       title?: string;
-      privacy?: 0 | 1 | 2;
+      privacy?: number;
```
CompuIves, line 773: Hmm, shouldn't this be `0 | 1 | 2`? I thought it's nice to encode in the type that it can only be one of these three values.

CompuIves (changes requested):
one question about types

packages/app/src/app/overmind/namespaces/editor/actions.ts
danilowoz: I changed that because there is a mismatch type between GraphQL and our types. Unfortunately, the backend can't return a specific type like `0 | 1 | 2` but `number`, making the application use the less restrict one. 

Alternatively, I could only cast the place with the mismatch type to `0 | 1| 2` instead of changing the main function parameter.

CompuIves (approved):

PR #8467, opened 2024-06-03
feat: deprecate legacy editor usage

alexnm:
Legacy editor code is still in place, this PR just bypasses the experiment flag so all users are redirected to the new editor even if they opted out of it from the user settings screen.


christianalfoni (approved):

danilowoz (dismissed):
Do it!

christianalfoni (approved):

PR #8481, opened 2024-06-20
feat: sandpack service worker

danilowoz:


packages/app/src/sandbox/worker/sw.ts
```diff
@@ -0,0 +1,210 @@
...
+  const req = event.request.clone();
+  const parsedUrl = new URL(req.url);
+
+  /**
+   * TODO: figure out a way to filter bundler requests and
+   * local requests. Now, it only supports `public`
+   */
+  if (
+    parsedUrl.origin !== self.location.origin ||
+    !parsedUrl.pathname.startsWith('/public')
```
danilowoz, line 187: @DeMoorJasper and @CompuIves, one of the essential parts of the service worker is to intercept requests only made by the sandbox in the iframe and ignore all other requests, like static files by the bundler. 

However, I'm struggling to find an efficient way to ignore bundler requests from among the ones I want to catch. For example, the bundler requests files from `static/**/*` and from root, which makes it hard to check each file should be caught by the SW. For now, as a temporary solution, I'm considering all files coming from `/public` to belong to the sandbox and should be intercepted, while all others are ignored. 

Ideally, we should put all bundler static files inside another unique folder, like `__csb_static`, then we can easily ignore everything coming from this folder. But the webpack configuration is so overwhelming, and I don't know if it can cause side effects in regular sandboxes.

packages/app/src/sandbox/worker/utils.ts
```diff
@@ -0,0 +1,163 @@
...
+import { invariant } from 'outvariant';
+// @ts-ignore
+import SandpackWorker from 'worker-loader?publicPath=/&name=sw.[hash:8].worker.js!./sw';
+import { CHANNEL_NAME, IWorkerPingMessage } from './types';
+
+/**
+ * TODO: Find a way to get the SW file name from webpack
+ */
+const __SERVICE_WORKER_BUNDLE_NAME = SandpackWorker.toString()
+  .replace('function() {\n  return new Worker("/" + "', '')
```
danilowoz, line 12: @DeMoorJasper, this is a hack to get the service-worker file URL. How can I properly get the URL from webpack without any hack? I found a way in line 5 to import the file, but I need this same path in a variable to use I feel like.

packages/app/src/sandbox/worker/sw.ts
danilowoz: Well, I've been trying to tweak the webpack to bundle everything inside `__csb_static`, but this seems a massive work, and I don't think we should do it right (too many files to change and check). Once we validate this Service Worker solution and after deprecating the legacy editor, I will be more confident about making these changes. So, for now, let's go for the temp solution (restrict requests coming from `public` folder) and evaluate it with beta testers.

packages/app/src/sandbox/worker/index.ts
```diff
@@ -0,0 +1,111 @@
...
+    $channel: CHANNEL_NAME,
+    $type: 'worker/init',
+  };
+  worker.postMessage(workerInitMessage, [workerChannel.port2]);
+
+  return worker;
+});
+
+const parentPortPromise = new DeferredPromise<MessagePort>();
+window.addEventListener(
```
DeMoorJasper, line 33: This could potentially be a breaking change, but not sure if it would have a real impact?

DeMoorJasper (dismissed):
Looks good, just left one comment for something we might need to keep an eye out for regarding breaking change

alexnm (approved):

alexnm (approved):

PR #8512, opened 2024-07-11
fix: get max credits

danilowoz:


aj-foster (approved):
Sorry, this is why you don't let backend people review frontend code.

danilowoz:
No worries, this was bad! I didn't check exactly when these keys were used

PR #8538, opened 2024-08-08
fix: Fix redirect

christianalfoni:
I am not sure how we force the V2 redirect and then override with the new url query param of `editorMode=v1`. But I make an assumption here that the beta experiment is always true now.


alexnm:
not needed anymore

PR #8541, opened 2024-08-13
fix(template): duplicate issue

danilowoz:


packages/app/src/app/components/Create/TemplateList.tsx
```diff
@@ -41,7 +41,7 @@ export const TemplateList = ({
         <TemplateGrid>
           {templates.map(template => (
             <TemplateCard
-              key={template.id}
+              key={template.id + template.browserSandboxId}
```
danilowoz, line 44: This was causing duplicated templates

CompuIves (approved):

PR #8554, opened 2024-08-22
FIX: set undefined collection id instead of 'Drafts'

sannek:
Closes a recurring API Sentry Error.

TIL that if you have something like this:
```
<option key="drafts" value={null}>
  Drafts
</option>
```

It does _not_ actually give you `null` when you try accessing the value but it uses the text that's inside the tag. This explains why we keep getting API requests that have `collection_id: "Drafts"`. (I was suspicious and confirmed in the console)/

To make it explicit I added a value of `drafts` that we can check against, but open to alternatives! 

The reason that it's so intermittent and tricky to reproduce is that this _only_ happens when you select a different folder and then go back and select `Drafts` -- If you never select anything _other_ than `Drafts` it remains correctly initialised.

1. Step A
2. Step B
3. Step C

## Checklist

- [ ] Documentation
- [ ] Testing 
- [ ] Ready to be merged
      
### DIFF
--- a/packages/app/src/app/components/Create/CreateBox/CreateBoxForm.tsx
+++ b/packages/app/src/app/components/Create/CreateBox/CreateBoxForm.tsx
@@ -198,10 +198,12 @@ export const CreateBoxForm: React.FC<CreateBoxFormProps> = ({
                   icon={() => <Icon name="folder" size={12} />}
                   value={collectionId}
                   onChange={({ target: { value } }) => {
-                    setCollectionId(value);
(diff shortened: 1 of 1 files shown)


CompuIves (approved):
Well... TIL! Didn't know it would default to the key if value is `null`, but it makes sense. It probably only supports strings because ultimately this becomes HTML. `null` would just omit the `value`. This is a good approach!

danilowoz (approved):

PR #8590, opened 2024-08-30
fix: create sandbox url crash

alexnm:


sannek (approved):
A client PR I feel like i understand enough to approve 💪

PR #8601, opened 2024-09-10
fix(sandpack): load transpiled file

danilowoz:


package.json
```diff
@@ -76,6 +76,7 @@
     "prepare": "husky install"
   },
   "dependencies": {
+    "postcss-url": "^10.1.3",
```
CompuIves, line 79: This one will need to go to `packages/app/package.json`

packages/sandpack-core/src/npm/dynamic/fetch-protocols/utils/tar-store.ts
```diff
@@ -26,7 +26,13 @@ export class TarStore implements FetchProtocol {
     tarContents.forEach(tar => {
       normalized[tar.name.replace(/^[^/]+/, '')] = {
         // TODO(@CompuIves): store buffers rather than strings for binary files
-        content: Buffer.from(tar.buffer).toString(),
+        content: Buffer.from(tar.buffer).toString(
```
CompuIves, line 29: Maybe add a second todo comment, to both explain why we do this and that we want to default to binary format in the future.

CompuIves (changes requested):
Some small nits, but looks good in general!

CompuIves (approved):

PR #8605, opened 2024-09-11
fix(dashboard): folders with same name but different parents

CompuIves:
When creating a folder structure in the dashbboard like

- `/test/a/b`
- `/test2/b/b`

Then if you would open `/test/a`, it would show two `b` folders in the overview. It was a small logic error, which is fixed in this PR.


alexnm (approved):
Nice fix!

PR #8637, opened 2024-10-14
fix: Typeform URLs

filipelima18:

Broken URLs replaced with new ones.


filipelima18:
@alexnm looks like some required tests are failing - can you please check?

PR #8658, opened 2024-11-11
feat: add support for signin in to cursor

CompuIves:
Adds support for signing into Cursor


christianalfoni (approved):
Makes sense to me 😄

PR #8683, opened 2025-01-10
fix(sandpack): prefer import over require for exports resolving

CompuIves:
Similar to how we should nowadays prioritize `module` over `main`, we should also prioritize `import` over `require`. It will be a bit slower (as we need to transpile esmodules to commonjs to run it in the browser), but these days all bundlers prioritize esmodules over commonjs, and so not defaulting to it on CSB introduces some subtle bugs.


DeMoorJasper (approved):
lgtm

PR #8684, opened 2025-01-10
fix: upgrade react devtools

CompuIves:
Upgrades React DevTools to make them work with React 19. Fixes https://github.com/codesandbox/codesandbox-client/issues/8673


CompuIves:
I can't repro on staging anymore, merging this so we can test in prod.

PR #8846, opened 2025-11-27
chore: upgrade to Node 18

tristandubbeld:
Upgrades the codebase from Node 16 to Node 18.

## Changes

- Updated all Node version specifications (package.json engines, .nvmrc, Dockerfiles, CI config)
- Updated @types/node to ^18.0.0 in packages that use it
- Updated CONTRIBUTING.md to reflect Node 18 requirement
- Upgraded webpack from v4 to v5 in codesandbox-browserfs to support Node 18's OpenSSL 3.0 natively

## Why Node 18?

Node 18 is the current LTS version and provides better performance, security updates, and modern JavaScript features. This upgrade ensures we're using a supported Node version.

## Webpack 5 Upgrade

Upgraded webpack in codesandbox-browserfs from v4 to v5 to eliminate the need for the deprecated `--openssl-legacy-provider` flag. Webpack 4 uses legacy OpenSSL algorithms that are incompatible with Node 18's OpenSSL 3.0. The upgrade includes:

- webpack: ~4.31.0 → ^5.88.0
- webpack-cli: ~3.3.2 → ^5.1.0
- webpack-dev-server: ~3.3.1 → ^4.15.0
- source-map-loader: ~0.2.3 → ^4.0.0
- Updated library configuration to webpack 5 format
- Replaced `-p` flag with `--mode production` in build scripts

## Additional Fix

Added resolution to use local workspace version of @codesandbox/components in sse-loading-screen to prevent it from pulling the old npm version that still requires Node 16.


tristandubbeld:
Need to upgrade webpack first.

